PR #4, opened 2015-01-03
Adding Python-binding for find_candidate_object_locations

vinhkhuc:
Hi Davis,
This PR is for adding find_candidate_object_locations() to dlib's Python API. An example of how to use it is also included. There are a couple of things that I'd like to mention:

1) As shown in the comment, pykvals in the Python API is different from kvals in the C++ code. pykvals is actually a Python tuple (start, end, num) for linspace(start, end, num). I'm not sure whether dlib provides a Python object for matrix_range_exp.

2) Locations of candidate objects (rects) should be the result of find_candidate_object_locations_py() rather than its argument. However, I think it's better to be consistent with dlib's C++ API.

Please take a look and let me know whether I need to change something.
Thanks.

### DIFF
--- a/python_examples/find_candidate_object_locations.py
+++ b/python_examples/find_candidate_object_locations.py
@@ -1,0 +1,19 @@
+#!/usr/bin/python
+# This example shows how to use find_candidate_object_locations()
+
+import dlib
+from skimage import io
+
+image_file = '../examples/faces/2009_004587.jpg'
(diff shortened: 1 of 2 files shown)


davisking:
That's great, thanks for the patch.  I just merged it in.

vinhkhuc:
Thanks for the merge, Davis! Dlib has been very useful for me :)

PR #8, opened 2015-02-26
Locate ATLAS on RHEL/CentOS 7

nxwhite-str:
For whatever reason, the standard yum-installed ATLAS libraries were renamed for RHEL/CentOS 7.0.

### DIFF
--- a/dlib/cmake_find_blas.txt
+++ b/dlib/cmake_find_blas.txt
@@ -145,6 +145,18 @@ if (UNIX)
       mark_as_advanced( atlas_lib cblas_lib)
    endif()
 
+   # CentOS 7 atlas
+   if (NOT blas_found)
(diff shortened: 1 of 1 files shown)


davisking:
Sweet, thanks for the patch.

nxwhite-str:
Just an update:  this works, but isn't quite right.  CentOS 7 ATLAS implements the full BLAS/LAPACK functionality in two separate libraries: libsatlas (single-threaded), libtatlas (mutlithreaded).  My changes link to both libraries, but libtaltas is first, so the linker discards the symbols from libsatlas.  There's a big thread at https://fedorahosted.org/fpc/ticket/352 describing the current mess in Fedora between netlib/ATLAS/OpenBLAS library scemes.

davisking:
Ha, ok.  As long as it works then that's fine.

PR #10, opened 2015-03-16
Add capability to get detection confidences and weight index in python

jackculpepper:
- Added a second version of `run_detector_with_upscale()` so it can return additional vectors (via pass by reference)
- Modified the code that actually runs the detector -- I call the version that returns rect_detections
- Added code to split rect_detections into three vectors: rectangles, scores, and weight_indices
- Added a python example program that shows how to use the new bindings
 - Rather than loading the image with skimage, I used Pillow because Pillow does not depend on scipy which takes a long time to build.

### DIFF
--- a/python_examples/face_detector_scores_idx.py
+++ b/python_examples/face_detector_scores_idx.py
@@ -1,0 +1,13 @@
+
+from PIL import Image
+import numpy as np
+import dlib
+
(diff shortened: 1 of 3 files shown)


davisking:
Great, thanks for the code.

Cheers,
Davis

PR #11, opened 2015-03-22
python3 support

zsteinmetz:
changes raw_input() to input()

### DIFF
--- a/python_examples/train_object_detector.py
+++ b/python_examples/train_object_detector.py
@@ -117,7 +117,7 @@ for f in glob.glob(os.path.join(faces_folder, "*.jpg")):
     win.clear_overlay()
     win.set_image(img)
     win.add_overlay(dets)
-    raw_input("Hit enter to continue")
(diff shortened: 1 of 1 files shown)


davisking:
This doesn't work quite right in python2.7.  You get a "SyntaxError: unexpected EOF while parsing" when the user hits enter.  I looked around but didn't find anything that will work right in both python2 and 3.  But it seems like there should be some way to just wait for a keypress.  Any ideas?

zsteinmetz:
Oh sorry, I thought input() is backward compatible. I added some more lines to avoid the error in 2.7.

davisking:
Yeah, that's the simplest thing I could find too.  But it's still overly complex and distracting as part of an example program.  So I just added a dlib.hit_enter_to_continue() and replaced all the raw_input() calls with that.  So I'm not going to merge this pull as it's fixed now but definitely thanks for pointing this out.

Cheers,
Davis

PR #12, opened 2015-04-09
Fix dlib tests for c++11

Abai:
c++11 std::mutex clashes with dlib::mutex when:

using namespace std;
using namespace dlib;


davisking:
Indeed, thanks for the PR.

Cheers,
Davis

PR #13, opened 2015-05-20
Add a Python interface for the Correlation Tracker

patricksnape:
This adds a very simple Python interface for the correlation tracker. It also handles adding the drectangle class and a small update to the GUI code so that drectangles can be visualized.


davisking:
Awesome.  I'll merge this in later tonight when I get home.

PR #18, opened 2015-06-25
Made probability weighting in gibbs sampler dependent on parents prob…

okirmis:
…abilities instead of the children's

To be honest, I'm not 100% sure, but I think it makes more sense to use the parent nodes' probabilities for weighting the result as they values are used to look up the probability for the current node in the probability table.

Please correct me if I am wrong as I use this modified version for a university project.

### DIFF
--- a/dlib/bayes_utils/bayes_utils.h
+++ b/dlib/bayes_utils/bayes_utils.h
@@ -1137,8 +1137,8 @@ namespace dlib
                     set_node_value(bn, n, i);
                     samples(i) = node_probability(bn, n);
 
-                    for (unsigned long j = 0; j < bn.node(n).number_of_children(); ++j)
-                        samples(i) *= node_probability(bn, bn.node(n).child(j).index());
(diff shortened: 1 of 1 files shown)


davisking:
Why would this be the right thing to do?  Also, it breaks the bayesian network unit tests and causes the Gibbs sampler to produce the wrong outputs.

okirmis:
As I said, I am not 100% sure about this. But why do we multiply it with the probability of the children when we want to create a probability for the node's state which depends on its parents' states?

What tool did you use to calculate the results you compare the the output to in the unit test? I'd like to give it a try.

davisking:
The calculation depends on the parents as well as the children, the why is complicated.  You should google bayesian networks and read about d-separation.  I calculated the correct value with the junction tree algorithm.

okirmis:
Yes, as written in the "Edit" section, the source I used for the algorithm was updated an the error was fixed. The PDF I linked there also describes it like you implemented it.

PR #19, opened 2015-07-10
make dlib less arogant

col-b:
Often it's the case that non-linear optimization code (such as dlib) is very useful for performing complex operations in low-level processes, of potentially a massive multi-threaded application controlling autonomous vehicles, potentially in the air. A bad argument in a two iterations of the optimization should not suddenly drop the expensive vehicle to the ground.

You can never fully know the scope outside of your application. It's never ok for a library to call exit/abort/assert(false)!

### DIFF
--- a/dlib/error.h
+++ b/dlib/error.h
@@ -257,16 +257,12 @@ namespace dlib
             static bool is_first_fatal_error = true;
             if (is_first_fatal_error == false)
             {
-                std::cerr << "\n\n ************************** FATAL ERROR DETECTED ************************** " << std::endl;
-                std::cerr << " ************************** FATAL ERROR DETECTED ************************** " << std::endl;
-                std::cerr << " ************************** FATAL ERROR DETECTED ************************** \n" << std::endl;
-                std::cerr << "Two fatal errors have been detected, the first was inappropriately ignored. \n"
-                          << "To prevent further fatal errors from being ignored this application will be \n"
-                          << "terminated immediately and you should go fix this buggy program.\n\n"
-                          << "The error message from this fatal error was:\n" << this->what() << "\n\n" << std::endl;
-                using namespace std;
-                assert(false);
-                abort();
+                std::cerr << " *** FATAL ERROR DETECTED ***" << std::endl
+                          << " Multiple fatal errors have been detected, and they continually are ignored." << std::endl
+                          << " This may be a buggy program that you'd like to fix, or you may really know" << std::endl
+                          << " what you are doing." << std::endl
+                          << std::endl
+                          << " The error message from this fatal error was:\n" << this->what() << "\n\n" << std::endl;
             }
             else
             {


theidealist:
+1

davisking:
The only way to reach that abort() statement is if multiple fatal errors have been detected (e.g. out of bounds array accesses).  Moreover, as the message in the code points out, the only way that can happen is if the containing application is ignoring the errors and continuing to execute code which has gotten itself into an invalid state.

Think about it like this, would you disable memory segmentation so that code which runs outside the bounds of an array doesn't result in an application terminating segmentation fault?  No, you wouldn't (or hopefully you don't).  Forcing the application to continue running after it's gotten itself into an undefined state is not a useful thing to do.  What will happen is that it will crash somewhere else, or worse, continue running but give the wrong outputs.  In either case it will be very hard to debug.  If you are really unlucky the underlying error will go unnoticed for a long time, only surfacing late in the development process when it is more costly to fix.  

So no, I'm not silencing the fatal error output.

col-b:
I sort of view it as the caller's responsibility and right to determine what error results in terminating the application. In order to ignore your error, the user must explicitly ignore the thrown exception. At this point it's the user's fault for shooting themselves in the foot. By all means, that's their choice. You shouldn't help them debug their code.

Comparing to the OS is a little different as the OS will do whatever it can to keep your application running. If you reach outside your bounds, that's more of an encroachment on other application's space and memory. It's also a security risk. I don't think the comparison is valid. Besides, the comparison would be more like saying if I segfault, the entire machine should shut down, not just the offending application. That makes me feel like it's 1990 again.

In my particular use case, I'm interfacing with a user in an application where only one small sub-task is to do non-linear optimization. If, they accidentally select values that throw this error, I'm happy to report up to them that they are stupid and should choose better numbers. I'm not ignoring the error, I'm elevating to the place where action can be taken to correct it (i.e. in this case, the user) and that's exactly what exceptions are designed to do, elevate the error to the point where action can be taken to correct the error. Only sometimes should that mean termination, and that's the user of the lib's decision, not the lib itself. It should be my choice as the user of the library to decide if the program should exit or not. Should they user again choose to use bad values, exiting the application with no stack unwinding or state saving is simply bad design, and arrogant to assume understanding of the entire state of the system by a single low-level optimization library.

Again, I'm not ignoring the error, I'm providing a proper method for recovery from it, but I just can't do that 2 times in a row...? Why is it ok for a single handling of the exception, but not a second... i.e. why not just call exit the first time??? If the calling code was wrong once, what's to assume that it will not be called again. Why use exceptions at all, just assert right at all the errors (man, feels like 1990 again).

My only choice now is to 1) dynamically load / unload the lib to reset the local static, or 2) provide redundant checking of the user's input, which you check anyway, or 3) use a different non-linear optimization lib, but quite frankly, this one is quite nice except for this small design choice.

What about at least making this kill-my-program-after-an-easily-recoverable-error thing a configurable option -- wait, that's what exceptions are, my choice to ignore errors and (potentially) screw myself in the process.

Seems like a pretty reasonable pull request to me...

theidealist:
I can tell from the DLIB documentation this is something you feel passionately about and so, while I doubt there is a chance we can change your mind, I would make the following brief counter-arguments:

- Exception handling is just that -- handling for exceptional circumstances.  I love your two question analysis on whether or not to throw an exception.  I think it's dead on.  I would also agree with you that there are some errors that are non-recoverable.  Where I think we would disagree is on who gets to define what those "non-recoverable" errors are.  Library writers are emphatically not in that group.  Operating system writers are emphatically in that group.  Probably language standard writers are also in that group.  Library writers, however, must avoid at all costs either causing a non-recoverable error, or themselves simulating a non-recoverable error.  It is not in their purview to do so. 

-- Consider further the case where DLIB is a tertiary dependency.  That is, DLIB is actually not a direct dependency of my application but is a dependency of a dependency.  Perhaps the intermediate dependency is actually just a plugin, not even a core capability of the system.  Why, then should it *ever* directly invoke an all-stop?  DLIB is a *library*, it is indeed arrogant and in poor form to exit the program just because the library thinks it is the only thing important that could possibly be done in this process.

-- Why is only one exception permitted?  Shouldn't by this logic it exit immediately?  Why give the user one chance but not two?  Usually arbitrary decisions like this indicate a poor design, in my experience.

-- Lastly, I would take your segmentation fault analogy a little further and draw a similarity between a program and its outer operating system and a library and its outer executable.  The library should be able to inform the executable of a problem but should *never* be able to crash the executable, just as a program should be able to inform the outer operating system of a problem (through an unhandled exception, call to exit(), abort(), terminate(), etc.), but never be able to crash it.

davisking:
dlib::fatal_error should only be thrown for when an unrecoverable error has occurred.  Have you seen it come out of dlib in a case that you think was recoverable?  Where?  

Here is another example:

```matrix<double> m(3,3);
m(6,7) = 9;
```
What should happen?  

The reason DLIB_ASSERT throws an exception rather than calling assert()/abort() the first time is because in many environments the diagnostic output of doing that is much less clear or more difficult to trace than an exception being thrown.  However, the second check is there to catch people who are trying to ignore non-recoverable errors like the m(6,7) = 9 above.

You are missing the point of the fatal_error.  It's not indicating that an error has been detected in dlib. It's telling you that dlib has detected an error in the calling application.  Just like when the operating system detects an error in your application when you go out of bounds of an array, it terminates that application with a message instructing you to fix your application.

col-b:
why do you feel compelled to help me debug my application? is it dlib's job to debug my application--I thought that was my job? If have to actually choose to ignore the exception. That's a choice I made to say, "dlib, you don't know when it's actually time to quit, I think I can handle this error more gracefully." Even if it is the right thing to quite, maybe I want the stack to unwind, call my destructors, and save the state of things so I can actually debug the program.

If I provide dlib a starting value outside of its search range through a gui, is that unrecoverable, or is that just a case where the I should notify the user, "hey, that's not a good value, pick something between these two values."?

Then, they do the same thing for another parameter. Instead of helping them and saying... "nope, still not quite right" dlib just crashes the entire multi-threaded application that was potentially hours into computing some other non-linear optimization problem. Doesn't seem useful or desirable.

Seems like exceptions are *exactly* what you want, and instead of assuming you know the context in which you are being called, and inferring that this means terminate, you should allow that decision to be made by the caller, not the callee. That's what exceptions provide. Percolate up the stack, properly unwinding things until the exception is handled, which *may* terminate the program if I, the developer, can't find a way to properly handle the exception. Forcing termination and calling my program buggy is, well, not really a good way to handle the error.

davisking:
What function in dlib are you calling that is throwing fatal_error?  None of the optimization functions will trigger the fatal error based on any kind of iteration limit or search range limit.  Have you actually had a fatal_error thrown twice on you or is this whole discussion hypothetical?  

Also, generally speaking, you won't get a fatal error unless dlib is compiled in debug mode.  Are you using dlib in debug mode?  Don't do that unless you are debugging your application.  Compile it in release mode, it will run much faster.  However, it should be pointed out that in release mode the checks are disabled.  So if you call into dlib with invalid arguments (e.g. you index outside the bounds of a dlib::matrix) the program will either crash someplace or will give invalid outputs.

col-b:
You call both `assert(false)` (which compiles away in release) and `abort()` (which does not). Besides, again, you're trying to help me debug my application. I *want* my application in debug mode as I'm trying to debug it. I actually am running in RelWithDebInfo. I can't just build dlib as a .so in release anyway since I'm not linking in dlib since this is a template header-only library. So, are you telling me that I should compile my entire application in release mode? That's crazy. You're (in this context at least) a library author. Why do you feel compelled to put these type of constraints on your users, assuming they cannot recover from an error like this? You are not out of bounds, you are not in undefined behavior, you are not overflowing the stack, you are clearly handling this error--you're just choosing to use logic which ends my application to do it.

This is not hypothetical discussion BTW. The following is a snippet where I first ran into the problem, and what I had to do to avoid the problem. Another colleague asked what optimization lib I use, and I pointed him to dlib, but felt compelled to warn him about this poor design choice. In the following, both the number of parameters to be optimized, the starting values, and their search ranges are dynamically set by the user of the application.

```
// make sure our params, mins, and maxes all match up
if ( (m_optimizationParams.size() != m_optimizationParamsMax.size()) ||
     (m_optimizationParams.size() != m_optimizationParamsMin.size()) )
    throw( std::runtime_error("Number of params/min/max doesn't match up") );

// make sure the mins are less than the params
for ( uint ii(0) ; ii<m_optimizationParams.size() ; ++ii )
    if ( m_optimizationParams(ii) <= m_optimizationParamsMin(ii) )
        throw( std::runtime_error("The starting value cannot be less than the minimum value") );

// make sure the maxes are greater than the params
for ( uint ii(0) ; ii<m_optimizationParams.size() ; ++ii )
    if ( m_optimizationParams(ii) >= m_optimizationParamsMax(ii) )
        throw( std::runtime_error("The starting value cannot be greater than the maximum value") );
    
// run the optimizer
double rv(0.0);
if ( m_optimizationParams.size() == 1 )
{
    rv = dlib::find_min_single_variable(ErfSingle(m_params, erf),
                                        m_optimizationParams(0),
                                        m_optimizationParamsMin(0),
                                        m_optimizationParamsMax(0));
...
```

davisking:
No, really, the DLIB_ASSERT statement in find_min_single_variable() is only enabled when compiled in debug mode.  Specifically, this only happens when ENABLE_ASSERTS, DEBUG, or _DEBUG is #defined.  So if you compile in visual studio in Release or RelWithDebInfo then the DLIB_ASSERT will be off and no fatal_error will thrown and assert() and abort() will never be called.  This is true no matter how you call find_min_single_variable().  

This means you shouldn't rely on fatal_error being thrown to provide you with input validation from your user.  You must do as you show in this code snippet, explicitly validate your user inputs yourself however you think is appropriate.

If you fail to call a dlib function correctly, dlib will inform you via the mechanism you are complaining about.  But only if compiled in debug mode, which it sounds like you aren't doing.  So there is nothing to worry about, assert() will never be called.  You also seem to be suggesting that you would compile dlib's code in debug mode for a deployed application.  That is a very bad idea as it will be extremely slow due to all the DLIB_ASSERT statements.  

Or are you saying that, while you were debugging your application, you had some difficulty where, during a single run inside the debugger, 2 DLIB_ASSERTS failed and it was somehow difficult to figure out how to fix the bug?

col-b:
1) I'm not running in visual studio, I'm running in linux, gcc 4.8.4.
2) I did see the double assert or I would never have looked into the code to find the issue. It was easy to search for the helpful print that told me my code was buggy and thus refusal to continue computation on a 3rd attempt to get inputs correct.
3) I always compile my code in RelWithDebInfo. That's the best of both worlds, debug symbols for stack traceability and core file usefulness, with release optimizations. 

When I have time, I'll put together a simple sample program that exacerbates the problem and verify that I'm actually getting to the abort() call, when compiled in RelWithDebInfo. I'll add it as a sample program in this fork and notify you to take a peak.

This will have to wait for next week now as it is now the weekend.

davisking:
Just turning on debug info should not enable it.  It can only be enabled via those preprocessor defines. 

I realize this whole thing seems excessive but I get emailed daily by people using dlib who have basically no idea what they are doing and this kind of stuff makes a difference in usability for such people.  In fact, dlib didn't start out with the "kill your program on the second ignored fatal error" behavior.  I added it after I kept seeing people write try{} catch() { // do nothing} blocks that caused bugs that *were being detected* by dlib to go unnoticed and unfixed.  Since adding this to dlib this kind of user error has ceased.  

I also don't want people to rely on fatal_error being thrown for the correct behavior of their program (e.g. as some kind of user input validation).  It would be easy to write a program and to only test it in debug mode, which always generates fatal_errors.  Then when debug is turned off for the final product suddenly things don't work because the fatal_errors go away.  Having this kind of belligerent behavior in dlib also prevents that.

theidealist:
After thinking about this over the weekend, I've got say this behavior really surprises me.  Your mantra throughout seems to be that surprise is bad, and that you should make it easy to use the library correctly and hard to use it incorrectly.  I agree with you 100% there.

Throwing an exception once -- but not twice -- is definitely a significant departure from the norm and therefore surprising.  Throwing an exception that is dependent on compilation flags is another significant departure from the norm.  When that conditional exception is thrown from code that is generally *not* used as a normal shared library but instead compiled inline as part of my library or executable and is further influenced by the manner in which *my* code is compiled, the surprise factor jumps up again.

I get your desire to lessen your support by catering to users "who have basically no idea what they are doing".  I'd much rather you be spending your time making DLIB even more awesome than having to debug other people's code for them.  But I think the cost here is a significant increase in the surprise factor, and making it easy to use incorrectly.

Anticipating your response, I think you would say a segfault-on-second-exception makes it harder to use incorrectly, because in your mind an exception means you are using it incorrectly.  If that is the case, why wait for the second exception?  It's inconsistent.  Either the error is truly fatal (and you continue to believe DLIB is the most important thing a user could being doing with the process) and the process should exit immediately.  Or, the error is not truly fatal and was detected by logic and is therefore recoverable.

Your library is not the operating system.  It's neither your responsibility -- nor should it be in your purview -- to close the user's application for doing something "wrong".  Further, doing so conditionally based on compiler flags is *highly* irregular.  Consider std::vector for expected, normal behavior.  Users may use ::at() if they want bounds checked access, or ::operator[]() if they are willing to accept the consequences and want the speed.  Nowhere is there macro induced behavioral changes.  Nowhere is their forced shutdown if you use ::operator[] out-of-bounds *twice*.  I'm not saying DLIB needs to duplicate API, I'm saying DLIB needs to decide which one of these it's providing, and under no condition does it call ::asset(false) or ::abort() after the *second* exception.

davisking:
So what should happen when this dlib code is executed?

    matrix<double> m(3,3);
    m(6,7) = 9;

theidealist:
Exactly one of two things should happen when that code is executed:

1.) An exception is thrown to indicate out-of-bounds (probably std::out_of_range), as in:

std::vector<double> v(3);
v.at(6) = 9;

2.) Undefined behavior, similar to that of std::vector's square bracket operator, as in:

std::vector<double> v(3);
v[6] = 9;

I believe either of these is respectable behavior.  My personal preference (if you're asking it) would be for option 2, as C++ users should get used to having enough rope to hang themselves.  It comes with the territory and, while you have to be careful not to shoot yourself in the foot, the flip side is in performance and flexibility.

davisking:
Well, I'm not taking the precondition checks out of dlib.  It's one of the
main dlib features that makes debugging easier.  Alternatively, leaving
them on outside debug mode is a huge performance hit so they can't be on
all the time either.  It should also be noted that dlib initially used
assert() to indicate these errors but switched to using exceptions because
of complains from embedded developers working on platforms where it's much
easier to log an exception's message than it is to redirect the output of
assert() to somewhere a human can see it. So exceptions it is.

Finally, the reason there is the second check followed by abort() is to
prevent someone from accidentally writing code that depends on those
exceptions. In fact, that's what happened to @col-b. He had code that tried
to use them for user input validation and got the nasty dlib message saying
he shouldn't do that. If I had removed that second check he would have
instead unknowingly sent his application out, in release mode where the
debugging checks are gone, and users would have been greeted with
segmentation faults or other bad behavior. The only reason this was brought
to his attention is because of the "second check then abort()" feature you
are complaining about.  However, without that bad things would have
happened.

col-b:
I guess it's unfortunate to me that you've had to deal with enough user-side errors (or ignoring errors) that you are now jaded and feel compelled to validate user's code for them. I can imagine a path that lead you to this decision point, and I (we) feel it has compelled you to make a poor, surprising API choice that is a pretty clear departure from the standard practices motivated by the developers of the c++ language.

To be clear though, my initial use of the code with bad inputs during testing was that an exception was thrown. I did *not* ignore the exception, and was not surprised by it as it seemed like a very standard approach (I thought) for dlib to validate the expected inputs. I was in fact pleased to see that inputs were being validated and it was very easy for me to find the problem. I was able to easily stack trace it to find exactly the path that caused it, where it happened, and why. Furthermore, I made the design decision in my own application that I could propagate this error to the user by notifying my user(s) that those were bad inputs and help them take corrective action. That is not ignoring an exception, that is handling an exception. If you can write code to handle an exception, you should do so, otherwise you should just let it propagate to the top of the call stack and if nothing can handle the exception, then the code will indeed exit--that's the standard error handling practice for exceptions. I spent (wasted) some considerable time adding in the necessary handling of this exception to (I thought) properly handle the error and continue on.

After having taking this corrective exception handling approach, I then proceeded to test and was very surprised by the use of a local static to keep track of how many exceptions were thrown, and then to have my application terminated because more than one of these exceptions were thrown, with insulting print implying that my code was "buggy". That was indeed surprising behavior, and even though I felt there was a method for handling this exception, I was not permitted to do so since the second time abort() was being called.

Having the throwing of an exception be conditional based on the compiler flags is indeed surprising behavior, and yes, I would have relied on the exception catching bad user inputs in a release build, because I would have in no way expected the input validation to be conditional on compiler flags. How many other things in there are conditional on compiler flags? Maybe a better method than trying to switch on release/debug would be to have an explicit defined DLIB_DEBUG that handles this switching for you. That way I could choose to have all my code compiling in debug, while dlib would be in "release".

Your documentation does indeed detail the pre-conditions, and so I am in fact required, based on the contract defined by your function, to have the code snippet previously posted to validate my inputs - this is fine, you're right, I should have that there anyway. I, again, was misguided by the initial throwing of an exception, thinking that user-input was validated internally and therefore I should not have to validate it (i.e. avoid additional unnecessary duplicate input validation). An out-of-bounds segfault would have been much more straight forward for me to find and fix this problem (the lack of me validating the user's inputs). There is, however, noted omission in the documentation defining the contract that the behavior will significantly change from a release build to a debug build. This change in behavior is pretty non-standard and again, surprising and should at a minimum be loudly documented at the function-level.

So, I guess we should consider this issue more or less closed. You're clearly not going to change this behavior (please however consider the use of something like DLIB_DEBUG as a #define instead of just using debug/release to switch) and I clearly need to have the input validation in place. However, I'll have to suffer with sub-optimal code as I'm actually trying to debug other things in our system which require me to have debug builds on. BTW - it seems pretty preposterous that you'd suggest that your users compile all (in our case 500K lines) of our codebase in release just to get sensical behavior out of the one or two calls to dlib.

Also, I'd be curious to see the profiler outputs that prove that it's actually a performance hit to have input validation exception throwing on in release mode. Today's modern compilers are pretty good at optimizing this type of thing, and I don't believe you actually incur any performance degradation unless you actually have to throw. And, besides, when you have to throw, you're not concerned about performance since you're not actually able to compute anything, and when you don't have to throw, you don't get degraded performance. Also, I still have to do the validation check outside your code anyway... So, any performance improvement I get using dlib in release, I lose because I have to apply the same check prior to calling. You've just moved the (minimal) inefficiencies to the other side of the fence.

davisking:
I've been very patient with you guys, trying to explain things to you.  But you have no idea what you are talking about so it's tough.  In particular, you keep saying that the way dlib behaves is some kind of oddity, but in fact its use of software contracts is standard practice.  To help you understand, here are some examples I suggest you study.

In Visual Studio, the runtime libraries (including things like the STL) have both a debug and release version.  If you compile in release mode there are no checks on things like std::vector::operator[].  However, if you compile in debug mode Visual Studio uses a different version of the STL that includes many runtime checks for things like bounds checking on std::vector::operator[].  In their case, they check them with asserts and programs that violate those checks are terminated, but only in debug mode.  You can also read about how design by contract is done in a huge number of other contexts and languages on wikipedia: https://en.wikipedia.org/wiki/Design_by_contract.  One notable excerpt is the following:

>Contract conditions should never be violated during execution of a bug-free program. Contracts are therefore typically only checked in debug mode during software development. Later at release, the contract checks are disabled to maximize performance.

I would also suggest reading the Effective C++ books by Scott Meyers.  While not specifically about contract programming, they nevertheless contain a lot of good information about it. To get you started, here is a relevant passage about contract programming from Effective Modern C++ discussing contracts and how you might check them with exceptions:

>Now suppose that f(std::string) has a precondition: the length of its std::string parameter doesn't exceed 32 characters.  If f were to be called with a std::string whose length is greater than 32, behavior would be undefined, because a precondition violation by definition results in undefined behavior. f is under no obligation to check this precondition, because functions may assume that their preconditions are satisfied. (Callers are responsible for ensuring that such assumptions are valid.)  
>
>But suppose that f's implementer chooses to check for precondition violations. Checking isn't required, but it's also not forbidden, and checking preconditions could be useful, e.g., during system testing.  Debugging an exception that's been thrown is generally easier than trying to track down the cause of undefined behavior. 

To summarize, the idea of preconditions/contracts is a fundamental concept in software development.  It is widely understood that if you violate a function's preconditions that you invoke undefined behavior.  While I appreciate your enthusiasm, it is not sensible to complain on github about this.  The right thing to do is to write your software so that you don't violate function preconditions, thereby avoiding undefined behavior.  You should also read some software engineering books to better understand these general design principles that underlie most modern software libraries.  Reading the material I mentioned above would be a good place to start.

andrewcapodieci:
I'm a bit confused by your references to contract conditions. If it's widely understood that violating a function's preconditions invokes undefined behavior, why did you feel the need to explicitly define the behavior with an abort? Isn't that going against the definitions you're quoting?

This is a fascinating discussion to me by the way. I've had a similar problem with dlib exiting my code and it's very interesting to see both sides; I really am trying to learn here so I'd appreciate it if you could continue this discussion.

davisking:
Undefined behavior is just that, undefined.  Anything can happen.  Far and away the most common thing to do is to check contracts in some kind of debug mode and abort the program if the check fails, but to have "wildly undefined and probably really bad" behavior in non-debug mode.

For an even more extended discussion you could get a book like Object-Oriented Software Construction by Bertrand Meyer.  It's a little dated but is one of the classics of contract programming and discusses the subject at great length.

theidealist:
@andrewcapodieci, it is indeed common to have checks on input. One does not need to quote giants in the field like Steve Meyers to back that up. It is even common to have these checks implemented using assert (instead of exceptions or error return values), so that they can be compiled away when the user compiles with NDEBUG defined (this, among other things, is what ms visual studio does in the default release configuration). What is wrong here is that dlib uses exceptions as asserts, muddying the waters to ostensibly make debugging easier. You will not find this practice duplicated anywhere else that I've seen in my decade-plus of c++ development. Combined with a few other subtleties (like invoking exit() directly and the manner in which the library is intended to be linked), this really falls under the category of 'trap' in my *opinion* which would be shared, I think, by people like Meyers.

@davisking, you have been patient with us and I thank you for that. I'm saddened that we could not convince you of the error herein but I thank you again for the generally excellent library and especially for making it open source. I wish you all the best and apologize if somehow this got more personal or unprofessional than it should have on my part.

davisking:
I really would change it if someone pointed out a use case that demonstrated how calling assert() directly rather than using the current behavior was better.  Here "better" needs to be something concrete like a specific example where the current behavior leads to an undiagnosed bug but calling assert() directly does not.  Just saying "I don't like it" or "it just isn't done that way" isn't a reason.  

Right now, there are real dlib users that benefit from the current behavior.  Here is a specific example: when the program you are developing is expected to run remotely rather than locally on your desktop it can be difficult to see the output of an assert() call.  Moreover, maybe you want to redirect the error message from DLIB_ASSERT() to some logging or unit test server's database.  Implementing these things is very easy right now, you just write a catch block that sends the error message out however you like.  

As I mentioned in a previous post.  Initially DLIB_ASSERT() behaved just like the standard assert().  Personally, that's how I liked it.  But I kept receiving requests to make it throw an exception so that the output could be more easily redirected when running in background or embedded processes.  Those users clearly explained their situations and how it improved usability for them, so I made the change. 

You haven't yet told me a use case that demonstrates why it would be a good idea to change it back and wIthout that I'm not changing it.  In fact, it's not even totally clear to me what you are asking for.  Should DLIB_ASSERT() just call assert()?  Should it just throw an exception? Should it always be enabled?  Only be enabled in debug mode?  Each of these opinions seem to have been articulated in this thread.

theidealist:
At this point, I've tried to make the best argument I can. I think the behavior is surprising, misleading and a significant departure from standard approaches in the name of catering to a user base that should be able to accomplish their goals here in a general way outside of DLIB (do they not use any other libraries?).  Therefore, although I am tempted to continue, I relent.

As far as the exact nature of the "fix" proposed, I think you're right. We have jumped around a bit on this thread as we've come to understand your intent and the history behind the current implementation. In light of that, and notably without further inspection of the code, I'd offer one more alternative solution: more expressibility in the DLIB_ASSERT macro. Maybe a rename to DLIB_ASSERT_OR_THROW, which could be set to either ASSERT mode, or THROW mode, with ASSERT mode being able to be explicitly compiled out via *NDEBUG* (as per the language standard, instead of per Microsoft visual studio default build configuration conventions), and THROW mode not being compiled away and not invoking exit directly on second hit. Further, I'd recommend that the default mode be ASSERT mode unless set by the user at compile time via something like '-DLIB_ASSERTION_MODE_THROW'.

col-b:
Thanks for referencing some of my favorite authors/books. However, the fact that you felt compelled to do so means that I have not been clear in my arguments, and for that I appologize. So, I'm going to try and be as clear here as possible.

1) I am very familiar with contractual based programming and I further applaud your effort in using and documenting the interface to dlib. It's well beyond the quality of a typical open source project and I think I've been remiss in not mentioning how nice/easy it was to integrate into our system. Not verifying the inputs matched the stated contract was certainly a mistake on my part. I would expect undefined behavior in that case, and certainly that is not what I got--instead it was well-defined, just not documented.

2) The single surprising, non-intuitive issue we have with dlib is summarized by the pull request. The calling of assert/exit on the second time an exception is thrown. This is non-standard behavior, I'm sure the referenced giants would agree.

My recommendation for addressing this issue would be as follows, in order of most preferred to least preferred.

Suggestion 1:
An exception is thrown when input assumptions are violated regardless of compiler flags (i.e. debug/release, NDEBUG... whatever). If this truly has been profiled (since we're quoting the giants now... i.e. no premature optimization here?) and you are certain this is a performance problem, then this may be undesirable, however, I believe that in general you will find that it does not cause performance degradation, only incurring some overhead when an exception is actually thrown (which is not a concern here since the program just exits when this happens --- talk about a performance degradation).

Suggestion 2:
An assert is used in "release", and an exception is thrown in "debug", always, no changing behavior in between the first call and the second call. Also, "release" and "debug" should be flags specific to DLIB as @theidealist has suggested. This allows me to compile my code in debug/release independent of how DLIB is compiled.

Suggestion 3:
If you feel you must maintain the current behavior, at a minimum, I would request that this non-standard behavior (throw first, assert/exit second, differences in debug/release) be well documented for each function using this logic. Currently the contract implies that a single exception will be thrown (i.e. bobyqa_failure), and no mention of a fatal exception, and certainly no mention of an abort/exit begin called on second throw of this particular exception.

Suggestion 4:
Do nothing, leave it as is, this is in fact your lib, and the beauty of github is that we are free to fork and add any desired changes/enhancements we want. You are of course providing this lib free of charge and are in no way obligated to justify your decisions to a couple of random guys on the internet. :)

Once again, thank you for supplying such a quality lib to the community. I apologize if this thread has made you feel that we are unsatisfied with the library in general. In fact, this being the only nit, otherwise the lib is performing quite well.

davisking:
Contracts in dlib are explained in many places in the documentation.  I don't know how many times it's written that violating a function requirement leads to undefined behavior, but it's a lot.  Notably, in the introduction (http://dlib.net/intro.html) and in the "how to contribute" page.

Contract checking really does make dlib slow.  I regularly enable them and it is normal for programs that used to run in seconds to now run in minutes.  Obviously the specific slowdown depends on the program.  But the slowdown is severe, so they can't be on all the time.  Even just enabling bounds checking on the matrix or array2d objects slows image processing operations by a factor of at least 4 on most systems.  Enabling all the contract checking in an entire image processing stack just makes it worse.

In any event, I'm not going to make it possible for client code to persistently violate contracts.  Client code that violates contracts should be fixed because it is, by definition, buggy.  If I changed dlib to, for example, always throw an exception on contract violation there would be huge numbers of programmers writing code that used precondition violations as user input validation or other bizarre forms of flow control.  

One of the goals I have with dlib is to make using it educational.  In particular, users should learn about contract programming since it is a fundamental element of software development and yet most programmers I have met are unaware of it, and it isn't as if it's some kind of fringe idea.  It's one of the core ideas of modern computer science and software engineering.  For example, the C++ standard documents the behavior of the standard library with contracts, where a precondition violation means undefined behavior.  So I really want to encourage users to do and understand contract programming right.  That means, first and foremost, that it's unacceptable to violate a contract.

The standards committee is also working on adding contracts into C++ as a language feature for C++17 (see http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4293.pdf).  Notably, if you read that document, after all the years of work on it there is still debate about what should happen if a contract is violated.  Everyone involves agrees that a contract violation in a program is a bug in client code and should not be allowed to persist and that checking contracts all the time has unacceptable performance implications.  It is even likely that compilers will generate a compile time error when they are able to prove a contract would be violated.  However, if an exception is thrown or the program is directly terminated when a run-time check fails is still part of the discussion.  I'm mentioning this because the issues around what should happen are a lot less straight forward than you guys imagine.  

Ultimately, I suspect that the default behavior of a failed precondition check in "contract check mode" when contracts are eventually part of the language will be to do something like call std::terminate().   When it's finally standardized I'll update dlib to be in line with the standard, but until then I'm going to leave it as is.

davisking:
@xsacha cuda runtime error exceptions are not indicative of contract violations, so it is ok to catch them. There is only one exception type in dlib that is thrown for contract violations, and it isn't even always thrown.  Sometimes a contract violation aborts the program.  And in release mode, a contract violation could do anything, like cause a segmentation fault.

PR #20, opened 2015-07-13
Replace add_definitions with CMAKE_CXX_FLAGS in cmake file

Abai:
When using add_definitions erroneous flags are passed to dll resource compiler under MinGW. This breaks the build for anything linked to dlib, using a dll rc and MinGW.

The proper way is to append to CMAKE_CXX_FLAGS. For a discussion on this see:

http://www.cmake.org/pipermail/cmake/2011-July/045532.html

For CMake 2.8.12 and higher the proper way is add_compile_options(-Wreturn-type) OR target_compile_options(dlib PRIVATE -Wreturn-type)


davisking:
Cool, thanks for the patch.

PR #24, opened 2015-08-06
Remove redundant include

tzutalin:
Include "unicode.h" twice in serialize.h. Remove the redundant one.


davisking:
Thanks for the patch :)

Issue #28, opened 2015-08-18
Submit for PyPI


dashesy:
Do you plan to submit the package for [PyPI](https://pypi.python.org/pypi) (Python Package Index)? It will bring in many more Python programmers in the scientific computing community, who otherwise are less willing to compile from the source.

davisking:
I'm not going to but you are welcome to do it if you are interested :)

dashesy:
Thanks for the nice project! 
I am adding a `setup.py` [here](https://github.com/dashesy/dlib/commits/setup_py) it works with `bdist_egg`, `bdist_wheel`, `build`, ... but not `sdist` yet (have to put source files above `setup.py` during build).  
Will do a PR when everything is ready. I can also submit your package for PyPI with your permission, if you do not want to do that yourself or do not have time, but it should not be that difficult with a working `setup.py` in hand.

davisking:
Sweet.  Let me know when it's ready and I'll try it out.  

Is submitting to PyPI very involved?  I'm not at all familiar with any of this stuff.

dashesy:
Frankly I have not done that myself, that requires having a worth-while package and I do not have one :)

According to the documents [here](https://wiki.python.org/moin/CheeseShopTutorial) once initial channels are set up and ready, the package owner can do e.g. `python setup.py sdist upload` for any new release.
After that users can use `pip install dlib`. For users on Windows and OSX having a wheel channel (`bdist_wheel`) helps them not needing any compilation from sources (setting up `cmake`, ...), for Linux users `sdist` channel is fine.

davisking:
Yeah I've read a little about PyPI but always got the impression it wasn't going to work well for binary packages.  I also get emailed constantly by people who have weird different types of python setups (e.g. system installed python vs. mac ports vs. brew vs. anaconda vs. whatever) and they are all incompatible with each other at a binary level.  Is pip really going to deal with all that somehow?  Does it have a compile farm setup that creates binaries for all the different python configurations?

dashesy:
The source distribution (`sdist`) is exactly like compiling from source. 
[Wheel packages](http://pythonwheels.com/)  are based on a new [standard](https://www.python.org/dev/peps/pep-0427/) and distinguish different ABI. For example when I make a `bdist_wheel` I get this `dlib-18.17-cp27-none-linux_x86_64.whl`, which is a wheel for Linux on a 64bit Python2.7.
[...]

PR #29, opened 2015-08-20
Setup py

dashesy:
As discussed in Issue #28  `setup.py` can be used to invoke `cmake` as usual and then package the extension into egg or wheel or source distribution. Additional options can be passed to control `cmake` build with similar option for example `--yes DLIB_NO_GUI_SUPPORT` will set `DLIB_NO_GUI_SUPPORT`. 
To build from source `cmake` is needed, but wheels can be used for binary distribution.


davisking:
Awesome.  I tried a few things like running `python setup.py sdist`, `python setup.py build`, and `python setup.py install` on Ubuntu and they worked.  However, on Ubuntu, when I ran `python setup.py bdist_wheel` I got a error: invalid command 'bdist_wheel' message.  Shouldn't that have done something?


I also tried it on windows 7 where I have the 64bit version of python 3.4.3 installed and got these errors.  Any ideas?


C:\Users\davis\Documents\swdev\temp\python>python setup.py install
running install
running bdist_egg
running build
Detected Python architecture: 64bit
Removing build directory C:\Users\davis\Documents\swdev\temp\python\./tools/python/build
Configuring cmake ...
  File "setup.py", line 431, in <module>
    'Topic :: Software Development',
  File "C:\Python34\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Python34\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Python34\lib\site-packages\setuptools\command\install.py", line 67, in run
    self.do_egg_install()
  File "C:\Python34\lib\site-packages\setuptools\command\install.py", line 109, in do_egg_install
    self.run_command('bdist_egg')
  File "C:\Python34\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "setup.py", line 379, in run
    self.run_command("build")
  File "C:\Python34\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "setup.py", line 272, in run
    self.build_dlib()
  File "setup.py", line 343, in build_dlib
    if run_process(cmake_cmd):
  File "setup.py", line 161, in run_process
    close_fds=_ON_POSIX, preexec_fn=os.setsid)
AttributeError: 'module' object has no attribute 'setsid'

C:\Users\davis\Documents\swdev\temp\python>python setup.py build
running build
Detected Python architecture: 64bit
Removing build directory C:\Users\davis\Documents\swdev\temp\python\./tools/python/build
Configuring cmake ...
  File "setup.py", line 431, in <module>
    'Topic :: Software Development',
  File "C:\Python34\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Python34\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "setup.py", line 272, in run
    self.build_dlib()
  File "setup.py", line 343, in build_dlib
    if run_process(cmake_cmd):
  File "setup.py", line 161, in run_process
    close_fds=_ON_POSIX, preexec_fn=os.setsid)
AttributeError: 'module' object has no attribute 'setsid'

dashesy:
`invalid command 'bdist_wheel' ` means you should install wheels `pip install wheels`.

I guess `AttributeError: 'module' object has no attribute 'setsid'` means I need to take that out on Windows. I added a patch to see if that helps. I have a Windows in VM, as well as a Mac Mini I use to make packages, I will be looking at those tomorrow when I am back :)

dashesy:
Ok, now I have wheel for 32bit Windows `dlib-18.17-cp27-none-win32.whl`, will try to also build one for OSX today to make sure it works.

dashesy:
I just created `dlib-18.17-cp27-none-macosx_10_10_x86_64.whl` too, so it works on OSX too.
Also to install boost in OSX in the sane distribution (Homebrew): `brew install boost --with-python` then `brew install boost-python` is enough.

davisking:
Sounds like it's coming along :)

I just tried it on windows again and get a different error now though.  It's shown below:

C:\Users\davis\Documents\swdev\temp\python>python setup.py install
running install
running bdist_egg
running build
Detected Python architecture: 64bit
Removing build directory C:\Users\davis\Documents\swdev\temp\python\./tools/python/build
Configuring cmake ...
  File "setup.py", line 424, in <module>
    'Topic :: Software Development',
  File "C:\Python34\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Python34\lib\distutils\dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "C:\Python34\lib\site-packages\setuptools\command\install.py", line 67, in run
    self.do_egg_install()
  File "C:\Python34\lib\site-packages\setuptools\command\install.py", line 109, in do_egg_install
    self.run_command('bdist_egg')
  File "C:\Python34\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "setup.py", line 372, in run
    self.run_command("build")
  File "C:\Python34\lib\distutils\cmd.py", line 313, in run_command
    self.distribution.run_command(command)
  File "C:\Python34\lib\distutils\dist.py", line 974, in run_command
    cmd_obj.run()
  File "setup.py", line 258, in run
    self.build_dlib()
  File "setup.py", line 336, in build_dlib
    if run_process(cmake_cmd):
  File "setup.py", line 162, in run_process
    _log_buf(buf)
  File "setup.py", line 134, in _log_buf
    log.info(line)
  File "C:\Python34\lib\distutils\log.py", line 44, in info
    self._log(INFO, msg, args)
  File "C:\Python34\lib\distutils\log.py", line 33, in _log
    msg = msg.encode(encoding, "backslashreplace").decode(encoding)
AttributeError: 'bytes' object has no attribute 'encode'

dashesy:
that seems to be Python3, yes I will have to test that too :)

dashesy:
Ok it should now work with Python3 too, to test it in my distro with Python3 I ran this: `python3 setup.py bdist_wheel`.

davisking:
Nice work.   How about for 64bit python?  I just tried that on windows and it tried to build it in visual studio's 32bit mode which doesn't work when 64bit python is installed.

dashesy:
I added an option to pass to cmak's `-G` to generate the build system. Also used running Python's architecture to have cmake select the right architecture. 
Yes, since `setup.py` generates the ABI strings whatever ABI it has should match the ABI of the generated artifact. 
IOW, to make a 64bit library, one should be able now to run `setup.py` with a 64bit Python. I do not have a 64bit Windows to test this however.

davisking:
Are you using windows XP?  Any recent version of windows is 64bit.

Also, I just tried this by running:  `python setup.py -G "Visual Studio 12 2013 Win64" build` and got this error (I'm using 64bit python 3.4.3 on windows 7):
  File "setup.py", line 119, in <module>
    options, cmake_config, cmake_path, cmake_extra = _get_options()
  File "setup.py", line 77, in _get_options
    _cmake_extra.append('-G {gen}'.format(gen=arg.trim()))
AttributeError: 'str' object has no attribute 'trim'


You don't have windows 7?

dashesy:
sorry! I pushed a fix the error was not really related to 64bit but rather some basic string parsing.

Even though `-G` should work, the default should select the right build depending on the arch of Python, I have an image for Windows 7, will need to install VS, ... on it so it might take time but I try to test these options today.

dashesy:
Finally set up a 64bit Windows 7 and now it builds when `-G "Visual Studio some_version Win64"` is passed to it. I am reading cmake's code to see how generator names get populated to do this step automatically.

dashesy:
Now for 64bit Python on Windows it mimics cmake's resolution to get the build generator but selects the Win64 target. So by default `-G` will not be required for most users.

davisking:
Nice.  I just tried it and it worked perfectly :).  I really like the visual studio -G auto-detection.  That's pretty great.

One final thing before I merge it.  Can you put the new files into tools/python?  We can put a comment in the top level README.txt saying they are there and how to install the python bindings. I mainly want to move the files out of the root so that C++ users of dlib don't get confused and think that the setup.py is something they need to be concerned with or that there is some python based install process for the main C++ API.

dashesy:
Great! Actually I first added them to `tools/python` but the problem is `sdist` requires `setup.py` to in parent of the  source files so in 	ae5c624 I moved it to the root. 

That means from `tools/python` during build I will have to check if it is not being setup from `sdist` already, copy all the source files from parent to a `src` directory and build from there. Distinction of setup from `sdist` vs from git will make it fragile. If you definitely cannot tolerate the `setup.py` in the root, let me know and I work on a solution towards this, unless you have a better way.

Alternatively we can add a line in the Readme to tell C++ users they should ignore `setup.py` :) 
and who knows maybe Python binding (after adding binding for the rest, ...) brings enough traction to make it the main target.

davisking:
Hmm, can you make setup.py always copy the source into a sub-directory before it does anything, regardless of if sdist is running?  Then it should work correctly in every case right?

Ha, na, the C++ API is always going to be the primary API.  There is a huge amount of stuff in dlib that uses C++ features as part of the API that just don't translate into python.  And Python is really slow so there are things that just don't work in python due to performance reasons.

dashesy:
`setup.py` can also be run from `sdist` in that case it will be in the root of a unpacked zip archive that pip unzips in some temporary location. So it cannot e.g. copy from `../../dlib/*` to `src`. The solution would be somehow distinguish this from when `setup.py` is running from a git repo, at the same time it should always try to keep the `src` and actual repository in sync. Another solution is to detect if it is inside a git repo (has `.git`) pull the repository to `src`. In any case, it can be done but would be fragile because it falls outside of the usual usage.

Yes Python is slow, that is why I almost never run any tight loops in Python. For the most parts I use `numpy`/`scipy` or a good library like `dlib` or `scikit` to do the heavy lifting and glue them in Python. Most algorithms get enough advantage from this vectorization that they are good enough. When absolutely required I do it in Cython which is basically running C/C++ from Python.

The advantage of Python is that I can quickly get something, and save on development time, if the solution is working but slow will make it faster by doing it in C.

davisking:
Oh, I see.  Well, I guess we can leave it in the root and I'll figure something out if I get a lot of confused users asking about it.  Maybe it won't be a problem. 

I'm going to merge it in unless there is anything else you want to tweak.

dashesy:
Great! and if we see a drawback we can move it later. I cannot wait, been waiting to have this then before working on other bindings (all for Python). I noticed for example that binding `point_transforms`  I wont need to also depend on `scikit-image` for image warping and at the same time it will be an additional option to try and compare against.

I added the fact that it supports Python 3, otherwise it should be good. 

What do you think about PyPI? do you want to do it or am I allowed to work on that? I think this is a good library that many people will enjoy having.

davisking:
Yeah, nice work :)  I just merged it in.  

The more bindings the better.  It's also possible to bind in image loading as well to totally avoid scikit-image.  I would have done that in the first place but I assumed most people were using scikit-image and would want to use it along with dlib but in retrospect that's probably not true.  So feel free to add more bindings, especially ones that reduce dependencies.  Just don't break backwards compatibility with previous dlib releases. 

Sure, if you want to be responsible for uploading dlib releases to PyPI then that would be super.  You did write the whole setup.py after all so it makes sense :)

davisking:
Oh and when you make the binaries it's probably a good idea to compile them with --yes USE_SSE4_INSTRUCTIONS since that makes the face detector and related functionality much faster.  SSE4 has been in CPUs since about 2007 so it's probably safe to assume any user will have it.

PR #33, opened 2015-09-09
minor syntax changes for C++ compliance

mikeseven:


davisking:
Thanks for the patch.  Does the float/double switch avoid a compiler warning on some compiler?

Also, why the try/catch block that swallows the exception?  Did you mean to include that in the patch?

mikeseven:
On Sep 9, 2015 5:09 PM, "Davis E. King" <notifications@github.com> wrote:
>
> Thanks for the patch. Does the float/double switch avoid a compiler
warning on some compiler?

Yes and since all these are doubles, it's more appropriate.

>
> Also, why the try/catch block that swallows the exception? Did you mean
to include that in the patch?

Yes because otherwise it creates an error in msvs 2015. It has to do with
the macro throwing an exception in a destructor.

davisking:
Oh right, good catch about the destructor.  However, the right thing to do is to add an exception specification so that the error message is not hidden.  Otherwise, calling exit() is basically the same as what the compiler will do automatically which is call std::terminate().

So add "throw (std::exception)" to the end of the destructor declaration so the compiler knows this destructor is allowed to throw and I'll merge the pull request (and don't forget to remove the try/catch) :)

mikeseven:
Yes makes sense. Exit() was really a quick way ;-(

--mike

mikeseven:
ok re-throw is implemented and works on MSVS, GCC, Clang. I'm obliged to put ... instead of named exception since VC++ doesn't support named exception specification in throw() yet.

davisking:
Looks good except for the change to turn on vector instructions by default.  Not everyone has a CPU that supports AVX so it can't be the default (e.g. my still quite usable laptop doesn't have AVX).  SSE4 is also still likely to be unsupported on some dlib user's chips (e.g. atom processors don't have it).

mikeseven:
True with ancient windows models of course ;-)
But these days they are quite extinct and it's so easy for users to forget
to turn them on that this leads in general to bad performance. Likewise if
one add a bit of openmp, it's easy to get 4x performance, even on a $300
windows laptop, which I doubt is the customer of dlib...

--mike

davisking:
Not everyone is running big machine learning models and stuff like that.  I
know of plenty of applications that use dlib on low power embedded
processors.

mikeseven:
Me too but here we are talking windows. If it was Linux, I'd completely
agree.

I know enough (embedded) machines running Windows for less that $100 with
these vector instructions (one of my target BTW).

Anyway, not an issue for me. I'll keep my performance enhancement on my
repo.

--mike

davisking:
No worries.  So are you going to submit a pull request with just the other changes?

mikeseven:
OK found the issue with throw. So here is an update tested on gcc ubuntu 15.04), clang (MacOSX 10.11), MS VC++ (Windows 10).

davisking:
Why not just use the throw(...) version?  I would much rather have that than #ifdef statements in the code.  Also, why did you make the setting of -DENABLE_ASSERTS use FORCE?

And there are a whole lot of line changes in the cmake files now.  Most look like non-changes but it's hard to review.  Can you undo those changes to make review easier and also don't forget to take out the SSE4 and AVX defaulting to on.  As I said before, I'm not making the default behavior broken even if it's only for a minority of dlib users.

mikeseven:
Because throw(…) doesn’t work with GCC 4.9.x.
I’d prefer it too but that’s the only way I found to make it work across gcc, clang, VC++.

mikeseven:
I didn’t change enable_assert or anything else. I checked, I wonder if this is a cr/lf issue between the various systems???
ok I’ll remove the vector instructions by default.

davisking:
Ah, yes, throw(...) is a MS extension.  I just tried adding throw (std::exception) in visual studio 2013 and it worked without error.  I also think I've seen it work in older versions of visual studio but maybe I'm mistaken.  What version of visual studio did you find didn't accept it?

Also, you should be able to add cmake statements to your own project's CMakeLists.txt files that default enable AVX instructions.

PR #46, opened 2015-10-05
Added the C++11 noexcept specifier to the matrix class for C++11 conformant compilers.

frederikaalund:
**Rationale:** The `throw` exception specifier is [deprecated in C++11](http://en.cppreference.com/w/cpp/language/except_spec).

**Breaking Changes:** None as far as I can see. I've added an `#if #else` statement to maintain backwards compatibility with C++98.

Note that I haven't checked the remaining library. There may be other places in dlib where you can use the same approach. The logic can easily be hidden in a macro.

### DIFF
--- a/dlib/matrix/matrix.h
+++ b/dlib/matrix/matrix.h
@@ -1605,7 +1605,18 @@ namespace dlib
 
             literal_assign_helper(const literal_assign_helper& item) : m(item.m), r(item.r), c(item.c), has_been_used(false) {}
             explicit literal_assign_helper(matrix* m_): m(m_), r(0), c(0),has_been_used(false) {next();}
-            ~literal_assign_helper() throw (std::exception)
+            ~literal_assign_helper()
+// If C++11 or later.
+// Reference: http://www.stroustrup.com/C++11FAQ.html#11
+#if __cplusplus > 199711L
+				// In C++11 destructors are noexcept by default unless declared otherwise.
+				// Reference: http://en.cppreference.com/w/cpp/language/noexcept_spec
+				noexcept(false)
+// C++98 or earlier. Note that partly conforming compilers will take this path too. E.g., MSVC14 and earlier.
+#else
+				// Use the now deprecated throw specifier
+				throw(std::exception)
+#endif
             {
                 DLIB_CASSERT(!has_been_used || r == m->nr(),
                              "You have used the matrix comma based assignment incorrectly by failing to\n"


davisking:
What is the issue that this pull request solves?

frederikaalund:
> What is the issue that this pull request solves?

The primary motivation is C++11 conformance. Switching to `noexcept` now is a preemptive change. I can understand if you don't want to merge this pull request until compilers actually starts warning you about using the deprecated `throw` specifier. Personally, I prefer to change as soon as possible so that future users never get such a warning. As far as I can tell, the reason to use the exception specifier in the first place was to remove warnings introduced by (partly) C++11-conformant compilers (e.g., MSVC14) due to destructors suddenly becoming non-throwing by default.

Also, `throw(...)` introduces a [run-time exception check](http://stackoverflow.com/questions/13841559/deprecated-throw-list-in-c11) against the given list of exceptions (the `...` part). `noexcept` is dynamically checked as well but the check is binary (throwing or non-throwing) resulting in minimal branching (and better performance).

Note that all exception specifiers (i.e., both `throw` and `noexcept`) are part of the functions signature. Changing from `throw` to `noexcept` is a change of signature and thus a change of ABI. However, this is not a problem for dlib since everything is header only.

davisking:
I suspect it will be a long time until compilers issue warnings for this.
Additionally, my long term plan is to begin to require a C++11 compiler to
use dlib starting around 2017 or 2018 to give all dlib users plenty of time
to update their compilers. At that time I'll go over dlib and tweak things
as appropriate.  Ultimately, I don't want to end up with a source base full
of conditional compilation statements to support old compilers.  If there
is an issue right now, like a compiler warning, then doing whatever is
right to avoid it is fine.  But I don't want to add macros and conditional
compilation when there isn't any problem.

So if you want to submit an issue that contains a todo list for a C++11
dlib that includes this and whatever else makes sense that's fine with me.
Then once I think it's reasonable to require a C++11 compiler I'll make
sure those changes get put into dlib.  But for now I want to leave it alone
unless there is a pressing need.

mikeseven:
I respect your decision but all compilers now are C++11 and some like msvc don't even have a switch, they just compile in C++11 mode.
Personally, I hate seeing lots of warnings or even errors when using third party libraries and C++1x compilers tend to be rather verbose.
My fork does solve these issues, add speed up switches and support for openBLAS on windows, resulting in significant performance increase. As time permits, I also add openmp support.

Given your decision, we are unlikely to see these additions in dlib but feel free to merge them when you are ready.

Cheers,

--mike

On Oct 5, 2015 7:48 AM, "Davis E. King" <notifications@github.com> wrote:
I suspect it will be a long time until compilers issue warnings for this.
Additionally, my long term plan is to begin to require a C++11 compiler to
use dlib starting around 2017 or 2018 to give all dlib users plenty of time
to update their compilers. At that time I'll go over dlib and tweak things
as appropriate. Ultimately, I don't want to end up with a source base full
of conditional compilation statements to support old compilers. If there
is an issue right now, like a compiler warning, then doing whatever is
right to avoid it is fine. But I don't want to add macros and conditional
compilation when there isn't any problem.

So if you want to submit an issue that contains a todo list for a C++11
dlib that includes this and whatever else makes sense that's fine with me.
Then once I think it's reasonable to require a C++11 compiler I'll make
sure those changes get put into dlib. But for now I want to leave it alone
unless there is a pressing need.

frederikaalund:
Thanks for reviewing. I can completely understand why you don't want these kind of compile-time conditionals in your code when they don't solve any current issues.

> Additionally, my long term plan is to begin to require a C++11 compiler to
use dlib starting around 2017 or 2018 to give all dlib users plenty of time
to update their compilers.

Sounds fine to me!

davisking:
@mikeseven Are you sure you mean to reply to this pull request rather than the one you submitted a while ago?

mikeseven:
Yes both are related.

On Oct 5, 2015 12:57 PM, "Davis E. King" <notifications@github.com> wrote:

@mikeseven<https://github.com/mikeseven> Are you sure you mean to reply to this pull request rather than the one you submitted a while ago?

davisking:
@frederikaalund No problem :)

PR #54, opened 2015-11-03
Change relative path of configure files in CMakeLists.

vrichter:
Using CMAKE_PROJECT_DIR instead of CMAKE_SOURCE_DIR/../dlib
makes the build process more portable.

1. If dlib is used as subdirectory of another project CMAKE_SOURCE_DIR will point to a different path then expected in the original code. 

2. If for some reason the dlib directory is renamed configuring would break which is at least more restrictive than necessary.


davisking:
Nice catch, thanks :)

PR #59, opened 2015-11-30
removing deprecated IplImage to support opencv3

ttoeniges:
The new stable openCV version only have very limited IplImage support. I removed the IplImage dependency completely.
It is quite a hard cut but I think it is necessary if new openCV versions should be supported.

### DIFF
--- a/dlib/opencv/cv_image.h
+++ b/dlib/opencv/cv_image.h
@@ -4,7 +4,6 @@
 #define DLIB_CvIMAGE_H_
 
 #include <opencv2/core/core.hpp>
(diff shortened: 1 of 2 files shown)


davisking:
I've used it with opencv 3 and it was working.  Is the current dlib code not compiling with the latest opencv?

ttoeniges:
That is strange... It seems, that I messed up my installation with different openCV environments. A clean installation seems to solve the problem.
To sum up, you are right, no need for changes.

davisking:
Cool :)

PR #68, opened 2016-01-14
sys.stdout.encoding instead of latin-1 in setup.py

yukoba:
Please use sys.stdout.encoding instead of latin-1 in setup.py.
This is necessary for non English OS.


davisking:
Thanks for the PR :)

PR #69, opened 2016-01-14
Add pragma guards around deprecated auto_ptr to prevent GCC warnings

### LINKED ISSUE #67: auto_ptr is deprecated in >= C++11
`auto_ptr` are deprecated in recent versions of C++ (c++11 and above).

`dlib` uses them in four files:
```
dlib/smart_pointers/shared_ptr_thread_safe.h
dlib/smart_pointers/shared_ptr_thread_safe_abstract.h
dlib/smart_pointers/shared_ptr_abstract.h
dlib/smart_pointers/shared_ptr.h
```

While not critical, it causes recent compilers to emit deprecation warnings every time one includes dlib in a project, which is annoying.

This is a more serious issue for users who want to compile their code with `-Werror`.

Remediation options include:
- putting the offending code behind a guard (`#if __cplusplus <= 199711L`)
- removing the offending code
- removing completely dlib's custom smart pointers to use standard ones (`std::shared_ptr` and `std::unique_ptr`) -- this would raise the minimum compiler requirements to C++11.
- selectively ignore these warnings by surrounding offending code with:
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
//...
#pragma GCC diagnostic pop
```
(note that this is more like a workaround, and besides, a GCC-only one)

severin-lemaignan:
Fixes #67

### DIFF
--- a/dlib/smart_pointers/shared_ptr.h
+++ b/dlib/smart_pointers/shared_ptr.h
@@ -294,6 +294,8 @@ namespace dlib
             shared_node->ref_count += 1;
         }
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
         template<typename Y>
         explicit shared_ptr(
             std::auto_ptr<Y>& r
@@ -308,6 +310,7 @@ namespace dlib
             shared_node->del = new default_deleter;
             data = r.release();
         }
+#pragma GCC diagnostic pop
 
         shared_ptr& operator= (
             const shared_ptr& r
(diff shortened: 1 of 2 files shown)


davisking:
Cool.  You don't need to add them to the _abstract.h files since they are not processed by the compiler and are just documentation for the user (and they make it look cluttered).  But other than that the PR is ready :)

severin-lemaignan:
@davisking Fixed

davisking:
Super, thanks for the PR :)

PR #71, opened 2016-01-20
Decode message only if encoding is known

paulinus:
When python does not know the encoding of stdout, sys.stdout.encoding
is None.  Then calling decode(None) raises an exception.  We just
skip decoding when the encoding is unknown.


davisking:
Cool, thanks for the PR

Issue #65, opened 2016-01-07
Comparable/hashable (d)rectangles in Python


hbredin:
As of now, `dlib.drectangle` and `dlib.rectangle` instances are not comparable (nor hash-stable):

```python
>>> import dlib
>>> x1 = dlib.drectangle(0, 1, 2, 3)
>>> x2 = dlib.drectangle(0, 1, 2, 3)
>>> assert x1 == x2, 'Rectangles differ'
AssertionError: Rectangle differ
>>> assert hash(x1) == hash(x2), 'Hashes differ'
AssertionError: Hashes differ
```

Therefore, they cannot be used as dictionary keys for instance, nor in a set.
Solving this issue can be solved by implementing `__hash__` and `__eq__` methods in Python API.

I am willing to help but am not familiar with interfacing C and Python API so I don't really know where to start...

uetoyo:
Python port is made by Boost Python Library, so you should start here...
http://www.boost.org/doc/libs/1_60_0/libs/python/doc/html/index.html
see...
https://github.com/davisking/dlib/blob/master/tools/python/src/rectangles.cpp

patricksnape:
@hbredin #73 fixes your equality issue. However, `rectangle` and `drectangle` are not immutable and therefore they should not hash to the same value according to the Python hash semantics.

davisking:
Sweet, thanks.

hbredin:
Great. Thanks a lot.

PR #73, opened 2016-02-03
Drectangle equality

patricksnape:
Addresses #65. Simply added the `(self == self)` Macro syntax and add equality operators to `drectangle` in the main dlib source - copied from rectangle.

Also, sneak in two other changes:

  - Make `shape_predictor_training_options` serializable and printable.
  - Add `__version__` attribute to the dlib Python module to be compatible with PEP 396. Only issue is that the string is hard coded - maybe someone can think of a better way to read it from the `CMakelists.txt` at compile time (@davisking)

### DIFF
--- a/dlib/geometry/drectangle.h
+++ b/dlib/geometry/drectangle.h
@@ -195,6 +195,19 @@ namespace dlib
             return *this;
         }
 
+        bool operator== (
+            const drectangle& rect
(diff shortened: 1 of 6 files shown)


davisking:
Sweet, thanks another patch.  I also updated it a little so the version information comes from CMake.

PR #77, opened 2016-02-11
detect OS X, and query sysconfig for libdir

jimreesman:
I was having trouble getting `python setup.py install` to build a usable library on OS X. Loading the shared object would seg fault. I use a brewed Python, but cmake would default to using the system python lib under /usr/lib. Forcing a non-default -DPYTHON_LIBRARY is one solution that seems to work. There may be other, better ways to accomplish this but this change works for me.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -466,6 +466,19 @@ class build(_build):
         if sys.version_info >= (3, 0):
             cmake_extra_arch += ['-DPYTHON3=yes']
 
+        log.info("Detected platform: %s" % sys.platform)
+        if sys.platform == "darwin":
+            # build on OS X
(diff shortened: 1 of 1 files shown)


davisking:
I tried running this on a mac with python installed via brew and got this error:

    File "setup.py", line 479, in build_dlib

    py_lib = os.path.join(get_config_var('LIBDIR'), 'libpython'+py_ver+'.dylib')

    NameError: global name 'get_config_var' is not defined

Any ideas?

jimreesman:
yes. sorry bad diff missing the import. let me update it.

davisking:
Sweet, thanks for the patch :)

PR #83, opened 2016-02-26
Support use of dlib a cross-compiled ExternalProject

nocnokneo:
This pull request contains a couple changesets to get dlib to build:

1. as and External Project
2. with a cross-toolchain (I've tested with the linaro-multilib-2014.06-gcc4.9 ARM toolchain)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,2 @@
+cmake_minimum_required(VERSION 2.8.4)
+add_subdirectory(dlib)
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -132,6 +132,13 @@ if (NOT TARGET dlib)
          stack_trace.cpp
          )
 
+      set(dlib_needed_libraries)
+      if(UNIX)
+         set(CMAKE_THREAD_PREFER_PTHREAD ON)
+         find_package(Threads REQUIRED)
+         set(dlib_needed_libraries ${dlib_needed_libraries} ${CMAKE_THREAD_LIBS_INIT})
+      endif()
+
       # we want to link to the right stuff depending on our platform.  
       if (WIN32 AND NOT CYGWIN) ###############################################################################
          if (DLIB_NO_GUI_SUPPORT)
@@ -140,9 +147,6 @@ if (NOT TARGET dlib)
             set (dlib_needed_libraries ws2_32 winmm comctl32 gdi32 imm32)
          endif()
       elseif(APPLE) ############################################################################
-         find_library(pthreadlib pthread)
-         set (dlib_needed_libraries ${pthreadlib})
-
          if (NOT DLIB_NO_GUI_SUPPORT)
             find_package(X11 QUIET)
             if (X11_FOUND)
@@ -178,9 +182,6 @@ if (NOT TARGET dlib)
 
          mark_as_advanced(pthreadlib xlib xlib_path x11_path)
       else () ##################################################################################
-         find_library(pthreadlib pthread)
-         set (dlib_needed_libraries ${pthreadlib})
-
          # link to the nsl library if it exists.  this is something you need sometimes 
          find_library(nsllib nsl)
          if (nsllib)
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -432,7 +432,7 @@ if (NOT TARGET dlib)
                    ARCHIVE DESTINATION lib)
        endif()
 
-       install(DIRECTORY ${CMAKE_SOURCE_DIR}/ DESTINATION include/dlib
+       install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/ DESTINATION include/dlib
                FILES_MATCHING PATTERN "*.h"
                REGEX "${CMAKE_CURRENT_BINARY_DIR}" EXCLUDE)
 


davisking:
Cool, thanks for the PR.

But why the CMakeLists.txt file in the top level?  Is that needed for some use case?

nocnokneo:
The [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) module is designed to work best with projects that adhere to established conventions. One of these conventions (more like a de facto standard at this point) is that a project should be buildable from a top-level CMakeLists.txt file. Without this, the `CONFIGURE_COMMAND` to `ExternalProject_Add` must be manually specified to override the default and you lose the ability to use all the other `CMAKE_*` options for the configure step of `ExternalProject_Add`.

davisking:
Why can't you use the one in the dlib folder directly?

nocnokneo:
It's not that I _can't_ use the one in the dlib folder. It's just that it makes life more difficult (as explained above) for me and for anyone else trying to use dlib as an ExternalProject.

davisking:
I haven't used ExternalProject before.  My impression was that no one used it because the documentation is horrible.  :)

Can you post a simple project that uses dlib via ExternalProject?  I want to make sure it compiles it in a sensible way (e.g. makes proper use of dlib/config.h) and so forth before merging this PR.

nocnokneo:
Hahah. I know what you mean about the documentation, although I wouldn't go quite as far as calling it horrible. I've seen much worse :-)

Here a minimal standalone CMakeLists.txt:

```cmake
cmake_minimum_required(VERSION 2.8)
include(ExternalProject)
ExternalProject_Add(dlib
  GIT_REPOSITORY https://github.com/nocnokneo/dlib.git
  GIT_TAG support-cross-compiled-externalproject
  CMAKE_ARGS
    -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
  )
```

Assuming you save that to `/tmp/dlib-consumer/CMakeLists.txt` you can test with:

```bash
mkdir /tmp/dlib-consumer-build
cd /tmp/dlib-consumer-build
cmake /tmp/dlib-consumer
make
```

davisking:
That just compiles dlib and installs it to an odd sub folder.  What is a
real usecase for this?  I assume it's in some larger project that uses
dlib.  How does that work?

nocnokneo:
That's right, the real use case if for a larger project that uses dlib. Here's a minimal but complete example of such a project:

**CMakeLists.txt**

```cmake
cmake_minimum_required(VERSION 2.8)
include(ExternalProject)

set(DLIBCONSUMER_USE_SUPERBUILD ON CACHE BOOL
  "Download/Configure/Build all dependencies before building dlibconsumer")

if(DLIBCONSUMER_USE_SUPERBUILD)
  # Base directory for all external project downloads, source, binaries, etc
  set(EP_PREFIX "${CMAKE_BINARY_DIR}/ExternalProjects")
  set_property(DIRECTORY PROPERTY EP_PREFIX "${EP_PREFIX}")

  #
  # dlib
  #
  ExternalProject_Add(dlib
    GIT_REPOSITORY https://github.com/nocnokneo/dlib.git
    GIT_TAG support-cross-compiled-externalproject
    CMAKE_ARGS
      -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
    )
  ExternalProject_Get_Property(${proj} INSTALL_DIR)
  set(dlib_DIR "${INSTALL_DIR}/lib/cmake/dlib")

  ExternalProject_Get_Property(dlib binary_dir)
  install(SCRIPT ${binary_dir}/cmake_install.cmake)

  #
  # Recursive call to our self
  #
  ExternalProject_Add(dlibconsumer
    SOURCE_DIR ${CMAKE_SOURCE_DIR}
    DEPENDS dlib
    CMAKE_ARGS
      -DDLIBCONSUMER_USE_SUPERBUILD:BOOL=OFF
      -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
      -Ddlib_DIR:PATH=${dlib_DIR}
    )

  ExternalProject_Get_Property(dlibconsumer binary_dir)
  install(SCRIPT ${binary_dir}/cmake_install.cmake)

  #
  # End of superbuild
  #
  return()
endif()

project(dlibconsumer)
find_package(dlib 18.18.99 REQUIRED)
add_executable(dlibconsumer main.cpp)
target_link_libraries(dlibconsumer dlib::dlib)

# Workaround for incomplete dlib::dlib imported target:
target_include_directories(dlibconsumer SYSTEM PUBLIC ${dlib_INCLUDE_DIRS})

install(TARGETS dlibconsumer DESTINATION bin)
```

**main.cpp**

```c++
#include <iostream>
#include <dlib/matrix.h>

using namespace dlib;
using namespace std;

int main()
{
    matrix<double,3,1> y;
    matrix<double> M(3,3);
    M = 54.2,  7.4,  12.1,
        1,     2,    3,
        5.9,   0.05, 1;
    y = 3.5,  
        1.2,    
        7.8;
    matrix<double> x = inv(M) * y;
    cout << "x:\n" << x << endl;
}
```

Of course a real world superbuild becomes a little more complex than this, but this shows the idea.

davisking:
Out of curiosity, is that the simplest way to use external project?  I ask because that is a pretty complex set of steps to do what is the moral equivalent to just caling find_package().

Also, I'm dithering on adding a CMakeLists.txt to the root folder because I don't want dlib users to be confused by it.  I get daily questions from students and others who, never having programmed before, make very basic mistakes.  This is even though both the README and "how to compile" page shows you exactly what to type to build the example programs and even though the example [CMakeLists.txt](https://github.com/davisking/dlib/blob/master/examples/CMakeLists.txt) file is very simple.  

I can pretty much guarantee an increase in confused questions by putting a CMakeLists.txt file in the root. :(

What's the difference between the example CMakeLists.txt you posted and what is needed without CMakeLists.txt in the dlib root?

nocnokneo:
> Out of curiosity, is that the simplest way to use external project? I ask because that is a pretty complex set of steps to do what is the moral equivalent to just caling find_package().

I agree that there are some improvements that CMake could do to make this simpler. but this is not at all the equivalent of just calling `find_package`. In fact, this example still uses `find_package` to do what that command is made for.

The last 9 lines are what a simple project without superbuild support would need. dlib would need to be find-able in one of the standard `find_package` search locations. And this approach is still possible (with `-DDLIBCONSUMER_USE_SUPERBUILD:OFF`) and useful for certain situations (e.g. for Linux distro package maintainers)

The rest of the lines are to support the Superbuild (the idea of downloading, extracting, patching, configuring, and building any number of external dependencies which may even have inter-dependencies on each other). The `install(SCRIPT...)` lines are only needed if you want to have an install target at the *superbuild level* that installs dlibconsumer *AND* all it's dependencies (useful for creating binary release packages that come with batteries included). The `EP_PREFIX` lines were just to address your comment about the "odd sub folder" organization; they are not required.

> Also, I'm dithering on adding a CMakeLists.txt to the root folder because I don't want dlib users to be confused by it. I get daily questions from students and others who, never having programmed before, make very basic mistakes. This is even though both the README and "how to compile" page shows you exactly what to type to build the example programs and even though the example CMakeLists.txt file is very simple.
>
> I can pretty much guarantee an increase in confused questions by putting a CMakeLists.txt file in the root. :(

I have to say that even as an experienced CMake user new to dlib, it took me a few minutes to figure out the unconventional organization (two top-level CMake projects in one repository and neither one in the repository root). I actually think that a better and simpler fix than 378431e would be to just have one top-level CMakeLists.txt that builds everything by default (dlib and examples) but has an option to disable examples (e.g. with `-Ddlib_BUILD_EXAMPLES=OFF`). If that's something your open to discussing, I can put together a separate pull request that shows what I'm talking about. I can also include some fixes to properly export the `dlib::dlib` target.

> What's the difference between the example CMakeLists.txt you posted and what is needed without CMakeLists.txt in the dlib root?

In this very simple example, it would look something like:

```cmake
ExternalProject_Add(dlib
  GIT_REPOSITORY https://github.com/nocnokneo/dlib.git
  GIT_TAG support-cross-compiled-externalproject
  CONFIGURE_COMMAND
    ${CMAKE_COMMAND} <SOURCE_DIR>/dlib
    -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
  )
```

Which doesn't look all that different, but it's much less complete and robust. Taking a snippet from ExternalProject.cmake, you'd loose all the logic from the else block below:

```
  get_property(cmd_set TARGET ${name} PROPERTY _EP_CONFIGURE_COMMAND SET)
  if(cmd_set)
    get_property(cmd TARGET ${name} PROPERTY _EP_CONFIGURE_COMMAND)
  else()
    get_target_property(cmake_command ${name} _EP_CMAKE_COMMAND)
    if(cmake_command)
      set(cmd "${cmake_command}")
    else()
      set(cmd "${CMAKE_COMMAND}")
    endif()

    get_property(cmake_args TARGET ${name} PROPERTY _EP_CMAKE_ARGS)
    list(APPEND cmd ${cmake_args})

    # If there are any CMAKE_CACHE_ARGS or CMAKE_CACHE_DEFAULT_ARGS,
    # write an initial cache and use it
    get_property(cmake_cache_args TARGET ${name} PROPERTY _EP_CMAKE_CACHE_ARGS)
    get_property(cmake_cache_default_args TARGET ${name} PROPERTY _EP_CMAKE_CACHE_DEFAULT_ARGS)

    if(cmake_cache_args OR cmake_cache_default_args)
      set(_ep_cache_args_script "${tmp_dir}/${name}-cache-$<CONFIG>.cmake")
      if(cmake_cache_args)
        _ep_command_line_to_initial_cache(script_initial_cache_force "${cmake_cache_args}" 1)
      endif()
      if(cmake_cache_default_args)
        _ep_command_line_to_initial_cache(script_initial_cache_default "${cmake_cache_default_args}" 0)
      endif()
      _ep_write_initial_cache(${name} "${_ep_cache_args_script}" "${script_initial_cache_force}${script_initial_cache_default}")
      list(APPEND cmd "-C${_ep_cache_args_script}")
    endif()

    get_target_property(cmake_generator ${name} _EP_CMAKE_GENERATOR)
    get_target_property(cmake_generator_platform ${name} _EP_CMAKE_GENERATOR_PLATFORM)
    get_target_property(cmake_generator_toolset ${name} _EP_CMAKE_GENERATOR_TOOLSET)
    if(cmake_generator)
      list(APPEND cmd "-G${cmake_generator}")
      if(cmake_generator_platform)
        list(APPEND cmd "-A${cmake_generator_platform}")
      endif()
      if(cmake_generator_toolset)
        list(APPEND cmd "-T${cmake_generator_toolset}")
      endif()
    else()
      if(CMAKE_EXTRA_GENERATOR)
        list(APPEND cmd "-G${CMAKE_EXTRA_GENERATOR} - ${CMAKE_GENERATOR}")
      else()
        list(APPEND cmd "-G${CMAKE_GENERATOR}")
      endif()
      if(cmake_generator_platform)
        message(FATAL_ERROR "Option CMAKE_GENERATOR_PLATFORM not allowed without CMAKE_GENERATOR.")
      endif()
      if(CMAKE_GENERATOR_PLATFORM)
        list(APPEND cmd "-A${CMAKE_GENERATOR_PLATFORM}")
      endif()
      if(cmake_generator_toolset)
        message(FATAL_ERROR "Option CMAKE_GENERATOR_TOOLSET not allowed without CMAKE_GENERATOR.")
      endif()
      if(CMAKE_GENERATOR_TOOLSET)
        list(APPEND cmd "-T${CMAKE_GENERATOR_TOOLSET}")
      endif()
    endif()

    list(APPEND cmd "${source_dir}")
  endif()
```

davisking:
> 
I have to say that even as an experienced CMake user new to dlib, it took me a few minutes to figure out the unconventional organization (two top-level CMake projects in one repository and neither one in the repository root). I actually think that a better and simpler fix than 378431e would be to just have one top-level CMakeLists.txt that builds everything by default (dlib and examples) but has an option to disable examples (e.g. with -Ddlib_BUILD_EXAMPLES=OFF). If that's something your open to discussing, I can put together a separate pull request that shows what I'm talking about. I can also include some fixes to properly export the dlib::dlib target.

Did you not see the build instructions in the [README](https://github.com/davisking/dlib/blob/master/README.txt) or the [web page](http://dlib.net/compile.html)?  I try to make it very easy.  For example, I want the dlib example [CMakeList.txt](http://dlib.net/examples/CMakeLists.txt.html) to be so simple even someone new to CMake will understand it immediately.

I'm certainly open to improving the build scripts, and adding things needed to support more CMake 3.0 features is cool (provided everything still works in 2.8.4).  However, I'm not sure I understand why you would want the top level CMake file to build the examples.  Wouldn't that make your external build scripts build them by default? I doubt people who are using ExternalProject would want that behavior.

Also, consider that I get questions [like this one](http://stackoverflow.com/questions/35574967/how-to-use-dlib-in-projects) pretty much every day.  So while it's definitely important to make dlib convenient for expert users, I want to make sure it stays accessible to beginners (to the extent that that is possible).  I also want to encourage users to statically compile dlib into their applications.  So that should be the default recommendation in the documentation.  I say this because, in my experience, very few developers, even professionals, understand issues like C++'s [one definition rule](https://en.wikipedia.org/wiki/One_Definition_Rule) or that [mixing runtimes on windows](http://siomsystems.com/mixing-visual-studio-versions/) is bad.  Currently, the setup in the dlib example CMakeList.txt file avoids any possibility of these issues arising.  Even despite this, these kinds of user errors are so frequent (but only from non-cmake users) I've even put stuff inside the code to try and detect ODR violations or people doing bad things like writing `#include <dlib/all/source.cpp>` in their files (their compiler will drop them here https://github.com/davisking/dlib/blob/master/dlib/dlib_basic_cpp_build_tutorial.txt). 

But all that said, yes, please do submit PRs that improve the user experience for experts as well and thanks for this PR :)

nocnokneo:
Thanks for merging. When I get a chance I'll put together a pull request for you to take a look at.

I did see the README with three sections:

> COMPILING DLIB C++ EXAMPLE PROGRAMS
> COMPILING DLIB Python API
> RUNNING THE UNIT TEST SUITE

But none of these was what I was looking for. I just wanted "COMPILING DLIB". If everything is built under one top-level CMakeLists.txt, the build instructions become simpler since the procedure is always the same. The user just has to change options via cmake-gui/ccmake. For example, `DLIB_BUILD_EXAMPLES` (which should be defined in the CMakeLists.txt with a helpful doc string). 

Same for building shared vs static libs. It's fine to default to build static libraries by default, but the build should use the standard CMake variable for choosing how the library gets built: `BUILD_SHARED_LIBS`.  Defining variables in your CMakeLists.txt like `option(BUILD_SHARED_LIBS "Build dlib as a shared library" OFF)` accomplishes two things:

- Sets the default value you'd like to use for this CMake built-in variable
- Makes the variable show up in ccmake/cmake-gui

davisking:
Cool.

Well, still, consider all dlib users.  The question users ask is "how can I write a program that uses dlib?", not "how can I compile dlib?".  The answer to that must be as obvious as possible.

PR #90, opened 2016-03-30
Python/Simple object detector trainer. Added upsample_limit option

e-fominov:
When fine-tuning object detector, I need to run training may times. With some combination of options, trainer starts upsampling images 1-2 times that makes trainer to crash due to available RAM limitations.
Original trainer had hard-coded upsample limit value == 2. 
I changed this to options, with default value == 2

With this feature, we can control RAM usage limiting upsample function


davisking:
Cool, thanks for the PR :)

PR #91, opened 2016-04-01
MSVC x64 SSE2 automatical support

e-fominov:
Some processors (Atoms/Pentiums) does not support AVX but support SSE2
When I compile face detector with MSVC, x64 - compiler gives me SSE2 support
To enable it in Dlib, i need to put "DLIB_HAVE_SSE2" into preprocessor definitions, otherwise face detector will be 2-3x slow
When building samples with CMAKE, I have this definition by default, but I can forget about it in real project, and I didnt find dlib's documentation about enabling only SSE2 without AVX

With this PR I am enabling SSE2 support for all x64 builds in MSVC, even when AVX is not enabled

### DIFF
--- a/dlib/simd/simd_check.h
+++ b/dlib/simd/simd_check.h
@@ -22,7 +22,7 @@
                 #define DLIB_HAVE_AVX
             #endif
         #endif
-        #if defined(_M_IX86_FP) && _M_IX86_FP >= 2 && !defined(DLIB_HAVE_SSE2)
+    #if (defined( _M_X64) || defined(_M_IX86_FP) && _M_IX86_FP >= 2) && !defined(DLIB_HAVE_SSE2)
             #define DLIB_HAVE_SSE2
         #endif
     #else
--- a/dlib/simd/simd_check.h
+++ b/dlib/simd/simd_check.h
@@ -22,7 +22,7 @@
                 #define DLIB_HAVE_AVX
             #endif
         #endif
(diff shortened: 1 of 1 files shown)


davisking:
Normally the `/arch:SSE2` switch needs to be given to visual studio to use SSE2 instructions.  When it is those preprocessor statements check the `_M_IX86_FP` macro which is set by `/arch:SSE2` and will automatically turn on SSE2 use in dlib.  

Is this behavior different on the version of visual studio you are using?  What version are you trying this on.  Or maybe `_M_IX86_FP` is never set for 64bit builds in visual studio.  Visual studio has never been very consistent in their setting of these kinds of macros so I wouldn't be surprised.

e-fominov:
I am using Visual Studio 2013.
/arch:SSE2 and /arch:SSE works only in x86 builds

for x64 builds, only /arch:AVX is available (MSVC 2015 adds /arch:AVX2)
if I try to compile /arch:SSE2 with x64 build, I am getting this error:
*cl : Command line warning D9002 : ignoring unknown option '/arch:SSE2'*

_M_IX86_FP is undefined in x64 builds

here is information about _M_IX86_FP for x86: https://msdn.microsoft.com/en-us/library/7t5yh4fd.aspx 
and here is x64: https://msdn.microsoft.com/en-us/library/jj620901.aspx (no word about _M_IX86_FP)

i have tested face detector - it starts working 2x faster if I #define DLIB_HAVE_SSE2 with no changing any compiler command line options for x64 build

davisking:
Sweet, sounds good then.  Thanks for the PR :)

PR #92, opened 2016-04-04
Face detector sample time measure

e-fominov:
I have some ideas on object detector performance optimization. 
To see any difference we need to measure the performance of object detector
I modified this sample with displaying face detector processing time

I think it will be usefull for everybody. Especially for those who tells that object detection is not fast enough (Debug build, forgot AVX, etc..)

### DIFF
--- a/examples/face_detection_ex.cpp
+++ b/examples/face_detection_ex.cpp
@@ -40,6 +40,7 @@
 #include <dlib/image_processing/frontal_face_detector.h>
 #include <dlib/gui_widgets.h>
 #include <dlib/image_io.h>
+#include <dlib/timing.h>
 #include <iostream>
 
 using namespace dlib;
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  However, I would rather not add this here as the purpose of the examples is to be concise educational documents rather than useful programs in their own right.  Since the timing statements don't help the reader understand the face detector they shouldn't be added to this particular example.

e-fominov:
Sure, thats why i made it as separate PR. May be I should create completely new sample for performance measurement? 
I want to try using simd8f for FHOG (now it's simd4f-based), and I need some good tool to compare performance

davisking:
To measure the performance it would be better to call the fhog feature
extraction code directly, as shown in the fhog_ex.cpp example.  But in any
case, useful tool programs don't belong in the examples folder.  The
examples are supposed to be more like essays on how to use dlib.

You certainly could create an example program that shows how to use the
timing statements in dlib and use fhog feature extraction as the example
thing being timed. However, it's got to be an actual example and therefore
should read like a short essay, have a single topic, be short and simple
enough to be easily comprehended, have an introduction, not meander into
topics not relevant to timing, etc.  Programs like that tend not to be
useful tools so I would recommend not putting the code you write for
testing and debugging into something for the examples folder.

PR #93, opened 2016-04-04
Using float instead of double for FHOG computations

e-fominov:
I found that FHOG module mixes doubles and floats in many places. This makes compiler harder to optimize the code. And we do not use more than 4-5 significant digits in float values
I think, we can use float instead of double here. All result HOG features were already float

My measurements (based on my previous PR measurement method):
MSVC 2013, Windows 7, 64bit, i7/2.5Ghz:
1) Release, x64 /O2 /arch:AVX : **1482** msec - this is master branch, if added AVX
2) Release, x64 /O2 : **1624** msec - master branch as-is
3) Release, x64 /O2 (without my recent PR#91 and without /DDLIB_HAVE_SSE2) : **6782** msec (much slower than 2)
4) Release, x64 /O2 (with PR#91 and without /DDLIB_HAVE_SSE2) : **1624** msec (same as 2)
5) Release, x64 /O2 (with PR#91 fhog-float branch): **1517** msec (6.6% faster than 2)
6) Release, x64 /O2 /arch:AVX  (fhog-float branch) : **1256** msec  (**15% faster** than 1)

GCC 5.2.1 (Ubuntu), x64 (same PC as Windows)

7) Release, x64, -O3 -mavx : **1227** msec
8) Release, x64, -O3 -mavx (fhog-float branch) : **1136** msec (8.8% faster than 7)

I tried other samples - they still compile and run well on Windows and Ubuntu

### DIFF
--- a/dlib/image_transforms/fhog.h
+++ b/dlib/image_transforms/fhog.h
@@ -709,7 +709,7 @@ namespace dlib
                 for (; x < visible_nc; x++) 
                 {
                     matrix<float, 2, 1> grad;
-					float v;
+                    float v;
                     get_gradient(y,x,img,grad,v);
 
                     // snap to one of 18 orientations
--- a/dlib/image_transforms/fhog.h
+++ b/dlib/image_transforms/fhog.h
@@ -21,16 +21,16 @@ namespace dlib
 
     namespace impl_fhog
     {
-        template <typename image_type>
+        template <typename image_type, typename T>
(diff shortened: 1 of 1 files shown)


davisking:
Cool.  Did you run the unit tests for the fhog feature extraction?  It's important that the output fhog features not change by any significant amount since there is a lot of code out there using trained models based on the existing dlib code.  So we need to make sure the outputs are still the same.

davisking:
I just ran all the tests and it looks like it's all good.  Thanks for the PR :)

PR #94, opened 2016-04-05
Added simd8 usage on FHOG features computation

e-fominov:
FHOG module is based on simd4 and works well. I am adding simd8 usage if DLIB_HAVE_AVX is defined

MSVC 2013, Windows 7, x64 - performance is about **3-4%** better (not much, but little better)
GCC 5.2, Ubuntu x64 - performance is about **12 %** better (i think, it is much better)

So, with my previous PR's, the overall performance is expected to be 20% better on AVX processors - as on Linux, as on Windows

BTW, GCC/Ubuntu build works about 20% faster than MSVC/Windows

For performance testing i am running 100 iterations of detector with 2007_007763.jpg from examples/faces folder. MSVC time is about 12 seconds, GCC - 9.7 seconds on my laptop

Regression test:
Release\dtest.exe --test_fhog

Testing Finished
Total number of individual testing statements executed: 175037700
All tests completed successfully

Is there a better way to organize this code? May be split into separate generic functions? I dont like how #ifdef is used

### DIFF
--- a/dlib/image_transforms/fhog.h
+++ b/dlib/image_transforms/fhog.h
@@ -142,7 +142,140 @@ namespace dlib
             len = select(cmp,tlen,blen);
         }
 
-    // ------------------------------------------------------------------------------------
+        // ------------------------------------------------------------------------------------
+
+        template <typename image_type>
+        inline typename dlib::enable_if_c<pixel_traits<typename image_type::pixel_type>::rgb>::type get_gradient(
+            const int r,
+            const int c,
+            const image_type& img,
+            simd8f& grad_x,
+            simd8f& grad_y,
+            simd8f& len
+            )
+        {
+            simd8i rleft((int)img[r][c - 1].red,
+                (int)img[r][c].red,
+                (int)img[r][c + 1].red,
+                (int)img[r][c + 2].red,
+                (int)img[r][c + 3].red,
+                (int)img[r][c + 4].red,
+                (int)img[r][c + 5].red,
+                (int)img[r][c + 6].red);
+            simd8i rright((int)img[r][c + 1].red,
+                (int)img[r][c + 2].red,
+                (int)img[r][c + 3].red,
+                (int)img[r][c + 4].red,
+                (int)img[r][c + 5].red,
+                (int)img[r][c + 6].red,
+                (int)img[r][c + 7].red,
+                (int)img[r][c + 8].red);
+            simd8i rtop((int)img[r - 1][c].red,
+                (int)img[r - 1][c + 1].red,
+                (int)img[r - 1][c + 2].red,
+                (int)img[r - 1][c + 3].red,
+                (int)img[r - 1][c + 4].red,
+                (int)img[r - 1][c + 5].red,
+                (int)img[r - 1][c + 6].red,
+                (int)img[r - 1][c + 7].red);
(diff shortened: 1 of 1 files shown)


davisking:
Cool.  Did you try always running it with the simd8 code?  Is it slower when AVX isn't enabled?

e-fominov:
I made some more testing. Added MSVC2015. All testing are now done on fhog_ex sample with one 2007_007763.jpg image, upsampled several times (lower time is better)

OS|Compiler|Compiler SIMD mode|Dlib simd class used|Time RGB24, s|Time GRAY8, s
----|-----------|---------------|--------------------------|----------------|----------------
Windows7 x64|MSVC2015 x64|AVX|simd8 with __mm* functions|11.8|8.0
Windows7 x64|MSVC2015 x64|AVX|simd4 with __mm* functions|12.6|10.0
Windows7 x64|MSVC2015 x64|AVX|simd8 with 2xsimd4|13.8|11.4
Windows7 x64|MSVC2015 x64|SSE2|simd4 with __mm* functions|19.3|14.4
Windows7 x64|MSVC2015 x64|SSE2|simd8 with 2xsimd4|18.4|13.4
Windows7 x64|MSVC2013 x64|AVX|simd8 with __mm* functions|13.4|9.4
Windows7 x64|MSVC2013 x64|AVX|simd4 with __mm* functions|13.6|10.5
Windows7 x64|MSVC2013 x64|AVX|simd8 with 2xsimd4|15.5|12.8
Windows7 x64|MSVC2013 x64|SSE2|simd4 with __mm* functions|21.3|15.2
Windows7 x64|MSVC2013 x64|SSE2|simd8 with 2xsimd4|20.7|15.7
Ubuntu 15 x64|GCC 5.2 x64|AVX|simd8 with __mm* functions|9.8|**6.2**
Ubuntu 15 x64|GCC 5.2 x64|AVX|simd4 with __mm* functions|9.8|8.3
Ubuntu 15 x64|GCC 5.2 x64|AVX|simd8 with 2xsimd4|10.0|7.2
Ubuntu 15 x64|GCC 5.2 x64|SSE2|simd4 with __mm* functions|19.6|15.1
Ubuntu 15 x64|GCC 5.2 x64|SSE2|simd8 with 2xsimd4|19.0|13.4

As I said, I upsampled image several times - to make near 8000x5000 size and runned 10 iterations. On small images, < 2 MP there will be no difference for simd4/simd8 - this is important for fhog_ex sample where we are building HOG features only for one pyramid level
With face detection we are building image pyramid and processing one image many times, so we will get profit from simd8 instructions
Most interesting results are on GRAY8 images (array2d<unsigned char>) - benefit from simd8 on GCC is very high
Using simd8 instead of simd4 everywhere in fhog module to make it work as 2xsimd4 is an open question. I think results will vary much on different CPU/Compiler/OS

davisking:
Nice analysis.  So it looks like it's generally a win to use simd8 all the time rather than switching based on AVX support or not.  I ran some tests on my computers and simd8 was also generally a little bit faster regardless of AVX. Since using simd8 all the time avoids needing to use #ifdef in the code and is faster can you modify this PR to use simd8 all the time and not use any #ifdef?

e-fominov:
Done. Now only simd8 is used for fhog

davisking:
Sweet, thanks for another nice patch :)

PR #96, opened 2016-04-07
Fix msvc2015 build dtest

### LINKED ISSUE #95: Error building DTEST on MSVC 2015
compiling by *cmake .. -G"Visual Studio 14 2015"*:
**dlib\dlib\test\multithreaded_object.cpp(186): error C2694: `anonymous namespace'::test4_c2::~test4_c2(void) throw(std::exception)
'override': overriding virtual function has less restrictive exception specification than base class virtual member function 'base'**

https://msdn.microsoft.com/en-us/library/e4x95xf8.aspx

MSVC 2015 fails to build a test (MSVC2013 works good)

failed destructor code:

```c++
~test4_c2() throw (std::exception)
{
            DLIB_TEST(number_of_threads_registered() == 2);
            DLIB_TEST(number_of_threads_alive() == 2);
            DLIB_TEST_MSG(is_running() == false,"is_running(): " << is_running());
            stop();
            DLIB_TEST(number_of_threads_registered() == 2);
            DLIB_TEST(is_running() == false);
            wait();
            DLIB_TEST(number_of_threads_registered() == 2);
            DLIB_TEST(number_of_threads_alive() == 0);
            DLIB_TEST(is_running() == false);
        }
```

base class destructor code
```c++
    multithreaded_object::
    ~multithreaded_object (
    )
    {
        try
        {
            DLIB_ASSERT(number_of_threads_alive() == 0,
                   "\tmultithreaded_object::~multithreaded_object()"
                   << "\n\tYou have let a multithreaded object destruct itself before terminating its threads"
                   << "\n\tthis: " << this
            );
        }
        catch (std::exception& e)
        {
            std::cerr << e.what() << std::endl;
            assert(false);
            abort();
        }
    }
```
do we really need *throw (std::exception)* ? if no - I can make PR

e-fominov:
Fixes #95 
Tested on GCC 5.2, MSVC2013, MSVC2015

### DIFF
--- a/dlib/test/multithreaded_object.cpp
+++ b/dlib/test/multithreaded_object.cpp
@@ -183,7 +183,7 @@ namespace
             DLIB_TEST(is_running() == false);
         }
 
-        ~test4_c2() throw (std::exception)
+        ~test4_c2()
         {
             DLIB_TEST(number_of_threads_registered() == 2);
             DLIB_TEST(number_of_threads_alive() == 2);
--- a/dlib/test/multithreaded_object.cpp
+++ b/dlib/test/multithreaded_object.cpp
@@ -187,7 +187,7 @@ namespace
         {
             DLIB_TEST(number_of_threads_registered() == 2);
             DLIB_TEST(number_of_threads_alive() == 2);
-            DLIB_TEST_MSG(is_running() == false,"is_running(): " << is_running());
+            DLIB_TEST(is_running() == false);
             stop();
             DLIB_TEST(number_of_threads_registered() == 2);
             DLIB_TEST(is_running() == false);


davisking:
We need the throw (std::exception) because destructors are noexcept by default in C++11 (see http://en.cppreference.com/w/cpp/language/noexcept_spec). Without it, if any of those tests fail in test4_c2 the program will be abruptly terminated.

So the right thing to do is to disable this visual studio warning in just this test file with a #pragma.

e-fominov:
I have tried to find #pragma for changing this in MSVC2015, but cant find it. This is not a warning, but error that breaks compilation process

I have read a link from C++11 standard you provided and there is one interesting paragraph:
>If any declaration of a function has a noexcept specification that isn't noexcept(false), other declarations of the same function have to have the same noexcept-specification. Similar restrictions apply to overriders of a virtual function and to initialization and assignment of pointers to functions (target entity has to have identical or more permissive exception specification).

This means, if the destructor of base class is *noexcept*, derived classes cant change exception specification and we cant use *throw (std::exception)* in derived classes destructors. I just cant understang why GCC 5.2 successfully compiles this code. May be this will change in a future

Here is one more link where same situation is discussed http://www.cplusplus.com/forum/beginner/172392/#msg857541
 
One more thing that I see - the only reason to use throw (std::exception) is:
```c++
            DLIB_TEST_MSG(is_running() == false,"is_running(): " << is_running());
```
And really we do not need using this type of marco, we will not get any additional information from its message. There are a lot of similar checks for is_running with simple (no msg) DLIB_TEST macro - even in the same destructor

davisking:
Oh right, sorry, I answered too quickly assuming it was a warning.  

Anyway, both DLIB_TEST and DLIB_TEST_MSG throw exceptions when they fail.  That's how they communicate the failure to the testing framework and also avoid further execution of the code they are in which might otherwise crash.

I also fixed this months ago but accidentally made the changes in a private dlib branch (I'm adding deep learning tooling to dlib).  I'm going push the updated branch in a few days which will fix this so don't worry about it :)

PR #97, opened 2016-04-07
Regression test for corellation tracker

e-fominov:
I am interested in using corellation_tracker. But it is extremely slow with my data
I want to improve its performance, but before doing it we need to create regression tests

This test simply uses 4 frames from examples/video_frames folder and checks if tracking is the same as recorded one. I dont want to generate images, I think that using real data will make better results

Do I need to add more frames? This test works now < 1 sec, but cpp file is 81kb, each frame is about 17 kb

Tested with GCC 5.2, MSVC2013x64, MSVC 2015x64 (with #96 ) and MSVC2015 x86 (with #96)

### DIFF
--- a/dlib/test/CMakeLists.txt
+++ b/dlib/test/CMakeLists.txt
@@ -33,6 +33,7 @@ set (tests
    conditioning_class_c.cpp
    conditioning_class.cpp
    config_reader.cpp
+   corellation_tracker.cpp
    crc32.cpp
    create_iris_datafile.cpp
    data_io.cpp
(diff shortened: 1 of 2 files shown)


davisking:
Cool, it's good to have this regression test to make sure the tracker isn't catastrophically broken by any code changes.  

However, if you want to modify the tracker then you need to test the new code on a large benchmarking dataset to see how it performs.  The most popular benchmark is this one: http://www.votchallenge.net/vot2015/dataset.html.  It's a bit of work but you will never really know if it's working better or worse without this kind of testing.

PR #98, opened 2016-04-07
Corellation tracker options

e-fominov:
Default behaviour of tracker didnt changed
I added some options to corellation_tracker constructor to control the size of filters and other params
And I made new update_noscale function that will track only object position without tracking scale

This small changes makes tracker work 20x faster on my data, where scale didnt change at all and filter size should be small
I am not creating separate struct for options, but it can be done if needed
I need options to be parameters to change them at runtime with contructing new tracker object. This will make the final user able to change them

### DIFF
--- a/dlib/image_processing/correlation_tracker.h
+++ b/dlib/image_processing/correlation_tracker.h
@@ -123,12 +123,12 @@ namespace dlib
         double update_noscale(
             const image_type& img,
             const drectangle& guess
-            )
+        )
         {
             DLIB_CASSERT(get_position().is_empty() == false,
                 "\t double correlation_tracker::update()"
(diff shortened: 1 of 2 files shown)


davisking:
Cool, thanks for another PR.  Make the constructor explicit so there isn't a default conversion from unsigned long to to correlation_tracker and it's good to go :)

e-fominov:
Now constructor is explicis. I also made some changes:
1. filter_size and num_scale_levels params are now automatically calculated to be a power of 2
2. added scale_pyramid_alpha param to have 100% control on all variables

Now tracker is ready for fine-tuning and experiments

davisking:
Cool, thanks for the PR :)

PR #108, opened 2016-05-08
Augment build-system, allow for multilib installation, support pkg-config discovery

SoapGentoo:
* Add **pkg-config** file, allowing for build-system agnostic dependencies
  This is important, as most build systems are not CMake and therefore asking users to employ CMake merely for one dependency is not optimal. In addition, I have versioned the .pc file, such that the library may be installed in parallel, should the API change substantially. This is considered good practice and done by all major libraries (GTK+ for instance). See also https://tecnocode.co.uk/2014/12/09/a-checklist-for-writing-pkg-config-files/

* Define `LIB_INSTALL_DIR` cache variable, allowing for multilib installations
  This is important for distributions like Gentoo, where we might install 32-bit and 64-bit libraries simultaneously. Nevertheless, the default remains `lib/`, so unless the user specifies `-DLIB_INSTALL_DIR`, nothing changes.

* Discover BLAS and LAPACK via pkg-config if possible
  This is more modern and more robust, but just the first attempt at detecting BLAS+LAPACK.

* Fix incorrect toolchain variables in "dlib/test/makefile"
  This fix has been carried for a while in the Gentoo tree, and we'd like to upstream it.

Thank you for a great library, and I hope you'll accept my changes, opening dlib for a much larger audience with the help of pkg-config.

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -79,6 +79,7 @@ if (NOT TARGET dlib)
    set (DLIB_LINK_WITH_SQLITE3_STR
    "Disable this if you don't want to link against sqlite3" )
    #set (DLIB_USE_FFTW_STR "Disable this if you don't want to link against fftw" )
+   set (LIB_INSTALL_DIR lib CACHE STRING "Install location of libraries (e.g. lib32 or lib64 for multilib installations)")
 
    option(DLIB_ISO_CPP_ONLY ${DLIB_ISO_CPP_ONLY_STR} OFF)
    toggle_preprocessor_switch(DLIB_ISO_CPP_ONLY)
@@ -308,6 +309,7 @@ if (NOT TARGET dlib)
          if (PNG_FOUND AND LIBPNG_IS_GOOD)
             include_directories(${PNG_INCLUDE_DIR})
             set (dlib_needed_libraries ${dlib_needed_libraries} ${PNG_LIBRARY})
+            set(REQUIRES_LIBS " libpng")
          else()
             # If we can't find libpng then statically compile it in.
             include_directories(external/libpng external/zlib)
@@ -343,6 +345,7 @@ if (NOT TARGET dlib)
                external/zlib/uncompr.c
                external/zlib/zutil.c
                )
+            set(REQUIRES_LIBS "")
          endif()
          set(source_files ${source_files}
             image_loader/png_loader.cpp
@@ -575,8 +578,8 @@ if (NOT TARGET dlib)
            install(TARGETS dlib dlib_shared
                    EXPORT dlib 
                    RUNTIME DESTINATION bin # Windows (including cygwin) considers .dll to be runtime artifacts
-                   LIBRARY DESTINATION lib
-                   ARCHIVE DESTINATION lib)
+                   LIBRARY DESTINATION "${LIB_INSTALL_DIR}"
+                   ARCHIVE DESTINATION "${LIB_INSTALL_DIR}")
        else()
            install(TARGETS dlib
                    EXPORT dlib 
@@ -601,7 +604,7 @@ if (NOT TARGET dlib)
 
        ## Config.cmake generation and installation
 
-       set(ConfigPackageLocation lib/cmake/dlib)
+       set(ConfigPackageLocation "${LIB_INSTALL_DIR}/cmake/dlib")
        install(EXPORT dlib
             NAMESPACE dlib::
             DESTINATION ${ConfigPackageLocation})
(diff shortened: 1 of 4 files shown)


davisking:
Cool.  Shouldn't it also add things like libjpeg, openblas, etc. to the REQUIRES_LIBS variable?  When I try it it only adds libpng.  Also, the file is named dlib-1.pc.  Why not dlib.pc?

SoapGentoo:
The main problem is that pkg-config support (as is the case with dlib right now) is not completely ubiquitous in the ecosystem. libjpeg, openblas etc. do not have pkg-config support. I'm currently on an updating spree, where I push pkg-config into as main libraries as possible, such that everyone can simplify their build systems. Once these libraries have full pkg-config support, yes, they then should be added to the `REQUIRES_LIBS` variable as you suggested. A stopgap solution is to add the linker arguments for these libraries to the `Libs:` flag, but this is more brittle, as it makes dlib responsible for the linker flags of other packages.

Concerning the `dlib-1.pc`: this is good practice and gives a rough versioning of the API. Imagine you update dlib in an incompatible way. What are users supposed to do that have developed to the older API? A lot of programs won't switch over night, and by having the new API use dlib-2.pc, the old and new library can be installed in parallel. This is more flexible and forward looking. All major libraries do it (GTK+, Pango, etc...) and it has 0 cost. See for instance:
https://developer.gnome.org/programming-guidelines/stable/parallel-installability.html.en

davisking:
Sounds good.  As for the naming, why not use the dlib version number then?

SoapGentoo:
Mainly because that would make consuming the library too messy, as it would require updating downstream build systems every minor release (due to the version being part of the identifier). People can always request a minimum (or maximum) version with pkg-config, like `dlib-1 >= 18.15`, but this doesn't guarantee parallel installability. Given the current state of the library, I assume most interfaces to be pretty solid, not? Hence a _major_ versioning scheme should be preferred. Or said differently, version as much as needed (backwards incompatible API changes) but as little as possible (i.e., dlib-18.17 and dlib-18.18 is too much and will again be too much of a hassle to be usable).

davisking:
Well, it's a C++ library that heavily uses templates.  So pretty much every
change is going to potentially break ABI compatibility since most of the
code is in header files.  You definitely need to link against the specific
dlib version you compiled with.

SoapGentoo:
The pkg-config file versioning doesn't care about **ABI** stuff (which is another whole issue, and pretty much impossible to keep stable with C++ anyway). I'm talking about **API** versioning. Say I have a program extensively using dlib today. Will it still _compile_ with the next version of dlib? Or will I have to adjust everything? The idea is to change to dlib-2.pc once my program doesn't compile against it anymore. (What actually constitutes _"doesn't compile"_ is not clearly defined. Changing one minor argument? Maybe not. Changing 20% of all function signature? I would say yes). Anyway, putting the whole version in the filename is not a tractable way, as consumers cannot know the exact version being compiled against.

davisking:
Ok, sounds good then as long as it doesn't lead to people trying to mix ABIs.  One final thing though, is there a reason not to name the file dlib/dlib.pc.in and have cmake rename it to dlib-1.pc?

SoapGentoo:
Mixing ABIs is semi-impossible on most Linux distributions and OS X, as you need to go our of your way to build a system that is compiled with two toolchains (say, GCC 4.8 and 4.9). You can also have it named `dlib/dlib-1.pc.in` and then produce `dlib/dlib-1.pc` after running cmake over it. Tell me which way you prefer, and I'll amend it accordingly.

davisking:
Well, I mean compiling an application with dlib v1 and then switching out the .so file for a new version without recompiling the application.  That obviously needs to be avoided, which it sounds like will be the case.

Yeah, definitely name it dlib/dlib.pc.in so I don't have to rename the file in source control every time the API changes (even though that isn't very often)

SoapGentoo:
@davisking, I've updated the commit, I hope that makes my PR mergeable :smile:

davisking:
Sweet.  Although I just noticed the #include path is wrong.  You can't put the dlib folder in the include path, it should be the folder containing the dlib folder :)

SoapGentoo:
Fixed :smile:

davisking:
Sweet, thanks for the PR :)

SoapGentoo:
Hey Davis,
thanks for merging. Given that the last proper release of dlib is some time back, would you consider making a cut soon? So we can push the pkg-config support into distributions.

davisking:
It should be soon.  I'm just finishing up the final bits of the new deep
learning tools in dlib.  Once that's good I'll label the next release.  It
should be some time this month.

PR #113, opened 2016-05-17
base_window::get_pos on windows don't accounted titlebar, converting …

marcoslima:
base_window::get_pos on windows don't accounted titlebar, converting the client area to screen coordinates. Now, using GetWindowRect, position is returned correctly.


davisking:
Thanks for the PR.   However, it's not supposed to account for that.  The coordinate is relative to the client area so that other code can overlap windows in a controlled way.  For example, by creating popup menu windows that overlay the correct parts of a parent window.

PR #114, opened 2016-05-17
Concat layer

e-fominov:
I am working on inception layer implementation for using with Dlib's DNN. This PR introduces new grp<> layer type that can be used for inception layer creation

This layer have one input from subnetwork that is propagated into several subnetworks inside group and after that their result is merged into one output

I made also working example with inception layer inside to show how it works. Compared to mnist_ex, it learnes in less iterations, but with more time

This is a very first implementation, but completely working (as CPU, as GPU). If it fits the concepts of Dlib, I will make some better example and also regression tests
The biggest restriction for now - it can't be connected to the input layer directly, only to some computational layers


davisking:
Cool.  I looked it over quickly (will look more later), and noticed that you called host_write_only() but didn't assign to every memory location.  If you aren't going to write to every location you have to use host().  Also, unless I'm not understanding, this line is wrong: https://github.com/davisking/dlib/pull/114/files#diff-7bda83d17616598a79a24bc54d484181R806.  Shouldn't the loop range take the offset into account?  Doesn't the loop try to write to everything but the offset might be non-zero.

I didn't test it yet, but does the layer<N>(net) function work with this core layer type?

e-fominov:
Idea of concat_depth is to merge samples from:
out1: [a][b][c][d]
out2: [e][f][g][h]
into
out: [ae][bf][cg][dh]

square brackets are samples
here will be two calls for concat_depth:
```
concat_depth(out, 0, out1) // put content of each sample from out1 into the beginning of each sample in out
concat_depth(out, 1, out2) // put content of each sample from out2 into the end of each sample in out
```

split_depth produces out1 and out2 from combined out

depth_concat layer is calling concat_depth function in a way where all data in result tensor will be overwritten. but if somebody will call it later with not overwriting all data - it can be a problem
is there any performance overhead of host() over host_write_only() ?

layer<11>(net) - will work
layer<tag1>(net) - will work even if tag is after inception layer, but **not inside group**

davisking:
If everything is running either on the CPU or all on the GPU then there isn't any difference between host() and host_write_only().  So it's not a big deal in this case.

I'm still concerned about these split and cat functions.  It still seems like the code has indexing problems. The comments make it sound like the sample_offset is indexing into the k dimension (and therefore should be in the range 0 <= sample_offset < k()), but in the code sample_offset is just added to the pointer.  So it's just indexing the floats from the beginning of the tensor.  It would definitely be preferable that it be indexing along the k dimension :)

Also, I'm pretty sure you can implement this using the skip layers.  That way you wouldn't need to add a bunch of additional machinery to core.h, which is something I want to be as minimal as possible since it's full of a bunch of arcane template magic.  So for example, you should be able to do something like this:

template <typename SUBNET> using block1 = relu<con<5,3,3,1,1,SUBNET>>;
template <typename SUBNET> using block2 = relu<con<8,5,3,1,1,SUBNET>>;
template <typename SUBNET> using block3 = relu<con<20,5,3,1,1,SUBNET>>;

template <typename SUBNET> using inception = concat<tag1,tag2,tag3, tag1<block1<skip4< tag2<block2<skip4< tag3<block3<  tag4<SUBNET>>>>>>>>>>;

using net_type = inception<inception<inception<input<matrix<float>>>>>;

The only thing missing is a concat layer.  But that can be implemented by writing something conforming to the example layer interface defined in layers_abstract.h.  I put a lot of difficult template magic into core.h so that the layer interface could be as simple as possible :).   So you should be able to implement this using the easier interface :)

e-fominov:
Yes, you are right, I will change interface to use correct ofsset in range in the range 0 <= sample_offset < k()) and change to data.host() instead of write_only

I cant understand your answer about skip layers - github thinks that angle brackets are some kind of markup and erased all your code. Can you rewrite it using "Insert code" function?
I have checked the code of add_skip layer, looks like it works for tags. I can't imagine how to put a group of two layers into it

I was trying to create this grouping(concat) with EXAMPLE_COMPUTATIONAL_LAYER_ from layers_abstract.h, but was unable to make it because add_layer has this line inside:
`        const static size_t num_computational_layers = subnet_type::num_computational_layers + 1;
`
And it means that I will have only one solver for all my concatenation
I can improve my group layer to support "tags" inside it

Now I see some other minor problems, that are not done inside group:
1. Making it work for input layer as subnet will be tricky, as the same you have separate add_layer for this, may be we will need separate group for input layer subnet. This means that _inception<input_ will not work. But as workaround - _inception<tag1<input_ - will work
2.  Constructor with layer detail arguments (set of constructors) is not done yet. Do we need it?
```
       template <typename ...T>
        add_layer(
            T&& ...args)
```

And it can be done much more simple if use C++14 with lambdas supporting for auto keyword, but this can lead to compiler support problem

So the main question - to be or not? :)  

If yes - I will finish it, make regression test and improve example. If no - I will wait for some concatenation sample/sopport from you

davisking:
Oops, sorry about the awful formatting.  It's fixed now.

e-fominov:
So, as I understand, the main idea is not to hold layers inside concat layer (like my group is working now), but keep references on them with tags and make layer construction and iteration with existing core functions

Both approaches will work, but group layer holding layers inside has a bit more clear syntax. I will try to make concat layer with this syntax (skip layers) and compare the results
What is better for understanding?
concat<std::tuple<>, SUBNET> or concat<l1, l2.... ln, SUBNET> ?
While SUBNET is not concatted, but used as input

davisking:
That's right.  This is a little more syntax at the end, but way less syntax since you don't need to write hundreds of lines of complex stuff in core.h.  Generally speaking, I want there to be a core set of tools in core.h that let anyone easily define any kind of network they want without needing to understand or mess with the implementation details in core.h.  This means core.h needs to have the right primitives that let you define anything else.  I think the tag and skip tools let you do that.  I'm sure there are some things that can be improved, this is still a new dlib feature, but it's important that the way people add new layers and/or define new network architectures isn't by editing core.h.  If that's what is required then the DNN API in dlib will have been a failure in my opinion.  That's not to say that we can't improve things in core.h.  But I don't think we need a special concatenation tool in core.h


Also, here is how an inception template could be written in a cleaner way:

```
// Maybe this declaration could be in dlib somewhere since it's something people might want to reuse.
template <template<typename>class B1, template<typename>class B2, template<typename>class B3, typename SUBNET> using inception3 = concat<tag1,tag2,tag3, tag1<B1<skip4< tag2<B2<skip4< tag3<B3<  tag4<SUBNET>>>>>>>>>>;

// Then a user can make inception blocks by defining the sub blocks
template <typename SUBNET> using block1 = relu<con<5,3,3,1,1,SUBNET>>;
template <typename SUBNET> using block2 = relu<con<8,5,3,1,1,SUBNET>>;
template <typename SUBNET> using block3 = relu<con<20,5,3,1,1,SUBNET>>;
// and then defining their inception block
template <typename SUBNET> using incept = inception3<block1,block2,block3,SUBNET>;

// and then the network type is
using net_type = incept<incept<incept<input<matrix<float>>>>>;
```

The only problem with this is that I need to make a mechanism for defining globally unique tags, since in the above if the blocks have tag1 in them for example, the wrong thing will happen.  There are a few ways to deal with that and I'll add something soonish after I think about it a little more.

e-fominov:
This implementation has *concat* layer and *inception* typedefs. tensor merge/split functions replaced with one copy_tensor function. 
Changed example dnn_inception_ex to use new inception typedefs based on concat
Also to exclude some possible errors I defined new tags to use in inceptions - so user will be able to use normal tags with no conflicts inside inception layer
All functions are added to regression tests suite

How to improve the code inside dtest in dnn_tester? My implementation works, but doesnt look good
In my test I need to check if *backward* works corretly. To make it i need to call back_propagate_error and check if gradients are correct, but after backward calculation completes, all gradients becomes zero (stalled). To solve this - I am adding dnn_tester to friends of add_layer 
But dnn_tester is in anonymous namespace and can't be a friend. So I added one temporary class in dlib namespace inside test/dnn.cpp (dnn_tester)
What could you recommend to make code better in dnn_tester?

davisking:
Sweet, I've only had a little bit today to look at this but it looks really nice.  

To test it I would either create a concat layer object on its own and call forward and backward and verify the gradients that way.  Or you could make a simple network object and call the network's get_final_data_gradient() method.  Either way is fine.  Normally the test_layer() function would do the trick but it can't handle layers that are as complex as this one. 

Also, can you make the SUBNET type be the last template argument for the concat template?  I'm not sure if it's possible to do that using variadic templates.  If not then maybe there should be concat2, concat3, etc. explicit overloads.  I know it's a little less elegant but having the order reversed from the other layers will definitely confuse users.

e-fominov:
Ok, I changed concat syntax into concat1, concat2....
I cant make variadic here because of two reasons:
1. variadic declaration must be the last in template params
2. SUBNET is typename, whyle TAG's are template<typename>, so I cant merge them into one variadic list

dnn_test made more readable

davisking:
Looks great.  Can you make the tests work without the friend declaration in core.h?

e-fominov:
Yes, I found a way to do it.

davisking:
Awesome.  I really like how this looks.  Your code is very clean and good compared to a lot of other things I often see :)

However, I was just testing it and found that it doesn't compile on Ubuntu 14.04 with g++ 4.8.4.  It gives an error in layers.h about the sizeof...(TAG_TYPES).  This seems to be clearly an error in this version of gcc, but nevertheless, this is the latest LTS Ubuntu officially supported by cuda so we need to work around it.  RedHat and CentOS use similar versions of gcc as well and I know many industrial users of dlib who use such systems.

Can you switch the sizeof statements with a recursive counter based on something like the concat_helper_impl template?  That seems like a straightforward way to work around this gcc bug.  Then I'll merge the PR :)

e-fominov:
Done, my kung fu is growing with you :) 
Thanks for testing, I dont have gcc 4.8.4 Developing with MINGW 5.3 and testing with Ubuntu 14/gcc 5.0 with Titan-X

davisking:

PR #115, opened 2016-05-18
Removing trailing whitespace

joelg236:
I'm sorry to be the one that does this, but when browsing dlib I noticed that there's a ton of trailing whitespace littered around. No big deal at all, it's just a "clean" thing.

I totally realize that there's over a thousand files to verify here which sucks, but I only performed this command (zsh style globbing):

    sed -i -r 's/([^ \t]+)[ \t]+$/\1/' **/*.h **/*.cpp

You can verify at home by doing the same and diffing the diffs.

    $ sed -i -r 's/([^ \t]+)[ \t]+$/\1/' **/*.h **/*.cpp && git diff | sha1sum
    f563b2b105161c170c41090b2ba425150125ca3c  -

Should be the same sha1. Obviously, you have to trust that regex to do the correct thing, but it's pretty basic. Let me know if this kind of thing isn't important, not a big deal by any stretch like I said.


joelg236:
Note that the sha1 is based on master being `617ffba`

davisking:
Thanks for the PR, but this change disrupts the change control history and makes it difficult to see what was really changed when.  Which is something that is sometimes useful to know.  It would be one thing if it fixed bugs but having less white space in the code doesn't matter in any practical use case.

PR #125, opened 2016-05-30
Added getter for trainer::train_one_step_calls

e-fominov:
When training with train_one_step(), I want to make some extra testing every 1000 calls (example value). I want to use counter provided by trainer, because it is serialized into syncronization file and restored from it if process is restarted


davisking:
Good idea.

PR #126, opened 2016-06-02
Shape predictor trainer optimizations

e-fominov:
I need to train some shape predictor and I found that this process is very slow for me - up to 5-6 days with 64 GB RAM usage. Also training process uses a lot of RAM

With this PR I made this changes:
1. training_sample::feature_pixel_values was float-based. Now I am using the type provided by image_traits. In the most cases - it is unsigned char leading to lower RAM usage (about 15% less summary with all other changes). Train_shape_predictor_ex uses 400 features inside each sample (1600 bytes was, now - 400)
2. some parts of trainer uses the shape difference training_sample::target_shape - current shape. And this differenct is recalculated for the same sample many times. I made caching for it inside training_sample::diff_shape. This increases RAM usage a little (+68*2*4 = 544), and improves performance for about 20%
3. After caching differences I found that some parts are now easy to run parallel and I made it work with parallel_for. When I am running it with 8 threads on i7, about 70% of CPU is used (because not all parts are paralleled) and training time reduces 2-3x

It is also possible to remove caching the difference of shapes and use much less RAM (about 20-30% less) depending on the task. 

I have tested this with train_shape_predictor_ex with increasing oversampling_amount to 3000, but I think we need some more time for testing

What do you think? Will such kind of optimizations be usefull?

### DIFF
--- a/dlib/image_processing/shape_predictor.h
+++ b/dlib/image_processing/shape_predictor.h
@@ -80,7 +80,7 @@ namespace dlib
                 i = 0;
                 while (i < splits.size())
                 {
-                    if (feature_pixel_values[splits[i].idx1] - feature_pixel_values[splits[i].idx2] > splits[i].thresh)
+                    if ((float)feature_pixel_values[splits[i].idx1] - (float)feature_pixel_values[splits[i].idx2] > splits[i].thresh)
                         i = left_child(i);
                     else
                         i = right_child(i);
@@ -235,7 +235,7 @@ namespace dlib
 
     // ------------------------------------------------------------------------------------
 
-        template <typename image_type>
+        template <typename image_type, typename feature_type>
         void extract_feature_pixel_values (
             const image_type& img_,
             const rectangle& rect,
@@ -243,7 +243,7 @@ namespace dlib
             const matrix<float,0,1>& reference_shape,
             const std::vector<unsigned long>& reference_pixel_anchor_idx,
             const std::vector<dlib::vector<float,2> >& reference_pixel_deltas,
-            std::vector<float>& feature_pixel_values
+            std::vector<feature_type>& feature_pixel_values
         )
         /*!
             requires
@@ -449,6 +449,7 @@ namespace dlib
             _num_test_splits = 20;
             _feature_pool_region_padding = 0;
             _verbose = false;
+            _num_threads = 0;
         }
 
         unsigned long get_cascade_depth (
@@ -601,6 +602,15 @@ namespace dlib
             _verbose = false;
         }
 
+        unsigned long get_num_threads (
+        ) const { return _num_threads; }
+        void set_num_threads (
+                unsigned long num
+        )
+        {
+            _num_threads = num;
+        }
+
         template <typename image_array>
         shape_predictor train (
             const image_array& images,
@@ -657,13 +667,17 @@ namespace dlib
                 << "\n\t you can't have a part that is always set to OBJECT_PART_NOT_PRESENT."
             );
 
+            // creating thread pool
+            thread_pool tp(_num_threads);
 
-
-
+            // determining the type of features used for this type of images
+            using image_type = typename std::remove_const<typename std::remove_reference<decltype(images[0])>::type>::type;
+            using pixel_type = typename image_traits<image_type>::pixel_type;
+            typedef typename pixel_traits<pixel_type>::basic_pixel_type feature_type;
 
             rnd.set_seed(get_random_seed());
 
-            std::vector<training_sample> samples;
+            std::vector<training_sample<feature_type>> samples;
             const matrix<float,0,1> initial_shape = populate_training_sample_shapes(objects, samples);
             const std::vector<std::vector<dlib::vector<float,2> > > pixel_coordinates = randomly_sample_pixel_coordinates(initial_shape);
 
@@ -684,17 +698,16 @@ namespace dlib
 
                 // First compute the feature_pixel_values for each training sample at this
                 // level of the cascade.
-                for (unsigned long i = 0; i < samples.size(); ++i)
-                {
+                parallel_for(tp, 0, samples.size(), [&](unsigned long i){
                     extract_feature_pixel_values(images[samples[i].image_idx], samples[i].rect,
-                        samples[i].current_shape, initial_shape, anchor_idx,
-                        deltas, samples[i].feature_pixel_values);
-                }
+                                                 samples[i].current_shape, initial_shape, anchor_idx,
+                                                 deltas, samples[i].feature_pixel_values);
+                });
 
                 // Now start building the trees at this cascade level.
                 for (unsigned long i = 0; i < get_num_trees_per_cascade_level(); ++i)
                 {
-                    forests[cascade].push_back(make_regression_tree(samples, pixel_coordinates[cascade]));
+                    forests[cascade].push_back(make_regression_tree(tp, samples, pixel_coordinates[cascade]));
 
                     if (_verbose)
                     {
@@ -741,7 +754,8 @@ namespace dlib
             }
         }
 
-        struct training_sample 
+        template<typename feature_type>
+        struct training_sample
         {
             /*!
 
@@ -756,15 +770,18 @@ namespace dlib
                 - present == 0/1 mask saying which parts of target_shape are present.
                 - rect == the position of the object in the image_idx-th image.  All shape
                   coordinates are coded relative to this rectangle.
+                - diff_shape == temporary value for holding difference between current
+                  shape and target shape
             !*/
 
             unsigned long image_idx;
             rectangle rect;
             matrix<float,0,1> target_shape; 
-            matrix<float,0,1> present; 
+            matrix<float,0,1> present;
 
-            matrix<float,0,1> current_shape;  
-            std::vector<float> feature_pixel_values;
+            matrix<float,0,1> current_shape;
+            matrix<float,0,1> diff_shape;
+            std::vector<feature_type> feature_pixel_values;
 
             void swap(training_sample& item)
             {
@@ -773,12 +790,15 @@ namespace dlib
                 target_shape.swap(item.target_shape);
                 present.swap(item.present);
                 current_shape.swap(item.current_shape);
+                diff_shape.swap(item.diff_shape);
                 feature_pixel_values.swap(item.feature_pixel_values);
             }
         };
 
+        template<typename feature_type>
         impl::regression_tree make_regression_tree (
-            std::vector<training_sample>& samples,
+            thread_pool& tp,
+            std::vector<training_sample<feature_type>>& samples,
             const std::vector<dlib::vector<float,2> >& pixel_coordinates
         ) const
         {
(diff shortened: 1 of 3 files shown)


davisking:
Sweet.  I looked it over and it seems reasonable to me.  I didn't test it though as I'm away on vacation this week.  I'll look at it more when I get back.  But if you want to test it you should train a landmarking model using this data: http://dlib.net/files/data/ibug_300W_large_face_landmark_dataset.tar.gz.  If you make a model with the default training parameters both before and after this change and get the same results then everything is good.

davisking:
I ran a bunch of tests on this today.  On a quad core i7, if I use 4 cores it's about 33% faster to train.  But for 1 core it's much slower, taking twice the time it normally does.  For 2 cores I got a slight slowdown but only a little.   Also, in each case the landmarking accuracy is now slightly worse, 0.5%.  It's small and might just be chance.   It doesn't seem like the code changes should result in any real changes to the behavior. I ran a few different tests and it seemed like it's persistently that much worse but it's hard to tell.  Also, the memory usage wasn't substantially different between the two versions.

So here is what I suggest.  Can you make it so using 1 thread doesn't have a slowdown relative to the previous version?  That could also use the lower RAM version you mentioned.  Also, can you make it compile without C++11 support?  I don't want to force C++11 usage yet for dlib users.  That does make the parallel_for code pretty ugly though.  

I'll probably require C++11 sometime around January 2017, so you could also wait until then to merge the PR if you want.  I know there are a bunch of corporate users of dlib who are stuck on old compilers and it takes them about 6 years to update :(.  Even 2017 is a little aggressive for some of them but not using C++11 in the bulk of dlib's code is really starting to grate on me.  So 2017 is the date I'm officially switching dlib to C++11.

e-fominov:
Thanks for the tesing. My servers are still busy. Planning to re-check it on this week
I will find some good way to solve this problems

davisking:
Cool.

I should also mention that the main holdup on C++11 is visual studio.  If this code works in visual studio 2013 maybe we can include it.  Do you know what version range of visual studio supports these features?

e-fominov:
Fixed compatibility for Visual Studio. Tested VS2010, VS2015, MINGW 5.3. 
Fixed performance leak in single thread mode (reason of problem - optimizer and CPU cache). Now trainer.set_num_threads(0 or 1) will make it work no slower than original implementation (really 20% faster)

I am not finished checks with ibug - waiting for complete.
0 and 1 thread mode now have no difference in the testing scores with original version
2+ thread mode has some difference in the 4th digit normally. This is because rounding and limited float 32 precision. I can make it work with no difference, but performance affects very much

My measurements on the examples\train_shape_predictor_ex (VS2015, x64)

|Threads|Time, seconds|Training error|Testing error|
----------|------------------|-----------------|----------------
|original | 171 sec|2.40105e-05|0.085379|
|new, 0-1|148 sec|2.40105e-05|0.085379|
|2          |96 sec|0 |0.0856668|
|3          |70 sec|2.40105e-05|0.0856295 |
|8          |59 sec|2.40105e-05 | 0.0858176|

difference between 3 and 8 threads is small, because sample uses small amount of data. testing results on ibug dataset - tomorrow
in 8 thread mode, CPU utilization on i7 is about 60-70%, because some parts are hard to make parallel

davisking:
Sweet, nice work.  Let me know when it's ready.

I've also been thinking about C++11 and think waiting until 2017 is
excessive.  I'll make the next release of dlib soon, which has some C++11
in it and also includes updated cmake scripts that enable it by default.
But right now all the header files that exist in v18.18 still only require
C++03.  Let's see if anyone complains about the cmake changes or compiler
support for C++11 stuff in the DNN tooling.  If nothing bad happens then
I'll make C++11 required for the following dlib release and that release
can include the updated shape predictor.

e-fominov:
IBug results:
-----------------------------------------------------
|Method|Time|Training error|Testing error|
|---------|-------|-----------------|----------------|
|original|08h09m|0.0712444|0.0709394|
|new, 1 thread|7h 41 min|0.0712444|0.0709394|
|new, 4 threads|4h 18 min|0.071225|0.0708772|

Tested on Ubuntu + GCC 5.3. Memory usage is near identical
This code works too on VS 2010 and VS 2015

About C++11, DNN and Visual Studio - something is changing now in VS - new Uptade 3 for VS 2015 inproved  C++ support, but test_for_c++11 still does not compile
And NVidia published new Cuda 8 with support of VS 2015 Update 1 (Up2+ not supported)
VS 2015 compilation for DNN now has only a little amount of errors (only for layer_has_inpace... - like overloaded functions), so may be it will be possible to make VS 2015 compile code soon, after Cuda 8 will support VS 2015 Up3

davisking:
Awesome, this is really nice work.  Did you try setting the block_size to a smaller value so that you get more than one block per thread?  I would expect that oversubscribing the thread pool to be faster, this way, if one of the threads finishes before the others it doesn't sit idle as long because there are probably other blocks to process.  So maybe having 8 times as many blocks as threads is often good.

e-fominov:
Just checked with 8x blocks. Calculation time increased 2x. I am splitting very small parts of algorithm into threads, so there is too much overhead on thread management. And more than 4 threads will not give any profit. So 4 thread makes calculations 2x faster and this is the best I can make now. May be some situations with bigger trees or more features can have better results from more threds.

To get better results I need to rewrite full algorithm and this is not what we need now. Even 2x faster is a great profit

davisking:
Ok, sounds good.  Yeah, 2x is definitely a good speedup.

patricksnape:
@davisking Just a note about C++11 - if you make C++11 a hard requirement then you will lose Python 2.7 support as 2.7 still requires MSVC 2008.

davisking:
Are you sure it doesn't work on newer MSVC versions?  I know about all the issues with different MSVC runtimes and all that, but they don't surface if the parts of the Python C API touched by boost.python are properly opaque.  I don't know if that's the case though, but I have noticed that using dlib compiled with different versions of MSVC always seemed to work fine.  That might have been chance though.

patricksnape:
As far as I know from having been involved in conda-forge is that there are some extremely prickly things (like file handles) that can cause segmentation faults even when not crossing the Python/C++ boundary (depending on the ordering of file handles being created etc - note that I don't think they even have to be shared due to some crazy global state).

This would only be on Windows though. So maybe you could decide to just drop support for 2.7 (and 3.4 since that requires MSVC 2010) on Windows which probably wouldn't be the end of the world. Or at least conda would drop support for those versions since binary compatibility is very important to us. Anyone willing to build from scratch could probably just build with MSVC 2015+ and hope!

davisking:
I think MSVC 2010 will run the code in this PR at least.  I do want to
start using more C++11 in dlib but would probably limit it to what is
supported in MSVC 2010 or 2012 at the absolute latest (except for the DNN
module which requires full C++11 support, but python users don't care about
that).

Do you have a sense for how common Pyhton 2.7 is on windows?  If it's still
wildly popular compared to 3.4 then we probably shouldn't drop support.  We
could let this PR hang here for a while until it's reasonable to drop 2.7.

patricksnape:
As far as I know Python 2.7 still holds the majority by a significant margin. However, the installation/build story on Windows is not friendly and so I think the vast majority of people actually just install pre-built binaries for Python packages (either conda or Christoph Gohlke's packages). So you could always just try and convince people that they should be using 3.x on Windows since lots of common packages now support it as well (OpenCV 3.x, Tensorflow, Caffe).

Official Python 2.7 support is dead as of 2020 - which is not too far in the future now.

davisking:
Hmm, well, what do you think @e-fominov?  Are you going to be particularly bothered if this PR isn't merged for maybe another 18 months?

patricksnape:
@davisking Excuse my naivety but what exactly in that code forces the use of C++11?

davisking:
Mainly the parallel_for loops that use C++11 lambdas.  They could be written by creating explicit function objects but it makes the code awful to read.  

Also, I'm not really sure it matter if we drop python 2.7 on windows right now anyway.  At least for myself, most of the things I'm interested in making over the next year are probably going to be related to dlib's deep learning toolkit which won't ever be exposed to python.  There would need to be some new feature in dlib that python 2.7 users on windows really wanted to use for this to matter.

patricksnape:
Ah right yes of course - so used to writing functional code that I had forgotten that C++ only recently got lambdas 😆

Given that you take a very strong line on what you do and don't support for dlib - I don't really foresee a major issue. It's only an issue on Windows Python 2.x and, as you mentioned, a user could always build dlib themselves with a non-compliant compiler and hope everything still works (which it may, it's just a crapshoot). I think this seems like a pretty nice addition and personally wouldn't be dogmatic about forcing support for Windows Python 2.7. 

My personal opinion on the 2vs3 debate is we should be moving to 3.x as soon as possible and anything we can do to encourage people to do that is a good thing.

e-fominov:
This code is tested VS2010, VS2015, MINGW 5.3, GCC 5.3.
Just now I tried to use Windows 7 x64, anaconda + python 2.7 + boost 1.59 and this PR from dlib. 
After running setup.py script - dlib was built successfull without errors (used VS2015 compiler found automatically) and shape predictor was trained successfull. 
No, there is no python interface for changing thread count, but trainer and the rest of lib works well

I dont have VS 2008 to test, but I am sure it will not compile because of lambdas.

davisking:
Yeah, I'm happy with this.   From now on we are using C++11 in dlib :)

But we should make sure that dlib works in visual studio 2012 at least.  The DNN stuff can require full C++11 support.  But for the rest of dlib, we should make sure it still works on a few of the recent versions of visual studio.

patricksnape:
😄 I can add the Python interface to choose number of threads some time next week probably.

davisking:
Sweet :)

davisking:
@patricksnape Where can I get visual studio 2010?  I have an old installer for visual studio 2010 express saved but I just tried to install it on windows 7 and after installing the SDK updates so I can build 64bit executables vc2010 is now completely broken (can't even compile this program "int main(){}").

How are you guys creating 64bit executables on windows?

patricksnape:
Visual Studio 2010 requires the Windows SDK 7.1 to build 64-bit programs. Installing it on Windows 7 is a nightmare. Please see [this guide](https://blog.ionelmc.ro/2014/12/21/compiling-python-extensions-on-windows/) to help you out.

davisking:
Oh joy :)

davisking:
Hmm.  I'm inclined to only support python 3.5 since that doesn't rely on a version of visual studio so old Microsoft has dropped support for it.  Is that going to be a big deal @patricksnape ?

patricksnape:
Not a big deal to me! I've moved to Python 3.5+ for all my work 😆. But on places like conda-forge we would thus only support Python 3.5 **for Windows (Unix would be unaffected I think)**. I imagine you might suddenly find a bunch of people on 2.7 complaining. However, I'm not averse to the 'stick' rather than the 'carrot' when it comes to forcing people into using Python 3!

Plus, you have a very convincing argument in that gaining the ability to develop in C++11 is probably more important than supporting a compiler from 2008.

davisking:
Cool, that's what we will do then :)

Issue #129, opened 2016-06-05
libjpeg: Use TRUE instead of true


AbdealiJK:
There are a lot of places in the source code where `true` is being used. While this is widely acknowledged and works, it would be a good idea to use TRUE. [This](https://travis-ci.org/AbdealiJK/file-metadata/builds/135429131#L687) travis build fails saying `cannot convert bool to boolean for ...`

The same issue was found in [lammps](https://sourceforge.net/p/lammps/mailman/message/32775346/) and they were able to fix it by using TRUE and Axel Kohlmeyer says:

> i think that is more likely due to the compiler.
> in any case, the suggested fix is correct and compatible with all
> compiler variants that i could test.

AbdealiJK:
@davisking Let me know if you agree. If yes, I can start making a PR

davisking:
In the C++ code?  I'm definitely not replacing true with TRUE.  true is part of the C++ standard, TRUE is not.

AbdealiJK:
Ah. Yes, indeed C++ standard is `true` not `TRUE`. Not sure what I was thinking when I was writing this ...

On reading http://news.povray.org/povray.unix/thread/%3Cweb.5145947644c9fa6cde1c639f0@news.povray.org%3E/ :
It seems that *sometimes* (depending on how libjpeg is compiled, or maybe in older versions?), it expects a `boolean` datatype which is custom defined by it with values TRUE and FALSE. `boolean` is an enumeration in jpeglib.

It seems in the libjpeg in [dlib/external/libjpeg/jmorecfg.h](https://github.com/davisking/dlib/blob/master/dlib/external/libjpeg/jmorecfg.h#L222) the TRUE is set to 1 and FALSE is set to 0. - but this is only for the static build.

In the dynamic build where it uses the libjpeg from the system, the `true` is not accepted sometimes as the TRUE is of type `boolean`.

davisking:
Yeah, if it were an enum that would be a problem.  I've never seen a version of libjpeg that used an enum though.  People do crazy things though.  So if you want to make a PR that replaces true and false in just the calls to libjpeg that's fine with me :)

PR #130, opened 2016-06-05
save_jpeg: Use TRUE instead of true

AbdealiJK:
In some verisons on jpeg, TRUE is an enum, and so `true`
fails because it is not of the enum's type. Now, all the
libjpeg calls use TRUE/FALSE.

Fixes https://github.com/davisking/dlib/issues/129


davisking:
Thanks for the PR :)

Issue #132, opened 2016-06-07
setup.py: cmake not found - Better error message


AbdealiJK:
It would be nice to specify methods to install cmake depending on the OS in https://github.com/davisking/dlib/blob/master/setup.py#L459 also.
Sometimes the issue isn't that it's not on the PATH (Especially for linux/osx where package managers do the needful) sometimes it's just that it's not installed.

PR #133, opened 2016-06-07
setup.py: Provide instructions to install cmake

AbdealiJK:
If cmake_path is not found, either cmake is not installed
or cmake is not in the PATH. Hence, we also give instructions
on how to install cmake if the path is not found.

Fixes https://github.com/davisking/dlib/issues/132

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -456,7 +456,42 @@ class build(_build):
         """use cmake to build and install the extension
         """
         if cmake_path is None:
-            raise DistutilsSetupError("Cannot find cmake in the path. Please specify its path with --cmake parameter.")
+            cmake_install_url = "https://cmake.org/install/"
+            message = ("You can install cmake using the instructions at " +
+                       cmake_install_url)
+            msg_pkgmanager = ("You can install cmake on {0} using "
+                              "`sudo {1} install cmake`.")
+            if sys.platform == "darwin":
+                pkgmanagers = ('brew', 'port')
+                for manager in pkgmanagers:
+                    if find_executable(manager) is not None:
+                        message = msg_pkgmanager.format('OSX', manager)
+                        break
+            elif sys.platform.startswith('linux'):
+                distname = None
+                try:
+                    import distro
+                except ImportError:
+                    import pip
+                    pip_exit = pip.main(['install', 'distro'])
+                    if pip_exit > 0:
+                        raise SystemExit('Pip was unable to install `distro`')
+                    import distro
+                distname = distro.id()
+                if distname in ('debian', 'ubuntu'):
+                    message = msg_pkgmanager.format(
+                        distname.title(), 'apt-get')
+                elif distname in ('fedora', 'centos', 'redhat'):
+                    pkgmanagers = ("dnf", "yum")
+                    for manager in pkgmanagers:
+                        if find_executable(manager) is not None:
+                            message = msg_pkgmanager.format(
+                                distname.title(), manager)
+                            break
+            raise DistutilsSetupError(
+                "Cannot find cmake, ensure it is installed and in the path.\n"
+                + message + "\n"
+                "You can also specify its path with --cmake parameter.")
 
         platform_arch = platform.architecture()[0]
         log.info("Detected Python architecture: %s" % platform_arch)


setup.py
```diff
@@ -456,7 +456,36 @@ def build_dlib():
...
+                                   .format(manager))
+                        break
+            else:
+                distname, _, __ = platform.linux_distribution()
+                distname = (distname or '').lower()
+                if distname in ('debian', 'ubuntu'):
+                    message = ("You can install cmake on Debian/Ubuntu "
+                               "using `apt-get install cmake`.")
+                elif distname in ('fedora', 'centos', 'redhat'):
+                    pkgmanagers = ("yum", "dnf")
```
jayvdb: this should be dnf, then yum, so latter versions with `dnf` do not suggest using `yum`

setup.py
```diff
@@ -456,7 +456,36 @@ def build_dlib():
...
+            if sys.platform == "darwin":
+                pkgmanagers = ('brew', 'port')
+                for manager in pkgmanagers:
+                    if find_executable(manager) is not None:
+                        message = ("You can install cmake on OSX using {0}"
+                                   " with the command `{0} install cmake`."
+                                   .format(manager))
+                        break
+            else:
+                distname, _, __ = platform.linux_distribution()
```
jayvdb: keep the original value of `platform.linux_distribution()` and then display that to the user?
It would mean your `message` variable could be constant...?

setup.py
```diff
@@ -456,7 +456,36 @@ def build_dlib():
...
+                distname = (distname or '').lower()
+                if distname in ('debian', 'ubuntu'):
+                    message = ("You can install cmake on Debian/Ubuntu "
+                               "using `apt-get install cmake`.")
```
jayvdb: All of these messages could use a `sudo`, to be slightly more helpful

setup.py
AbdealiJK: I tried doing that, but apparently on my Ubuntu 14.04:
```
In [10]: platform.linux_distribution()
Out[10]: ('debian', 'jessie/sid', '')
```
There is a bug at http://bugs.python.org/issue9514

If I use that variable directly on my system it'd show "debian" instead of "ubuntu" which may confuse users

I've seen other SO questions like [this](http://stackoverflow.com/questions/30109770/python2-7-pip-install-pymssql-fails-on-centos-6-3) Where centos is wrongly detected as RHEL.

So, I thought it would be a better idea to do this, as that value is not very reliable

jayvdb: :+1:
Note that this is deprecated in Python 3.5 https://docs.python.org/3/library/platform.html#platform.linux_distribution
Is there an alternative?

AbdealiJK: So, it seems they were in favour of deprecating it because it's implementation needed to be changed freqeuntly https://bugs.python.org/issue1322 and they recommend a pypi package for it because of this.
This is a package made from that code - https://pypi.python.org/pypi/distro

Perhaps we can add it to `setup_requires` and use it ?

jayvdb: Worth trying ; it is at least better than using a deprecated and incorrect `platform.linux_distribution`.  Then @davisking can decide whether he likes it or not.

Do mind this nasty .. https://github.com/nir0s/distro/blob/master/distro.py#L45

davisking:
@jayvdb If it refuses to run on non-linux platforms then we definitely can't require it.

jayvdb:
They have set it up so it can be required on any platform, but cant be imported on any platform.
We can work around that, by only importing after it has been confirmed that the platform is linux (of unknown flavour).

AbdealiJK:
Hey, so i've updated the code to use `distro`.
Also, *Just incase* (if the pip version is too old to support setup_requires, or whatever reason) I've made "distro" optional. i.e. If `distro` is not found, it just gives the link to the cmake installation instructions page.

I've never used `setup_requires` before, so if @jayvdb could check it's the right way to use it - that'd be great.

setup.py
```diff
@@ -456,7 +456,38 @@ def build_dlib():
...
+                    pkgmanagers = ("dnf", "yum")
+                    for manager in pkgmanagers:
+                        if find_executable(manager) is not None:
+                            message = msg_pkgmanager.format(
+                                distname.title(), manager)
+                            break
+            raise DistutilsSetupError(
+                "Cannot find cmake, ensure it is installed and in the path.\n"
+                + message + "\n"
+                "You can also specify its path with --cmake parameter.\n")
```
jayvdb: trailing \n in exception

jayvdb:
It looks right, but I'm not comfortable making that call on sight only.  I suggest checking a few older versions of `setuptools` and, separately, older versions of pip, doing `pip install -e .` and `pip install .`.

jayvdb:
See https://tahoe-lafs.org/trac/tahoe-lafs/ticket/2473 ; it appears that `setup_requires` may have some quirks.  Worth check to see whether they crop up on this project.  Possibly a better option is to fetch https://raw.githubusercontent.com/nir0s/distro/master/distro.py & compile it into a module on the fly only if it is needed when `cmake` wasnt found.

AbdealiJK:
I'm beginning to thing it may be a better idea to make a page in http://dlib.net/ which explains in detail about what to do in each platform and just show that link.

The reason for this is that, distros change things, for example `yum` -> `dnf` and now it seems Ubuntu is changing `apt-get` -> `apt` ? it makes it easy to change these instructions (It does not depend on the dlib release cycle). Rather than keep track of these in code, it would be easier to keep track of this in documentation.

@davisking @jayvdb Opinions ?

jayvdb:
Maybe `distro.py` could also hold code for determining the packaging manager... ?

AbdealiJK:
I don't think asking `distro` to do that is a good idea, because even if they detect the package manager, the rest of the command to install the package manager would be different. In our case above, yum, dnf, apt-get happen to use `install`, but others may have something else ... ?

Also, I've created https://github.com/nir0s/distro/issues/132 to directly ask the guys at `distro` what they think is a good way to use `distro` in setup.py.

davisking:
I would put the message in the code since the sort of users that need to be
told things like the name of their package manager generally won't read
documentation.  I very frequently get questions like "I read page X but
still can't do Y.".  Then I go look at page X and their question is
answered directly at the top of the page.  You then copy and paste the text
to the user and they go "oh, finally, thank you!".  Obviously they didn't
read the documentation at all.

AbdealiJK:
@davisking Point noted :D

I've updated the PR with a method which installs `distro` with `pip` if the package is not already installed. I thought this would be a better way than downloading the distro.py file directly.

davisking:
Does that require `setup python.py install` to be run as root?

AbdealiJK:
If 'python setup.py install' is being run with system python, the pip will be installing to system directories => both will be root.

If 'python setup.py install' is run in a virtual environment, then dlib and distro will be installed in the virtualenv and won't need sudo.

davisking:
Sounds good so long as commands like `python setup.py install --user` still
work.

jayvdb:
One scenario which i think fails now is a normal user running the following on a system without cmake or distro

  python setup.py build

or

  python setup.py test

The pip invocation needs --user in that case.

Also I think `ensurepip` should be loaded.

(Sorry i am without a computer to test atm)

AbdealiJK:
@jayvdb Yes, those would indeed fail.
I'm going to make another PR which checks the `pip` command's returnvalue and skip `distro` in such cases. It will simply give the URL in such cases rather than the recommended command.

I think users that use `setup.py` directly have a higher probability of reading the docs too :D So, they should probably not need this message

davisking:
Ha, maybe :)

PR #136, opened 2016-06-11
setup.py: Don't fail if pip fails in cmake err msg

AbdealiJK:
When cmake is not found, the error message suggests what to do to
install cmake. But if `distro` is not found and cannot be installed,
the distro specific error message cannot be shown. Hence, we simply
ignore this and continue on.

Also, make the pip install quiet so that if there is an error message
in that, the user does not get confused by it.


davisking:
Super, thanks for another PR :)

PR #139, opened 2016-06-21
DNN Visual Studio 2015 Update3 support

e-fominov:
Visual Studio 2015 Update 3 now has much better C++11 support, but not perfect :(
With this changes I successfully compiled and run dnn_mnist_ex in CPU mode
For CUDNN support on Windows platform we need to wait CUDA 8 next release


davisking:
Awesome!  Does the dnn_mnist_advanced_ex.cpp example compile as well? How about the unit tests?  There is a bunch of dnn stuff in there too.

e-fominov:
Yes, you are right, some problems with dtest - i will fix them. 
dnn_mnist_advanced_ex compiles well, but with millions of warnings (warning C4503 - 'identifier' : decorated name length exceeded, name was truncated https://msdn.microsoft.com/en-us/library/074af4b6.aspx)

So, you can plan support of VS2015 up3 in the next release. I will make the rest fixes today-tomorrow
Btw, on my machine compiling dnn_mnist_advanced_ex takes 3 GB RAM for compiling process. May be we should put this info somewhere in docs? Because somebody will try to build it on Raspberry Pi

e-fominov:
Now compiles all of them. But still have a problem - dnn_mnist_ex works good, but other dnn samples are crashing at runtime
If I simply add tag1<> anywhere into net definition inside dnn_mnist_ex - program compiles but crashes. 

I will try to localize problem. The problem is not tied with warning C4503 - small nets are crashing too

e-fominov:
Found the reason of crashing:

the code like
`        add_tag_layer() = default;`
does not work in MSVC
`        add_tag_layer() {};`

works

What do you think about this kind of replacement? (=default to {})

davisking:
Nice work. 

Yeah, replacing =default with {} is fine.

e-fominov:
Ok, I changed constructors. Now compiles and run well on VS and GCC. all dnn samples and dtest checked 
I didnt changed copy and move constructors and assignments - looks like they work as-is on VS

davisking:
Awesome.

What about putting the bigobj flag in dlib/cmake so it's always on?  Is there a drawback to doing that?  Also, why not put the warning suppression flag into a pragma statement in the code.  I've already got a lot of visual studio's inane warnings disabled with such pragmas :)

e-fominov:
Yes, we can safely put bigobj flag into them main cmake file. It has only one penalty - .obj files will not be readable by VS 2005
I changed samples to have #pragmas

e-fominov:
some minor formatting fixes done.

davisking:
Oh, I mean putting the pragma in core.h so users never see it at all.

e-fominov:
Moving to core.h does not help - too many header relations
I put it into dlib/dnn.h and it works good

davisking:
That's fine.  Is this ready to be merged or should I wait a bit?

e-fominov:
I think its ready

davisking:
Sweet.  Thanks for another PR :)

I also finished training an imagenet classifier (after months of testing on my GPU).  So I'm going to post an example program that shows how to use that along with the model file and then I'll tag dlib v19.0.  I'll do that this weekend.

e-fominov:
Wow! Good news! Are you planning to publish trainer code?  And how about future roadmap?

davisking:
I'll include the trainer code.  It's very similar to what is already in the
examples though so nothing too exciting.

The next thing I'm going to do is upgrade the object_detector to use a
CNN.  I think that will be pretty sweet.

davisking:
I just posed the imagenet model and example: https://github.com/davisking/dlib/blob/master/examples/dnn_imagenet_ex.cpp.  This model that comes with it took two weeks to train on my Titan X card :)

e-fominov:
Looks like this sample is too hard for MSVC. It does not print errors/message, but compilation time takes more than one hour (still waiting). Do we need some notice in comments about this? Other samples (even dnn_mnist_advanced) are compiled fast

And one more - as this sample takes crops with 227x227 size, it till not work good on ANY image. I think we should put some imagenet samples with correct image size for using example by users

davisking:
Wow, MSVC's C++11 support is amazingly bad.  I'm content to wait for the
next version of MSVC to come out and see if it gets better.  But if you are
still interested in trying to find more clever workarounds then that would
be cool.  But this one sounds like it's going to be hard to work around. :(
 So don't feel obligated.

It will work on any image.  If the image is too small it will upsample as
appropriate.

e-fominov:
The example dnn_imagenet_ex didnt compiled - compiler was working 12 hours with no progress. 
How about adding something like?:

`cmake .. -DEXPERIMENTAL_MSVC_DNN=ON`

And disable MSVC by default

davisking:
I put the cmake check for cxx_decltype_incomplete_return_types back into
cmake last night because of this.  So it should be behaving reasonably now.

davisking:
Which by that I mean disabling visual studio :)

PR #140, opened 2016-06-23
Add detection threshold adjustment to object detection python interface

nxwhite-str:
Here is a small set of changes that adds adjust_threshold as an optional parameter to the object_detector python interface.  It also updates the face_detector.py example to exercise it.

There are some irrelevant commits that I reverted in here - let me know if you want me to squash these into one, but I think github merge has an option for that now.


davisking:
Cool, thanks for the PR :)

PR #143, opened 2016-06-25
Add cmake files for android build

lilac:
Borrow cmake files from opencv, to make it buildable using Android's ndk toolchain.

### DIFF
--- a/android_build/AndroidNdkGdb.cmake
+++ b/android_build/AndroidNdkGdb.cmake
@@ -1,0 +1,96 @@
+# Copyright (c) 2014, Pavel Rojtberg
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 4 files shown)


davisking:
Have these files been adapted to dlib in some way?  Or are they simply copies of files from another project?

lilac:
@davisking They are copied from https://github.com/taka-no-me/android-cmake but I verified that it works without problems. I actually did not modify any of those files.

lilac:
If you think that's no worth to be copied to this project. Feel free to close it.

davisking:
Yeah, people who want to use it can just get it from the original project :)

lilac:
But there is no any documentation about that. Having it checkin have the benefits or supporting more platform without much effort.
If not added, I wish there is a documentation about this somewhere. That would save others' time.

davisking:
That's true.  But copying the files doesn't address the lack of
documentation.  If you wrote a blog post or something like that which
documented how to setup an android build using cmake that would be super.

PR #151, opened 2016-07-01
Faq changes for DNN

e-fominov:
Added DNN questions

Unfortunately, I cant check if it can build docs, when runnig "./makedocs" i get this error:

> abort: there is not Mercurial repository here (.hg not found)

But xml syntax is correct


davisking:
Cool.  I just merged it after tweaking the text a bit :)

PR #155, opened 2016-07-10
Spport to discontiguous Numpy arrays

mljli:
One problem is the function name `get_numpy_ndarray_parts` suggests that its a generic function. While I have to check for pixel type's consistency where  I assume the array to be no more than 3-D before copying the data. Otherwise, Python core dumps.  I know we could assign the pointer outside the function but that means we need to change the function prototype. So is this PR acceptable?

### DIFF
--- a/dlib/python/numpy.h
+++ b/dlib/python/numpy.h
@@ -7,12 +7,14 @@
 #include <dlib/error.h>
 #include <dlib/algs.h>
 #include <dlib/string.h>
+#include <dlib/array.h>
+#include <dlib/pixel.h>
 
 // ----------------------------------------------------------------------------------------
 
 template <typename T>
 void validate_numpy_array_type (
-    boost::python::object& obj
+    const boost::python::object& obj
 )
 {
     using namespace boost::python;
@@ -30,26 +32,23 @@ void validate_numpy_array_type (
 
 // ----------------------------------------------------------------------------------------
 
-template <typename T, int dims>
-void get_numpy_ndarray_parts (
-    boost::python::object& obj,
-    T*& data,
+template <int dims>
+void get_numpy_ndarray_shape (
+    const boost::python::object& obj,
     long (&shape)[dims]
 )
 /*!
     ensures
-        - extracts the pointer to the data from the given numpy ndarray.  Stores the shape
-          of the array into #shape.
+        - stores the shape of the array into #shape.
+        - the dimension of the given numpy array is not greater than #dims.
 !*/
 {
     Py_buffer pybuf;
-    if (PyObject_GetBuffer(obj.ptr(), &pybuf, PyBUF_ND | PyBUF_WRITABLE ))
-        throw dlib::error("Expected contiguous and writable numpy.ndarray.");
+    if (PyObject_GetBuffer(obj.ptr(), &pybuf, PyBUF_STRIDES ))
+        throw dlib::error("Expected numpy.ndarray with shape set.");
 
     try
     {
-        validate_numpy_array_type<T>(obj);
-        data = (T*)pybuf.buf;
 
         if (pybuf.ndim > dims)
             throw dlib::error("Expected array with " + dlib::cast_to_string(dims) + " dimensions.");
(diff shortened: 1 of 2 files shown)


davisking:
Couldn't you add another check to avoid the core dump?  

Also, you can't return a pointer to a stack local object (e.g. https://github.com/davisking/dlib/pull/155/files#diff-9836e22ceeab652f0f5f4ad9da01a482R127).  That pointer is not valid after `get_numpy_ndarray_parts` terminates.

mljli:
Then how? Without dynamic allocation, how to return a valid pointer? I thought about replacing `_data` with a `dlib::array` instance but was stuck with initializing `dlib::array` with a raw pointer in the case when the numpy.ndarray is C-contiguous.

davisking:
There are lots of options.  I would look at the place that uses this
function to get context.  But the most obvious thing to do is to pass the
array2d in by reference so that its lifetime is longer than the single
function invocation.

mljli:
I end up using `PyMem_Malloc` to leave memory management to the Python interpreter. The first reason is I can't figure out how to construct `dlib::array` with a raw pointer. The second is `dlib:array` frees its memory in its destructor and if that memory is borrowed from a Python object I am not sure if it will cause problems.

mljli:
You mean potential memory leak? Haven't tested that. If so, what about
calling PyMem_Free in the numpy_image's destructor?

davisking:
It's a leak if it never gets freed, which appears to be the case.

Calling PyMem_Free in numpy_image's destructor is also almost certainly an
error since you will sometimes be calling it on numpy's memory block, and
will get a double free when numpy or the python interpreter goes to free
it.  Unless there is some mechanism to tell it that you freed the memory
already.

In any case, far and away the simplest thing to do is to not do manual
memory management.  Using an array2d as I suggested avoids all these
difficulties.

mljli:
Sorry for not seeing your update. So what you mean is if no copying occurs, `_data` will be `pybuf.buf` and `array2d` is not used. Otherwise, `_data` points to `array2d`? In that way, `array2d` should be a member of `numpy_image` as well as `_data`. It's somewhat weird...OK, I'll code it later.

davisking:
Yeah, that's what I mean.  Some object needs to own the memory, i.e. be responsible for destroying it.  In C++ people call this RAII (https://en.wikipedia.org/wiki/Resource_Acquisition_Is_Initialization) and it's how all code in dlib should be written.

mljli:
Thank you for explaining it. I noticed that at [here](https://github.com/davisking/dlib/blob/master/tools/python/src/object_detection.cpp#L100) it copys a `numpy_rgb_image` to a `array<rgb_pixel>`. Since `numpy_*_image` already have those generic image interfaces and `find_candidate_object_locations` only converts it to a `const_image_view`. Is that copying unnecessary?

davisking:
No problem.

Yeah, that copy is redundant and could be removed.

mljli:
For performance concern, since constructing a `numpy_*_image` involves potential data copying now, I created a new function to get an array's shape and use that function to check if a numpy.ndarray is grayscale or RGB. Though this make the code a little bit complex and ugly.

I only did some simple tests and don't know if that is enough. [Here](https://github.com/davisking/dlib/files/365059/test_dlib.txt) is the unit test script I used.

dlib/python/numpy.h
```diff
@@ -114,5 +108,97 @@ void get_numpy_ndarray_parts (
...
+            throw dlib::error("Expected numpy.ndarray with " + dlib::cast_to_string(dlib::pixel_traits<T>::num) + " channels.");
+
+        if (PyBuffer_IsContiguous(&pybuf, 'C'))
+            data = (T*)pybuf.buf;
+        else
+        {
+            contig_buf.resize(pybuf.len);
+            if (PyBuffer_ToContiguous(&contig_buf[0], &pybuf, pybuf.len, 'C'))
+                throw dlib::error("Can't copy numpy.ndarray to a contiguous buffer.");
+            data = (T*)&contig_buf[0];
```
davisking: You shouldn't need this cast to T*.  It's better to not have casts when you don't need them.

dlib/python/numpy.h
```diff
@@ -30,26 +32,22 @@ void validate_numpy_array_type (
 
 // ----------------------------------------------------------------------------------------
 
-template <typename T, int dims>
-void get_numpy_ndarray_parts (
+template <int dims>
+void get_numpy_ndarray_shape (
```
davisking: Why are there two versions of get_numpy_ndarray_shape().  It seems like you should only need the const version.

dlib/python/numpy.h
```diff
@@ -114,5 +108,97 @@ void get_numpy_ndarray_parts (
...
+            throw dlib::error("Expected numpy.ndarray with " + dlib::cast_to_string(dlib::pixel_traits<T>::num) + " channels.");
+
+        if (PyBuffer_IsContiguous(&pybuf, 'C'))
+            data = (const T*)pybuf.buf;
+        else
+        {
+            contig_buf.resize(pybuf.len);
+            if (PyBuffer_ToContiguous(&contig_buf[0], &pybuf, pybuf.len, 'C'))
+                throw dlib::error("Can't copy numpy.ndarray to a contiguous buffer.");
+            data = (const T*)&contig_buf[0];
```
davisking: Shouldn't need this cast either.   C style casts like this are dangerous since they will silently cause undefined behavior in many cases.  It works here, but for example, if the type inside contig_buf were changed in the future this cast would likely compile without any errors and produce a program with undefined behavior.  So it's good to avoid this when it's not needed.

dlib/python/numpy_image.h
```diff
@@ -70,11 +73,9 @@ class numpy_rgb_image
     numpy_rgb_image (boost::python::object& img) 
     {
         long shape[3];
-        get_numpy_ndarray_parts(img, _data, shape);
+        get_numpy_ndarray_parts(img, _data, _contig_buf, shape);
         _nr = shape[0];
         _nc = shape[1];
-        if (shape[2] != 3)
```
davisking: Isn't this check still needed to see if the python image is 3 channel?

davisking:
Cool.  It's almost there it seems.  I added a few comments to the diff on github though about a few problems.

dlib/python/numpy_image.h
mljli: Nope. It's checked in `get_numpy_ndarray_parts` to ensure successful data copying. Anyway, do you have any suggestion on how to reduce the complexity? (i.e. to reuse as more code as possible for both checking and constructing)

davisking: Oh, I see.  In that case, the if statements in get_numpy_ndarray_parts() should be adjusted to make sure dims==3 for this case.  Right now they might give the wrong result for a grayscale image that just happened to have 3 columns.

mljli: I compare `pixel_traits<T>::num` to the last dimension only when it is greater than 1. So it should work with grayscale image as well.

davisking: Right, but what if dims==2?  Then it does the wrong thing.

mljli: get_numpy_ndarray_shape() ensures dims>=pybuf.ndim.

davisking: Ah, so it does.  Still, the code is a little hard to follow since reasoning
about its correctness requires pulling together these different little
details from various parts of the code.  It's better if the bodies of
functions are self evidently correct just own their own.  This isn't always
possible, but adding a little more to the if statements in some of these
functions can make them obviously correct independent of the specific error
handling happening in other functions.  Also, it is likely that other
people will edit these functions at some point and they will probably not
consider their effects on code that calls them.  For example, if the error
handling behavior of get_numpy_ndarray_shape() were to change then the
error handling behavior of the function we are talking about could easily
become incorrect.

mljli: Yeah. A lot of thought was given to it. Since the ndarray is very likely to be non-contiguous (e.g. ROI of an image), copying the data twice is obviously unwise. I will rethink about it.

davisking: The code is good how it is :)

I'm just saying maybe add a little more error handling to some of the
functions even though it's redundant within the scope of the entire
program.

mljli:
Updated.

davisking:
Cool, thanks for the PR :)

PR #156, opened 2016-07-11
fix shape_predictor dlib::searialize and dlib::desearialize error

vsooda:
move friend method searialize and desearialize implement outside the shape_predictor, so that we can use the  not only `desearialize` but also `dlib::desearialize`.

in fact, in the current master branch, if use as follow:

```cpp
istringstream in("model_string");
dlib::deserialize(sp, in);
```

The compiler will report ` error: no matching function for call to ‘deserialize(dlib::shape_predictor&, std::istringstream&)’` error.

this pull request fix this bug.

### DIFF
--- a/dlib/image_processing/shape_predictor.h
+++ b/dlib/image_processing/shape_predictor.h
@@ -398,26 +398,9 @@ namespace dlib
             return full_object_detection(rect, parts);
         }
 
-        friend void serialize (const shape_predictor& item, std::ostream& out)
-        {
-            int version = 1;
-            dlib::serialize(version, out);
-            dlib::serialize(item.initial_shape, out);
-            dlib::serialize(item.forests, out);
-            dlib::serialize(item.anchor_idx, out);
-            dlib::serialize(item.deltas, out);
-        }
-        friend void deserialize (shape_predictor& item, std::istream& in)
-        {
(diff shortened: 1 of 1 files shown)


davisking:
Are you sure?  Serialize is spelled wrong in that error message.

vsooda:
sorry, I have not pasted wrong error message. I have fix it. You can re produce the result, just add the two line code above in `face_landmark_detection_ex.cpp`

davisking:
Yeah, so it does :)

Anyway, those functions need to be declared inline to avoid linker errors.  Do that and I'll merge the PR.

davisking:
Cool.  I'm pretty sure it needs to be on the function bodies, not the friend declarations though.

vsooda:
what do you mean?  the inline declare must in the head of implement? I have try that, and it can compile successfully.

davisking:
I mean the functions themselves must be declared inline or you will get errors at link time about multiply defined symbols.  You aren't getting this error now because your program has only one .cpp file that includes this code. But if you had a multi-cpp file program you would get linker errors.

vsooda:
I am not familiar with "inline". I have google it. Thank you explaining this.

davisking:
No problem

PR #169, opened 2016-07-26
Make dlib compile with WIN32_LEAN_AND_MEAN defined

rggjan:
This is the only file that doesn't compile when `WIN32_LEAN_AND_MEAN` is defined. Since I define this flag project wide and just include `dlib/all/source.cpp`, it would be cool to make this work.


davisking:
Cool, thanks for the PR :)

PR #170, opened 2016-07-26
Added support for CIELab color space pixels.

sutr90:
I have created new type of pixel. This pixel represents color in CIELab color space.

### DIFF
--- a/dlib/pixel.h
+++ b/dlib/pixel.h
@@ -1241,6 +1241,18 @@ namespace dlib
             assign_pixel_helpers::assign(dest,temp);
         }
 
+        template < typename P1, typename P2 >
+        typename enable_if_c<pixel_traits<P1>::hsi && pixel_traits<P2>::lab>::type
+        assign(P1& dest, const P2& src)
+        {
+            rgb_pixel temp;
+            // convert lab value to our temp rgb pixel
+            assign_pixel_helpers::assign(temp,src);
+            // now we can just go assign the new rgb value to the
+            // hsi pixel
+            assign_pixel_helpers::assign(dest,temp);
+        }
+
     // -----------------------------
         // dest is an lab pixel
         template < typename P1>
@@ -1291,6 +1303,20 @@ namespace dlib
 
             // now we can just go assign the new rgb value to the
             // hsi pixel
+            assign_pixel_helpers::assign(dest,temp);
+        }
+
+        template < typename P1, typename P2 >
+        typename enable_if_c<pixel_traits<P1>::lab && pixel_traits<P2>::hsi>::type
+        assign(P1& dest, const P2& src)
+        {
+            rgb_pixel temp;
+
+            // convert hsi value to our temp rgb pixel
+            assign_pixel_helpers::assign(temp,src);
+
+            // now we can just go assign the new rgb value to the
+            // lab pixel
             assign_pixel_helpers::assign(dest,temp);
         }
     }
--- a/dlib/pixel.h
+++ b/dlib/pixel.h
@@ -1262,8 +1262,8 @@ namespace dlib
         typename enable_if_c<pixel_traits<P1>::lab>::type
         assign(P1& dest, const unsigned char& src)
         {
-            dest.a = 0;
-            dest.b = 0;
+            dest.a = 128;
+            dest.b = 128;
             dest.l = src;
         }
 
@@ -1272,8 +1272,8 @@ namespace dlib
         typename enable_if_c<pixel_traits<P1>::lab && pixel_traits<P2>::grayscale>::type
         assign(P1& dest, const P2& src)
         {
-            dest.a = 0;
-            dest.b = 0;
+            dest.a = 128;
+            dest.b = 128;
             assign_pixel(dest.l, src);
         }
 
--- a/dlib/pixel.h
+++ b/dlib/pixel.h
@@ -88,9 +88,9 @@ namespace dlib
                 - This type of pixel represents the Lab color space.
                 - num == 3
                 - has_alpha == false
-                - basic_pixel_type == signed char
-                - min() == -128
-                - max() == 127
+                - basic_pixel_type == unsigned char
+                - min() == 0
+                - max() == 255
                 - is_unsigned == false
             - else
                 - grayscale == true
@@ -214,14 +214,14 @@ namespace dlib
         ) {}
 
         lab_pixel (
-                signed char l_,
-                signed char a_,
-                signed char b_
+                unsigned char l_,
+                unsigned char a_,
+                unsigned char b_
         ) : l(l_), a(a_), b(b_) {}
 
-        signed char l;
-        signed char a;
-        signed char b;
+        unsigned char l;
+        unsigned char a;
+        unsigned char b;
     };
 
 // ----------------------------------------------------------------------------------------
@@ -486,10 +486,10 @@ namespace dlib
         const static bool hsi = false;
         const static bool lab = true;
         const static long num = 3;
-        typedef signed char basic_pixel_type;
-        static basic_pixel_type min() { return -128;}
-        static basic_pixel_type max() { return 127;}
-        const static bool is_unsigned = false;
+        typedef unsigned char basic_pixel_type;
+        static basic_pixel_type min() { return 0;}
+        static basic_pixel_type max() { return 255;}
+        const static bool is_unsigned = true;
         const static bool has_alpha = false;
     };
 
@@ -878,6 +878,7 @@ namespace dlib
             L is between 0 and 100
             a is between -128 and 127
             b is between -128 and 127
+            RGB is between 0.0 and 1.0
         */
         inline Lab RGB2Lab(COLOUR c1)
         {
@@ -940,19 +941,20 @@ namespace dlib
                 var_Z = (7.787 * var_Z) + (16.0 / 116);
             }
 
-
-            c2.l = (116 * var_Y) - 16;
-            c2.a = 500 * (var_X - var_Y);
-            c2.b = 200 * (var_Y - var_Z);
+            //clamping
+            c2.l = max(0.0, (116.0 * var_Y) - 16);
+            c2.a = max(-128.0, min(127.0, 500.0 * (var_X - var_Y)));
+            c2.b = max(-128.0, min(127.0, 200.0 * (var_Y - var_Z)));
 
             return c2;
         }
 
         /*
-            Calculate RGB from HSL, reverse of RGB2HSL()
-            Hue is in degrees
-            Lightness is between 0 and 1
-            Saturation is between 0 and 1
+            Calculate RGB from Lab, reverse of RGB2LAb()
+            L is between 0 and 100
+            a is between -128 and 127
+            b is between -128 and 127
+            RGB is between 0.0 and 1.0
         */
         inline COLOUR Lab2RGB(Lab c1) {
             COLOUR c2;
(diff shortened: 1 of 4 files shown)


dlib/pixel.h
```diff
@@ -771,6 +865,158 @@ namespace dlib
...
+
+        struct Lab
+        {
+            double l;
+            double a;
+            double b;
+        };
+        /*
+            Calculate Lab from RGB
+            L is between 0 and 100
```
davisking: Is this comment correct?  Because code that uses this function seems to assume L has the range -128 to 127.  In any case, the full range of values of l should be used so there isn't loss of dynamic range when using LAB pixels.

dlib/pixel.h
```diff
@@ -962,6 +1241,84 @@ namespace dlib
...
+        }
+
+
+        template < typename P1, typename P2 >
+        typename enable_if_c<pixel_traits<P1>::lab && pixel_traits<P2>::grayscale>::type
+        assign(P1& dest, const P2& src)
+        {
+            dest.a = 0;
+            dest.b = 0;
+            assign_pixel(dest.l, src);
```
davisking: There are a number of places where grayscale pixels are assigned to L, but L is signed and assign_pixel will perform clipping.  So for example, 8bit grayscale values are in the range [0,255], and if I assigned a grayscale pixel with a value of 255 to a LAB pixel it would get saturated to L==127.  Then when the LAB pixel is assigned back to grayscale it will become 127 rather than 255.

davisking:
Sweet, thanks for the PR.  I added some comments to the code.   The main thing to do is to make sure conversions between pixels types are as invertable as possible.   So some of the code needs to be updated, I think, and a few more tests need to be added to check that more extreme pixel values convert correctly (e.g. grayscale 255 and 0 to/from LAB as well as similar extreme values in other color types).

dlib/pixel.h
sutr90: Hmm, this is tough one. The L is lightness in percent, by definition the range is [0;100] [see second paragraph](https://en.wikipedia.org/wiki/Lab_color_space#CIELAB). I have chosen *signed char* so the pixel is internally consistent.

Your observation is correct though. There are significant losses in the dynamic range of L. **I will check some other implementations out there to see how do they work.** 

Scaling and shifting the L from [0;100] -> [-128;127] would be a solution, but then the actual value would not be usable, because the expected values are in [0;100].

Another option would be to switch to *float* but I'm not sure if the 4x increase in memory consumption is not overkill.

dlib/pixel.h
sutr90: Yes this is definitely incorrect, I will fix it when I decide how to represent the L value correctly.

dlib/pixel.h
davisking: Sounds good.  I wager the best thing to do is to scale it to [0,255].   :)

sutr90: Ok, I see how it is done with the hsi_pixel. The representation is correct, it's just shifted and scaled to fit "unsigned char". So when accessing the members directly, you have to do the shift and scale first. 

What I'm trying to say, is that you cannot assign values to the lab_pixel using the struct members directly, but you cannot do that with hsi_pixel either:
`//RGB #FF00AA (magenta)
rgb_pixel magenta();
magenta.r = 0xff;
magenta.g = 0x0;
magenta.b = 0xaa;

// HSL[320, 100%, 50%] (same magenta)
hsi_pixel magenta();
magenta.h = 320; // not possible (use 320/360 * 256)
magenta.s = 100; // incorrect
magenta.i = 50; // incorrect
`
So I will do it in same way as the HSI works.

davisking: Right, that sounds good.  It's all about packing it into 8bits somehow.

sutr90:
I have converted the Lab pixel to *unsigned char*. But there is still some bug. In following images you can see the artifacts: [http://imgur.com/a/ToZPy](http://imgur.com/a/ToZPy)

The original image contains all possible RGB values, the second is result after conversion to my Lab and back. I will investigate and report back.

sutr90:
I have tested the code and there is definitely some glitch in my compiler. I'm using GCC from MinGW on Windows and it produces the glitch I have mentioned previously. 

I have tested the conversion functions separately and they are correct. Then I have compiled my code on MSVC and the glitch is gone. I have then tested it on Linux with GCC and again, it is working without the glitch.

So as far as I know the commit 933ab00 is correct and working implementation of Lab pixel in dlib.

davisking:
So you mean the images shown in http://imgur.com/a/ToZPy only have the funny colors in mingw?  The other compilers work correctly?

davisking:
Also, can you add some unit tests for extreme values of l, a, and b like 255 and 0?

sutr90:
Yes the problem is only with lab_pixel only in mingw. I have tested GCC 4.8:
```
COLLECT_GCC=g++
Target: mingw32
Configured with: ../../../src/gcc-4.8.1/configure --build=mingw32 --enable-languages=ada,c,c++,fortran,lto,objc,obj-c++ --enable-libgomp --enable-lto --enable-graphite --enable-libstdcxx-debug --enable-threads=posix --enable-version-specific-runtime-libs --enable-fully-dynamic-string --enable-libstdcxx-threads --enable-libstdcxx-time --with-gnu-ld --disable-werror --disable-nls --disable-win32-registry --disable-symvers --enable-cxx-flags='-fno-function-sections -fno-data-sections -DWINPTHREAD_STATIC' --prefix=/mingw32tdm --with-local-prefix=/mingw32tdm --with-pkgversion=tdm-2 --enable-sjlj-exceptions --with-bugurl=http://tdm-gcc.tdragon.net/bugs
Thread model: posix
gcc version 4.8.1 (tdm-2)
```
and 
```
COLLECT_GCC=g++
Target: i686-w64-mingw32
Configured with: ../gcc-5.3.0/configure --prefix=/mingw32 --with-local-prefix=/mingw32/local --build=i686-w64-mingw32 --host=i686-w64-mingw32 --target=i686-w64-mingw32 --with-native-system-header-dir=/mingw32/i686-w64-mingw32/include --libexecdir=/mingw32/lib --with-gxx-include-dir=/mingw32/include/c++/5.3.0 --enable-bootstrap --with-arch=i686 --with-tune=generic --enable-languages=c,lto,c++,objc,obj-c++,fortran,ada --enable-shared --enable-static --enable-libatomic --enable-threads=posix --enable-graphite --enable-fully-dynamic-string --enable-libstdcxx-time=yes --disable-libstdcxx-pch --disable-libstdcxx-debug --enable-version-specific-runtime-libs --disable-isl-version-check --enable-lto --enable-libgomp --disable-multilib --enable-checking=release --disable-rpath --disable-win32-registry --disable-nls --disable-werror --disable-symvers --with-libiconv --with-system-zlib --with-gmp=/mingw32 --with-mpfr=/mingw32 --with-mpc=/mingw32 --with-isl=/mingw32 --with-pkgversion='Rev5, Built by MSYS2 project' --with-bugurl=https://sourceforge.net/projects/msys2 --with-gnu-as --with-gnu-ld --disable-sjlj-exceptions --with-dwarf2
Thread model: posix
gcc version 5.3.0 (Rev5, Built by MSYS2 project)
```
both are Mingw, though different packagers and versions, produce artifacts.

On the other hand GCC on Ubuntu
```
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/5/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 5.4.0-6ubuntu1~16.04.1' --with-bugurl=file:///usr/share/doc/gcc-5/README.Bugs --enable-languages=c,ada,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-5 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-5-amd64/jre --enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-5-amd64 --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-5-amd64 --with-arch-directory=amd64 --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --enable-objc-gc --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.1)
```
and MSVC 2015 on Linux
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24210 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.
```
are without the artifact.

I can reproduce the issue no matter what combination of input and output formats I use, so it's most probably not a libpng or libjpg issue.

I will investigate the MinGW issue first and do the missing tests later.

davisking:
Sweet!  Let me know when it's ready  :)

sutr90:
The additional tests are in. :+1:

davisking:
Super.  That covers lab to rgb, but those new tests need to convert the rgb
value back to lab and make sure it turns back into the original lab value
(or nearly so).

sutr90:
That is not possible afaik. The problem is that the conversion between Lab and RGB is not invertible. Consider the first test:
```
lab_pixel(0,0,0) [Lab(128,-128,0)] to RGB -> (0, 64, 194)
RGB(0, 64, 194) to Lab -> Lab(33,36,-71)[lab_pixel(84,164,56)]
```
but 
```
Lab(33,36,-71)[lab_pixel(84,164,56)] to RGB -> (0, 64, 194)
```

The problem is that multiple Lab values are mapped to same RGB value, and only some of the Lab values are actually invertible.

davisking:
Ok. How about checking that rgb->lab->rgb is approximately invertable?  Surely that should work reasonably.

sutr90:
Yes, that would work. I will add those 8 tests (for limit values of RGB). And keep the current tests so the Lab limits are under control.

sutr90:
Hmm I have added the RGB->Lab->RGB tests and it's no good. The loss of precision on 8-bit is too big. I have measured error up to 9 on single channel (e.g. R input = 0; R output = 9). If I "upgrade" the lab_pixel to **short** I can get the error almost to 0, with occasional 1. 

_What do you think, is the extra byte per channel per pixel justifiable?_

davisking:
It's not going to be a perfect mapping.  The central test is to map that
image with all the RGB values to LAB and then back and see that it looks
correct to your eye.  Which is what you already did if I understand
correctly.  The unit tests should just check that the inversion is
sufficiently approximate.

sutr90:
The forth-and-back tests are in. I have checked OpenCV implementation and it produces same artefacts, when using 8bits per channel.

davisking:
Sweet.   Thanks for the PR :)

PR #209, opened 2016-08-31
Get the number of iterations of find_min/max_box_constrained optimization

cbeauthi:
Proposed changes in order to obtain the number of iterations performed by find_min_box_constrained and find_max_box_constrained functions of optimization.h


davisking:
This change breaks existing code since it changes the public API.  If you want to include this kind of feature in the dlib main repository then it would be best to add counters to the stopping strategies since they get called once per iteration.  Then you could query the strategy object to get the count.

PR #212, opened 2016-08-31
Add dnn_trainer::train_one_step iterator signature

jpblackburn:
Add an overload of dnn_trainer::train_one_step that takes a pair of
iterators rather than a std::vector.


davisking:
Super, thanks for the PR :+1:

PR #213, opened 2016-09-01
Adding specific parameters interface on fc_ layer

### LINKED ISSUE #208: How to initialize a the bias vector on a fc DNN layer?
I've had some success in implementing a RNN over dlib's DNN framework (I expect to publish it soon), but to correctly implement an LSTM, I need the bias of forget gate set initially to a large value (as [this paper](http://jmlr.org/proceedings/papers/v37/jozefowicz15.pdf) says on section 2.2), otherwise the LSTM loses its power to track large time dependencies. In dlib's terms, this an `fc<>`. Is there an easy way to set how this value will be initialized?

lvella:
Added interface to resolve #208. Is it fit for merge?

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -969,6 +969,30 @@ namespace dlib
             tt::gemm(1,sub.get_gradient_input(), 1,gradient_input,false, w,true);
         }
 
+        alias_tensor_instance get_weights()
+        {
+            return weights(params, 0);
+        }
+
+        alias_tensor_const_instance get_weights() const
+        {
+            return weights(params, 0);
+        }
+
+        alias_tensor_instance get_biases()
+        {
+            static_assert(bias_mode == FC_HAS_BIAS, "This fc_ layer doesn't have a bias vector "
+                "to be retrieved, as per template parameter 'bias_mode'.");
+            return biases(params, weights.size());
+        }
+
+        alias_tensor_const_instance get_biases() const
+        {
+            static_assert(bias_mode == FC_HAS_BIAS, "This fc_ layer doesn't have a bias vector "
+                "to be retrieved, as per template parameter 'bias_mode'.");
+            return biases(params, weights.size());
+        }
+
         const tensor& get_layer_params() const { return params; }
         tensor& get_layer_params() { return params; }
 
--- a/dlib/dnn/layers_abstract.h
+++ b/dlib/dnn/layers_abstract.h
@@ -529,6 +529,62 @@ namespace dlib
                 - #get_bias_weight_decay_multiplier() == val
         !*/
 
+        alias_tensor_const_instance get_weights(
+        ) const;
+        /*!
+            ensures
+                - returns an alias of get_layer_params(), containing the weights matrix of
+                  the fully connected layer.
+                - #get_weights().num_samples() is the number of elements in input sample,
+                  i.e. sublayer's output's k * nc * nr.
+                - #get_bias().k() == #get_num_outputs()
+                - if get_bias_mode() == FC_HAS_BIAS:
+                    - #get_layer_params().size() == (#get_weights().size() + #get_biases().size())
+                - else:
+                    - #get_layer_params().size() == #get_weights().size()
+        !*/
+
+        alias_tensor_instance get_weights(
+        );
+        /*!
+            ensures
+                - returns an alias of get_layer_params(), containing the weights matrix of
+                  the fully connected layer.
+                - #get_weights().num_samples() is the number of elements in input sample,
+                  i.e. sublayer's output's k * nc * nr.
+                - #get_bias().k() == #get_num_outputs()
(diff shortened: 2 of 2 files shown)


davisking:
But this doesn't initialize the bias value.  You need to do something that changes the behavior of the setup() method.  That's where the initialization happens.

lvella:
But having a way to get the the bias vector, I just derived fc_<num_outputs, FC_HAS_BIAS> and reimplemented setup().

In the end, this is just to ensure the documented contract, because if I called get_layer_params() and created the alias tensor myself, I would have no guarantee that the private implementation of fc_ wouldn't change (ij a future version)

lvella:
But having a way to get the the bias vector, I just derived fc_<num_outputs, FC_HAS_BIAS> and reimplemented setup().

In the end, this is just to ensure the documented contract, because if I called get_layer_params() and created the alias tensor myself, I would have no guarantee that the private implementation of fc_ wouldn't change (in a future version, maybe) change the position of the biases inside the params tensor.

davisking:
Yeah, it's nice to have.  Just pointing out that it doesn't address the bias issue you were having :)

However, are you sure `template <typename = std::enable_if<bias_mode == FC_HAS_BIAS>>` does that you want?  I don't think that will cause a compiler error when called.  Did you test it?  In any case, it's almost certainly more clear to simply put a static_assert inside the function body that bias_mode has the right value.

lvella:
I've tested the method yesterday on a small test sample. It worked, but I couldn't reproduce it today, I changed something while coding and it is certainly wrong now. Anyway, a static_assert seems indeed a much better solution. Let me change it.

lvella:
Oh, by the way, the rnn code is in https://github.com/lvella/dlib-rnn

It is very raw: I can't lower the loss below a certain value, and I don't know why, maybe my model is wrong, the trainer setup is wrong, or there is a bug. I was trying to teach it Shakespeare, as from this blog post: http://karpathy.github.io/2015/05/21/rnn-effectiveness/ and next step would be to try to reproduce closely what the author's program do.

I don't know how slow it is compared to others libs, because I don't have Cuda to test it on GPU, but it would be most certainly slower (if I was certain it worked) because it uses a lot of memory.

The memory problem is, as you said in another place, because I had to unroll the full DAG. But I only had to do it because of this particular "ensure" in the `backward()` method in `EXAMPLE_COMPUTATIONAL_LAYER_`:

> Moreover,  this was the most recent call to forward().  This means that
> forward() is allowed to cache intermediate results so they can be used
> during the backward computation.

But in a RNN, forward must be called many times for the same layer before backpropagation starts, so I had to copy the subnetwork for each sample, to ensure each `backward()` is called after the corresponding `forward()`.

On the bright side, the building of the network and the rnn topology works much like the rest of dlib, and I was able to implement all 5 RNN types described in [this paper](http://jmlr.org/proceedings/papers/v37/jozefowicz15.pdf) by simply staking different layer types (some of which I've implemented). I tried each of them during a limited time, but all off them dropped the loss with training.

You may want to take a look on it to see if the strategy used to implement RNN could be merged in dlib itself. If you are interested, I am willing to rework the code to match dlib's style, testing and documentation guidelines.

lvella:
Changed to static_assert.

davisking:
Sorry for the delay.  I've been a combination of sick and super busy with other stuff :(

Anyway, that's pretty cool.  I know about the unrolling problem.  I thought a lot about how to support non-DAG networks when designing the API.  It almost supports it, but there are some things that would need to be different (like the thing you mention) to really support non-DAG graphs but those things have negative performance impacts for important DAG architectures like CNNs.  Also, I think the static network definition is a little funny for RNNs.  

At various points I considered the sort of API you setup.  That's why there are functions there that let you get the final gradient and control the loss and gradient backprop actions (which I saw you were using in your code).  Like it would be cool if you could define processing blocks with the dlib template API and then wrap them up in some non-DAG RNN thing.  But there needs to be a way to do it that allows for parameter sharing among unrolled nodes in the RNN.  Some libraries deal with this by pushing the ownership of the parameters outside the layer details objects, but I don't like that because it complicates the API a user needs to use to define those layers.  

One possible approach to parameter sharing would be to add some kind of "tie" function to the tensors.  So maybe you would say, tensor1.tie(tensor2); and that would cause them to both be some kind of alias to the same underlying tensor.  That is, from that point on rensor1 and tensor2 would be have the same state with the same memory buffers associated with them.  So you could create your RNNs by unrolling things but you would tie all the parameter tensors in corresponding layers.

This way, the layer details interface is not made complex, and indeed doesn't even know about this.  It also still allows for layer implementations to do some internal caching if that's appropriate for speed.  I have to think about this a little more to make sure  there isn't some reason why it doesn't work.  Or maybe if there is some better way in general.

Also, there is the question of what is the API a user uses to define RNNs.  That is totally unaddressed in the current dlib API.  I would be more in favor of a runtime based API for that than a compile time API.  Anyway, it's a big undertaking :)

Issue #223, opened 2016-09-09
Cannot build on OSX 10.11.6 with CUDA 7.5 and XCode 7.3.1


grefab:
Building dlib with CUDA support on OSX does not work (at least for me). Here is what I've tried:
- Building with clang from XCode 7.31 (Apple LLVM version 7.3.0 (clang-703.0.31) ): cudnn_test_build fails: Does not support thread_local (see http://stackoverflow.com/a/29929949/72082, my output here: https://gist.github.com/grefab/dfe4e64355d8bf13312ee446f7b346a4)
- Building with clang 3.8 from homewbrew: Build of cuda_test_build fails: That clang version is not supported by nvcc.
- Building with clang 3.7 from homebrew: cuda-test-build fails: Multiple definitions of pow (see https://gist.github.com/grefab/b6399975368d87bb8324748e1cc336dd) I'm not sure what to do about that.
- Building with GCC from homebrew (gcc-6 (Homebrew gcc 6.2.0) 6.2.0): cuda_test_build fails: nvcc fatal: GNU C/C++ compiler is no longer supported as a host compiler on Mac OS X.

Any ideas what I've missed and/or could be done to get dlib with CUDA support build on OSX?

davisking:
Maybe you need clang 6 from x code (
http://docs.nvidia.com/cuda/cuda-getting-started-guide-for-mac-os-x/#axzz4JkoHKEy1).
In any case, to use dlib with cuda you need a compiler supported by NVIDIA
and the compiler needs to support C++11.

grefab:
Meh. That won't help. C++11 support of Apple's current clang is incomplete, i.e. it does not support `thread_local`. Versions before that are no better. Compiler extensions (`__thread`, `_Thread_local`) are not the same (I tried replacing `thread_local` by them: `cudnn_dlibapi.cpp:166:13: error: '__thread' variables must have global storage`, `cudnn_dlibapi.cpp:115:13: error: '_Thread_local' variables must have global storage`, so unless I figure out a way to get rid of `thread_local` in my fork of dlib completely (which I really dont'want) it seems that I have to wait for XCode 8 with a version of clang that really supports C++11.

davisking:
Well, you should be able to use the dlib::thread_specific_data instead.
You declare them statically in a function and it works basically the same
way as thread_local.

davisking:
[...]

PR #224, opened 2016-09-09
Fixed detection of C++11 features

grefab:
Detecting C++11 capabilities did not work on OSX 10.11.6, XCode 7.3.1, Apple LLVM 7.3.0, cmake 3.6.1. First check revealed that CMAKE_CXX_KNOWN_FEATURES was empty, so subsequent matches failed. After reading http://public.kitware.com/pipermail/cmake/2014-November/059182.html changed handling to fetch property instead of relying on variable. Works on my machine.

### DIFF
--- a/dlib/cmake_utils/use_cpp_11.cmake
+++ b/dlib/cmake_utils/use_cpp_11.cmake
@@ -52,15 +52,16 @@ if (CMAKE_VERSION VERSION_LESS "3.1.2")
    endif()
 else()
    # Set a flag if the compiler you are using is capable of providing C++11 features.
-   if (";${CMAKE_CXX_COMPILE_FEATURES};" MATCHES ";cxx_rvalue_references;" AND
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the patch.  Did this get CUDA working for you in OS X (#223)?

grefab:
Nope, this is what allowed me to try to build dlib's cuda component in the first place. As long as C++11 detection did not work, dlib did not even try to build cuda.

davisking:
Ah, too bad.  I'm not sure what the deal is with clang and `thread_local`.  There is a similar problem in Ubuntu.  I think it was supported in previous versions of clang but now it isn't.  I've read the rationale about ABI problems with gcc libraries.  But still.  It's kinda ridiculous that clang went backwards on C++11 for a while.

grefab:
Yep, I don't get it either. Still, there is this little bit of satisfaction that at least the problem is now understood. :) Out of curiosity: Which setup do you use?

davisking:
I use gcc on Ubuntu most of the time.

PR #230, opened 2016-09-12
Fixed cudnn_hint_path detection for OSX where CUDA_CUBLAS_LIBRARIES is a list, not a single string.

grefab:


davisking:
Awesome.  Thanks for the PR :+1:

PR #234, opened 2016-09-20
Fix for USE_AVX_INSTRUCTIONS and MKL build

dashesy:
Currently, doing `cmake` for `dlib` without examples, i.e. running

``` bash
    cd dlib
    mkdir build
    cd build
    cmake .. -DUSE_AVX_INSTRUCTIONS=1
```

Will result in 

> CMake Warning:
>  Manually-specified variables were not used by the project:
> 
>    USE_AVX_INSTRUCTIONS

Another problem is that MKL does not find `cblas_dot` because it is in `mkl_rt` and therefore `CHECK_FUNCTION_EXISTS(cblas_ddot HAVE_CBLAS)` will result in:

> -- Looking for cblas_ddot
> -- Looking for cblas_ddot - not found

In addition I added a `MKL_PATH` to override search path for MKL when it is known.


davisking:
Did you try to set CMAKE_PREFIX_PATH to a location that contained the MLK? That should work and is in general what you do to solve this problem.

dashesy:
Yes `CMAKE_PREFIX_PATH` will work, but we have our own `CMakeList` that pulls multiple cmake projects, and tweaking `CMAKE_PREFIX_PATH` is like playing with a global variable. Whereas we know where MKL is located because it is part of the repo, and not installed in usual places.

davisking:
Ok.  How about making cmake look in an environment variable?  Anything
other than another cmake option since that clutters cmake-gui.

dlib/cmake_utils/cmake_find_blas.txt
```diff
@@ -58,21 +62,23 @@ if (UNIX)
     check_type_size( "void*" SIZE_OF_VOID_PTR)
 
     if (SIZE_OF_VOID_PTR EQUAL 8)
-        set( mkl_search_path
```
davisking: Why not just add $ENV{MKL_PATH} to the list of paths searched?  That would be much clearer, no if statements.  I'm not sure these statements are even right.  Above, mkl_search_path is set.  Do these if statements really fall through correctly if MKL_PATH is not set to something?

davisking:
I just added some comments.   

I actually added them over a week ago but didn't understand what github's new 'review' system is doing.  Apparently there are now a bunch of extra clicks needed to make comments on a change log public.  It kinda sucks.

dlib/cmake_utils/cmake_find_blas.txt
dashesy: It works but your suggestion is much cleaner

dashesy:
I fix and test everything then prepare another PR

Issue #235, opened 2016-09-21
Warnings


dashesy:
In our project we run with warnings as errors, and it is C++14.
I can see some warnings could be easily fixed, for example:

> warning C4297: 'dlib::thread_pool::~thread_pool': function assumed not to throw an exception but does
> note: destructor or deallocator has a (possibly implicit) non-throwing exception specification

Also it seems the presence of so many consts in `numeric_constants.h` make it unavoidable to clash, for example `e` will result in warning:

> optimization_line_search.h(135): warning C4459: declaration of 'e' hides global declaration

That can be fixed by renaming `e` to `err` but perhaps hiding constants in `numeric_constants` in another level of namespace like

``` C++
    namespace dlib 
    {
    namespace constants 
    {
    ...
    }
    }
```

could be a better approach.

Do you think it is useful if I send some PR for cleaning up these warnings?

davisking:
Visual studio has a lot of inane warnings.  That warning about constants is
silly.  I already have a bunch of pragma statements that disable some
visual studio warnings in dlib/algs.h.  You can add to that list if you
want.

Although some warnings you are getting might be reasonable to fix with code
changes.  It depends on what they are though.  For example, putting a call
to exit() in the thread_pool's destructor could be a reasonable change.  Or
maybe even declaring the destructor as a throwing destructor, although it's
definitely a bad user error if there are exceptions at that point.  My
preference is to leave the code as it is.   But I could be persuaded here
to use exit() to avoid the warning.

I also test dlib at the highest warning levels in gcc and clang and don't
get warnings.  I also never have warnings in gcc that I need to disable.
But visual studio is ridiculous.   There is even a warning in visual studio
(which I've disabled in algs.h) that says parts of the C++ standard library
are deprecated and you are advised to use something windows specific.

e-fominov:
Hi, i have just updated to GCC 6.2 (MinGW-based) and I see this warnings when building dlib (latest source from git):
In file included from C:/work/dlib/dlib/threads.h:22:0,
                 from C:/work/dlib/dlib/sync_extension/sync_extension_kernel_1.h:6,
                 from C:/work/dlib/dlib/sync_extension.h:6,
                 from C:/work/dlib/dlib/gui_core/gui_core_kernel_1.h:34,
                 from C:/work/dlib/dlib/gui_core/windows.h:4,
                 from C:/work/dlib/dlib/gui_core.h:12,
[...]

PR #236, opened 2016-09-21
Fixed warning about winsock2.h in windows build

e-fominov:
Fixed warning about sockets from #235


davisking:
Awesome, thanks :)

PR #238, opened 2016-09-21
namespace for numeric constants

dashesy:
Initial attempt to enclose constants in namespace.
Used `constants::` when only few constants where needed, but `using` in the function when more needed. 
In the cpp files, using added to the top of the file.

### DIFF
--- a/dlib/geometry/point_transforms.h
+++ b/dlib/geometry/point_transforms.h
@@ -911,7 +911,7 @@ namespace dlib
             set_rowm(proj,2) = trans(Z);
 
             width = num_pixels/2.0;
-            dist_scale = width/std::tan(pi/180*camera_field_of_view/2);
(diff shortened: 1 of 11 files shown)


davisking:
Does that fix the warnings?

dashesy:
No all of them.
Here is another:
In `rectangle.h` function `inline bool operator<` the input parameter name is `b` which is same as the field with the same name in that class:

> warning C4458: declaration of 'b' hides class member
> note: see declaration of 'dlib::rectangle::b'

I think in this case for example, the `b` in the argument list should be just renamed to `rect`

dashesy:
I can send more PR while I encounter them, or find a way to reveal them all and one PR?

davisking:
You might as well wait until the PR covers all the warnings that are bothering you :)

PR #239, opened 2016-09-21
first check large number in elsif

dashesy:
This is to use initializer lists with MSVC 2015.
Also I wonder why [this commit](https://github.com/davisking/dlib/commit/48df23a0e065e01c5ba176d4999fffb0d9a29e02) disabled the C++11 with MSVC? It is quite capable of compiling C++14.


davisking:
Good catch.  

As for VC2015, it definitely didn't support all of C++11 as of a month or two ago.  I would be very surprised if their C++14 support was very good considering that.

PR #241, opened 2016-09-22
Update fhog_object_detector_ex.cpp

mhaghighat:
I think you are decreasing the size of the images using pyramid_down. If you were increasing, how would it be supposed to get trained for smaller faces?


davisking:
No, the example is correct.

PR #246, opened 2016-09-23
add set_state to kalman filter for the case we know the initial state

dashesy:
Sometimes initial state is known, also sometimes the state should be corrected (like if the model has known boundaries).


davisking:
That's a good idea.  Can you update the documentation in the abstract.h file as well?  Make sure it's clear how the "state" in this function relates to the distinct predict and current states.

PR #258, opened 2016-09-28
fixed warning about unsigned-signed comparison in imgtool

grisevg:
I hope no one figures out my evil plan of becoming a contributor, by doing the smallest possible change.


davisking:
Ha, cool, thanks for the patch :)

PR #259, opened 2016-09-28
added --working-dir argument for imgtool

grisevg:
Using `--working-dir` user can store "xml descriptor" separately from image folder or even mix and match different image folders and "xml descriptor" files.

### DIFF
--- a/tools/imglab/src/cluster.cpp
+++ b/tools/imglab/src/cluster.cpp
@@ -140,7 +140,9 @@ int cluster_dataset(
     image_dataset_metadata::dataset data;
 
     image_dataset_metadata::load_image_dataset_metadata(data, parser[0]);
-    set_current_dir(get_parent_directory(file(parser[0])));
+    
+    const string working_dir = get_option(parser, "working-dir", get_parent_directory(file(parser[0])).full_name());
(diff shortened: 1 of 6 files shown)


davisking:
This seems somewhat perverse.  You are making it so you can have XML files that contain relative paths but those paths aren't correct relative to the XML file.  Why do you want to do that?  That seems like an organizational nightmare.

grisevg:
If anything, to me it makes more sense to use the current path `imglab` was executed in as root for relative paths. Using parent directory of the XML file seems a bit confusing to me.

davisking:
If you give someone a dataset (an XML file and images), and the XML file
doesn't contain the correct paths to the images, how are they supposed to
open it?

grisevg:
That's a good point. I better just fix my XML generator and project structure.

davisking:
Yeah, make it so the datasets are self contained :)

I also added the `convert_imglab_paths_to_relative` and `copy_imglab_dataset` scripts to the imglab folder recently that make moving datasets around more convenient.  What the first script does is obvious.  The second one is useful because you might have a folder with a bunch of xml files in it and you want to copy just the part of that whole folder tree corresponding to one of the xml files.

PR #261, opened 2016-09-29
Fix building on win32 with 32bit architecture

Satchitananda:
Now PYTHON_INCLUDE_DIR and PYTHON_LIBRARY setting up correctly for both 32/64 bit architectures


davisking:
Cool, thanks for the PR :)

PR #269, opened 2016-10-03
DNN Samples modified for VS 2015 compilation

e-fominov:
With this changes I have successfully compiled ALL examples with VS 2015 Up3 with CUDA support. For training dnn_imagenet_train_ex i used x64 compiler, as x86 is not able to compile because of not enough heap size (C1002 error)

May be some renaming should be made

### DIFF
--- a/examples/dnn_imagenet_train_ex.cpp
+++ b/examples/dnn_imagenet_train_ex.cpp
@@ -40,25 +40,36 @@ template <int N, typename SUBNET> using ares_down = relu<residual_down<block,N,a
 
 // ----------------------------------------------------------------------------------------
 
+template <typename SUBNET> using level1 = res<512,res<512,res_down<512,SUBNET>>>;
+template <typename SUBNET> using level2 = res<256,res<256,res<256,res<256,res<256,res_down<256,SUBNET>>>>>>;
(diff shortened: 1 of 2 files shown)


examples/dnn_imagenet_train_ex.cpp
```diff
@@ -40,25 +40,36 @@ template <int N, typename SUBNET> using ares_down = relu<residual_down<block,N,a
...
-                            res<512,res<512,res_down<512,
-                            res<256,res<256,res<256,res<256,res<256,res_down<256,
-                            res<128,res<128,res<128,res_down<128,
-                            res<64,res<64,res<64,
-                            max_pool<3,3,2,2,relu<bn_con<con<64,7,7,2,2,
+                            level1<
+                            level2<
+                            level3<
+                            level4<
+                            max_pool<3,3,2,2,relu<affine<con<64,7,7,2,2,
```
davisking: The training network should use all bn_con layers instead of affine layers.

davisking:
Sweet.  Thanks for figuring this out.

grisevg:
@e-fominov, big thanks for the PR. Could you change `affine` to `bn_con` or give edit right to @davisking so he could change it?

davisking:
I merged the changes manually when I updated a few things a moment ago.  So no worries.

PR #271, opened 2016-10-05
Unreachable code, fixed compiler warning

rggjan:


davisking:
Oh yeah, that is weird.  Oops :)

But really, the change should be to remove the first return and keep the second.  Otherwise the code still looks odd.

rggjan:
Both of them would work, I guess :)

davisking:
Right, but if we are changing it we should make it nice :)

rggjan:
Depends on if nice=short/stylish or nice=readable/understandable ;)

davisking:
It's always readable.  

In any event, I want the code to be readable.  I would make the change myself but then I have to reject the PR.  Or you can make the change and I'll merge the PR.  Either way is fine with me :)

davisking:
Sweet, thanks for the PR :)

PR #275, opened 2016-10-07
find_cudnn modified for cudnn installed a user directory.

elda27:
I have tried the dlib 19.1 with Visual Studio 2015 Update 3, CUDA 8 and cuDNN 5.1. 
I failed configuration of CMake because cuDNN is not found when I installed cudnn in %USERPROFILE%/library/cuDNN/ .
Of course, I added the above path to CMAKE_PREFIX_PATH.

Unzipped cuDNN 5.1 has following directories and files.

(Root directory of cuDNN 5.1)
├─bin ─ cudnn64_5.dll
├─include ─ cudnn,h
└─lib ─ x64 ─ cudnn.lib

So, I added x64 to PATH_SUFFIXES on line 14 in its source code to solve this problem.


davisking:
Sounds good, thanks for the PR :)

Issue #281, opened 2016-10-10
error when build dlib on windows


ZhaoWangFu:
I cloned the newest dlib from github and use CMake3.5.0 to biuld it on windows7 64bit.
And i want to use dlib in a Qt project.
And What i need is the face landmark detection function in Dlib.

Here is My local environment:
**PS D:\dlib-master\dlib_build> gcc -v
Using built-in specs.
COLLECT_GCC=C:\MinGW\bin\gcc.exe
COLLECT_LTO_WRAPPER=c:/mingw/bin/../libexec/gcc/mingw32/4.9.3/lto-wrapper.exe
Target: mingw32
Configured with: ../src/gcc-4.9.3/configure --build=x86_64-pc-linux-gnu --host=mingw32 
--prefix=/mingw --disable-win32-registry --target=mingw32 --with-arch=i586 
--enable-languages=c,c++,objc,obj-c++,fortran,ada --enable-static --enable-shared 
--enable-threads --with-dwarf2 --disable-sjlj-exceptions --enable-version-specific-runtime-libs 
--enable-libstdcxx-d**

And Then i use CMake-GUI to configure and generate the Makefile;

In CMake GUI, the generater is set to MinGW Makefiles, and use default native compilers.
The configure and generate step goes well EXCEPT not found CUDA.

Configure log info:
**The C compiler identification is GNU 4.9.3
The CXX compiler identification is GNU 4.9.3
Check for working C compiler: C:/MinGW/bin/gcc.exe
Check for working C compiler: C:/MinGW/bin/gcc.exe -- works
Detecting C compiler ABI info
Detecting C compiler ABI info - done
Detecting C compile features
Detecting C compile features - done
Check for working CXX compiler: C:/MinGW/bin/g++.exe
Check for working CXX compiler: C:/MinGW/bin/g++.exe -- works
Detecting CXX compiler ABI info
Detecting CXX compiler ABI info - done
Detecting CXX compile features
Detecting CXX compile features - done
C++11 activated.
CUDA_TOOLKIT_ROOT_DIR not found or specified
Could NOT find CUDA (missing:  CUDA_TOOLKIT_ROOT_DIR CUDA_NVCC_EXECUTABLE CUDA_INCLUDE_DIRS CUDA_CUDART_LIBRARY) (Required is at least version "7.5")
*** cuDNN V5.0 OR GREATER NOT FOUND.  DLIB WILL NOT USE CUDA. ***
[...]

PR #282, opened 2016-10-10
Fix for #281

e-fominov:
Added `#include <thread>`. This can help to exclude possible misunderstanding of MinGW x86 comiplation problems in a future


davisking:
Thanks for the PR :)

Issue #280, opened 2016-10-10
shape_predictor.h missing parallel_for_extension.h include


semeyerz:
#126 from @e-fominov introduced threaded computation for shape prediction.

Since then the build of my project using dlib fails with:
`include/dlib/image_processing/shape_predictor.h:805:13: error: ‘thread_pool’ has not been declared`

adding parallel_for_extension.h include to shape_predictor fixes this.

davisking:
Thanks for reporting this.  It should be fixed now.

PR #283, opened 2016-10-10
Added threads.h into shape_predictor.h

e-fominov:
fix for #280 . train_shape_predictor_ex example compiles well even without this PR. this PR just clarifies the dependancy


davisking:
I was just about to push this :)  lol, thanks for another PR :)

PR #291, opened 2016-10-14
optimization - poly_min_extrap fixed for a specific case

mateuszbaran:
If temp is nonzero but a subnormal then temp2 matrix may contain infinities which may cause temp assignment in line 225 to be a NaN (as a difference of two positive infinities).


davisking:
Super, thanks for the PR :)

PR #292, opened 2016-10-14
pass CUDA_HOST_COMPILER variable to cuda cmake tests

hannometer:
In Arch Linux I had the problem that the default gcc (v6.2.1) was too new and incompatible with CUDA 8. There is an official gcc5 package, which is compatible with cuda. But the host compiler flag was not passed to the cmake cuda tests, so it always reported, that the cuda tests could not be compiled.


davisking:
Cool, thanks for the PR.

Issue #118, opened 2016-05-21
Travis builds


AbdealiJK:
It would be nice to set up some sort of continuous integration tests like Travis, CircleCI, etc.

I couldn't find any CI services being used. If there is one, please put it in the docs (http://dlib.net/other.html#dlib_testing_suite) or readme so that it is easy to find

davisking:
I run my own server that tests dlib automatically.

PR #294, opened 2016-10-14
Travis CI Integration scripts

e-fominov:
This PR introduces minimal Travis-CI integration scripts. This was earlier discussed here #118 and didn't got enough attention
I am now using Travis with one other project and I really see that it is usefull to see if incoming PR breaks code or not
Travis will make automatic compilation of dtest (and execute it with --runall) and all examples (no running - build only) on each incoming PR and PR change. Author of PR will show integration result immediately with no need to as Davis to check his code.
This does not replace existing way to test dlib on existing servers, but can be usefull as an additional automatic CI tool visible to PR authors

How to integrate:
1. register on https://travis-ci.org/ with using github account (Free for open-source)
2. "My repositiories" -> press "+" -> select repository (press switch) and press "settings" 
3. enable "Build only if .travis.yml is present" and "Build pull requests"
4. If you want to check direct pushes to a master branch - enable  "Build pushes"  too
5. test with sample PR

After that every new PR and changes on existing PR will get additional testing results information
Testing time is about 15 minutes long

I can extend this scripts to make automatic python and tools test-building too
Cuda is not supported now, but i think its possible to test cuda + cudnn compilation too. Running CUDA tests will not be possible
Its possible to add Clang compiler testing too
May be - mingw/linux
Visual Studio - now not possible, but may be in a future
Code coverage measurement - possible as next steps

If you will want to remove it in a future - delete .travis.yml file and dlib/travis directory

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,50 @@
+language: cpp
+
+packages: &gcc5_pkgs
+  - gcc-5
+  - g++-5
+  - libstdc++6
+
+matrix:
+  include:
+    # build and run all tests
+    - compiler: gcc
+      env:
+        - GCC_VER=5
+        - ADDRESS_MODEL=64
+        - BUILD_SYSTEM=cmake
+        - PATH=$HOME/bin:$PATH
+        - VARIANT=test
+      addons: &ao_gcc5
+        apt:
+          sources: 
+            - ubuntu-toolchain-r-test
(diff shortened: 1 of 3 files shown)


davisking:
Sweet.  I'll try it out.  Does github automatically display the rest of testing on a PR or do we need to add something to github?

PR #298, opened 2016-10-16
removed compiler warnings

odedharth:
warnings that appear on xcode


dlib/sockets/sockets_kernel_2.cpp
```diff
@@ -461,7 +461,7 @@ namespace dlib
         // setup a timeval structure
         timeval time_to_wait;
         time_to_wait.tv_sec = static_cast<long>(timeout/1000);
-        time_to_wait.tv_usec = static_cast<long>((timeout%1000)*1000);
+        time_to_wait.tv_usec = (int)static_cast<long>((timeout%1000)*1000);
```
davisking: Come one :)

That's not a reasonable way to change this.  Also, the type of tv_usec isn't int, it's suseconds_t.

PR #324, opened 2016-11-09
Add a dnn_mmod_ex_test file to easilty test a dnn

thaiat:


davisking:
Thanks for another PR.

However, there is already an example program for this: http://dlib.net/dnn_mmod_face_detection_ex.cpp.html.  Also, this seems like more of a tool than an example.  Everything in the examples folder should be thought of as essays that introduce or explain some aspect of dlib useage.  So their primary goals need to be about being educational, easy to read, and so forth.  That is somewhat at odds with making useful programs in their own right, which in general belong in some other location like the tools subfolder, or more often in a different repository dedicated to a specific application rather than in the main dlib repo.

PR #327, opened 2016-11-11
Modify to support both train-images-idx3-ubyte and train-images.idx3-ubyte styles of the filename.

MambaWong:


davisking:
Thanks for the PR.  

However, if you run the command `gunzip train-images-idx3-ubyte.gz` you will get `train-images-idx3-ubyte` as output.  Not `train-images.idx3-ubyte`.  Although some decompression programs other than gunzip do produce `train-images.idx3-ubyte`.  But most users use gunzip, so the current names are the preferred ones.

davisking:
Although, if you want to change the PR so that it looks for both styles of filename that would be good :)

davisking:

PR #328, opened 2016-11-11
Expose python binding for running multiple detectors at once

thaiat:
This exposes a static method `run_multiple` on `dlib.fhog_object_detector` that accepts a list (python list) of detectors

The results have the same form as for the `run` method

Example:
```python
detector_1= dlib.fhog_object_detector("object_detector1.svm")
detector2 = dlib.fhog_object_detector("object_detector2.svm")
detectors = [detector1, detector2]
retval = dlib.fhog_object_detector.run_multiple(
    detectors, 
    image, 
    upsample_num_times=0,  
    adjust_threshold=0.0)
```

### DIFF
--- a/tools/python/src/simple_object_detector_py.h
+++ b/tools/python/src/simple_object_detector_py.h
@@ -12,8 +12,7 @@
 namespace dlib
 {
     typedef object_detector<scan_fhog_pyramid<pyramid_down<6> > > simple_object_detector;
-    typedef scan_fhog_pyramid<pyramid_down<6> > image_scanner_type;
-    
+
     inline void split_rect_detections (
         std::vector<rect_detection>& rect_detections,
         std::vector<rectangle>& rectangles,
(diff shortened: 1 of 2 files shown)


tools/python/src/simple_object_detector_py.h
```diff
@@ -12,7 +12,8 @@
 namespace dlib
 {
     typedef object_detector<scan_fhog_pyramid<pyramid_down<6> > > simple_object_detector;
-
+    typedef scan_fhog_pyramid<pyramid_down<6> > image_scanner_type;
```
davisking: I think you should be able to use the simple_object_detector everywhere here, so this new typedef isn't needed.  It would also make it clear that all these pieces of code must operate on the same type, simple_object_detector, which I'm pretty sure is the case.

davisking:
Awesome, thanks for the PR :)

PR #334, opened 2016-11-14
Remove initializer lists support for Apple/libstdc < 4.02

caseymcc:
Remove initializer lists support for Apple (libstdc++ version used does not include std::initializer_list). When compiling on Mac an older version of libstdc++ can be linked in that does not included std::initializer_list. Same issue can be seen with Qt (https://git.merproject.org/faenil/qtbase/commit/387f75c39bfad4439a6bd013089c2cd216cda5ea), basically the same fix.

### DIFF
--- a/dlib/assert.h
+++ b/dlib/assert.h
@@ -47,6 +47,10 @@
 #   define DLIB_HAS_INITIALIZER_LISTS
 #endif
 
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the PR :)

Issue #288, opened 2016-10-12
Semantic Segmentation Functionality


davidmascharka:
I'm interested in using dlib for semantic segmentation. I think the only necessary features would be:
- Loss function - it doesn't look like loss_multiclass_log would work for this
- "Unpooling" upsampling layer - there are a few ways to do this

Is this something you're interested in supporting? I'm planning to add this functionality regardless, just wanted to check in. 

Regarding implementation, would the upsampling be something you would want added to the pooling class or its own upsample class?

davisking:
Those are definitely things we want in dlib.  As you say, you just need a new loss layer and some kind of upsampling layer.  Definitely create new layer classes rather than adding more stuff into existing classes.

qmaruf:
I want to contribute. From where should I start?

davisking:
You mean you want to implement the things @davidmascharka asked about specifically?  Or something else?

qmaruf:
Yup. The feature @davidmascharka asked for.

davisking:
I would start by implementing an upsampling layer.

reunanen:
Should [`loss_mean_squared_multioutput`](https://github.com/davisking/dlib/blob/master/dlib/dnn/loss.h#L1225-L1340) basically work for binary semantic segmentation tasks? I do understand that it wouldn't be an ideal choice for classification – but other than that?

I too really want to do semantic segmentation using dlib. In fact I am almost desperate: I have already created a network that can classify individual image patches of size 28x28 or so. I can then convert any image to a list of patches, and classify each patch. This more or less gives the desired result, however it is terribly slow of course. But as said, I am desperate...

Any improvement over the above approach would be great for me. I would be willing to contribute also, however I'm not quite on a level where I can just start typing. If the stars were aligned however, I might be able to take `loss_multiclass_log` and `loss_mean_squared_multioutput`, and combine these two to a `loss_multiclass_log_multioutput` (or something like that), if that makes sense? (I am looking at the loss layer mainly because it looks like there's already some significant progress on #336.)

davisking:
You could do that. But I would code up some loss from a recent paper that got good results. I doubt MSE is what gives the best results in recent papers.

davidmascharka:
I've had good results in segmentation from the mean of a pixelwise multiclass log loss. That might be a way to start -- adapting [`loss_multiclass_log`](https://github.com/davisking/dlib/blob/master/dlib/dnn/loss.h#L245) to take the mean over pixels.

[...]

PR #336, opened 2016-11-16
Add upsampling functionality

davidmascharka:
Addresses half of Issue #288. I copied the `pooling` class to base this on and modified that. 

A couple considerations: 
- I'm not qualified to provide a GPU implementation, so that's not currently an option. What is the best way to handle this?
- I copied the idea behind the pooling class' `does_max_pooling` to allow easy modifications to add other upsampling techniques. I plan to add a "sparse" upsampling where the index of the max-pool value is set to the correct input value as in SegNet. This looks a little odd right now, but that was my motivation behind leaving it there.

Very small example using the class:

```
#include <dlib/dnn.h>
#include <iostream>

using namespace std;
using namespace dlib;

int main(int argc, char** argv) try
{
    matrix<float,2,2> sample_data;
    sample_data = 1.0, 2.0,
                  3.0, 4.0;
    block_upsample<2,3,input<matrix<float>>> upsample_tester;
    cout << sample_data << endl;
    cout << mat(upsample_tester(sample_data), 4, 6);
}
catch(std::exception& e)
{
    cout << e.what() << endl;
}
```

Running this should give the output:

```
1 2
3 4

1 1 1 2 2 2
1 1 1 2 2 2
3 3 3 4 4 4
3 3 3 4 4 4
```

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -711,7 +711,6 @@ namespace dlib
         static_assert(_repeat_c >= 1, "The number of columns to repeat must be >= 1");
     public:
 
-
         block_upsample_(
         )
         {}
@@ -737,46 +736,21 @@ namespace dlib
             return p;
         }
 
-        block_upsample_ (
-            const block_upsample_& item
-        )
-        {
-            // this->bu is non-copyable so we have to write our own copy to avoid trying to
-            // copy it and getting an error.
-        }
-
-        block_upsample_& operator= (
-            const block_upsample_& item
-        )
-        {
-            if (this == &item)
-                return *this;
-
-            // this->bu is non-copyable so we have to write our own copy to avoid trying to
-            // copy it and getting an error.
-            return *this;
-        }
-
         template <typename SUBNET>
         void setup (const SUBNET& /*sub*/)
         {
+            bu.setup_block_upsampling(_repeat_r, _repeat_c);
         }
 
         template <typename SUBNET>
         void forward(const SUBNET& sub, resizable_tensor& output)
         {
-            bu.setup_block_upsampling(_repeat_r!=0?_repeat_r:sub.get_output().nr(), 
-                                    _repeat_c!=0?_repeat_c:sub.get_output().nc());
-
             bu(output, sub.get_output());
         } 
 
         template <typename SUBNET>
         void backward(const tensor& computed_output, const tensor& gradient_input, SUBNET& sub, tensor& /*params_grad*/)
         {
-            bu.setup_block_upsampling(_repeat_r!=0?_repeat_r:sub.get_output().nr(), 
-                                    _repeat_c!=0?_repeat_c:sub.get_output().nc());
-
             bu.get_gradient(gradient_input, computed_output, sub.get_output(), sub.get_gradient_input());
         }
 
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -770,7 +770,7 @@ namespace dlib
             deserialize(version, in);
             long repeat_r;
             long repeat_c;
-            if (version == "max_pool_1")
+            if (version == "block_upsample_1")
             {
                 deserialize(repeat_r, in);
                 deserialize(repeat_c, in);
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -701,6 +701,149 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <
+        long _repeat_r,
+        long _repeat_c
+        >
+    class block_upsample_
+    {
+        static_assert(_repeat_r >= 1, "The number of rows to repeat must be >= 1");
+        static_assert(_repeat_c >= 1, "The number of columns to repeat must be >= 1");
+    public:
+
+
+        block_upsample_(
+        )
+        {}
+
+        long repeat_r() const { return _repeat_r; }
+        long repeat_c() const { return _repeat_c; }
+
+        inline point map_input_to_output (
+            point p
+        ) const
+        {
+            p.x() = p.x()*repeat_c();
+            p.y() = p.y()*repeat_r();
+            return p;
+        }
+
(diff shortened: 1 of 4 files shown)


dlib/dnn/tensor_tools.h
```diff
@@ -978,6 +978,90 @@ namespace dlib { namespace tt
...
+        ) { impl(dest, src); }
+        /*!
+            requires
+                - is_same_object(dest,src) == false
+                - setup_block_upsampling() has been called.
+            ensures
+                - #dest.num_samples() == src.num_samples()
+                - #dest.k() == src.k()
+                - #dest.nr() == src.nr() * window_height
+                - #dest.nc() == src.nc() * window_width
```
davisking: These comments aren't right.  E.g. what is window_height?

dlib/dnn/layers.h
```diff
@@ -701,6 +701,149 @@ namespace dlib
...
+            p.x() = p.x()/repeat_c();
+            p.y() = p.y()/repeat_r();
+            return p;
+        }
+
+        block_upsample_ (
+            const block_upsample_& item
+        )
+        {
+            // this->bu is non-copyable so we have to write our own copy to avoid trying to
```
davidmascharka: Should I implement a copy procedue here or let the compiler generate one if somebody calls it in code?

davisking:
It looked like none of your code required a user defined copy constructor.

davidmascharka:
Thanks for the check. I'm going to work on adding some testing for the layer and getting famililar enough with CUDA to competently implement the GPU version. Should I leave this open as a staging area or close and resubmit when it's implemented?

davisking:
Cool, no problem.  You can leave the PR open until you are ready.

You should also look at the CUDA code in dlib/dnn/cuda_dlib.cu.  There are
a lot of examples of good CUDA coding patterns.  In particular, the
`grid_stride_range` makes writing parallel for loops very clean and easy in
CUDA.  A combination of `grid_stride_range` and `grid_stride_range_y` would
make for a very straight forward implementation of block based upsampling.
You should also read this nvidia post about grid stride loops which
explains a lot of the relevant background information:
https://devblogs.nvidia.com/parallelforall/cuda-pro-tip-write-flexible-kernels-grid-stride-loops/.
  Beyond that, Professional CUDA C Programming is a great book on CUDA.

andremira:
Hi davidmascharka,

In the layers.h there is a small issue with the name that you use for deserialize and serialize.

davidmascharka:
@andremira you're absolutely right. Thanks!

OranjeeGeneral:
Can anybody say what's the status with this? As I was looking into writing my own upsampling/deconvolution layer potentially as I need it to for building Fully Convoluted Networks in dlib, but if this pull request is close to happen I might abandon it.

davidmascharka:
@OranjeeGeneral I'm working on a CUDA implementation of the upsampling layer. A deconv would be worth adding if you wanted to focus there. That would definitely help me out as well.

decrispell:
@OranjeeGeneral  @davidmascharka  FYI @elda27 also expressed interest in implementing a transpose convolution / deconv layer in issue #372

OranjeeGeneral:
@davidmascharka 

Are you sure your backwards pass is correct? 

Shouldn't it basically down sample can't really see that in the code? 

I am just asking as I am thinking about embarking on an upsample code of my own which is a bit more flexible then yours (so it allows nearest neighbour, as well as linear filtering or filling the gap with zeros)

davidmascharka:
@OranjeeGeneral I believe it is. I'm adding up the gradient at each of the pixels that got upsampled. If

```
a | b
c | d
```
transforms to 
```
a | a | b | b
a | a | b | b
c | c | d | d
c | c | d | d
```
then the gradient coming back along each of these would be
```
partial a/partial f | partial a/partial f | partial b/partial f | partial b/partial f
partial a/partial f | partial a/partial f | partial b/partial f | partial b/partial f
partial c/partial f | partial c/partial f | partial d/partial f | partial d/partial f
partial c/partial f | partial c/partial f | partial d/partial f | partial d/partial f
```
and you would end up with (at the front of the upsampling layer):
```
4(partial a/partial f) | 4(partial b/partial f)
4(partial c/partial f) | 4(partial d/partial f)
```
Since the upsample layer contributed to the error along each of the upsampled pixels (because it controlled what went there) it gets the full gradient of all the pixels in each upsampled block.

I'd be excited to see upsampling finally make it in (haven't had time to dive into the CUDA code yet), so I'll definitely be following along.

OranjeeGeneral:
Thanks makes sense must have read the loop wrong. 

Well I think I've got forward pass running on my upsampler for both CPU/CUDA allowing for zero filing, nearest neighbor (which is similar to yours) and bilinear. 

I am just wondering if for backward I should have a reverse weighting of the gradients or not.

davisking:
I'm closing this PR since a transpose/deconvoltion layer (from this PR https://github.com/davisking/dlib/pull/476) is now part of dlib.

OranjeeGeneral:
Cool thanks for the merge :)

But that's not quiet the same I am afraid. 

It would be good to have a basic upsample layer as well as it differs in its behaviour to transpose convolution they do produce different results and in different use cases sometimes the one is better than the other.

If you like I can submit a pull request I've got an upsample layer (with both CUDA/CPU) with different upsampling strategies

davisking:
You could just set the cont_ layer to have a 0 learning rate to create a static upsampling layer (with an appropriate parameter selection).  Although that's a little bit cumbersome from a usability point of view.  

I get the impression this PR isn't being worked on anymore.  But if you want to submit an upsampling layer then that would be great :)

davidmascharka:
That's great! It looks like #476 could solve what I had originally wanted to do with this. A little hacky to get this exact functionality but closing this is totally reasonable.

davisking:
Yeah, it's kinda hacky.  Really if one of you guys wants to make a PR with a dedicated layer for this that's probably better (or finish this one) :)

PR #337, opened 2016-11-18
Fix misleading-indentation warnings for unicode.h

stefan-it:
This PR fixes > 100 misleading-indentation warnings when compiling all test cases on GCC 6.2 for the `unicode.h` implementation.

E.g. the following code:

```cpp
if ( ch < 0x610 )return false;if ( ch < 0x616 )return true;
```
raises a misleading-indentation warning. I guess this PR needs a proper to not introduce a regression, as many lines are changed!

### DIFF
--- a/dlib/unicode/unicode.h
+++ b/dlib/unicode/unicode.h
@@ -19,7 +19,7 @@ namespace dlib
     typedef uint32 unichar;
 
 #if defined(__GNUC__) && __GNUC__ < 4 && __GNUC_MINOR__ < 4
-    struct unichar_traits 
+    struct unichar_traits
     {
         typedef dlib::unichar    char_type;
         typedef dlib::unichar    int_type;
@@ -30,7 +30,7 @@ namespace dlib
         static void assign(char_type& c1, const char_type& c2) { c1 = c2; }
         static bool eq(const char_type& c1, const char_type& c2) { return c1 == c2; }
         static bool lt(const char_type& c1, const char_type& c2) { return c1 < c2; }
-        static int compare(const char_type* s1, const char_type* s2, size_t n) 
+        static int compare(const char_type* s1, const char_type* s2, size_t n)
         {
             for (size_t i = 0; i < n; ++i)
             {
@@ -85,7 +85,7 @@ namespace dlib
         }
 
 
-        static int_type not_eof(const int_type& c) 
+        static int_type not_eof(const int_type& c)
         {
             if (!eq_int_type(c,eof()))
                 return to_int_type(c);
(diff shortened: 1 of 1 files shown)


davisking:
What tool is giving you warnings about this code?

stefan-it:
E.g.:

```bash
[  1%] Built target dlib_all_source_cpp
[ 39%] Built target dlib
[ 40%] Building CXX object CMakeFiles/dtest.dir/tester.cpp.o
[ 40%] Building CXX object CMakeFiles/dtest.dir/main.cpp.o
[ 41%] Building CXX object CMakeFiles/dtest.dir/example.cpp.o
[ 41%] Building CXX object CMakeFiles/dtest.dir/active_learning.cpp.o
In file included from /mnt/Repositories/dlib/dlib/../dlib/map/../unicode.h:6:0,
                 from /mnt/Repositories/dlib/dlib/../dlib/map/../serialize.h:160,
                 from /mnt/Repositories/dlib/dlib/../dlib/map/map_kernel_1.h:11,
                 from /mnt/Repositories/dlib/dlib/../dlib/map.h:6,
                 from /mnt/Repositories/dlib/dlib/test/tester.h:8,
                 from /mnt/Repositories/dlib/dlib/test/tester.cpp:5:
/mnt/Repositories/dlib/dlib/../dlib/map/../unicode/unicode.h: In function 'bool dlib::is_combining_char(T)':
/mnt/Repositories/dlib/dlib/../dlib/map/../unicode/unicode.h:304:13: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
             if ( ch < 0x483 )return false;if ( ch < 0x48A )return true;
             ^~
/mnt/Repositories/dlib/dlib/../dlib/map/../unicode/unicode.h:304:43: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the 'if'
             if ( ch < 0x483 )return false;if ( ch < 0x48A )return true;
```

The Travis CI uses GCC 5. For GCC 6 the `-Wmisleading-indentation` was added to `-Wall`, see [here](https://gcc.gnu.org/gcc-6/porting_to.html) - so that is the reason, why we don't see these kind of messages in the CI builds :)

davisking:
Hmm, ok.  I'm not really sure this PR improves the readability of the code though.  There are patterns in the processing that are obvious in the current layout but become hidden in the regular indenting.  Also, don't you get this warning from other parts of dlib?  It seems like a pretty aggressive warning.

stefan-it:
Yeah, it doesn't improve the code readability :( 

On the other side I redirected all build messages + warnings to a file and grepped for the "misleading" warnings: these kind of messages appeared 19880 times (!) on the build output  - so when using GCC >= 6 it will be much harder to identify "normal" warnings.

davisking:
We should probably just disable this warning with a #pragma in the code.  It wouldn't be the first time I've disabled a bad warning in such a way.  So if you want to make such a PR that would be cool, but I don't want to do something that damages code readability, especially when it makes big rifts in source control history.

PR #338, opened 2016-11-19
unicode: Disable misleading-indentation warnings with GCC pragma

stefan-it:
This PR introduces a GCC pragma for disabling misleading-indentation warnings in the `is_combining_char()` method in `unicode.h`.

On GCC >= 6 the `-Wmisleading-indentation` option was added to `-Wall` and this would cause about 19880 warnings in `unicode.h`.

As discussed in #337 it is ok to disable these warnings via a special GCC `#pragma`.

### DIFF
--- a/dlib/unicode/unicode.h
+++ b/dlib/unicode/unicode.h
@@ -289,7 +289,10 @@ namespace dlib
     }
 
 // ----------------------------------------------------------------------------------------
-
+#if defined(__GNUC__) && __GNUC__ >= 6
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmisleading-indentation"
+#endif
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.

You need to put a guard around the pragma statements to make sure they are only applied to GCC since other compilers will complain or error out on them.  For example, like this: https://github.com/davisking/dlib/blob/master/dlib/matrix/matrix_assign_fwd.h#L8

stefan-it:
I force pushed a guard (GCC >= 6) around the pragma statements, I hope it's ok now :)

Btw.: I tested the compilation on `clang` *3.9* and these kind of `misleading-indentation` didn't appear.

davisking:

PR #342, opened 2016-11-22
Add MKL DFTI FFT bindings.

dunkpa:
Hi Davis,

I recently added bindings to the MKL DFTI library for FFT computation. I'm not 
sure if you want to add MKL specifics to DLIB, but if you're happy to do so, 
then I hope these will be useful. Using DFTI speeds things up for me when using
the correlation_tracker.

I've tested using the version of MKL that comes with Intel Parallel Studio 2017, 
on Ubuntu 16.04.

Tests all pass when built with either -DDLIB_USE_MKL_FFT=ON or -DDLIB_USE_MKL_FFT=OFF;

    (p) pal181@prodata-ub-ph:build (dunka/master) $ ./dtest --test_fft

    Testing Finished
    Total number of individual testing statements executed: 7732
    All tests completed successfully

Regards,
Duncan Palmer

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -451,8 +451,8 @@ if (NOT TARGET dlib)
       endif()
 
 
-      if (DLIB_USE_BLAS OR DLIB_USE_LAPACK)
-         # Try to find BLAS and LAPACK 
+      if (DLIB_USE_BLAS OR DLIB_USE_LAPACK OR DLIB_USE_MKL_FFT)
+          # Try to find BLAS, LAPACK and MKL
          include(cmake_utils/cmake_find_blas.txt)
 
          if (DLIB_USE_BLAS)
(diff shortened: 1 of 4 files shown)


davisking:
Cool, that's great.  Shouldn't it require use of BLAS the way you have set it up in cmake?  I don't see any new cmake code for linking (nothing sets dlib_needed_libraries).  So it seems like the cmake code that enables this should be inside the if statement about using BLAS.  

Other than that this looks super though :)

dunkpa:
Looks like I cut a corner there..

I've now made the MKL_FFT and BLAS configurations independent. A side-effect of the change is that MKL won't be used for BLAS of LAPACK unless mkl_version.h is found in addition to the MKL libraries. This is because I require the MKL headers, and am making the assumption that if MKL development libraries are present on the system, then the headers should also be there.

davisking:
You can't assume the MKL headers are available.  There are plenty of times I've seen the MKL without the headers around.  So don't modify the existing MKL BLAS/LAPACK behavior.

dunkpa:
Oh - that's a bit anti-social.. My latest commit removes this assumption.

davisking:
Sweet.  Thanks for the PR :)

Issue #152, opened 2016-07-03
[Feature Request] Mean Square Loss Function 


kestrelm:
Hello,

It will be great if Mean Square Loss was part of your default library (For Linear Regression).
I know you can write your own loss layers but this is a very commonly used feature for DNN so it will be fantastic if it could be included by default.

Can you give a brief description of how the gradients are calculated in your API for the loss layer class?
For Mean Square Loss, I am assuming the local gradient should be calculated something like this:
https://github.com/nyanp/tiny-cnn/blob/master/tiny_cnn/lossfunctions/loss_function.h (under mse)

How do we go about implementing Mean Square Loss ourselves in your API? I read your docs about it but it wasn't clear about the gradient calculation portion.

Thanks!

davisking:
Indeed, that's something that will be added before the next release.  

The instructions for adding more loss layers are here: http://dlib.net/dlib/dnn/loss_abstract.h.html#EXAMPLE_LOSS_LAYER_

dennisfrancis:
This my first time here. If nobody else is working on this, I would like to give this a shot. Thanks.

davisking:
Go for it.  I've gotten side tracked into other things so I'm not adding it.  It's easy to do though.

dennisfrancis:
Hi @davisking, I created a loss_mean_squared loss layer in my forked repo and tried running a very simple one variable linear regression ( y = mx + c ) using it (see https://gist.github.com/dennisfrancis/444dd3e918f61f898f9d1551eeb8cde0 ). Noticed that the slope coefficient is calculated correctly, but not the intercept term. I am using a fc layer before the loss layer which is supposed to add bias terms as I understand from the documentation. 

I also tried using fc_no_bias instead with manual insertion of 1's in the input data to act as bias terms, but did not work. Could you please give me some hint on what I may be missing ? Thanks !

davisking:
I don't know.  Maybe you aren't seeing what you expect because of weight decay, because the gradients are wrong, or some other thing.  I'm too busy right now to look over your repository, test your code, and figure it out :)

dennisfrancis:
Ok, I figured that out. What I was missing was the learning rate multiplier for bias needs to be tuned separately. I will soon give a pull request which implements the mse loss layer with test cases.

PR #344, opened 2016-11-23
feature_addition : Added a mean squared loss layer to DNN with test case

dennisfrancis:
Added mean squared loss layer "loss_mean_squared" to
DNN as requested in https://github.com/davisking/dlib/issues/152
Also added test case of a simple linear regression with one variable
that uses this layer.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1305,70 +1305,70 @@ namespace dlib
             typename SUB_TYPE,
             typename label_iterator
             >
-            void to_label (
-                const tensor& input_tensor,
-                const SUB_TYPE& sub,
-                label_iterator iter
-                ) const
+        void to_label (
+            const tensor& input_tensor,
+            const SUB_TYPE& sub,
+            label_iterator iter
+        ) const
+        {
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+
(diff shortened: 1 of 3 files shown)


davisking:
Awesome, looks good :)

Some parts have inconsistent indenting due to the presence of tab characters.  Can you fix the code so the indenting is uniform?  Then the PR will be ready to merge.

dennisfrancis:
Thanks for pointing out, I have made the change.

dlib/dnn/loss.h
```diff
@@ -1294,6 +1294,113 @@ namespace dlib
...
+                             output_tensor.nc() == 1 &&
+                             output_tensor.k() == 1);
+                DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+
+                const float* out_data = output_tensor.host();
+                for (long i = 0; i < output_tensor.num_samples(); ++i)
+                {
+                    *iter++ = out_data[i];
+                }
+            }
```
davisking: Tabbing is still weird :)

dennisfrancis: hmm, looks like my emacs settings indents templates in a particular way. I have manually adapted to dlib style. Hopefully I fixed all of them :)

davisking:

Issue #363, opened 2016-12-19
The "simple CUDA program test" fails on Visual Studio when run from another CMake script


reunanen:
Just learned that dlib exists. It looks amazing!

I'm trying to build [today's version](https://github.com/davisking/dlib/commit/bf36a846de88ba39b82b36db6d59910aebbe4910) on Visual Studio 2015 Update 3 on Windows 10 Pro.

The basic build process fails to compile the simple CUDA program:

```
D:\dlib\build>cmake -G "Visual Studio 14 2015 Win64" ..
-- The C compiler identification is MSVC 19.0.24215.1
-- The CXX compiler identification is MSVC 19.0.24215.1
-- Check for working C compiler: D:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/x86_amd64/cl.exe
-- Check for working C compiler: D:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/x86_amd64/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: D:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/x86_amd64/cl.exe
-- Check for working CXX compiler: D:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/x86_amd64/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- C++11 activated.
-- Searching for BLAS and LAPACK
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of void*
-- Check size of void* - done
-- Found CUDA: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v8.0 (found suitable version "8.0", minimum required is "7.5")
-- Looking for cuDNN install...
-- Building a CUDA test project to see if your compiler is compatible with CUDA...
-- *** CUDA was found but your compiler failed to compile a simple CUDA program so dlib isn't going to use CUDA. ***
-- *** cuDNN V5.0 OR GREATER NOT FOUND.  DLIB WILL NOT USE CUDA. ***
-- *** If you have cuDNN then set CMAKE_PREFIX_PATH to include cuDNN's folder.
-- Configuring done
[...]

PR #364, opened 2016-12-19
Fix Visual Studio build by not setting CUDA_PROPAGATE_HOST_FLAGS to OFF

reunanen:
Problem: Visual Studio linker errors resulting from mixed /MT and /MD settings

Solution: don't set CUDA_PROPAGATE_HOST_FLAGS to OFF when using Visual Studio

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -510,7 +510,10 @@ if (NOT TARGET dlib)
             # -std=c++11 option if you let it propagate it to nvcc in some
             # cases.  So instead we disable this and manually include
             # things from CMAKE_CXX_FLAGS in the CUDA_NVCC_FLAGS list below.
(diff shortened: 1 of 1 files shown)


davisking:
Cool.  Does this fix this issue (https://github.com/davisking/dlib/issues/363) or is it only a partial fix?

reunanen:
No, this is another issue – completely orthogonal, I think.

davisking:
Ok, sounds good.  Thanks for the PR :)

davisking:
Does putting a similar if MSVC on the CUDA_HOST_COMPILER variable fix the other issue?

reunanen:
I think it should. I'll create a PR for that too.

Thanks for the amazing library! Will hopefully have some

davisking:
Thanks, glad you like dlib :)

PR #374, opened 2016-12-23
Added new test case for split method (unicode non-breaking space)

stefan-it:
Hi,

the current test cases for the `split` method use wide-strings, but a test case e.g. for unicode non-breaking space is missing.

So this PR introduces a new test case for a unicode non-breaking space delimiter and uses some (German) umlauts in the text to be splitted.


davisking:
Super, thanks for the PR :)

PR #376, opened 2016-12-26
Add ninja build commands to README.md

tzutalin:


davisking:
Thanks for the PR.  However, I don't want to clutter the readme with notes about different CMake generators.  There are a lot of them and users who want to use them can read about the available generators in the CMake documentation.

I also just compiled the dlib tests with ninja and it took the same time as regular make.

tzutalin:
Ok. Got it

PR #377, opened 2016-12-27
Modify setup.py so that it can be installed under conda environment

hiiwave:
Since `cmake` is not able to find Python-libraries under anaconda environment, 
`-DPYTHON_LIBRARY `and `-DPYTHON_INCLUDE_DIR` flags should be appended to `cmake` command with  respective folders.
I did this by appending them to `cmake_extra_arch`.


davisking:
Thanks for the PR.  It looks good.  Is there a reason not to add these paths for all python builds?  It seems like they might help locate the correct python files in other situations as well.  I haven't done any testing though.

hiiwave:
To add these paths for all python builds? How to do that? 
Actually I'm not familiar with python environment variables. I was just not able to install dlib today and finally found this solution workable. =P

davisking:
Ah, ok.  It looks like the code might work with other python builds.  But I
need to test it as I'm not familiar with the python environment variables
either.  In any case, thanks for the PR, I'll merge it and figure out if
the trick applies to other python builds later.

PR #386, opened 2017-01-02
Fixed issue with uninitialised variables in thread_pool_extension. 

rquk:
Fixed issue with uninitialised variables. There are 2 places where std::exception_ptr eptr is not initialised. One in the CPP and one in the header file of thread_pool_extension.

### DIFF
--- a/dlib/statistics/running_gradient.h
+++ b/dlib/statistics/running_gradient.h
@@ -165,7 +165,7 @@ namespace dlib
                     return 0.5;
             }
             value = (value-mean)/stddev;
-            return 0.5 * erfc(-value / std::sqrt(2.0));
+            return 0.5 * std::erfc(-value / std::sqrt(2.0));
         }
 
         double n;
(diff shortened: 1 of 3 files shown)


davisking:
Did you get warnings about this?  According to http://en.cppreference.com/w/cpp/error/exception_ptr 

> A default-constructed std::exception_ptr is a null pointer

rquk:
Hi,

 

According to that you are correct it does not need initialising. I did not have a compile error, but a runtime error which went away when I initialised it. The RadStudio compiler I’m using, uses Dinkumware for the STD library. It may be an old one which does not support C++11. 

 

It’s defined like this:

 

typedef void *exception_ptr;

 

So does need initialising. Your choice as to whether or not you include my changes.

 

Many thanks for your time.

 

Regards,

 

Richard

davisking:
Thanks that's good to know about.  We definitely want this PR, so thanks for it :)

PR #400, opened 2017-01-16
Fix imglab readme typo in cd path

frankiek3:


davisking:
Super, thanks for the PR :)

PR #402, opened 2017-01-18
Move serialized face detector data into dlib library from header

dmpetroff:
Move raw serialized data into separate *.cpp file which is linked into static and shared dlib libraries improving both compile and runtime performance of get_frontal_face_detector function.
Note: this only boosts face detector creation, not detection speed.


fenollp:
How is runtime perf better? What are your numbers?

dmpetroff:
@fenollp runtime performance of face detection is unaffected. Only get_frontal_face_detector itself executes faster.

davisking:
Thanks for the PR.  It's definitely faster for detector creation.  The only thing I'm wondering about is using base64 instead of many individual hex codes.  Have you tried using the original base64 method but just without compression?  I bet the speed is more or less the same since base64 decoding is so fast.  The program that generated the base64 code is the examples/file_to_code_ex.cpp which is straightforward to experiment with.

radioneko:
base64 is fast, most of the time is spend inside decompression. But performing base64→binary conversion in every call will make data copy & memory allocations that aren't needed at all.

I can modify file_to_code_ex.cpp to optionally produce files in "new" format (raw binary + const std::string + function that returns const std::string&) and leave current base64 + compress as default behavior.

davisking:
I'm just wondering about the resulting binary sizes.  Some users might complain about the size of their executables going up.  So if the difference in speed between binary/base64 is some small percent but the binary executable size is different by 2x or 3x then it's probably worth keeping it in base64 (but without compression).  

Don't submit a modified file_to_code_ex patch, I just mean you can use it to accomplish this :).  That program is supposed to be a little tutorial more than a full featured program (as is the case with all the examples).

dmpetroff:
Hmm, currently because of global string constant every binary linked with dlib will have its size increased by 157kb.
This can be easily fixed by getting rid of static std::string variable (done).
Was:
- dynamic and static dlib libraries are larger by 157kb
- binaries linked statically/dynamically with dlib will be larger by 157kb and will always have 157kb of `obj_detector_dat` copy in `obj_detector_s` at runtime
- call to `get_serialized_frontal_faces` costs nothing because it returns reference to static std::string

Now:
- dynamic and static dlib libraries are larger by 157kb
- object linked with static library will only have serialized face detector data included when `get_serialized_frontal_faces` is called from application (ie only when they use dlib's frontal face detector). No long-living extra copies of detector data are made
- each call to `get_serialized_frontal_faces` makes temporary copy of it.

davisking:
I'm still not sure about doing it exactly this way.  Why not keep everything the same as it is currently except remove the compression?  This way, only programs or libraries that #include the header file will have their sizes increased.  So neither the dlib shared or static library files will have this data in it.  Also, I'm likely to add more saved models like this to dlib in the future and I don't want to keep increasing the size the shared library file, which is what would happen if the data was in a .cpp file each time.

dmpetroff:
Shared library size is usually not a big problem, because on all modern platforms shared libs are mmap-ed so data that is never used will not be physically loaded into memory. Actually the same is true for executables: they are mmap-ed, not loaded.
Size of static library is also not an issue, because it stays on developer size and during link phase only models that are actually used are linked into resulting binary.

I'm not sure if base64 effort worth it, because long binary data still takes time to compile while runtime penalties from compiled-in blobs are really negligible. But the choice is definitely up to :)

P. S. also note that splitting base64 string is inefficient because multiple calls to `std::ostringstream::operator<<` causes string growing with several memory reallocations, so it's more efficient to ask compiler (preprocessor actually) to concatenate strings for us.

davisking:
I know.  I'm just anticipating complaints from users.  Hardly anyone cares about shared library size, but some do and some large fraction of those that do care are confused.  But some of them have legitimate reasons.

The reason the string is split in the code is to work around bugs in some compilers which would crash or hang for large binary blobs.  I'm not sure if that bug is still present in today's compilers, but it used to be an issue on some platforms.

pencoa:
I have read the dlib code for whole night. 
in the frontal_face_detector.h, there imports the object_detector.dat.
where is it come from?
does it have some correlation with the detector.svm from scan_fhog_pyramid?
but they have different suffix.

Issue #406, opened 2017-01-23
RGBA JPEG support


decrispell:
Is there any reason why 4-channel (rgba) images are guarded against in the jpeg loader? (link to source below)

https://github.com/davisking/dlib/blob/0f60ba365bce953ecc131c5d8a9d8429ca75c075/dlib/image_loader/jpeg_loader.cpp#L125-L126

There seems to be some debate as to whether an alpha channel is allowable within a valid JPEG image, but at least some jpeg libraries seem to handle it just fine regardless. (I am using the libjpeg-turbo that ships with Ubuntu 16.04).

If there is no good reason disallow it, I will submit a simple PR that enables it.

davisking:
Not as far as I am aware.  So if you want to submit a PR for it then that
would be super.  Please include a test image somewhere so I can see it
working though :)

decrispell:
After some more digging, it looks like libjpeg-turbo does in fact support reading RGBA pixel types, but the alpha channel is always guaranteed to be set to 255.  Writing RGBA JPEG images also seems to be supported, but I haven't been able to get that to work.

Relevant comments from libjpeg-turbo source:
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/a0b7de9a47358a19b31db447dbf02b488463eb7e/jpeglib.h#L231-L236

Neither the standard libjpeg included with dlib nor the most recent version appear to have any support for the RGBA color space whatsoever.

So, a symmetric write/read capability doesn't seem to be in cards at the moment.  Not failing when loading an image that has 4 planes still seems worthwhile, though.  At least helps me out, since I ran across a 4-plane image in an "in the wild" dataset that I needed to process.  Unfortunately, despite my best efforts, I have been unable to save my own 4 channel JPEG image and the one that I have I cannot distribute.

davisking:
No worries.   Also, you know PNG supports alpha channels as part of the PNG standard?  I use it all the time for that and it works great.

decrispell:
Yeah, I never would think to use JPEG to store an image with a transparency layer (lossy alpha channel sounds like a bad idea).  Not sure why or how the images in my dataset ended up coming back from libjpeg-turbo with 4 channels, but at least I can process them now - thanks!

davisking:
No problem :)

PR #409, opened 2017-01-24
add support for loading of RGBA JPEG images

decrispell:
Addresses the issue mentioned in #406 , i.e. dlib reports an error when libjpeg-turbo successfully loads a 4-plane RGBA JPEG image.


davisking:
Cool, thanks for the PR :)

Issue #397, opened 2017-01-13
Optimize dlib for POWER8 VSX


edelsohn:
Enable and optimize support for POWER8 VSX SIMD instructions on PPC64LE Linux to dlib/simd.

$$$$ Financial bounties available.  Any reasonable suggested value will be seriously considered.

I welcome contact / replies from developers in the dlib community who are interested to work on this project.

dmiller423:
@davisking / maintainers:

There are a few ways to accomplish this,  the best would be using gcc's vector code and allowing the compiler to handle the job.  This would work out of the box for at least 4 different base architectures.  They would not work with visual studio however, and the current sse intrinsics do.  

I could:
- Add vsx intrinsics with further preprocessor checks similar to SSE code there now.
- Split it into separate headers and implement the classes based on availability,  and the options there are many with one set of headers containing vector code or intrinsics.
- Use the header as definition only,  make cpp files with implementation that are built based on build time checks (runtime checks could be used, unnecessary though if compiler is already building a binary with support) 

  I would suggest using separate headers and the gcc vectors as it will have no impact on existing code,  and any new gcc with supported targets will see a nice boost.  

Please let me know what you prefer.

davisking:
I don't see how putting the code in cpp files could work out well since inlining is very important for the speed of these functions.  So a header only option is a must.  Separate headers for different architectures (for the code within the dlib/simd) folder is probably the cleanest way to go about it.   

Also, the most important benchmark for speed improvements is probably the fhog based object detector in dlib.  So that's what you should test to see if you are getting performance improvements. 

Anyway, I would welcome such a patch.  You will have to talk to @edelsohn to find out the $$$$ details :)

dmiller423:
Thanks for the quick response,  

I will go ahead with the inline'd headers then, 
i'll make sure to bench against the existing code with mentioned fhog obj. detector.

davisking:
Cool, no problem.   Do talk to @edelsohn about the price you want though.  He is the one offering the bounty :)

edelsohn:
I agree that inlining is important and the functions must be in headers, but I don't understand the objection to using GCC/LLVM vector operations.  In fact, SSE/AVX intrinsics (mmintrin.h) are implemented as vector operations, not direct SSE/AVX operations.

dmiller423:
I don't think there was any objection to it? 
Unless I missed it

edelsohn:
[...]

PR #413, opened 2017-01-26
POWER8 VSX Optimization

barkovv:
@davisking 
Resolves #397 
Code can be compiled with gcc or clang in similar to SSE way:
```
cd dlib/test
mkdir build
cd build
cmake ..  -DUSE_VSX_INSTRUCTIONS=1
cmake --build . --config Release
./dtest --runall
```

### Current problems:
- [x] "hash" test fails on either optimized or not optimized version (but for x86 everything is ok). I will fix it.
- [x] I don't know how measure profit of optimisation. Is  there some profiler tool or something else for that purpose in dlib? Any help is appreciated

PR #414, opened 2017-01-27
GCC/Clang compatible SIMD code ./dlib/simd/simd**_vec.h

dmiller423:


dlib/simd.h
```diff
@@ -5,7 +5,7 @@
 
 #include "simd/simd_check.h"
 
-#ifdef DLIB_HAVE_GCC_VECTOR
+#if 1 //def DLIB_HAVE_GCC_VECTOR
```
davisking: Did you mean to commit this?

dmiller423:
No sorry,  I actually uploaded a whole set of incorrect files and i'm fixing it now.
The product of having it on too many machines to test.

davisking:
No worries.

dmiller423:
Ok,  I think this is ready... sorry for the delay,  I ran into a few problems 🔢

davisking:
No worries, there is no rush :)

dlib/simd/simd4f.h
```diff
@@ -3,7 +3,6 @@
 #ifndef DLIB_sIMD4F_Hh_
 #define DLIB_sIMD4F_Hh_
 
-#include "simd_check.h"
```
davisking: Does this have to be removed?

dlib/simd/simd4f_vec.h
```diff
@@ -0,0 +1,104 @@
...
+        
+        
+        inline simd4f(float f) : x{f,f,f,f} { }
+        inline simd4f(float r0, float r1, float r2, float r3)
+             : x{r0,r1,r2,r3} { }
+
+        inline simd4f& operator=(const simd4f& v) { x = v.x; return *this; }
+        inline simd4f& operator=(const float& v) { *this = simd4f(v); return *this; }
+
+        inline vector float operator() () const { return x.v; }
```
davisking: This seems a little weird.  Why can't a conversion operator be used here like it is for the SSE versions of simd4f?  This would avoid having code like lhs(),rhs() later on that looks a little confusing.

dlib/simd/simd4i.h
```diff
@@ -3,7 +3,6 @@
 #ifndef DLIB_sIMD4I_Hh_
 #define DLIB_sIMD4I_Hh_
 
-#include "simd_check.h"
```
davisking: Does this need to be removed?

dlib/simd/simd8f.h
```diff
@@ -3,11 +3,9 @@
 #ifndef DLIB_sIMD8F_Hh_
 #define DLIB_sIMD8F_Hh_
 
-#include "simd_check.h"
```
davisking: Same question here.  People might include these files own their own so it would be good if they behaved reasonably in that case.

dlib/simd/simd8f_vec.h
```diff
@@ -0,0 +1,251 @@
+// Copyright (C) 2013  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_sIMD8F_Hh_
```
davisking: This whole file looks like a copy of simd8f.h.  Can't we just #include simd8f.h?

dlib/simd/simd8i_vec.h
```diff
@@ -0,0 +1,234 @@
+// Copyright (C) 2013  Davis E. King (davis@dlib.net)
```
davisking: This whole file looks like a copy of simd8i.h.  Can't we just #include simd8i.h?

dlib/simd/simd8f.h
dmiller423: I think the easiest course is for something to include dlib/simd.h,    it includes the checks and picks the appropriate version.   It makes the most sense,  and can be used on multiple architectures. I don't think it's going to hurt anything if it's included,  the ifdef in the header will keep it from being included twice.

dlib/simd/simd4f_vec.h
dmiller423: I will double check,  it may just be left over kludge from previous autovectorization code where returning a reference on a vector through a cast operator wasn't allowed.

dlib/simd/simd8f.h
davisking: I agree.  But I suspect there might be user code out there that already includes the original files.  So it's best to avoid possibly disruption.

dlib/simd/simd8f_vec.h
dmiller423: Yes,  and I wanted you to look at tell me what you wanted done with these files.
I can merge the 4i/4f_vec  into the regular 4i/4f  like   C/SSE/AVX code if you like.   I think it's cleaner separate if there is going to be SSE2/4,  AVX,  NEON, VSX.   There are a few things that could be done however.   

 Instead of adding per  extension all mingled,   they could be separated like the classes are.   and each area will have the all function definitions large defined area.

Each specific ARCH/ext  could use its own header like _vec

Or separate .cpp files,  (inlining can still be used,  it's applied at code generation level with optimizations IIRC)

The last option is the cleanest,  but would require some modification to the cmake build.





#elif   AVX
#elif  ....
}

dlib/simd/simd8f.h
dmiller423: Ok

dlib/simd/simd4f_vec.h
dmiller423: Ok,  they chose to use macros and the compiler doesn't seem to look past the initial type.
An explicit cast works when there is an overloaded cast operator of that type, but it's just as ugly.
I could make a macro LHS(lhs) ((vector T)lhs) and use  vec_macro(LHS,RHS) but I don't know how much better that is than  vec_macro(lhs(),rhs())  or vec_macro((vector T)lhs, (vector T)rhs).

   If you have any other ideas I'll gladly test a better implementation.

davisking: Na, if it doesn't work then doing lhs() is fine :)

dlib/simd/simd8f_vec.h
davisking: I would say keep the separate files, except for these that are copies of other files.  As for using .cpp files, that would be fine if all compilers performed link time optimization as well as GCC but I doubt that's the case.  And I don't want to require the use of CMake so anything that makes compilation easier for users is better.

On that note, if you want to split the current simd8 code into a C and SSE .h files that's fine with me.  Or leaving them alone and just #including the simd8 as needed is fine too.  My main concern is to avoid unnecessary code duplication.

dmiller423:
Hopefully this satisfies for merge,  If i find a way to make the macros accept my cast I will submit another pull at a later date.

dlib/simd/simd8f_vec.h
davisking: Ha, just saw your next change. :)   All in one file is fine too.  I think it's easy to read now.  There is the worry that it might become horrendous at some point in the future, but now it's good.  And it has the benefit at least that when new architectures are added it's difficult to accidentally forget to add some part of the simd*f API.

davisking:
Sweet, thanks for the PR :)

dmiller423:
Thank you for all the quick replies and review work 👍 
Can we close issue #397 ?

davisking:
No problem.

Yeah, we can close it as far as I'm concerned.  Is there anything else you had in mind @edelsohn?

edelsohn:
LGTM

edelsohn:
@davisking Could you add DLIB_HAVE_VSX and the auto configuration information to README.md?

davisking:
I don't have a machine to test this stuff on, but I assume DLIB_HAVE_VSX is automatically switched on when the appropriate gcc compiler optimization flag is enabled?  So nothing needs to be done as far as cmake is concerned.  Although we could add a cmake option.  @dmiller423 How were you enabling this?

dmiller423:
@davisking && @edelsohn :   Any reasonably new gcc and linux toolchain should be built for a target cpu,  there is no need to specify -maltivec or -mvsx (nor -m64).   If someone is cross compiling and builds a generic powerpc (multilib or not),  then it's their responsibility to make sure everything built adds flags they want to use.   Still they should use -march -mcpu -mtune  and not have to specify -maltivec -mvsx etc.

My code does a simple check for defines that the compiler outputs if it has support.
At that point if there is no support the binary the compiler builds probably won't work anyhow,  since gcc internally uses builtins for target arch/cpu/extensions.

I will make any changes you request though,  just let me know 👍

edelsohn:
I wanted to check because #413 referred to cmake with the explicit option.

dmiller423:
I don't want to comment on someone else code.

davisking:
The existing CMake settings are mainly there to make life easier for visual studio users.  With GCC enabling these things is automatic, but with visual studio it's a little funny and how you do it varies from one version of visual studio to another.  If it weren't for visual studio I wouldn't have added any of this stuff to cmake, since as @dmiller423  points out, gcc automatically does the right thing.

Issue #404, opened 2017-01-22
[Feature Request] Mean Square Loss Function for Multiple Output


orinocopl:
Hello,

There is support for Mean Square Loss in git version of dlib for further release ([Feature Request] Mean Square Loss Function #152). The main limitation of this solution is the number of outputs (only one is allowed) at this moment. The support of Multiple outputs is necessary for some tasks.

Thanks !

dennisfrancis:
Hi, 

I am working on this, will create a PR soon.
Thanks.

PR #426, opened 2017-02-04
feature_addition : Mean squared loss layer for multiple output (#404)

dennisfrancis:
* Generalized existing loss_mean_squared layer to support multiple outputs.
* Also updated the corresponding test case to test regression
  with multiple outputs.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1083,8 +1083,8 @@ namespace dlib
     {
     public:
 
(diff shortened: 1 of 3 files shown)


davisking:
Thanks for the PR.  It looks good.  But can you make it a separate loss class?  Many users will just want the scalar version and some are already using it, so not breaking their code is a good thing.

dennisfrancis:
Created a new pull request afresh.

PR #427, opened 2017-02-05
feature_addition : Mean squared loss layer for multiple output (#404)

dennisfrancis:
* Added separate loss_mean_squared_multioutput layer class to support multiple outputs.
* Also added a corresponding test case to test a single variable regression
  with multiple outputs.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1243,6 +1243,13 @@ namespace dlib
             DLIB_CASSERT(grad.nr() == 1 &&
                          grad.nc() == 1);
             DLIB_CASSERT(grad.k() == output_tensor.k());
+            const long k = output_tensor.k();
+            for (long idx = 0; idx < output_tensor.num_samples(); ++idx)
+            {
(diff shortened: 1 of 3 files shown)


dlib/dnn/loss.h
```diff
@@ -1186,6 +1186,118 @@ namespace dlib
...
+        }
+
+
+        template <
+            typename const_label_iterator,
+            typename SUBNET
+            >
+        double compute_loss_value_and_gradient (
+            const tensor& input_tensor,
+            const_label_iterator truth,
```
davisking: You need to have some assert statement to check that truth contains matrices of the correct dimension relative to the output tensor's size.

Also, the code assumes the truth matrices are column vectors.  This should be mentioned in the abstract documentation.

dennisfrancis: Yes, I missed that. Added in the latest commit. Thanks.

davisking:
Awesome, thanks for the PR :)

Issue #437, opened 2017-02-13
XQuartz and Anaconda Python building and library linking errors


kylemcdonald:
tl;dr: I fixed errors building dlib by disabling the GUI, and fixed errors linking to libmkl_rt.dylib by modifying the .so with `install_name_change`.

I'm using OSX 10.11 with XCode 8.2.1 and Anaconda. I have CUDA 8.0.55 and CuDNN 5.1.10.

```
$ brew cask install Caskroom/cask/xquartz
$ brew install boost boost-python
$ mkdir build; cd build; cmake .. ; cmake --build .
```

CUDA and CuDNN are successfully detected + built by `cmake ..`. But `cmake --build .` gives me an error pretty early on:

```
[  1%] Building CXX object dlib/CMakeFiles/dlib.dir/gui_widgets/fonts.cpp.o
In file included from /Users/kyle/dlib/dlib/gui_widgets/fonts.cpp:14:
/Users/kyle/dlib/dlib/gui_widgets/nativefont.h:313:21: error: use of undeclared identifier 'XAllocColor'
```

Instead of digging into this, I just disabled the GUI:

```
$ cmake .. -DDLIB_NO_GUI_SUPPORT:BOOL=ON
$ cmake --build .
```

And it build dlib successfully. Time to build Python support (specifying no GUI here too):

```
$ cd .. ; python setup.py install --yes DLIB_NO_GUI_SUPPORT
```

Success! But then...

```
$ python
kyle@kyle ~/D/L/dlib (master)> python
Python 2.7.12 |Anaconda 4.0.0 (x86_64)| (default, Jul  2 2016, 17:43:17) 
[GCC 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.11.00)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> import dlib
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/kyle/anaconda/lib/python2.7/site-packages/dlib-19.2.99-py2.7-macosx-10.6-x86_64.egg/dlib/__init__.py", line 1, in <module>
    from .dlib import *
[...]

PR #438, opened 2017-02-13
Remove anaconda from X11_INCLUDE_DIR

kylemcdonald:
This is one solution for getting dlib to build on Mac when both X11 and anaconda are installed. See #437 for more details. Tested on Mac with X11 and Anaconda installed.


davisking:
Awesome, that looks like a good way to solve the problem :)

anubhavshrimal:
The update didn't work on my mac.
As an alternative, I simply deleted the X11 folder in the `anaconda/include` folder before running `pip install dlib`.
It worked out for me.

davisking:
Huh.  You should figure out why the check added by this PR isn't excluding the anaconda files and submit a new PR ;)

PR #446, opened 2017-02-22
Remove extra semicolons

ageitgey:
Super tiny PR to remove some unneeded semicolons in the new python example that I noticed while testing it.

Feel free to close / reject this PR if you don't care.


davisking:

PR #455, opened 2017-02-27
Allow MINGW OpenBLAS usage

e-fominov:
This will allow linking to OpenBLAS when compiling with MinGW.
For example, on my system OpenBLAS is located in d:\lib directory. And compiling this way:
```
cmake .. -G"MinGW Makefiles" -DCMAKE_PREFIX_PATH=d:\lib
```
gives me dlib compiled with OpenBLAS

This PR only modified one line and everything works. Do you prefer moving MINGW into separate if-branch of CMakeLists.txt for clearer code?


davisking:

PR #456, opened 2017-02-27
Docs: readme.md added installation instructions

e-fominov:
Dlib can be easily used with CMake, I am adding some instructions how to install and use dlib. Hope this will reduce amount of such questions

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,38 @@
 
 Dlib is a modern C++ toolkit containing machine learning algorithms and tools for creating complex software in C++ to solve real world problems. See [http://dlib.net](http://dlib.net) for the main project documentation and API reference.
 
+Some common questions are answered in [FAQ](http://dlib.net/faq.html). The best way to get support is using [StackOverflow](http://stackoverflow.com/questions/tagged/dlib).
 
+## Installing dlib as a shared library into your system
+
+Dlib is designed to be compiled with [CMake](https://cmake.org/). After cloning dlib from git or downloading it's sources, go into dlib folder and type:
+
+```bash
+mkdir build
+cd build
+cmake .. 
+cmake --build . --target install
+```
+
+On some Linux systems last command should be run as `sudo` to get access to /usr/local folder:
+```
+sudo cmake --build . --target install
+```
+
+## Using dlib in your project
+After dlib is installed in your system, you can now use it with CMake like this:
+```CMake
+cmake_minimum_required(VERSION 2.8.4)
+find_package(dlib REQUIRED)
+
+include_directories(${dlib_INCLUDE_DIRS})
+link_directories(${DLIB_LIBRARY_DIRS})
+
+add_executable(dlibsample dlibsample.cpp)
+target_link_libraries(dlibsample ${dlib_LIBRARIES})
+```
+
+Some additional ways to compile and use dlib are described in [Documentation](http://dlib.net/compile.html)
 
 ## Compiling dlib C++ example programs
 


davisking:
I don't really want to encourage people to install dlib.  I'm pretty sure that will lead to even more confused questions.  That's not to say that these instructions aren't useful, but maybe they should be placed somewhere less prominent and with a huge disclaimer that says like "you don't need to install dlib to use it, but if you want to install it as a shared library then do this".  So maybe at the bottom of the how to compile web page.

I'm saying this because I'm really sure that 99% of people who post in the dlib forums don't really even understand what a shared library is or how the C++ compile+link process works. Hell, I've met a lot of professional software developers in real life who didn't understand linking.  The main reason I've always avoided any kind of dlib install process is because I think the step of producing a shared or static library file that's then used later is too error prone for most people.  For example, I've lost track of the number of times I've seen someone compile some part of their code to a library file (or download the file) in some mode or with some compiler or runtime, then try to use that compiled library with incompatible compiler switches, runtime, whatever.  Then they get confusing linker errors or maybe just program crashes and they have no idea what's happening.  


So for example, here are things that will happen if we put this instruction out where most users see it.  People will run it on some linux machine but not know to run `ldconfig`, which sometimes needs to be run after installing a shared library on linux.  Or they will install it on OS X and there will be some similar issue. Does OS X put /usr/local/lib into the linker load path by default?  I'm not sure.  

What's even worse is visual studio users.  The notion of installing a library is just totally broken in windows.  Each version of visual studio has 8 different compile modes (the combination of 64/32 bit, debug release, multi-threaded/not-multi-threaded) and all are incompatible with each other in subtle ways that sometimes show up as confusing linker errors but more often as weird runtime errors.  Practically every single visual studio user who tries to install dlib with this process we are suggesting is going to have a bad experience.  I've already setup the CMake scripts to only generate static shared libraries when you do the install on windows to avoid people needing to use .dlls, which requires updating the system PATH environment variable, which most users will not understand.  Static don't require updating PATH, so at least programs will be executable without PATH variable changes, so at least there is that positive :)

e-fominov:
Thanks for the explanation. I agree that some skills are required for using CMake, and not everybody have them.
I will move CMake instructions into compile.html documentation file - to have all of them in one place - i will update PR

Do you want to keep faq/stackofervlow text on the top of readme?

davisking:
The FAQ link is probably a good idea.  But the stackoverflow link should probably go or else people will probably post bug reports to stack overflow.  Although most of "the bug" reports we already get on github are just user error :(

e-fominov:
"How to compile Dlib" is the most common question on StackOverflow :( 
30% - using with Visual Studio
30% - OSX problems
30% - python "no image_window"
<10% - Linux users with GCC

I think we should extend the documentation with more screen-shots and compilation explanations. I will write some additional documents about using dlib with Visual Studio, CMake, MinGW and GCC on Linux and put into this PR a bit later

davisking:
Sounds good.  Just make sure the documentation stays concise and easily navigated.  Mainly because most people don't read documentation.  When they glance at the page, if the answer doesn't jump out and punch them in the face, they won't read it.  This is made worse and worse the longer a piece of documentation gets.  So it's easy to end up with *more* questions after adding more detailed instructions.  

I even get questions where people copy and paste the error message to me, and inside the error message is literally the answer to their question.  But they just didn't read it.   That's how bad people are at reading instructions.  

There is also a selection bias where the most ignorant and unmotivated people post the most, while the vast silent majority of users who will read simple instructions don't have problems. Right now, for most people who have developed software before, the http://dlib.net/compile.html page is pretty straight and to the point.  But if you make it really long with a bunch of tutorials on GCC, 'what is a compiler?', visual studio tutorials, and stuff for novices then you will run into the problem of more experienced people taking a look at that page and then not reading it because it looks like it's long winded or not containing the content they want (basic, how to compile instructions).  

So keep all that in mind.  Maybe the things you are talking about adding would be on a separate page but linked to from the 'how to compile page'.  Or located further down.  There is definitely a problem with people though where they land on a web page and immediately flick the scroll wheel, skipping the top content, since top content is often boilerplate.  So if there are suddenly screenshots and tutorials down below they might end up being what they read when their question was more easily answered by the top content. 

Anyway, yeah, this sounds good.  But writing documentation is harder than most people think because of all these human factors problems.  But give it a go :)

e-fominov:
So may be better will be adding "5-minute tutorial" section on the main page of www.dlib.net where we can describe how to use dlib with different compiler types?

davisking:
Maybe.  Do you think people don't find the how to compile page at all?

davisking:
My impression has always been that that's not the problem.

davisking:
Maybe change the how to compile page to show the cmake instructions.  And then something below that which says "if you don't want to use cmake then here are a list of non-cmake tutorials"

davisking:
Although, most of those "I can't compile dlib" questions come from people who obviously have never used visual studio/xcode/gcc befor either.  I think the real issue is how can we best convince them to use CMake?  CMake is a lot easier to use for a novice than navigating visual studio's GUI, for instance.  

So maybe if the cmake instructions included a very short example cmake file/tutorial.  I'm not sure.  The examples literally already contain a cmake file that's pretty simple and functionally a tutorial.  But that obviously doesn't convince these people to use CMake or to not be confused by it.

e-fominov:
I think that 5-minute tutorials will be useful for the users that are starting with dlib. What can we put into tutorials?
   1. how to compile and run face detection example with CMake
   2. how to compile, install and use dlib in your project with CMake
   3. Using Visual Studio IDE with dlib, how to deal with Release/Debug configurations with MSVC and CMake, using dlib with all/Source.cpp
   4. using dlib with OpenCV
   5. how to compile python library and use it, what to do with image_window issue
   6. compiling Dlib with MSVC and CUDA for DNN module
   7. Using QTCreator IDE with dlib
   8. XCode with dlib
    
Having this small tutorials should solve some user problems

e-fominov:
@davisking , I have created a sample tutorial. Please check if you want to have it on your website

https://drive.google.com/file/d/0B0_JKP9ZnGdMd2tCWEJmeXR0ams/view?usp=sharing

If this is Ok - I will write some more tutorials and we will find a way how to publish them

davisking:
I just read the tutorial and it's good and clearly written.  However, I'm pretty sure that putting it on the dlib web page will lead to more confused user questions rather than less.  Because, (1) it's long and there are lots of steps, and (2) visual studio's multiple runtimes are inherently complex and this way of setting up a project requires the user to be aware of them.  Except that the sort of people who need this help definitely aren't aware of the issue.

For instance, here is something that will definitely happen regularly.  Someone will follow these steps and generate a visual studio project.  It will be either the release or debug one.  Then they will open it in the visual studio IDE and click the compile button, as they should be able to do.  However, if the little dropdown in the IDE that picks release/debug doesn't match the release/debug command used to generate the project then everything will break in a way most users won't understand. 

There is also this general problem I've noticed where novice users don't want to use cmake.  In general, the people who use cmake don't have a problem.  Experienced people don't have a problem either.  It's just the people who are new to C++ programming and also, for some reason, refuse to use cmake, who have trouble compiling dlib.  These people would have trouble compiling anything though.  

So a better way to approach this is probably to write a general cmake tutorial that encourages people to use cmake.  And definitely encourages visual studio users to use dlib the way the [examples/CMakeLists.txt](https://github.com/davisking/dlib/blob/master/examples/CMakeLists.txt) file does, by just compiling it inside their project in one big project.  That way the whole issue of visual studio runtimes is gone.  


Also, part of the problem I have with this kind of thing is that [examples/CMakeLists.txt](https://github.com/davisking/dlib/blob/master/examples/CMakeLists.txt) is literally already a tutorial.  Yet somehow people look at that file and don't get it.  After years of doing this I've noticed that a lot of the time people simply don't read the text on the screen.  For instance, if you sit with someone, even a smart person, they will land on a page and immediately start scrolling without reading the text on the monitor.  So presentation matters a lot.  So I just now modified the example CMakeLists.txt file to include a huge ASCII art comment at the top that says "this a tutorial, read the comments!!!".  So hopefully that will catch the eye.  

There is still the issue of getting people to look at that file in the first place.  The links to it should probably be more common and prominent on the dlib web sites.

davisking:
I also just added a link from the front page README.  So maybe that will help :)

e-fominov:
examples/CMakeLists.txt is very imressing! I think people will like it. CMake is hard for Windows users who are using GUI applications most of time. You are right here, that we if someone can use CMake, he will not have problems with Dlib

so, how can we reduce the amount of issue requests from newbies?

davisking:
lol, yeah.  I think the ASCII art will help.

But yeah, still I bet there will be questions.  I'm not sure what to do. Also, consider, dlib is a software library, not an end user application.  You have to be able to program to use it.  But we have all these questions from people who obviously don't know how to program since they can't even compile a basic program in visual studio.  Suppose we somehow eliminate all the troubles for those people while compiling the examples.  How are they then going to write a program that does anything?   They will just post questions at that point like "my program won't compile?  now what?"

davisking:
As another example.  What kind of documentation do you think would make it easier for this guy to make progress: http://stackoverflow.com/questions/43017973/cant-include-the-jpeg-support-headers-in-a-dlib-cpp-file ?  :)

e-fominov:
I am closing this PR as existing documentation improved and I don't see the right way how to make it better. May be later we can find some more ideas

davisking:
Yeah, sounds good.  Thanks for working on this :)

PR #459, opened 2017-02-27
Modernize CMakeLists

avaucher:
# Aim

To summarize this PR, allows to simply use, in projects relying on dlib, 
```CMake
target_link_libraries(MyTarget dlib::dlib)
```
instead of 
```CMake
target_include_directories(MyTarget ${dlib_INCLUDE_DIRS}) # or just "include_directories(${dlib_INCLUDE_DIRS})
target_link_libraries(MyTarget ${dlib_LIBRARIES})
```
This makes using dlib in other project simpler and more concise, and also works with installed versions of dlib (CMake configures the targets automatically when installing them).
Furthermore it allowed to simplify some of the CMakeLists.txt of the dlib source code.

Note that with this PR it is still possible to import dlib the "old" way.

# More details

This PR makes use of CMake functions to specify the dependencies of CMake targets more precisely so that CMake also knows that it must include those dependencies when we link against them.
To be more precise, this relies mainly on `target_include_directories` and `target_link_libraries` with specification of `PUBLIC`, `PRIVATE` or `INTERFACE`.

I changed the CMakeLists.txt in order not to break anything else but it would probably be possible to push the change a bit more to have cleaner CMakeLists, which I didn't do.
For instance, one could also use such features to import the optional dependencies (fftw, mkl, lapack, ...).

*Disadvantage*: The new functionality requires a CMake version >= 2.8.12, which is larger than the 2.8.4 of the status quo. In my eyes, 2.8.4 is already very old and therefore I think it is ok to switch to 2.8.12, but otherwise I guess it would also be possible to allow for both.


davisking:
Awesome, thanks for the PR.  It all looks great, except I noticed that if I tried to use 
```
target_link_libraries(MyTarget dlib::dlib)
```
in the examples/CMakeLists.txt file it doesn't work.

davisking:
There are also some issues related to CUDA and include search path settings not getting set right.  I can push those to your PR branch if you give me access.

davisking:
While we are at it, is there a nice clean way to enable C++11 in new versions of cmake?

I'm tempted to make dlibConfig.cmake include dlib/cmake_utils/use_cpp_11.cmake so that C++11 get's enabled.  But if there is some better way to do it I'm all ears.

e-fominov:
@davisking , there is a way to enable standard:

```CMake
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
some explanation can be found [here](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/)

after adding this lines into CMakeLists.txt I found that CXX_FLAGS changed into
```CMake
CXX_FLAGS =   -std=gnu++11
```

avaucher:
# Non-working examples/CMakeLists.txt with dlib::dlib

Oh, true; in the build directory, you need only `dlib` and not `dlib::dlib`, as the namespace `dlib::` is only set for the installed version.

For `dlib::dlib` to also work in the build directory, one can add the following to the dlib CMakeLists.txt:
```CMake
add_library(dlib::dlib ALIAS dlib)
[...]
add_library(dlib::dlib_shared ALIAS dlib_shared)
```

# CUDA
How can I give you access to do changes? 

# C++11
`CMAKE_CXX_STANDARD` is indeed the usual way to do it, but it appears it was only introduced in CMake 3.1.3 (not sure). If you are ready to set this as a minimal version, this is doable. Note also [`cmake_compile_features`](https://cmake.org/cmake/help/v3.8/manual/cmake-compile-features.7.html) which is sometimes preferred to `CMAKE_CXX_STANDARD`, because it is also compatible with MSVC.

davisking:
You can allow me to commit to the PR by following these instructions: https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/.  But no need now, I just merged it into the main repo along with my changes.  

I added statements to dlibConfig.cmake to include dlib/cmake_utils/use_cpp_11.cmake, so C++11 gets enabled.  I know about `set(CMAKE_CXX_STANDARD 11)` but that turns out to be somewhat less reliable than you would expect.  You can see all the stuff I've had to do in use_cpp_11.cmake to make it work in various situations.

Also added that alias statement and fixed the CUDA issue.

Anyway, thanks for the PR.  Let me know if you guys find any compile issues with these new cmake scripts.  I wouldn't be surprised if there is some funny problem on some platform since that's usually the way this goes :/

e-fominov:
Great PR, thanks!

Successfull built dlib + sample project with:
1. MinGW 6.3 x64 + OpenBLAS / Windows 7 x64
2. MinGW 6.3 x86 / Windows 7 x64
3. MSVC 2015 x64 + CUDA 8 / Windows 7 x64
4. GCC 5.3 x64 + CUDA8 + OpenBLAS / Ubuntu 14.04

When building with MSVC and CUDA got some warnings:
```
  Building NVCC (Device) object dlib/CMakeFiles/dlib.dir/dnn/Release/dlib_generated_cuda_dlib.cu.obj
  CMake Warning (dev) at dlib_generated_cuda_dlib.cu.obj.Release.cmake:86 (list):
    Policy CMP0007 is not set: list command no longer ignores empty elements.
    Run "cmake --help-policy CMP0007" for policy details.  Use the cmake_policy
    command to set the policy and suppress this warning.  List has value =
    [C:/Program Files/NVIDIA GPU Computing
    Toolkit/CUDA/v8.0/include;C:/work/dlib/dlib/external/libpng;C:/work/dlib/dlib/external/zlib;C:/Program
    Files/NVIDIA GPU Computing Toolkit/CUDA/v8.0/include;].
  This warning is for project developers.  Use -Wno-dev to suppress it.

  cuda_dlib.cu

CUSTOMBUILD : nvcc warning : The -std=c++11 flag is not supported with the configured host compiler. Flag will be ignor
ed. [C:\work\dlib\b15\dlib\dlib.vcxproj]

CUSTOMBUILD : nvcc warning : The -std=c++11 flag is not supported with the configured host compiler. Flag will be ignor
ed. [C:\work\dlib\b15\dlib\dlib.vcxproj]
```

Not sure if this PR is a reason of warnings

davisking:
Cool, thanks for checking.  

I just tried using MSVC 2015 x64 with CUDA 8 and cmake 3.6 but didn't get that CMP0007 warning.  Maybe I need a different version of cmake to get it?  I did get the one about -std=c++11 which I just fixed.

e-fominov:
I was using CMake 3.7, but the warning should come from CMake 3+. This mean it is something specific to my computer. I will check

avaucher:
Thanks!
I justed checked it on my Linux CentOS 7 with CMake 3.6 and GCC 6.2.0 and it worked as expected.

PR #476, opened 2017-03-13
add initial version of convolutional transpose layer

OranjeeGeneral:
Here are my changes to add a convolutional transpose (or deconvolutional) layer


davisking:
Does this make output tensors bigger than the input tensor?  The comments say it just does a convolution where the output is <= the input's size.

OranjeeGeneral:
Sorry the comment is misleading, I probably did change it. 

But to answer your question yes it makes the output tensor bigger if you use stride that is larger than one. That's the whole point of it. 

It kind of acts as a reverse of convolution were the backward and forward pass are basically swapped. So you can use it kind of an up-sampling were instead of a fixed weight up-sampling it actually learns the weights.

davisking:
Ok, that's what I thought.  Just making sure since the comments say otherwise :)

dlib/dnn/cpu_dlib.cpp
```diff
@@ -1699,6 +1699,28 @@ namespace dlib
             }
         }
 
+        void tensor_conv::forward_without_setup(
```
davisking: This code looks the same as tensor_conv::operator().  Am I just not seeing it?  It looks identical.

OranjeeGeneral: Nope not quiet it doesn't set

last_stride_y = stride_y;
last_stride_x = stride_x;
last_padding_y = padding_y;
last_padding_x = padding_x;

 that's the only "setup" the CPU implementation seems to do.

You could call this one in the other one to make it nicer and reduce code duplication yes. 
I thought about that but left it out for now.

The reason why is there is a much larger difference in the CUDA version but I have to keep the same signatures as interface between the two have to stay the same.

OranjeeGeneral: Nope not quiet it doesn't set

last_stride_y = stride_y;
last_stride_x = stride_x;
last_padding_y = padding_y;
last_padding_x = padding_x;

 that's the only "setup" the CPU implementation seems to do.

You could call this one in the other one to make it nicer and reduce code duplication yes. 
I thought about that but left it out for now.

The reason why is there is a much larger difference in the CUDA version but I have to keep the same signatures as interface between the two have to stay the same.

OranjeeGeneral: Nope not quiet it doesn't set

last_stride_y = stride_y;
last_stride_x = stride_x;
last_padding_y = padding_y;
last_padding_x = padding_x;

 that's the only "setup" the CPU implementation seems to do.

You could call this one in the other one to make it nicer and reduce code duplication yes. 
I thought about that but left it out for now.

The reason why is there is a much larger difference in the CUDA version but I have to keep the same signatures as interface between the two have to stay the same

OranjeeGeneral: Nope not quiet it doesn't set

last_stride_y = stride_y;
last_stride_x = stride_x;
last_padding_y = padding_y;
last_padding_x = padding_x;

 that's the only "setup" the CPU implementation seems to do.

You could call this one in the other one to make it nicer and reduce code duplication yes. 
I thought about that but left it out for now.

The reason why is there is a much larger difference in the CUDA version but I have to keep the same signatures as interface between the two have to stay the same

davisking: Ok.  I'll look into it when you eventually want to merge.  

But I should point out now that I won't merge pull requests that have incorrect/missing comments/documentation.  I have some instructions about this here: http://dlib.net/compile.html.  I'm a pretty big fan of precise interface specifications :)

OranjeeGeneral: Fair enough your project your rules :). Will adopt.

davisking:
Looks like cuDNN now supports upsampling layers.  So you can probably just call cuDNN to do this.  I haven't looked at the specifics yet though.

OranjeeGeneral:
Thanks I looked through the 6.0 release notes. 

No what they added is the option of a dilated convolution which is not the same basically it looks like they changed the purpose of the scaling parameter. But you could add dilate factor in the convolution parameter. Many other Deep learning packages have this as an optional parameter already. I've seen it in Caffe and MatConvNet. However personally I have not found any real use for it (yet) neither have I come across anyone mentioning it in the literature. 

Besides for convolution transpose you can already use the CuDNN calls just the order of them is reversed.

Scanning further through the release notes. The interesting bits are probably the new RNN stuff, the reduce Tensor, ELU Activaton and maybe the Int8 type. But I wonder if that really is such a big deal. As binary or tertiary networks are far more interesting if you go small scale. But I guess if you want to add support for that you would have to rewrite tensor class to be of a template element type which looks like it might be a lot of work.

davisking:
Yeah, it's not too exciting.  Just making sure none of it is helpful for
this PR :)

daisyron:
Is there a schedule to complete this feature?

OranjeeGeneral:
This feature is complete and functional so you can pull it off or add the patches. 

I am also have a fully working up-sample layer (both CPU/GPU) and a spatial padding layer which I probably would put in a sub-branch of my dlib-fork in case anybody is interested.

davisking:
Sweet.  I'll look it over soon and get it merged :)

OranjeeGeneral:
@rdaiasy 
The clue is on the top of the pull request page :). But in case my fork is here 

https://github.com/OranjeeGeneral/dlib

dlib/dnn/tensor_tools.h
```diff
@@ -947,6 +947,24 @@ namespace dlib { namespace tt
...
+        void operator() (
+            resizable_tensor& output,
+            const tensor& data,
+            const tensor& filters
+       )
+       {
+          impl.forward_without_setup(output,data,filters);
+       }
+
+       void setup(
```
davisking: This whole PR looks great, except for this part where there are these two undocumented functions.  They also behave in a somewhat opposite way compared to the others, which makes the class interface as a whole not cohesive and hard to understand.  I get what it's doing, but interface complexity is one of the roots of software bugs, and so should be avoided.

Really the interface for this class should either require some kind of setup before use for all the functions uniformly, or not require setup at all, as it previously did.  Not a mixture of both, and especially not one that isn't documented in any way.  

So what about this.  There can be a setup that takes only the 4 stride and padding arguments.  Calling that function will be required before any other function is executed.  Then the rest of the functions can do an internal allocation, initialization, or whatever, if they need it, to make everything work.  This would mean removing the other operator() that takes stride and padding.  It seems like that would cover everything you want to do but provide for a uniform and documentable interface.  Does that sound right or am I missing something?

OranjeeGeneral: Thanks

Yes I agree with your statement on interface complexity, my feeling was I didn't want to change too much of the original code (the setup code was already there but didn't fit well when the order of things are swapped). I am rather conservative :). 

But changing the signature of the setup code to include the right set of parameters instead of relying to the current pre-assumptions sounds like a good compromise I will have a look later and see if I can adjust it.

dlib/dnn/layers.h
```diff
@@ -147,14 +147,16 @@ namespace dlib
...
+            
             conv(output,
                 sub.get_output(),
-                filters(params,0),
-                _stride_y,
-                _stride_x,
-                padding_y_,
-                padding_x_
-                );
+                filters(params,0));
```
davisking: Does this line need to be here?  It looks like it can be removed?

davisking: Oops, never mind.  Funny formatting in the diff I'm looking at.  Obviously this line needs to be here :)

davisking:
Sweet, thanks for adding this layer.  I just merged it to the main branch.  There were some reallocations and copies in the code that I removed so it should run faster now too.

PR #487, opened 2017-03-18
dets cleared twice in evaluate_detectors of scan_fhog_pyramid.h

ebroglio:
already cleared at the beginning


davisking:
Thanks for the PR

PR #500, opened 2017-03-24
remove miniconda

marcovzla:
There is a conflict between X11 and miniconda. The solution is the same as the one for anaconda.


davisking:
Cool, thanks for the PR :)

Issue #505, opened 2017-03-28
Cleanup functionality needed for win-platform for singletons (socket, global_clock)


sigbjorn:
**Hi!**

First of all,
  while writing something at dlib site, **thanks so much** for this high quality library !

  We are using dlib as part of the shyft open-source project, at https://github.com/statkraft/shyft.
  It has excellent quality, header-only, well documented, multiplatform, and very well maintained.

We encountered something that we think is win-platform specific issue that we tracked to 
the function

dlib::get_global_clock().

That function maintains the singleton for timer_global_clock(), and works as expected for all practical purposes. We use dlib socket/streams and server, and it most likely gets activated for those parts.

The only problem we have found is when  we use the dlib as part of a python extension library.
When we exit python,  the python interpreter
 hangs forever in a lock/unlock during delete of the timer_global_clock singleton object.

The debugger attached picks up the thread/callstack located to:
 line 168..169 in shared_ptr_thread_safe.h 
  shared_node->unlock(); 
  delete data;

The callstack indicates that it's during the extension-library unload process, maybe static object destructors are called.

More specific info:
 only on windows platform (ms c++ 2015&2017, python 3.4..3.5.3),
 We are not able to reproduce on any linux-platform nor python version.

We found a workaround, that could be formed as a pull-request:
We just exposed a new function:

dlib::delete_global_clock();

- that reset the global-thread-safe singleton shared-pointer.
  (it's available at the fork https://github.com/sigbjorn/dlib)

If  we invoke this function as part of the python atexit registered function in our extension-module,
everything cleans up nice, and no more hang-situations. 

I suspect the dlib author(s) could provide even better solutions to this, maybe  with no code-changes,
so I leave this as an issue/question to maybe get some hint's about better approaches for solving this.

Br.
Sigbjørn Helset
MSc, Chief Developer Statkraft/ maintainer of shyft

davisking:
Thanks, I'm glad you like dlib :)

If it's pausing it should be because one of the timer events is still running when you try to unload it.  Like maybe you have called `iosockstream::terminate_connection_after_timeout()` and the timer is still running.  Are you sure all the dlib objects have been freed before you unload it?  If that were the case this shouldn't happen.

sigbjorn:
Hi,

Thanks for immediate feedback, - 
I am looking into it to day, and yes, the only dlib-object, the iostream socket is deleted before we call exit. (verified by printouts).

During the socket/close/delete, a timer is created (and global-clock thread started), 
[...]

PR #512, opened 2017-04-02
Provide cleanup for timer_global_clock singleton, issue #505

sigbjorn:
# Overview

Ref to issue #505 , and our discussions and findings,
this PR provides a new function dlib::delete_global_clock() that 
can be used to ensure proper cleanup prior to process termination.

Without this cleanup function, the user will have no means to workaround this
windows specific issue.

Except for the new function, there is no changes to the interfaces.

If possible it would be nice to have a solution that was usage and platform independent,
but since the root cause here is the windows ExitProcess call handling of threads, it is hard
to figure out a solution that would work in all contexts and circumstances.

The tests are run on windows 10 x64, arch -linux, and ubuntu 16.04 platforms.

### DIFF
--- a/dlib/timer/timer.cpp
+++ b/dlib/timer/timer.cpp
@@ -24,42 +24,42 @@ namespace dlib
     timer_global_clock::
     ~timer_global_clock()
     {
-		// The only time this destructor is called is when 
-		//
-		// a) the process terminates
-		// b) the dynamic library(.so/.dll) is unloaded (could be a part of a))
-		// 
-		// in case of a)
-		//   windows: the process termination is especially painful, since threads are killed
-		//     before destructors of the process image .dll's are called.
-		//     Thus, for the windows platform, there is no threads running, so the only thing
-		//     to do here is just let the standard memberwise destructors run
-		//   linux: it's ok to just signal shutdown and wait for the running thread, to exit
-		//   
-		// in case of b)
-		//   windows:
-		//     if it's part of the termination process, a) applies
-		//     if its part of user doing manual load_library/unload_library
-		//     there is no (safe/robust)solution, but best practices are described here
-		// 		    https://msdn.microsoft.com/en-us/library/windows/desktop/dn633971.aspx
-		//     to support such a clean shutdown, you are required to make a call prior to
-		//     unload dll, that shutdown all the threads in the contained dll.
-		//     This could be done in this module by providing a global_delete_clock()
-		// 
-		// linux: the destructor for linux will do it's usual job regardless.
-		//
+        // The only time this destructor is called is when 
+        //
+        // a) the process terminates
+        // b) the dynamic library(.so/.dll) is unloaded (could be a part of a))
+        // 
+        // in case of a)
+        //   windows: the process termination is especially painful, since threads are killed
+        //     before destructors of the process image .dll's are called.
+        //     Thus, for the windows platform, there is no threads running, so the only thing
+        //     to do here is just let the standard memberwise destructors run
+        //   linux: it's ok to just signal shutdown and wait for the running thread, to exit
+        //   
+        // in case of b)
+        //   windows:
+        //     if it's part of the termination process, a) applies
+        //     if its part of user doing manual load_library/unload_library
+        //     there is no (safe/robust)solution, but best practices are described here
+        //          https://msdn.microsoft.com/en-us/library/windows/desktop/dn633971.aspx
+        //     to support such a clean shutdown, you are required to make a call prior to
+        //     unload dll, that shutdown all the threads in the contained dll.
+        //     This could be done in this module by providing a global_delete_clock()
+        // 
+        // linux: the destructor for linux will do it's usual job regardless.
+        //
 
-		#ifndef _WIN32
+        #ifndef _WIN32
         m.lock();// at this point, we have the lock (on windows global destructor in dll, it does not matter, there is just one(this) thread)
         shutdown = true;// signal to thread (while-loop) it's going to shutdown(regardless pre-state of shutdown)
-		if(running) 
-		{ // ok, the thread should be running, or is about to run
-			s.signal();// wake up thread, (this could lock on the windows implementation of signaller if in exit-process)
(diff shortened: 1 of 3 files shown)


davisking:
Thanks for the PR.  I agree that we need to do something about this.  I'm not super excited about adding a function users need to call though.  Partly because I'm still hoping there is a more invisible solution and partly because there are other parts of dlib that have thread_local variables.   In #505 you discovered that visual studio just leaks thread_local variables when a dll containing such variables is unloaded.  That's a huge problem and needs some kind of more general solution, which if found could be used to fix this as well.  

So I'm going to leave this PR open until we figure out something that deals with the more general thread_local leaking problem.  Although yeah, in the short term, you should probably do what you did to make your application work.

sigbjorn:
Hi,

Yes, I agree,  it's not an ideal solution, and there might be other and better workarounds as well.

The hard fact on the windows platform is that the threads are taken away silently.
Destructors are called (but threads are not alive then).

A solution more in your direction here would be to simply stop the worker-thread when the user remove the timers,
 then 
re-start the thread when the first timer is added (similar to the initial situation, first timer added).

I also tested out this approach, It do work, - but I left it out since it was somewhat more involving changing code that is working (ref. timer_global_clock::remove(..)
 .. at the end of the function, check if the tm.size() ==0, and then shutdown=true, wait().

If that is a solution that could be an acceptable as a PR, I can make the mod's needed, incl. tests to this one.

**Addressing the larger problem, the silent win-thread-kill in ExitProcess:**

The take-away here seems to be that we can't rely on any threads alive in any of the static destructors. As stated in doc, there is no signal given at this point, that we could hook on to.

thread-local storage(memory) is not the main issue here I think, since it's reclaimed anyway. The problem occurs if those objects maintains external resources not reclaimed by win-os.
On windows, clearly, one should stay away from tls for all objects except internal state/memory etc.

 Adding platform specific code to do the destructor-work for dlib statics can be done,
 like
 'if in ExitProcess..., then skip lock/waits etc. and just cleanup open *external* resources, since the memory etc. is recovered by the win-os anyway.)'

Another thing to checkout is the standard **atexit** function, but there is a hard requirement that the .dll's that register a callback to it-self, should still be loaded at the time when the at-exit is called.
(We basically can't promise that, since it's out of our control)

Without this issue resolved, we can't use the standard version dlib in any python extension module we create on the windows-platform;
 - it will hang any process that uses it, 
so I will do my best to figure out a solution that could be acceptable. 

If anybody else have other ideas, or even proven concepts that fixes this, -please let us know.

Br Sigbjørn

davisking:
Well, for example, there are thread_local variables in dlib/dnn that allocate GPU resources.  I'm not at all confident that windows somehow cleans up those resources on its own when the dll is unloaded.  I wager mightily that they are just leaked if their destructors don't get executed.

As for python on windows.  It works fine so long as you don't create any of these timers.  The normally exposed python API doesn't let you do that and I think you are the only user who has done this sort of thing so far.  But yeah, it still needs a fix.

I just spent some time trying to find out if visual studio #defines some macro to let you know you are being built as a DLL but couldn't find anything.  We could just make the code always use dlib/timer_heavy.h when compiled on windows, which is a basic timer object that has just one thread per timer instance and no global state.  That's a somewhat heavy solution though since the global timer thread makes timers a lot faster, so I would prefer it to only be used when necessary (e.g. in a DLL apparently).  Although if you were really concerned with execution performance you probably wouldn't be using visual studio in the first place :/  So maybe it's not a big deal to use the slower version on windows.

sigbjorn:
Hi,
After getting some experience with my current approach (did not work in all circumstances)
, and doing some more research,
 that is, reading microsofts own documentation about dll's: https://msdn.microsoft.com/en-us/library/windows/desktop/dn633971.aspx

I got to the conclusion that the most robust solution on the _WIN32 platform for the timer-specific issue is to do as little as possible in the global-desctructors.

Regarding the cudnn problem, and tls as you mention; on windows, this is a serious issue and enforces special considerations *if* that code is part of a .dll. 

According to microsoft:
"*Thread Synchronization in DllMain during Process Exit
  By the time DllMain is called at process exit, all the process’s threads have been forcibly cleaned up **and there is a chance that the address space is inconsistent.** Synchronization is not required in this case. In other words, the ideal DLL_PROCESS_DETACH handler is empty. "

So even trying do do anything  here will in general do things worse.

davisking:
I haven't found anything either.  In the end, it seems like you are right that we should just do nothing in the destructor on windows.  Thanks for looking into all this and figuring out how to solve it :)

If you remove the tabs and clean up a few things (e.g. remove the floating delete_global_clock prototype) from the PR I'll merge it.

dlib/timer/timer.cpp
```diff
@@ -24,11 +24,42 @@ namespace dlib
...
+        //     unload dll, that shutdown all the threads in the contained dll.
+        //     This could be done in this module by providing a global_delete_clock()
+        // 
+        // linux: the destructor for linux will do it's usual job regardless.
+        //
+
+        #ifndef _WIN32
+        m.lock();// at this point, we have the lock (on windows global destructor in dll, it does not matter, there is just one(this) thread)
+        shutdown = true;// signal to thread (while-loop) it's going to shutdown(regardless pre-state of shutdown)
+        if(running)
```
davisking: Oh and what about this?  I don't see why we don't keep the original code here (but disabled on WIN32).

sigbjorn: I will clean it up,  I think the if(running) was to avoid doing the sync stuff if it was not started. There is no explicit tests of this since it's in static desctructors anyway.

davisking:
Sweet, looks good.  But one more thing.  There is really only one file changed.  Can you remove the other spurious files? :)  

Then I'll merge the PR.

sigbjorn:
Yes, definitely, that's most clean approach.

sigbjorn:
ok, now rebased and aligned formatting. I left the comments about the windows-issues in the code, is that ok, or should we reduce / or reformulate it ?

davisking:
Super, looks good, thanks for all your work on this.  The comments are fantastic :)

Issue #513, opened 2017-04-02
Add additional pixel color-depth/bit-depth options


frankiek3:
Currently all the pixel types, like rgb_pixel, have 8-bits per color channel, except for grayscale.
It would make sense to have each type have a larger range as more and more images are available with wider color gamuts. Most newer cameras also have a 16-bit per channel output.

I am willing to add the changes, but I think I need some guidance.

Here is what I have so far (compiles but untested).

Changes to `dlib/pixel.h`
Each pixel structure will only accept unsigned types for the color channels. Also made backwards compatible. [modifications]:
```
    template <typename T = uint8, typename std::enable_if<std::is_unsigned<T>::value, T>::type = false>
    struct rgb_pixel_t
    {
        /*!
            WHAT THIS OBJECT REPRESENTS
                This is a simple struct that represents an RGB colored graphical pixel.
        !*/
        typedef T type;

        rgb_pixel_t (
        ) {}

        rgb_pixel_t (
            T red_,
            T green_,
            T blue_
        ) : red(red_), green(green_), blue(blue_) {}

        T red;
        T green;
        T blue;
    };
    typedef rgb_pixel_t<> rgb_pixel;
```

Aliases to be used for code cleanup [new]:
```
    template <typename P>
    using is_pixel_t = typename std::integral_constant<bool, std::is_same<P, rgb_pixel>::value
                                   || std::is_same<P, bgr_pixel>::value
                                   || std::is_same<P, rgb_alpha_pixel>::value
                                   || std::is_same<P, hsi_pixel>::value
                                   || std::is_same<P, lab_pixel>::value>;
    template <typename P>
    using is_grayscale_t = typename std::integral_constant<bool, std::is_same<P, unsigned char>::value
                                                             || std::is_same<P, unsigned short>::value
                                                             || std::is_same<P, unsigned int>::value
                                                             || std::is_same<P, unsigned long>::value
                                                             || std::is_same<P, char>::value
                                                             || std::is_same<P, signed char>::value
                                                             || std::is_same<P, short>::value
                                                             || std::is_same<P, int>::value
                                                             || std::is_same<P, long>::value
                                                             || std::is_same<P, int64>::value
                                                             || std::is_same<P, uint64>::value>;
    template <typename P>
    using is_float_grayscale_t = typename std::integral_constant<bool, std::is_same<P, float>::value
                                                             || std::is_same<P, double>::value
                                                             || std::is_same<P, long double>::value
[...]

PR #518, opened 2017-04-05
Add expanded pixel color channel depth options #1

frankiek3:
1st Stage
- Adds ability to specify pixel depth using template arguments
- Uses pixel_traits for color space convertions instead of relying on 8-bit
- Changes some formatting

### DIFF
--- a/dlib/pixel.h
+++ b/dlib/pixel.h
@@ -1,6 +1,6 @@
 // Copyright (C) 2006  Davis E. King (davis@dlib.net)
 // License: Boost Software License   See LICENSE.txt for the full license.
-#ifndef DLIB_PIXEl_ 
+#ifndef DLIB_PIXEl_
 #define DLIB_PIXEl_
 
 #include <iostream>
@@ -40,7 +40,7 @@ namespace dlib
 
             - bool has_alpha
 
-            - long num 
+            - long num
 
             - basic_pixel_type
             - basic_pixel_type min()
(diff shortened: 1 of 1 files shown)


frankiek3:
From #513

davisking:
Cool.  I assume you are still working on this.  So far it seems sensible.  But could you try to avoid making whitespace only changes?  I looked over it and it's harder to review than needed since a lot of the changes aren't really changes, but just modified whitespace.

Also, another thing to watch out for is that the existing image processing routines aren't disturbed.  For instance, it's likely that there are some routines with contracts that say something like "I'll take in any RGB pixel" and maybe there is something about them that only makes sense for 8bit pixels.  Finding that kind of thing will require some looking around and testing.  Also, if you are going to make it templated you might as well let RGB pixels contain floats since I know people will ask for that.  Although that might complicate things.

frankiek3:
Sorry about the whitespace changes if you want I'll seperate it into 2 commits; One for formatting and one for then changes.
I could also just do one with the real changes, but the new file seems to me to be more readable.

Notes:
I reworked the alpha blending, as the compiler should take care of promoting types.

For some of the color space conversations that used HSL2RGB and RGB2HSL, I only allowed 8-bit using std::is_same
It might be better to use an optional external lib like imagemagick if further conversations are necessary using the new types.

I didn't touch serialization​ yet, since I don't want to mess up backwords compatibility.

So far the changes don't use the expanded color depth. I'll still need to add the commits that do.

davisking:
Yeah, don't include formatting in this PR.  Maybe in another one later on.

I don't want to add any dependencies, especially for something as minor as color conversion.  And in any event, the current behavior needs to be preserved so existing code doesn't break, especially the unit tests which aren't passing for this PR.  

Also, one of the important things about assign_pixel() and it's related tools is that they are able to convert from any pixel type to any other.  So that's definitely required.  For instance, the reason load_image() works on any image type is because it calls assign_pixel() to store the pixel value into the output image.  But if some color space conversions aren't allowed then things like load_image() will fail in ways users won't understand and are generally irritating.  The need to support all possible combinations of color space conversions is one of the main reasons why I was, and am, a bit wary about trying to have templated pixels.  But if you get it all working that would be super :)

fenollp:
You can ignore whitespace on review on Github https://github.com/blog/967-github-secrets

PR #519, opened 2017-04-05
fix unneeded reference for cblas_strsm and cblas_dtrsm when DLIB_USE_…

goodloop:
fix unneeded reference for cblas_strsm and cblas_dtrsm when DLIB_USE_BLAS is not defined

### DIFF
--- a/dlib/matrix/matrix_trsm.h
+++ b/dlib/matrix/matrix_trsm.h
@@ -9,6 +9,7 @@ namespace dlib
 {
     namespace blas_bindings
     {
+#ifdef DLIB_USE_BLAS
 #ifndef CBLAS_H
         extern "C"
         {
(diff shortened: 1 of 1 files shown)


davisking:
What does this fix?

goodloop:
if DLIB_USE_BLAS is configured as OFF in CMake configuration, it will also try to find the symbol for   cblas_strsm and cblas_dtrsm

davisking:
I don't think it does.  Have you gotten a linker error because of this and
this fixes it?

goodloop:
yes, when I build with xcode on Mac

davisking:
Hmm, function prototypes shouldn't ever cause linker errors in C++.  So that shouldn't happen but ok :)  xcode does bad and funny things to C++ sometimes.

davisking:
Oh and thanks for the PR :)

PR #525, opened 2017-04-06
Fix asserts for debug builds in IN_PROJECT_BUILD.

mcourteaux:
My project is in trouble because of the way the dlib cmake uses the `ENABLE_ASSERTS` for debug builds for in_project_builds.

My proposal is to move this logic to the. Why is the current approach bad: it doesn't work the first time. The reason is rather complex, but I'll try to explain anyways:

```
if (DLIB_IN_PROJECT_BUILD)
   # Make sure ENABLE_ASSERTS is defined for debug builds, but only for uses
   # who are building an application.  If they are just building dlib as a
   # stand alone library then don't set this because it will conflict with the
   # settings in config.h if we did.
   if (NOT CMAKE_CXX_FLAGS_DEBUG MATCHES "-DENABLE_ASSERTS")
      set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DENABLE_ASSERTS" 
         CACHE STRING "Flags used by the compiler during C++ debug builds." 
         FORCE)
   endif()
endif()
````

is currently in the dlib CMakeLists.txt file. The compiler flags are adjusted properly for the current directory, as this is the way cmake works. So the flag is only added for dlib. Therefor, the parent project (my project, which uses dlib) doesn't have this flag, and therefore, the ENABLE_ASSERTS is only added to dlib files, and not to mine, which causes the inconsistency checked at compile time in the code for the asserts using the magic int.

Now, if you rerun, CMake, it updates the cache, which is somehow shared between the project and dlib. And now my project gets the compiler flag ENABLE_ASSERTS. Weird, I'd say, but this is what happens.

So, to fix this, this block should come in the `dlib/cmake` file, which has to be `INCLUDE()`'d instead of `ADD_SUBDIRECTORY()`'d. Therefore, the flags will apply to the parent project as well.

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -29,19 +29,6 @@ cmake_policy(SET CMP0023 OLD)
 
 include(cmake_utils/add_global_compiler_switch.cmake)
 
-
-if (DLIB_IN_PROJECT_BUILD)
-   # Make sure ENABLE_ASSERTS is defined for debug builds, but only for uses
-   # who are building an application.  If they are just building dlib as a
-   # stand alone library then don't set this because it will conflict with the
-   # settings in config.h if we did.
-   if (NOT CMAKE_CXX_FLAGS_DEBUG MATCHES "-DENABLE_ASSERTS")
-      set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DENABLE_ASSERTS" 
-         CACHE STRING "Flags used by the compiler during C++ debug builds." 
-         FORCE)
-   endif()
-endif()
-
 macro (toggle_preprocessor_switch option_name)
    if (${option_name})
       add_global_define(${option_name})
--- a/dlib/cmake
+++ b/dlib/cmake
@@ -97,6 +97,15 @@ endif()
 # created by cmake is always release by default.
 include(${dlib_path}/cmake_utils/release_build_by_default)
 
+# Make sure ENABLE_ASSERTS is defined for debug builds, but only for uses
+# who are building an application.  If they are just building dlib as a
+# stand alone library then don't set this because it will conflict with the
+# settings in config.h if we did.
+if (NOT CMAKE_CXX_FLAGS_DEBUG MATCHES "-DENABLE_ASSERTS")
+    set (CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DENABLE_ASSERTS"
+            CACHE STRING "Flags used by the compiler during C++ debug builds."
+            FORCE)
+endif()
 
 # Don't add dlib if it's already been added to the cmake project
 if (NOT TARGET dlib)

### CODE REVIEW

**Comment by mcourteaux** [comment]
Actually there is more trouble, I believe. As the DLIB_ASSERTS_ENABLED flag for cmake, modifies CMAKE_CXX_FLAGS, whereas the code I moved modifies CMAKE_CXX_FLAGS_DEBUG.

**Comment by davisking** [comment]
Are you sure?  The dlib/CMakeLists.txt file has been doing this kind of thing for years and it's always worked.  For instance, I just went and compiled the example programs and that bit of cmake code does put ENABLE_ASSERTS in the parent project on the first run and everything is correct.  This has been the case for every version of CMake I've used for almost a decade.

**Comment by mcourteaux** [comment]
Well, I'll try to reproduce the issue with the examples from dlib.
At least there is a problem with the `DLIB_ENABLE_ASSERTS`:
If I do this in a debug build of my project:

    SET(DLIB_ENABLE_ASSERTS         OFF   CACHE BOOL "" FORCE)

Dlib will add the `ENABLE_ASSERTS` nevertheless to the `CMAKE_CXX_FLAGS_DEBUG`.

**Comment by davisking** [comment]
Yeah, well, there is also stuff in the header files that enable the asserts
sometimes based on what you are doing.  For instance, visual studio defines
a _DEBUG macro when in debug mode and that automatically enables the
asserts regardless of any cmake setting.   I think that's good.

But it's got to be defined consistently obviously.  If this really is a bug
then we can fix it.  Maybe it's only active in newer versions of cmake I
haven't tested with.  Or maybe it's nothing :)

**Comment by mcourteaux** [comment]
In `examples/CMakeLists.txt`, I modified this, to simulate my project CMakeFile.

```
# Dlib ruins my compiler setting by forcing C++11, while my project uses C++14.
SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")

SET(DLIB_JPEG_SUPPORT           OFF   CACHE BOOL "" FORCE)
SET(DLIB_PNG_SUPPORT            OFF   CACHE BOOL "" FORCE)
SET(DLIB_GIF_SUPPORT            OFF   CACHE BOOL "" FORCE)
SET(DLIB_USE_CUDA               OFF   CACHE BOOL "" FORCE)
SET(DLIB_LINK_WITH_SQLITE3      OFF   CACHE BOOL "" FORCE)
SET(DLIB_USE_LAPACK             ON    CACHE BOOL "" FORCE)
SET(DLIB_USE_BLAS               ON    CACHE BOOL "" FORCE)
SET(DLIB_NO_GUI_SUPPORT         ON    CACHE BOOL "" FORCE)
SET(DLIB_ENABLE_ASSERTS         ON    CACHE BOOL "" FORCE)
# Tell dlib it's fine.
SET(COMPILER_CAN_DO_CPP_11 1)
include(../dlib/cmake)
```

And then make a release build like this:

    cmake -DCMAKE_BUILD_TYPE=Release -DUSE_AVX_INSTRUCTIONS=ON ../examples && make -j4

Results in the dlib library being compiled with ASSERTS, and the examples **without** ASSERTS, causing the inconsistency.

**Comment by davisking** [comment]
I just tried what you posted and it worked correctly in two versions of
cmake (3.5.1 and 2.8.12).  What version are you using?

**Comment by mcourteaux** [comment]
Latest, I guess: cmake 3.7.2
Super weird...

**Comment by mcourteaux** [comment]
Make sure you do this in a clean build folder as you test it. Because, if you run `cmake .` after you ran `cmake` for the first time, it does work.

**Comment by davisking** [comment]
Yeah, I did it with a clean build folder a few times.  Can you check if it
happens to you with older versions of cmake?

**Comment by mcourteaux** [comment]
I made a minimal working demo, showing the problem for me:

File `cmake_test/CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.7)
project(cmake_test)

set(MY_VAR "123")
option(MY_OPTION "test option" ON)

message("MY_VAR: ${MY_VAR}")
message("MY_OPTION: ${MY_OPTION}")
message("add_subdirectory")
add_subdirectory(sub_dir)
message("MY_VAR: ${MY_VAR}")
message("MY_OPTION: ${MY_OPTION}")
```

File `cmake_test/sub_dir/CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.7)
project(sub_project)

message("in subdir, changing values")
set(MY_VAR "123_sub" CACHE STRING "test var" FORCE)
set(MY_OPTION OFF CACHE BOOL "test option" FORCE)
```

Now, running `cmake ..` in a clean folder gives me this:

davisking:
That's what happens to me too, and that's always been the case.  It's just
that CMAKE_CXX_FLAGS has always been magic.  For instance, I use this macro
a lot
https://github.com/davisking/dlib/blob/master/dlib/cmake_utils/add_global_compiler_switch.cmake
which does what we are talking about here.  I've been using that thing for
years and years and it's always worked as expected.  The whole point of it
is to set flags globally as we are discussing and so far it's always
worked.  Humor me and try this on your machine with an older version of
cmake to see if it's not something crazy going on :)

mcourteaux:
Ye, I'm compiling cmake 3.5.1... I'm curious...

However, about that macro: ye, it's strange... I'll come back to you as soon as I tested this with cmake 3.5.1.

mcourteaux:
Okay, for the sake of completeness: here is the full console output.
https://pastebin.com/raw/k3jDXyGk

(Ctrl-F / Cmd-F for `--version` and for `--DENABLE_ASSERTS` to highlight important parts)
I first compile Release using cmake 3.7.2. The problem occurs. Then unlink this version of cmake: successfully unlinked: cmake not found. Then install version 3.5.1: successful as indicated by the --version. Compile Release again after clearing the folder completely. Error occurs as well. So no difference with the version of Cmake.

Soo... I guess you are not testing it exactly as I am. Have you carefully added this line to your CMakeLists of the examples folder?
```cmake
SET(DLIB_ENABLE_ASSERTS         ON    CACHE BOOL "" FORCE)
```

So basically, what I'm trying is: enable asserts in release mode.

mcourteaux:
For reference, the entire CMakeLists.txt from the examples folder on my machine is now:

```cmake
#
# This is a CMake makefile.  You can find the cmake utility and
# information about it at http://www.cmake.org
#


cmake_minimum_required(VERSION 2.8.12)

PROJECT(examples)

SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")

SET(DLIB_JPEG_SUPPORT           OFF   CACHE BOOL "" FORCE)
SET(DLIB_PNG_SUPPORT            OFF   CACHE BOOL "" FORCE)
SET(DLIB_GIF_SUPPORT            OFF   CACHE BOOL "" FORCE)
SET(DLIB_USE_CUDA               OFF   CACHE BOOL "" FORCE)
SET(DLIB_LINK_WITH_SQLITE3      OFF   CACHE BOOL "" FORCE)
SET(DLIB_USE_LAPACK             ON    CACHE BOOL "" FORCE)
SET(DLIB_USE_BLAS               ON    CACHE BOOL "" FORCE)
SET(DLIB_NO_GUI_SUPPORT         ON    CACHE BOOL "" FORCE)
SET(DLIB_ENABLE_ASSERTS         ON    CACHE BOOL "" FORCE)
SET(COMPILER_CAN_DO_CPP_11 1)
include(../dlib/cmake)

# Tell CMake to compile a program.  We do this with the ADD_EXECUTABLE()
# statement which takes the name of the output executable and then a list of
# .cpp files to compile.  Here each example consists of only one .cpp file but
# in general you will make programs that const of many .cpp files.
ADD_EXECUTABLE(assignment_learning_ex assignment_learning_ex.cpp)
# Then we tell it to link with dlib.
TARGET_LINK_LIBRARIES(assignment_learning_ex dlib::dlib)




# Since there are a lot of examples I'm going to use a macro to simply this
# CMakeLists.txt file.  However, usually you will create only one executable in
# your cmake projects and use the syntax shown above.
MACRO(add_example name)
   ADD_EXECUTABLE(${name} ${name}.cpp)
   TARGET_LINK_LIBRARIES(${name} dlib::dlib )
ENDMACRO()

# if an example requires GUI, call this macro to check DLIB_NO_GUI_SUPPORT to include or exclude
MACRO(add_gui_example name)
   if (DLIB_NO_GUI_SUPPORT)
      message("No GUI support, so we won't build the ${name} example.")
   else()
      add_example(${name})
   endif()
ENDMACRO()

# The deep learning toolkit requires a compiler with essentially complete C++11
# support.  However, versions of Visual Studio prior to October 2016 didn't
# provide enough C++11 support to compile the DNN tooling, but were good enough
# to compile the rest of dlib.  
if (NOT USING_OLD_VISUAL_STUDIO_COMPILER)
   add_example(dnn_metric_learning_ex)
   add_gui_example(dnn_face_recognition_ex)
   add_example(dnn_introduction_ex)
   add_example(dnn_introduction2_ex)
   add_example(dnn_inception_ex)
   add_gui_example(dnn_mmod_ex)
   add_gui_example(dnn_mmod_face_detection_ex)
   add_gui_example(random_cropper_ex)
   add_gui_example(dnn_mmod_dog_hipsterizer)
   add_gui_example(dnn_imagenet_ex)
   if (NOT MSVC)
      # Don't try to compile this program using Visual Studio since it causes the
      # compiler to run out of RAM and to crash.  Maybe someday Visual Studio
      # won't be broken :(
      add_example(dnn_imagenet_train_ex)
      add_example(dnn_metric_learning_on_images_ex)
   endif()
endif()

#here we apply our macros 
add_gui_example(3d_point_cloud_ex)
add_example(bayes_net_ex)
add_example(bayes_net_from_disk_ex)
add_gui_example(bayes_net_gui_ex)
add_example(bridge_ex)
add_example(bsp_ex)
add_example(compress_stream_ex)
add_example(config_reader_ex)
add_example(custom_trainer_ex)
add_example(dir_nav_ex)
add_example(empirical_kernel_map_ex)
add_gui_example(face_detection_ex)
add_gui_example(face_landmark_detection_ex)
add_gui_example(fhog_ex)
add_gui_example(fhog_object_detector_ex)
add_example(file_to_code_ex)
add_example(graph_labeling_ex)
add_gui_example(gui_api_ex)
add_gui_example(hough_transform_ex)
add_gui_example(image_ex)
add_example(integrate_function_adapt_simp_ex)
add_example(iosockstream_ex)
add_example(kcentroid_ex)
add_example(kkmeans_ex)
add_example(krls_ex)
add_example(krls_filter_ex)
add_example(krr_classification_ex)
add_example(krr_regression_ex)
add_example(learning_to_track_ex)
add_example(least_squares_ex)
add_example(linear_manifold_regularizer_ex)
add_example(logger_custom_output_ex)
add_example(logger_ex)
add_example(logger_ex_2)
add_example(matrix_ex)
add_example(matrix_expressions_ex)
add_example(max_cost_assignment_ex)
add_example(member_function_pointer_ex)
add_example(mlp_ex)
add_example(model_selection_ex)
add_gui_example(mpc_ex)
add_example(multiclass_classification_ex)
add_example(multithreaded_object_ex)
add_gui_example(object_detector_advanced_ex)
add_gui_example(object_detector_ex)
add_gui_example(one_class_classifiers_ex)
add_example(optimization_ex)
add_example(parallel_for_ex)
add_example(pipe_ex)
add_example(pipe_ex_2)
add_example(quantum_computing_ex)
add_example(queue_ex)
add_example(rank_features_ex)
add_example(running_stats_ex)
add_example(rvm_ex)
add_example(rvm_regression_ex)
add_example(sequence_labeler_ex)
add_example(sequence_segmenter_ex)
add_example(server_http_ex)
add_example(server_iostream_ex)
add_example(sockets_ex)
add_example(sockstreambuf_ex)
add_example(std_allocator_ex)
add_gui_example(surf_ex)
add_example(svm_c_ex)
add_example(svm_ex)
add_example(svm_pegasos_ex)
add_example(svm_rank_ex)
add_example(svm_sparse_ex)
add_example(svm_struct_ex)
add_example(svr_ex)
add_example(thread_function_ex)
add_example(thread_pool_ex)
add_example(threaded_object_ex)
add_example(threads_ex)
add_example(timer_ex)
add_gui_example(train_object_detector)
add_example(train_shape_predictor_ex)
add_example(using_custom_kernels_ex)
add_gui_example(video_tracking_ex)
add_example(xml_parser_ex)


if (DLIB_NO_GUI_SUPPORT)
   message("No GUI support, so we won't build the webcam_face_pose_ex example.")
else()
   find_package(OpenCV QUIET)
   if (OpenCV_FOUND)
      include_directories(${OpenCV_INCLUDE_DIRS})

      ADD_EXECUTABLE(webcam_face_pose_ex webcam_face_pose_ex.cpp)
      TARGET_LINK_LIBRARIES(webcam_face_pose_ex dlib::dlib ${OpenCV_LIBS} )
   else()
      message("OpenCV not found, so we won't build the webcam_face_pose_ex example.")
   endif()
endif()


if (DLIB_LINK_WITH_SQLITE3)
   add_example(sqlite_ex)
endif()
```

davisking:
Hmm.  I tried that CMakeLists.txt file you posted and still don't have any issue.  I also tried with the new cmake and it works fine.  Also, everything you are doing is in release mode so the debug settings were were talking about don't come into play anyway.  But you are saying changing the debug code in cmake fixes this?

davisking:
Hmm, apparently there is something magical about cmake on the mac that causes this to happen.  Doesn't happen on any other platform seemingly.  Anyway, it doesn't seem like this PR should fix this.  Does it for you?

mcourteaux:
I'm getting tired, and waisting this much time on this is getting annoying. Tomorrow, I'll restart my fixes, and make them as clean as possible and update this Pull Request.

davisking:
The code in github should fix this now.  Try it and let me know.   Also, you can tell cmake to enable whatever version of C++ you want with a statement like 

```
set(CMAKE_CXX_STANDARD 14)
```

and then everything works properly and you don't need to futz with dlib's C++11 variables.

PR #526, opened 2017-04-06
Fix asserts, super clean!

mcourteaux:
Wether asserts are enabled or not, is now only suggested by dlib, and not longer imposed. The user can override using SET(CACHE FORCE) or command line -DDLIB_ENABLE_ASSERT=ON/OFF anytime.

Apparently, vim stripped all of the trailing spaces as well. I suggest looking at the file here, if all the whitespace changes are too confusing.
https://github.com/mcourteaux/dlib/blob/master/dlib/CMakeLists.txt

(If you want I can make another pull request to strip a bunch of trailing spaces.)

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -46,12 +46,20 @@ macro (toggle_preprocessor_switch option_name)
    endif()
 endmacro()
 
+
+# Suggest asserts enabled or not. However, user can override
+# using SET(CACHE FORCE). This is ideal for in_project_builds, as this
+# preprocessor define will be adopted by the parent project.
+# The user can *always* override this setting, and is in control when needed.
+set (DLIB_ENABLE_ASSERTS_STR
+    "Enable this if you want to turn on the DLIB_ASSERT macro" )
+
 if (DLIB_IN_PROJECT_BUILD AND CMAKE_BUILD_TYPE MATCHES "Debug")
-   # Make sure ENABLE_ASSERTS is defined for debug builds, but only for uses
-   # who are building an application.  If they are just building dlib as a
-   # stand alone library then don't set this because it will conflict with the
-   # settings in config.h if we did.
-   enable_preprocessor_switch(ENABLE_ASSERTS)
+    # In Debug for in project builds, we **suggest** asserts ON.
+    option(DLIB_ENABLE_ASSERTS ${DLIB_ENABLE_ASSERTS_STR} ON)
+else()
+    # Otherwise, we **suggest** asserts OFF.
+    option(DLIB_ENABLE_ASSERTS ${DLIB_ENABLE_ASSERTS_STR} OFF)
 endif()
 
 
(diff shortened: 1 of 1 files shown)


mcourteaux:
So to conclude: the behaviour should remain the same as before, but now a user is in control of the assert "enabledness" if needed by using
```cmake
SET(DLIB_ENABLE_ASSERTS ON/OFF CACHE BOOL "" FORCE)
```

I tested this thoroughly and it seems to be working beautifully on my machine now. Note that this builds on your last changes using the `target_compile_options`.

davisking:
Cool, but can you take the white space changes out of the PR?

mcourteaux:
Fixed!

dlib/CMakeLists.txt
```diff
@@ -118,14 +125,6 @@ if (NOT TARGET dlib)
       set (DLIB_DISABLE_ASSERTS true)
       set (ENABLE_ASSERTS false) 
       disable_preprocessor_switch(ENABLE_ASSERTS)
-      # Never force the asserts off when doing an in project build.  Instead,
-      # let the debug/release mode setting toggle asserts on or off (or the
-      # DLIB_ENABLE_ASSERTS option obviously).  That is, even if the
-      # DLIB_ENABLE_ASSERTS option is off debug mode can still cause the
-      # asserts to turn on when using an in project build. 
-      if (NOT DLIB_IN_PROJECT_BUILD)
-         enable_preprocessor_switch(DLIB_DISABLE_ASSERTS)
```
davisking: I'm pretty sure we need this line. Is there a reason to take it out?

mcourteaux: So, you are saying one would **never** want to compile dlib with asserts as a standalone library?

mcourteaux: Oh, sorry, my bad! Misread it! Will commit now.

mcourteaux:
Voila, I think this should be it.

davisking:

mcourteaux:
Thank you very much for the fast replies throughout the day!

davisking:
No problem :)

davisking:
It occurred to me that we broke the normal assert on/off behavior for visual studio users.  I just pushed a change that fixes that.  I also took out the setting of the asserts based on cmake's debug/release build setting since it never toggles asserts when you change that setting in ccmake or cmake-gui since the scripts didn't force the setting to the cache.  Which could be changed so that it did but that would be weird too.

mcourteaux:
I am getting crazy... It worked perfect for the examples. But not for my project... Again inconsistency between dlib and my project with the assert flags... Trying to figure out why now... :(

mcourteaux:
Ye, your latest changes fixes it partly. However, still: it's broken if I explicitly ask for assertions to be enabled. The `-DENABLE_ASSERTS` flag is used on dlib source files, but not on mine. So the property somehow doesn't propagate back to my project `compile_target_options`. I could go with this "fix", but it's weird. I'll try to see what this now gives on the examples.

davisking:
Huh.  Propagating `compile_target_options` do dependent targets is one of modern cmake's main features.  So I would expect it to work.

mcourteaux:
I figured it out a minute ago: I'm using an object library (that's a cmake feature, that allows you to share object files between multiple targets). The target_compile_options doesn't propagate through the OBJECT library:
```cmake
INCLUDE("ext/dlib/dlib/cmake")
ADD_LIBRARY(FooObjects OBJECT ${SRCFILES})
ADD_EXECUTABLE(Foo $<TARGET_OBJECTS:FooObjects>)
TARGET_LINK_LIBRARIES(Foo dlib::dlib)
```

mcourteaux:
I'll try refactoring my project cmake to use a static library instead of an object library...

davisking:
I'm pretty sure that's how the dependency stuff in cmake is supposed to work.

mcourteaux:
Well, it seems like it's broken, and no-one solves it:
http://stackoverflow.com/questions/31290077/how-to-specify-imported-dependencies-of-object-library

davisking:
Yeah.  I would just use a static library.

PR #527, opened 2017-04-07
Add support for Python3.6 lib search with Cmake

y0ast:
A recent commit added support for python3.5 lib search with Cmake. This PR extends that to python 3.6.

Relevant old commit: https://github.com/davisking/dlib/commit/17a2c75ddffb13e915686803c5e3bf4cc1389a96


davisking:
I'm pretty sure that overwrites 3.5.  It probably needs to be a list.

y0ast:
Right, I didn't realize that and only tested on 3.6. New commit should solve it.

davisking:
Cool, thanks for the PR :)

Issue #536, opened 2017-04-14
Objects called nil collide with Apple's definition


roymacdonald:
Greetings,
I've just came across a very annoying problem while trying to compile on OSX, in an Xcode project that also uses Objective-C stuff, in particular an [openFrameworks addon I'm working on.](https://github.com/roymacdonald/ofxDLib).

There are several places where objects are called `nil` like [here](https://github.com/davisking/dlib/blob/master/dlib/graph_cuts/min_cut.h#L197).

Changing these objects name's will fix this problem. Is there any problem if I change these and PR?
Is there any reason for naming these `nil`?

BTW, I came with [this really ugly hack](https://github.com/roymacdonald/ofxDLib/issues/8) to overcome this issue but I rather avoid with a real fix.

All the best!

PR #537, opened 2017-04-14
Fixes issue with object named nil that collide with apple's definition.

roymacdonald:
Fixes issue https://github.com/davisking/dlib/issues/536

### DIFF
--- a/dlib/graph_cuts/min_cut.h
+++ b/dlib/graph_cuts/min_cut.h
@@ -162,9 +162,9 @@ namespace dlib
             g.set_label(sink_node, SINK_CUT);
 
             // used to indicate "no parent"
-            const unsigned long nil = g.number_of_nodes();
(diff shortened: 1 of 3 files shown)


davisking:
Cool, thanks for the PR.  Can you make the change to a variable that has the same meaning as nil.  num_nodes isn't a sensible name for that variable in the min cut code.  Something like no_parent would be much better.

roymacdonald:
Hi. Sure. No problem.
I used num_nodes as the variable was assigned with the number of nodes.
I'll fix it asap.

roymacdonald:
@davisking done. 
Are the other var names I chose fine?
BTW, dlib is superb. Awesome work you've done. :D 
best

davisking:
Those all look good.

Thanks, glad you like dlib :)

roymacdonald:
super!
Thanks.
cannot think anything else than "the internet is awesome." :D

PR #538, opened 2017-04-15
Suggesting to fix possible parenthesis issue in elastic net test

reunanen:
Noticed compiler warning `C4800: 'double': forcing value to bool 'true' or 'false' (performance warning)`.

Suggesting to change the test to what perhaps was the original intention.


davisking:
I somehow forgot to merge this PR a while ago.  Anyway, looks good, thanks for the PR :)

PR #539, opened 2017-04-16
CMake improvements

SoapGentoo:
Hi @davisking 

1. I'd like to get the `GNUInstallDirs` module into dlib. It sets all install paths to the GNU coding standard defaults (i.e. what is currently the case) but allows for users to change them. Furthermore, this gets rid of the previous custom `LIB_INSTALL_DIR` solution, which is not idiomatic and unsupported by Kitware. The current solution using `GNUInstallDirs` is more generalisable and supported by Kitware.
2. Furthermore, I've removed the installing of the license file. Most distributions do not install license files as they are redundant (it is specified in package metadata). Also, what license a software is under is independent of whether the license is installed alongside it.

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -7,6 +7,9 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(dlib)
 
+# Adhere to GNU filesystem layout conventions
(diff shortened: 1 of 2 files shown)


davisking:

PR #540, opened 2017-04-16
Add new loss layer for semantic segmentation (pixel-wise classification)

reunanen:
The intent of this PR is to address the other half of #288 by adding a new loss layer called `loss_multiclass_log_matrixoutput` (*). This new layer is similar to the `loss_multiclass_log` layer, which evaluates a single classification result for the whole image. The key difference is that the new layer evaluates a per-pixel classification result matrix.

Seems to work fine at least for my purposes, although not really battle-tested yet.

(*): Feel free to suggest a better name. I didn't want to call it `loss_multiclass_log_multioutput` though, as [originally planned](https://github.com/davisking/dlib/issues/288#issuecomment-294292178), because the new layer is not quite as similar to the `loss_mean_squared_multioutput` layer as I had initially thought. (I don't think `loss_mean_squared_multioutput` evaluates matrix output.)


dlib/dnn/loss.h
```diff
@@ -1341,6 +1341,174 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    // In semantic segmentation, 65535 classes ought to be enough for anybody.
+    typedef unsigned short matrixoutput_label_t;
+
+    // In semantic segmentation, if you don't know the ground-truth of some pixel,
+    // set the label of that pixel to this value. When you do so, the pixel will be
+    // ignored when computing gradients.
+    static const matrixoutput_label_t label_to_ignore = std::numeric_limits<matrixoutput_label_t>::max();
```
reunanen: Wasn't quite sure where to put this, so left it here.

dlib/dnn/loss.h
```diff
@@ -1341,6 +1341,174 @@ namespace dlib
...
+                    if (value > max_value) {
+                        label = static_cast<matrixoutput_label_t>(k);
+                        max_value = value;
+                    }
+                }
+                return label;
+            };
+
+            for (long i = 0; i < output_tensor.num_samples(); ++i, ++iter) {
+                *iter = matrix<matrixoutput_label_t>(output_tensor.nr(), output_tensor.nc());
```
davisking: This does a copy.  You should call .set_size() on the matrix instead.

dlib/dnn/loss_abstract.h
```diff
@@ -769,6 +769,68 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    // In semantic segmentation, 65535 classes ought to be enough for anybody.
+    typedef unsigned short matrixoutput_label_t;
```
davisking: I would just use uint16 for this and not even bother with a typedef to give it a special name.

davisking:
Cool, looks good generally.  Probably a better name is `loss_multiclass_log_per_pixel`.

Also, can you add a small unit test for it?  Just something basic like the multiclass regression one that runs fast and shows that it works.

reunanen:
@davisking Thanks for the guidance, have now tried to address all your points.

dlib/test/dnn.cpp
```diff
@@ -1888,9 +1888,225 @@ namespace
...
+        ::std::default_random_engine generator(16);
+        ::std::bernoulli_distribution coinflip(0.5);
+
+        using filter_type = con<num_classes, filter_height, filter_width, 1, 1, input<matrix<float>>>;
+
+        // Define a "truth" filter
+        filter_type truth_filter;
+        truth_filter(xtmp); // Set up the convolutional layer
+
+        // Generate training data
```
reunanen: There's some code duplication here (cf. the previous function). Generating the training data like this could be extracted to a separate function in order to dedupe, however it might also be good to keep the test cases independent in case someone wants to modify one, but not the other.

davisking: Yeah, leave them independent.

reunanen:
Unfortunately, it seems to me that there may be something wrong here. Not sure where exactly, and it is still possible that the problem is on the call site (*) and not in this PR. Will try to add more tests at some point, but it may well take quite a while before I can get myself to do it. (If anyone else feels like adding relevant tests – please go ahead!)

More specifically, I'd expect to get identical results, if I:

1. Use `loss_multiclass_log` with a sliding window, or
2. Call `loss_multiclass_log_per_pixel` with the full image as input.

However, in some cases option 2 appeared to give results that are almost fine, but still arguably worse (than the results I get from option 1).

Therefore, I'd recommend against merging this before there's at least a test that shows that options 1 and 2 give identical results. (They should, right?)

(*): Sorry – not FOSS. Belongs to my client.

davisking:
I'm not sure, it depends on what's going on at the call site :)

reunanen:
I guess the reason for the different results may be padding.

Consider forward propagation. In scenario 2 (call `loss_multiclass_log_per_pixel` with the full image as input), only the borders of the full image will be padded, right? On the other hand, in scenario 1 (use `loss_multiclass_log` with a sliding window) every input window will be automatically padded, right?

If so, then this may at least explain the non-identical results I'm getting.

If the above reasoning is correct, then the PR itself should be fine in this regard. Nevertheless, I wonder how I should approach fixing the problem at the call site. I assume I could disable all padding, but that would mean changing the net structure. On the other hand, I guess I could train with larger windows. Right now I'm training with crops that produce only 1 pixel in the output.

(I don't really need the results of these two scenarios to be identical; I started to looking at this only to understand why my results degraded after taking the code of this PR in use.)

davisking:
Yeah, you definitely need to account for padding if you want them to match.  I'm not sure what you are talking about specifically.  Trouble you were having making the unit tests work out, or something else on a real ML task that wasn't working right?   The code you have in here looks to me like it's computing the correct gradient so I would expect it to work.

reunanen:
I meant something else, on a real ML task.

davisking:
Ok, sounds good.  It seems like you are still refining the unit tests.  Let me know when you are happy when them and I'll merge the PR :)

reunanen:
I'm now happy with the unit tests.

davisking:
Sweet, thanks for the PR :)

davisking:
FYI, I also moved label_to_ignore into the loss class so it doesn't clutter the global namespace.

PR #555, opened 2017-04-26
Added history to clustering

mkdubik:
Hi, thanks for the awesome library!

I've added a feature to track the history of the bottom up clustering, with the original node id's in each of the nodes, similar to sklearn's 'children' attribute (http://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html).   

Here's a small output example of what I'm talking about, using a dissimilarity 16x16 matrix

https://gist.github.com/mkdubik/2417c70d55c0c2a34f1bd687b8062220

All the best,

### DIFF
--- a/dlib/clustering/bottom_up_cluster.h
+++ b/dlib/clustering/bottom_up_cluster.h
@@ -44,6 +44,28 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    struct bu_cluster
+    {
+        long id;
+        long left;
+        long right;
+        std::vector<long> orig;
+    };
+
+    inline std::ostream& operator<< (std::ostream& out, const bu_cluster& c)
+    {
+        out << "["<<c.id<<","<<c.left<<","<<c.right<<",[";
+        for (auto iter = c.orig.begin(); iter != c.orig.end(); iter++) {
+          out << *iter;
+          if (iter != c.orig.end() - 1)
+            out << ", ";
+        }
+        out <<"]]";
+        return out;
+    }
+
+    static std::vector<bu_cluster> history_;
+
     template <
         typename EXP
         >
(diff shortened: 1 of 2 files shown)


davisking:
Thanks for the PR.  This PR breaks the unit tests so it can't be merged.  It could be done as a separate function to avoid breaking the API.  You would also need to clearly write in the documentation what exactly this new output is.  Just saying "the history" isn't clear.

davisking:
I still don't understand what this does based on only reading the documentation.  Also, it's a somewhat computationally expensive operation which always executes even if the user doesn't ask for the history.  Users who don't want it shouldn't have to pay the computational cost.

mkdubik:
Right. I have made this requirement optional and improved the documentation (hopefully). It doesn't break the API and affect the computational time if not used. 

My motivation for this feature is that I'm implementing an algorithm that requires the dendrogram of the hierarchial clustering. This dendrogram along with another parameter,  gamma is used to choose the best partition of a network. The partition that maximizes the modularity.

dlib/clustering/bottom_up_cluster.h
```diff
@@ -44,24 +44,47 @@ namespace dlib
...
+        for (auto iter = c.orig.begin(); iter != c.orig.end(); iter++) {
+          out << *iter;
+          if (iter != c.orig.end() - 1)
+            out << ", ";
+        }
+        out <<"]]";
+        return out;
+    }
+
+    static std::vector<bu_cluster> history_;
```
davisking: You can't stuff a static variable like this into the header file.  I mean you can. But you shouldn't.

dlib/clustering/bottom_up_cluster_abstract.h
```diff
@@ -9,6 +9,31 @@ namespace dlib
...
 
 // ----------------------------------------------------------------------------------------
+    struct bu_cluster
+    {
+        long left;
+        long right;
+        long id;
+        std::vector<long> orig;
+    };
+    /*!  WHAT THIS OBJECT REPRESENTS:
```
davisking: I still don't understand what this is telling me.

mkdubik:
I don't see another way of introducing an optional parameter without breaking the API. Sure, you could introduce a new function but it would be a case of code duplication (unless I'm missing something here). 

I can't describe the functionality in simpler terms. What this parameter is is basically a dendrogram and if you don't know what that is, well there are a lot of data science books on that topic. I have decided instead to fork your project and move on because this is just a small part of my thesis. 

Again, thanks for the library!

davisking:
No worries.  Good luck with your thesis :)

PR #566, opened 2017-04-30
Update CMakeLists.txt

cclough:
Novice to C totally, embarrassed to admit that not switching to the build directory tripped me up. Compile seems to run, but no executable gets produced if still in examples directory.


davisking:
Oops, yeah, that's confusing.  Thanks for the PR :)

PR #574, opened 2017-05-04
Appveyor for Windows CI

andyli:
I've created a simple AppVeyor config for building the test and examples on Windows. The test is actually run, although I find them a little bit unstable - they sometimes fail.. e.g: https://ci.appveyor.com/project/andyli/dlib/build/5

Anyway, that can be improved separately later by e.g. having a fixed random seed.

Before merging this, please head to https://ci.appveyor.com, add this repo as a project.
You probably want to enable "Skip branches without appveyor.yml" in project setting.


davisking:
Cool, thanks for the PR.

I'm not sure why that test failed.  It doesn't on any other compiler and doesn't even fail for me when I test it in visual studio 2015.  That failure isn't attributable to randomness either since there is nothing random about that test and the magnitude of the error is very very large.  Something has probably gone horribly wrong with the code generation.

Issue #586, opened 2017-05-11
Deprecated C++ features actually removed from compiler


elelel:
Dlib uses some very old C++ features that are deprecated, like auto_ptr. Usually this led to compiler warnings that these features should not be used and this was 'fixed' just by quelling them:
https://github.com/davisking/dlib/blob/master/dlib/smart_pointers/shared_ptr.h#L13
Now under some scenarios these deprecated features became unavailable at all. For example, in Microsoft Studio 2017 when compiling with -std:c++latest option auto_ptr is simply non-existent identifier. 
This means DLib's header files can't be used in projects that use C++17 libraries/code unless it is compiled as a separate module wrapped into a compatible interface. Though the C++17 is not a standard yet, and neither it is known to me if Microsoft plans to get rid of auto_ptr for good, I think it still makes since to let people at least prototype in what might become C++17. I'd suggest that any features marked as deprecated in the standard should be used only when required (e.g. if an old compiler is detected) and not by default.
To solve this problem for at least in my own case I would like to replace the dlib's shared_ptr with stdlib's one. Is it safe to assume that I can just get away with replacing the implementation with standard one, or are there any behavioural difference between the two?

davisking:
There aren't any important differences between the two.  So if you want to
make a pull request that just completely replaces dlib's shared_ptr with
the std one that's fine with me.  You should probably put a using statement
in dlib's namespace though so that `dlib::shared_ptr` still works for old
code, but other than that I don't think there are any backwards
compatibility worries.

elelel:
I look through the smart pointers code and noticed that a conversion would not be that straightforward:

- dlib::shared_ptr interface is not fully compatible with std::shared_ptr. In particular, it exposes uinque() which is marked deprecated in C++17 for the reason it can't be exact in multithreaded environments. I don't know if they will also delete it as they did with auto_ptr
- auto_ptr is also used in shared_ptr_thread_safe which has no equivalent counterpart int stdlib. Perhaps switching auto_ptr to something like unique_ptr in the implementation makes sense, but I'm not sure about the current logic and interchangeability

davisking:
I know. But I suspect mightily that no one uses those features.  It's hard
to be sure.  But even so, everyone should switch to the std versions of
these things anyway.

The very few people who might be using them can just grab a copy of these
objects from an older version of dlib if they really want to keep using
them in the future.  WIth that in mind, it's probably best to avoid putting
any shared_ptr name in the dlib namespace.

[...]

PR #588, opened 2017-05-12
Replace shared_ptr/weak_ptr with STL's versions

elelel:
Current C++17 proposal mandates that some deprecated C++11 features should be removed, among which is std::auto_ptr which is used in DLib's smart pointer implementations. Using auto_ptr will prevent DLib from compile under default configuration when C++17 becomes default standard. To address the issue with dlib::shared_ptr and dlib::weak_ptr:

- All dlib::shared_ptr usage by the library is replaced with std::shared_ptr
- Tests and examples are modified accordingly
- smart_pointers.cpp test is excluded from CMakeLists.txt of test directory
- smart_pointers.h modified to include a warning that would help people to detect the usage and port their legacy programs, smart_pointers.h is removed from all internal includes
- <memory> header is included into the code when needed
- references to scoped_ptr which is not affected by C++17 changes are now made directly through smart_pointers/scoped_ptr.h

See also #586


davisking:
Nice work.  Thanks for the PR :)

PR #590, opened 2017-05-13
Replace shared_ptr_thread_safe with std::shared_ptr

elelel:
In an effort to make DLib compilable under C++17 shared_ptr_thread_safe is replaced with std::shared_ptr. See #586 and #588

### DIFF
--- a/dlib/gui_core/gui_core_kernel_1.cpp
+++ b/dlib/gui_core/gui_core_kernel_1.cpp
@@ -17,16 +17,17 @@
 #pragma comment (lib, "imm32.lib")
 #endif
 
-
+#include <cmath>
+#include <memory>
 #include <sstream>
+#include <vector>
+
 #include "../threads.h"
 #include "../assert.h"
 #include "../queue.h"
 #include "../sync_extension.h"
 #include "../queue.h"
 #include "../logger.h"
-#include <cmath>
-#include <vector>
 
 namespace dlib
 {
@@ -63,9 +64,9 @@ namespace dlib
 
     // ----------------------------------------------------------------------------------------
 
-        const shared_ptr_thread_safe<dlib::mutex>& global_mutex()
+        const std::shared_ptr<dlib::mutex>& global_mutex()
         {
-            static shared_ptr_thread_safe<dlib::mutex> m(new dlib::mutex);
+            static std::shared_ptr<dlib::mutex> m(new dlib::mutex);
             return m;
         }
 
@@ -118,7 +119,7 @@ namespace dlib
             // processing.
             thread_id_type event_thread_id;
 
-            shared_ptr_thread_safe<dlib::mutex> reference_to_global_mutex;
+            std::shared_ptr<dlib::mutex> reference_to_global_mutex;
 
             event_handler_thread(
             ) :
@@ -288,10 +289,10 @@ namespace dlib
         struct call_global_mutex { call_global_mutex() { global_mutex(); } };
         static call_global_mutex call_global_mutex_instance;
 
-        const shared_ptr_thread_safe<event_handler_thread>& global_data()
+        const std::shared_ptr<event_handler_thread>& global_data()
         {
             auto_mutex M(*global_mutex());
-            static shared_ptr_thread_safe<event_handler_thread> p;
+            static std::shared_ptr<event_handler_thread> p;
             if (p.get() == 0)
             {
                 p.reset(new event_handler_thread());
@@ -671,7 +672,7 @@ namespace dlib
             SetThreadPriority(hand,THREAD_PRIORITY_ABOVE_NORMAL);
             CloseHandle(hand);
 
-            shared_ptr_thread_safe<event_handler_thread> globals(global_data());
+            std::shared_ptr<event_handler_thread> globals(global_data());
 
             window_table_type& window_table = globals->window_table;
             HWND& helper_window = globals->helper_window;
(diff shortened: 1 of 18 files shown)


davisking:
Cool.  I'll look it over in a few hours hopefully.   The appveyor error is spurious.  I've seen it before and haven't figured out why that happens yet.  It only seems to happen on appveyor and only sometimes.  So you can ignore that.

elelel:
I think I've missed some includes, checking it by deleting the shared_ptr_thread_safe.h now and trying to compile.

elelel:
I think that fixes both shared_ptr and shared_ptr_thread_safe. I don't know why I missed them though. Maybe forgot to clean the cache and missed the warning or maybe they are not covered with tests/examples.

elelel:
Regarding the Appveyor test the problem, I would suggest explicitly importing the header file abs function is used from in every file that uses this function and also fully qualifying it with the namespace. C function abs is int-only, so if <cmath> overloads were not included the arguments may be autopromoted to ints from doubles which may lead to undesired results. So the reason for this particular case could be that I changed the order/set of headers included higher in the hierarchy somehow and now the wrong abs is used. That's just a guess though.

davisking:

davisking:
That's a good guess about abs().  I've had that problem be for to be sure.  But here it seems like something else.  This is the last time it happened: https://ci.appveyor.com/project/andyli/dlib/build/5 and there weren't any changes to any file in dlib but it was exactly the same error.   Also, the error is 62 which is way too large to be from integer rounding error, although it's conspicuous that it's an integer.  

In any case, I just added more outputs to the error message.  The next time it happens it should be more clear what's going on.

elelel:
Yes, it can't be that, because the arguments are not doubles actually. But there's another thing that could be fixed here with the same method - max. AppVeyour is used for Studio tests, and Studio header files declare max/min as a C macro. So as a rule of thumb it's always better to import std functions and use full qualifications, this way it at least fails to compile if it is a macro.

elelel:
Oh, I'm looking at wrong part of the test again :) there's no max/min there

davisking:
I know.  I've got all that stuff disabled.  But maybe something on appveyor's setup messes it up.  Still, the error seems to be intermittent.  The first test that failed then worked later on, even though it was exactly the same code.

davisking:
We will see what happens.  The messages I added should be a big clue as to what's happening next time.  Personally, I'm suspicious that visual studio is just broken somehow.

ruthans:
Hi,

I noticed that dlib (v19.8) doesn't compile under c++17 when I include dlib/gui_widgets.h  (I'm using visual studio 2017). After some further investigation I found that there are still a few references to shared_ptr_thread_safe in gui_core_kernel_1:

```
dlib::shared_ptr_thread_safe<gui_core_kernel_1_globals::event_handler_thread> globals; c:\dlib\gui_core\gui_core_kernel_1.h Line 172

const shared_ptr_thread_safe<event_handler_thread>& global_data()	c:\dlib\gui_core\gui_core_kernel_1.cpp Line 311

static shared_ptr_thread_safe<event_handler_thread> p; c:\dlib\gui_core\gui_core_kernel_1.cpp Line 314

```

To get around it, I followed the same fix as in PR #590. I.e. I replaced all occurrences of dlib::shared_ptr_thread_safe with std::shared_ptr and I also removed the include of shared_ptr_thread_safe.h in gui_core_kernel_1. After that, dlib compiles fine under c++17 and it seems to run fine as well. 

However, I then noticed in the comment for global_data() (gui_core_kernel_1.cpp) that there could potentially be problems using std::shared_ptr here instead of dlib::shared_ptr_thread_safe. 
I reckon this is the reason gui_core_kernel_1.h and gui_core_kernel_1.cpp were not updated and replaced with std::shared_ptr in this change?

davisking:
Yes, don't do that.  The comment explains why.

There shouldn't be any reason why dlib::shared_ptr_thread_safe can't be used in C++17.  A better approach would be to fix whatever is upsetting VC2017 about that class.

ruthans:
Apologies, I just saw that dlib::shared_ptr_thread_safe was indeed replaced with std::shared_ptr in this change but reverted later because of the found bug mentioned in my previous post. Please ignore it.

As it seems there is no easy way to use c++17 (safely) for a WIN32 env, I will revert the c++17 features I was using and go back to an older version for the time being.

elelel:
You don't have to give up C++17, you can just exclude gui_widgets from compilation if you are not using the part of Dlib, or anything that depends on it.

ruthans:
I realize that's an option, thanks for the advice elelel. In this case it would however be easier to remove the few c++17 specifics used rather than the dependency on gui_widgets.

Referring to Davis' point above "There shouldn't be any reason why dlib::shared_ptr_thread_safe can't be used in C++17. A better approach would be to fix whatever is upsetting VC2017 about that class.". 
I'm afraid I'm a bit of a c++ novice (coming from C#) but may I ask if you have any concern with replacing std::auto_ptr with std::unique_ptr in smart_pointers/shared_ptr_thread_safe.h? I don't see any issue with it myself reviewing the code. I know this was mentioned in #586 but I'm not sure exactly what the concern was/is, or if it was tried and failed.

elelel:
Replacing auto_ptr with unique_ptr is possible only if auto_ptr is not copied in the original code. The difference between auto_ptr and the unique_ptr is that the latter only has move semantics when used as a source of assignment, that what makes it unque. If it was never used in a code with copy semantics, I think it's safe to blindly replace it.

The other important thing if std::auto_ptr was reintroduced when rolling back my patches, then it's fully legitimate reason for the compiler to refuse compiling such code under C++17. The standard mandates such behaviour, and the only thing to fix whatever is upsetting VC2017 is to fix the standard. This is a rather new thing in C++ world, when a standard mandates removal of dangerous features.

davisking:
Oh right, I forgot auto_ptr was in there. I assume that's the error you are
getting?  I just removed that thing from the code.  If you get the latest
code from github it should all work in C++17 now.

ruthans:
Yes, that was the error, I should have mentioned that in the first post. I've pulled the the latest code and it works great under c++17. Thanks Davis!

davisking:
No problem, thanks for reporting this.

PR #591, opened 2017-05-13
Change type traits from inherited to explicit

elelel:
C++17 mandates deletion of inheritance-style traits from standard lib.
These are replaced with explicit typedefs

### DIFF
--- a/dlib/geometry/vector.h
+++ b/dlib/geometry/vector.h
@@ -1284,8 +1284,11 @@ namespace std
         Define std::less<vector<T,3> > so that you can use vectors in the associative containers.
     !*/
     template<typename T>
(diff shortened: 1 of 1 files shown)


elelel:
This is the last change to support C++17. At least it now works for my little project. Do you want me to change scoped_ptr to std::unique_ptr as we discussed in #586 ?

davisking:
Cool, thanks for another PR :)

Sure, if you want you can remove scoped_ptr too.  For that though you should add an alias template to the dlib namespace so that code that old user code that references dlib::scoped_ptr still compiles. I think that change should be transparent to users and not cause any frustrating incompatibilities.

PR #593, opened 2017-05-14
Migrate from dlib::scoped_ptr to std::unique_ptr

elelel:
See #586


dlib/test/smart_pointers.cpp
```diff
@@ -98,37 +98,37 @@ namespace
 
             DLIB_TEST_MSG(counter == 2,counter);
 
-            DLIB_TEST(p1);
```
davisking: Can you put a cast into the DLIB_TEST macro instead of putting them on each call?

dlib/smart_pointers.h
```diff
@@ -13,10 +13,21 @@
...
 #endif
 
-#include "smart_pointers/scoped_ptr.h"
+#include <memory>
+
 #include "smart_pointers/shared_ptr.h"
 #include "smart_pointers/weak_ptr.h"
 
+namespace dlib {
+  // Template alias for compatibility with clients using old dlib::scoped_ptr
```
davisking: Can you put this code into scoped_ptr.h instead?   Then keep the socket headers including scoped_ptr.h.  The reason being so that old dlib code that uses the sockets API will still compile.  With these current changes it looks like people who don't explicitly include the smart pointers header won't get the alias and then their socket code would fail to compile.

elelel: > With these current changes it looks like people who don't explicitly include the smart pointers header won't get the alias and then their socket code would fail to compile. 

Doesn't 0115f99 fix the problem with sockets? I've missed the sockets at first because CMake didn't want to recompile the socket tests and that came up only on Appveyor. Let me know if there's some other sockets usage and they're still need to be fixed

davisking:
Right. That's good but I mean people who have written scoped_ptr<connection> in their own code will have issues unless the scoped_ptr declaration makes it into their code, which means the sockets code needs to include it even though it doesn't use it.

elelel:
Ok, I'll do that. But by this logic any class that used to expose scoped_ptr in any part of non-private interface should reexport scoped_ptr.h through a local import

davisking:
Indeed. But I think only the sockets expose such interfaces. Can't check at the moment though because I'm posting from my phone. But I'm pretty sure there aren't other APIs that expose it.

davisking:

elelel:
In last Appveyor's log, besides the compiler running out of heap, there're C4503 warnings about exceeding name lengths. This could be because we replaced the old template with unique_ptr and perhaps is a real problem

davisking:
Na, it's because of an order of header inclusion change you made accidentally.  I've had a pragma in dlib that disables it for a while but the ordering change made it not trigger.  I fixed it and the warnings are gone.  

Also, the warning doesn't matter. See https://msdn.microsoft.com/en-us/library/074af4b6.aspx says that correct binaries are still produced even when this warning happens.

I also disabled parallel builds which seems to have stopped the out of memory problem.

elelel:
Cool! Then we're done it seems :) It's been pleasure to work with you and thanks for the legendary lib!

davisking:
You too, thanks for helping with dlib :)

Issue #595, opened 2017-05-16
Installed CMake target does not set include directory


tschw:
`target_link_library(my_project PRIVATE dlib::dlib)` does set the include directory as it's supposed to be.

The `INTERFACE_INCLUDE_DIRECTORIES` property should be set in `<install-root>/lib/cmake/dlib/dlib.cmake`, but the code for it is missing.

I'm not really sure what causes it, since `<src>/dlib/CMakeLists.txt` calls `target_include_directories` with `$<INSTALL_INTERFACE:include>` (that's what normally sets it). Maybe it's because it's listed under `INTERFACE` (in my projects, I use the `PUBLIC` section and I've never seen it fail - just as a guess - it's the only difference I can spot)?

tschw:
> Maybe it's because it's listed under `INTERFACE` 

Nope, tried that. Switching to `PUBLIC` doesn't fix the problem.

tschw:
Just had to pull it out of the `NOT DLIB_ISO_CPP_ONLY` conditional...

PR #597, opened 2017-05-16
Fix installed CMake targets when DLIB_ISO_CPP_ONLY (#595)

tschw:


davisking:
I assume you were enabling DLIB_ISO_CPP_ONLY?  I'm tempted to just remove the DLIB_ISO_CPP_ONLY option entirely.

tschw:
> I assume you were enabling DLIB_ISO_CPP_ONLY?

Yes, at first.

> I'm tempted to just remove the DLIB_ISO_CPP_ONLY option entirely.

I ended up just using different CMake options in my dependent build, so I'm fine without the patch.

However, I have found that option useful in the past, to quickly exclude most optional stuff: Dlib is rather header-heavy and only one of my components actually needed it, so I just compiled the parts of dlib plus every file that would include it in a single translation unit, so the headers would only need to be parsed once. That option (as a `#define`) was quite handy, but with this use case we're definitely in the "kids, don't try that at home" department.

davisking:
Ok.  I'll leave it in there :)

Anyway, thanks for the PR :)

PR #599, opened 2017-05-20
Move shape_predictor_trainer to a separate header

emptyVoid:
Moved shape_predictor_trainer to its own header in order to support the
use of shape_predictor with DLIB_ISO_CPP_ONLY defined (i.e. without
the API wrappers).


davisking:
Sweet, thanks for the PR :)

PR #603, opened 2017-05-24
Shape Predictor Members Protected

nidegen:
In order to derive shape predictor (we want to make a compressed dat file with custom serialization) we would need these to be protected. I believe it makes sense also generally to keep it that way.


davisking:
Why not submit a PR that changes the serialization code then? :)

nidegen:
currently, i serialize it with some other data. i will come back when i have time to untangle it:)

Issue #614, opened 2017-05-30
dnn_metric_learning_on_images_ex: Crash when using multi GPU and Loss increased.


sebcactus:
When using the example dnn_metric_learning_on_images_ex with 2 GPUs who have not the same amount of memory: 8GB and 11GB ;  the overall memory usage is 13GB which is nicely split evenly (~6500MB per GPU) on the two GPUs but would not fit on one GPU ; and the state is reloaded because the loss increased, I got the following issue
```
step#: 57758  learning rate: 0.1   average loss: 0.0182338    steps without apparent progress: 1619
Saved state to face_metric_sync
step#: 57843  learning rate: 0.1   average loss: 0.0183944    steps without apparent progress: 1622
step#: 57931  learning rate: 0.1   average loss: 0.0183579    steps without apparent progress: 1730
step#: 58019  learning rate: 0.1   average loss: 0.0188993    steps without apparent progress: 1887
step#: 58107  learning rate: 0.1   average loss: 0.0188605    steps without apparent progress: 1999
step#: 58195  learning rate: 0.1   average loss: 0.0184567    steps without apparent progress: 2122
step#: 58283  learning rate: 0.1   average loss: 0.0180592    steps without apparent progress: 2204
step#: 58371  learning rate: 0.1   average loss: 0.0185274    steps without apparent progress: 2275
step#: 58459  learning rate: 0.1   average loss: 0.0191554    steps without apparent progress: 2466
Loss has been increasing, reloading saved state from face_metric_sync
terminate called after throwing an instance of 'dlib::cuda_error'
  what():  Error while calling cudaMemcpy(dest.device_write_only(), src.device()+src_offset, num*sizeof(float), cudaMemcpyDeviceToDevice) in file /home/spiccand/git/dlib/dlib/dnn/gpu_data.cpp:62. code: 11, reason: invalid argument
Aborted (core dumped)
```

davisking:
Does this happen reliably, or just once?

sebcactus:
Systematically, when the Loss is increasing

davisking:
How are you creating the dnn_trainer?

davisking:
Well, actually, can you post a complete program and whatever is needed for
me to reproduce this problem?

sebcactus:
dlib version: commit 97ff8cb2a9256b78907b0b9ac3630b4a54490783
Here is the code.
The notable changes from your example:
- convert RGB to gray
- trainer devices {0, 1}
[...]

PR #629, opened 2017-06-06
Reinitialize averagers when saved sync file was reloaded.

plumtus:
Fix one bug in trainer that cause crash when using multi GPUs and loss increased. The details see https://github.com/davisking/dlib/issues/614.


davisking:
Sweet, thanks for the PR and finding this bug :)

PR #642, opened 2017-06-11
Patch 1

visionjo:
Check whether XQuartz was installed via brew  …
For UNIX builds, CMAKE now check whether xquartz was installed via brew cask. If so, a message is displayed suggesting to unistall brew cask version and reinstall from source. Then, NO_GUI_SUPPORT is set. 

Reasons for this pull request is that brew cask xquartz causes problems for python build (i.e., gui) and compiling single files with -X11 flag (somehow builds cpp examples, but not all work, e.g., face landmark detection)-- I checked that this was the case on my Macbook Pro (2016) running Sierra 10.12.5 (current version).

Let me know if there is a more efficient way of doing this (or coherent wrt your cmake file)... I am no expert (by any means) when it comes to CMake, but I do look for ways to learn more about it (so please, critique and or advise if there is something you would have done differently). 

Hope this helps :)


davisking:
Thanks for the PR :)

Just so I understand, the C++ examples did compile with the X11 version of brew but the GUI ones didn't run correctly?  What exactly happens?  If the C++ examples do work then the python API should be able to work as well.

PR #663, opened 2017-06-28
Upsampling and SpatialPadding DNN layers

OranjeeGeneral:
As mentioned earlier here implementation of two new DNN layers
Upsampling that allows you to upsample a tensor in three different ways:

*use nearest neighbour
*fill with zeroes
*use bilinear filtering

Another tool to build potential hourglass shape neural networks or encoder/decoder networks

The second one is a Spatial padding layer that allows a tensor to be padded in different ways

*Standard Zero padding
*Mirror padding
*Edge repeating padding


davisking:

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+            p.y() = p.y()*scale_y();
+            return p;
+        }
+
+        upsample_ (
+            const upsample_& item
+        )  
+        {
+            (void)item;
+            // this->conv is non-copyable and basically stateless, so we have to write our
```
davisking: There isn't a this->conv in this class.

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+        }
+
+        upsample_& operator= (
+            const upsample_& item
+        )
+        {
+            if (this == &item)
+                return *this;
+
+            // this->conv is non-copyable and basically stateless, so we have to write our
```
davisking: no this->conv

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+
+        template <typename SUBNET>
+        void forward(const SUBNET& sub, resizable_tensor& output)
+        {
+            unsigned int gnr = _scale_y * sub.get_output().nr();
+            unsigned int gnc = _scale_x * sub.get_output().nc();
+            unsigned int gnsamps = sub.get_output().num_samples();
+            unsigned int gk = sub.get_output().k();
+            output.set_size(gnsamps,gk,gnr,gnc);
+            output = 0;
```
davisking: Doing the =0 and a separate forward call will launch multiple GPU kernels.  It would be faster if ups.forward() didn't require it's input to be zero.  Is there a reason that isn't feasible?

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+            }
+        }
+
+
+        friend std::ostream& operator<<(std::ostream& out, const upsample_& item)
+        {
+            out << "upsample\t ("
+                << "scale_y="<<_scale_y
+                << ", scale_x="<<_scale_x
+                << ", method="<<_method
```
davisking: _method isn't going to print right since it has a char type.  Really, _method should be an integer, not a char type.

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+        }
+
+        spatialpadding_& operator= (
+            const spatialpadding_& item
+        )
+        {
+            if (this == &item)
+                return *this;
+
+            // this->conv is non-copyable and basically stateless, so we have to write our
```
davisking: no this->conv

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+
+        template <typename SUBNET>
+        void forward(const SUBNET& sub, resizable_tensor& output)
+        {
+            unsigned int gnr = sub.get_output().nr() + 2 * padding_y();
+            unsigned int gnc = sub.get_output().nc() + 2 * padding_x();
+            unsigned int gnsamps = sub.get_output().num_samples();
+            unsigned int gk = sub.get_output().k();
+            output.set_size(gnsamps,gk,gnr,gnc);
+            output = 0;
```
davisking: Same question here about why pad.forward() needs to have zeroed inputs.  Can't pad.forward() do an assignment rather than an accumulate to the output?

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+            pad.forward(output,sub.get_output(),_padding_y,_padding_x,_method);
+        } 
+
+        template <typename SUBNET>
+        void backward(const tensor& gradient_input, SUBNET& sub, tensor& params_grad)
+        {
+            (void)params_grad;
+            resizable_tensor temp;
```
davisking: You shouldn't allocate tensors here.  That will make the solver run much slower as GPU allocation overhead is very significant.  Why can't pad.backward() be written so that it doesn't need this temporary here?

dlib/dnn/layers.h
```diff
@@ -594,9 +594,334 @@ namespace dlib
...
+            }
+        }
+
+
+        friend std::ostream& operator<<(std::ostream& out, const spatialpadding_& item)
+        {
+            out << "spatialpadding\t ("
+                << "padding_y="<<_padding_y
+                << ", padding_x="<<_padding_x
+                << ", method="<<_method
```
davisking: Same issue with _method being a char.  It's not going to print correctly.  It would be better if _method was globally changed to int.

dlib/test/dnn.cpp
```diff
@@ -767,6 +767,53 @@ namespace
...
+        up.backward(gradient1,output1,2,2,1);
+        up2.backward(gradient2,output2,2,2,1);        
+        dlog << LINFO << "backward error: "<< max(abs(mat(gradient1)-mat(gradient2)));
+        DLIB_TEST_MSG(max(abs(mat(gradient1)-mat(gradient2))) < 1e-3, max(abs(mat(gradient1)-mat(gradient2))));
+        up.backward(gradient1,output1,2,2,2);
+        up2.backward(gradient2,output2,2,2,2);        
+        dlog << LINFO << "backward error: "<< max(abs(mat(gradient1)-mat(gradient2)));
+        DLIB_TEST_MSG(max(abs(mat(gradient1)-mat(gradient2))) < 1e-3, max(abs(mat(gradient1)-mat(gradient2))));
+    }
+
```
davisking: There should also be a few of the test_layer() calls for the new layers (put with the rest of the test_layer calls in this file).

dlib/dnn/layers.h
OranjeeGeneral: Same reason as with the upsample it is just for the zero padding and only is required for that mode.

dlib/dnn/layers.h
OranjeeGeneral: Okay yes makes sense. Will see I can't remember why I wrote it that way tbh. :)

dlib/dnn/layers.h
davisking: Maybe.  I find that most of my cuda kernels are limited more by how much and in what pattern I touch GPU memory, rather than by the particular operations inside the kernel.

dlib/dnn/cuda_dlib.cu
```diff
@@ -1172,6 +1177,476 @@ namespace dlib
...
+                                                size_t nr2,                                                
+                                                size_t nc2,
+                                                size_t scale_y, 
+                                                size_t scale_x)
+        {
+
+            for (auto r : grid_stride_range_y(0, nr))
+            {
+                auto srcR = r/scale_y;
+                auto srcP = in + srcR * nc2;
```
davisking: I just looked at the CUDA code.  These kernels would be a lot faster if, rather than launching many of them you processed the entire tensor with one grid stride loop.  You also don't need the grid_stride_range_y loop.  You will get better hardware occupancy with one grid stride loop over everything.  The only time you need 2D grid stride loops is when you need to do something complex that has hardware limits on how the warps are structured.  For instance, I think the only time I've needed 2D grid stride loops is when doing complex warp reductions, which you aren't using here (and don't need).

dlib/test/dnn.cpp
davisking: I just ran test_layer() on the bilinear upsampling layer and it doesn't output the right gradients :/

davisking:
FYI, I've got an experiment I wanted to run that needed a bilinear upsampling layer.  So I just added that to dlib.   Still need all the other layers in this PR though :)

OranjeeGeneral:
Okay great unfortunately have been busy with other things I will get back to this hopefully soon.

davisking:
No worries :)

OranjeeGeneral:
Okay I fixed up upsample_ with your changes, so the three modes are back, bilinear, nearest neighbour and fill with zeros.

I still have to adjust SaptialPadding to some of your change requests and add some tests

OranjeeGeneral:
Weird that test didn't seem to have failed on my checkout I usually always run the testsuite before pushing need to investigate hadn't time lately.

OranjeeGeneral:
Hmm I tried it again this time with a Visual Studio checkout I don't get this test failure weird. Unless wait a minute might it probably might be running the CPU version?

davisking:
Yes, these automatic tests run the CPU version.  I'm not sure any of these services support testing GPU code at all.

OranjeeGeneral:
Don't think this is worth persuing any longer. A general Dilation layer would be more useful. Closing this for now

PR #666, opened 2017-06-30
Problem: Visual Studio's vcpkgsrv.exe constantly uses a single CPU core

reunanen:
**Problem:** Visual Studio's `vcpkgsrv.exe` constantly uses a single CPU core, apparently never finishing whatever it's trying to do. Moreover, this issue prevents some operations like switching from Debug to Release (and vice versa) in the IDE. (Your mileage may vary.)

**Workaround:** Keep manually killing the `vcpkgsrv.exe` process.

**Solution:** Disable IntelliSense for some files. Which files? Unfortunately this seems to be a trial-and-error process.

NB: Visual Studio needs to be restarted for this fix to help. (Otherwise `vcpkgsrv.exe` keeps on popping up.)

### DIFF
--- a/dlib/dnn.h
+++ b/dlib/dnn.h
@@ -11,7 +11,18 @@
 
 #include "dnn/tensor.h"
 #include "dnn/input.h"
+
+// Problem:    Visual Studio's vcpkgsrv.exe constantly uses a single CPU core,
+//             apparently never finishing whatever it's trying to do. Moreover,
+//             this issue prevents some operations like switching from Debug to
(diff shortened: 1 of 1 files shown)


OranjeeGeneral:
Great this has been bugging me for a long time too!

davisking:
Lol, nice.  Thanks for the PR :)

reunanen:
Looks like this maybe wasn't quite enough for all cases. Probably need to disable IntelliSense for some more headers. Will look into it.

PR #678, opened 2017-07-03
Disable IntelliSense for all DNN unit tests

reunanen:
Disable IntelliSense even for the local ResNet declarations in the DNN tests.

(See #666.)

### DIFF
--- a/dlib/dnn.h
+++ b/dlib/dnn.h
@@ -11,7 +11,18 @@
 
 #include "dnn/tensor.h"
 #include "dnn/input.h"
+
+// Problem:    Visual Studio's vcpkgsrv.exe constantly uses a single CPU core,
+//             apparently never finishing whatever it's trying to do. Moreover,
+//             this issue prevents some operations like switching from Debug to
+//             Release (and vice versa) in the IDE. (Your mileage may vary.)
+// Workaround: Keep manually killing the vcpkgsrv.exe process.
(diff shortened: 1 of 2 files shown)


davisking:
You sure you don't want to just wrap that whole file in those statements?  I doubt anyone cares much about getting IDE completions in that code.

reunanen:
No, I'm not sure about anything :) ... But I think I would like to keep it focused, because this file may also serve as a kind of documentation for other code (where people may care a lot). I mean, I guess the whole `vcpgksrv.exe` problem is no big deal at all from the unit tests point of view, however solving it here may show an example how the issue can perhaps be circumvented in some production code as well. But up to you, really.

At least for me, the current version now seems to have solved the problem for the unit tests.

davisking:
It's a little less cluttered if it's around the whole thing, and less likely to trigger the problem later as well.  Also, I realize that sometimes people talk about unit test code being documentation, but I'm violently opposed to that idea.  Seeing code examples is not remotely sufficient to document an API, neither from a completeness (what all can it do?) nor from a correctness (what things are forbidden?) point of view.  So honestly, if there was a way I could intentionally sabotage attempts to use unit tests as sources of documentation I would do it :)

Anyway, if you put the guards around the whole file I'll merge the PR :).  At  the very least, it will prevent the problem from reoccurring when more tests are added later.

reunanen:
Yep, I don't disagree. Changed it like that.

davisking:

PR #685, opened 2017-07-06
Add new loss for weighted pixel inputs

reunanen:
Add new loss for arbitrarily weighted pixel inputs. May be useful e.g. to emphasize rare classes.

I don't know if this is interesting to anyone but myself. If this is a candidate to be merged, then I can add some test case and an entry to `loss_abstract.h`. If not, then the PR can well be closed – won't be a problem for me.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1549,11 +1549,11 @@ namespace dlib
             typename SUB_TYPE,
             typename label_iterator
             >
-        void to_label (
+        static void to_label (
             const tensor& input_tensor,
             const SUB_TYPE& sub,
             label_iterator iter
-        ) const
+        )
         {
             DLIB_CASSERT(sub.sample_expansion_factor() == 1);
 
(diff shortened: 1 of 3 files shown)


dlib/dnn/loss.h
```diff
@@ -1706,6 +1706,187 @@ namespace dlib
...
+
+            // In semantic segmentation, 65536 classes ought to be enough for anybody.
+            uint16_t label = 0;
+            float weight = 1.f;
+        };
+
+        typedef matrix<weighted_label> training_label_type;
+        typedef matrix<uint16_t> output_label_type;
+
+        // This function is identical to loss_multiclass_log_per_pixel_::to_label.
```
reunanen: Should this be deduplicated?

dlib/dnn/loss.h
```diff
@@ -1678,7 +1678,7 @@ namespace dlib
             std::string version;
             deserialize(version, in);
             if (version != "loss_multiclass_log_per_pixel_")
-                throw serialization_error("Unexpected version found while deserializing dlib::loss_multiclass_log_.");
+                throw serialization_error("Unexpected version found while deserializing dlib::loss_multiclass_log_per_pixel_.");
```
reunanen: At least this could be fixed in `master`.

davisking: Yeah, definitely.

dlib/dnn/loss.h
davisking: Yeah, that would be a good idea.

davisking:
Sure, it seems like a nice thing to have.  So if you want to do the things you suggested it would be a nice PR to merge :)

reunanen:
@davisking
Added the things now.

davisking:

PR #687, opened 2017-07-06
added check for libpython_version#m.dylib present in some virtual env…

IanPhilips:
…ironments


davisking:
Sweet, thanks for the PR :)

PR #690, opened 2017-07-08
Add per-pixel mean square loss

reunanen:
Add per-pixel mean square loss. Something like this may be useful for convolutional autoencoders, for example.

Yeah, I'm actually using this one too! But again: if this is not generally interesting, then there's no need to merge to master. On the other hand: if it is, then I can again add a test and some documentation.

For generality, I guess there could be a `loss_mean_squared_multioutput_per_pixel` as well. And why not an `rgb_pixel` version also. However, they are beyond the scope of this PR.


davisking:
It seems like something that might get used.  So if you want to make a full PR out of it I'm all for it :)

dlib/test/dnn.cpp
```diff
@@ -1997,6 +1997,76 @@ namespace
...
+
+        matrix<float> tmp(output_width, output_height);
+        for (int i = 0; i < num_samples; ++i)
+        {
+            const int model = distribution(generator);
+
+            for (int r = 0; r < output_height; ++r)
+                for (int c = 0; c < output_width; ++c)
+                    switch (model) {
+                    case 0: tmp(r, c) = r / output_height;
```
davisking: You probably meant to but break statements here right?

dlib/dnn/loss.h
```diff
@@ -1848,6 +1848,138 @@ namespace dlib
...
+
+            const float* out_data = output_tensor.host();
+            for (long i = 0; i < output_tensor.num_samples(); ++i, ++iter)
+            {
+                iter->set_size(output_tensor.nr(), output_tensor.nc());
+                for (long r = 0; r < output_tensor.nr(); ++r)
+                {
+                    for (long c = 0; c < output_tensor.nc(); ++c)
+                    {
+                        iter->operator()(r, c) = out_data[tensor_index(output_tensor, i, r, c, 0)];
```
davisking: I also just noticed that tensor_index goes sample,row,col,k.  It should really go sample,k,row,col to avoid confusion with the rest of the code since all other functions that deal in those values are ordered that way.

dlib/test/dnn.cpp
reunanen: Yes indeed I did! No wonder the tests passed with flying colors. =)

dlib/dnn/loss.h
reunanen: Oh, ok. Btw, is there already somewhere a function like `tensor_index` that I should be using instead?

davisking:

PR #691, opened 2017-07-08
Add methods pipe::has_blocked_dequeues and trainer::is_training.

reunanen:
The latter uses the former. It may be useful e.g. if you are doing online training and need to know when a new trained network is readily available (without having to block in get_net).

Again: I can add documentation, if this is seen generally useful and in line with the overall design of the library.

### DIFF
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -564,6 +564,11 @@ namespace dlib
             return train_one_step_calls;
         }
 
+        bool is_training() const
+        {
+            return !job_pipe.has_blocked_dequeues();
+        }
+
     private:
 
         void record_test_loss(double loss)
--- a/dlib/pipe/pipe_kernel_1.h
+++ b/dlib/pipe/pipe_kernel_1.h
@@ -82,6 +82,9 @@ namespace dlib
             unsigned long num
         )const;
 
+        bool has_blocked_dequeues (
+        ) const;
+
         void enable (
         );
 
@@ -664,6 +667,18 @@ namespace dlib
             unblock_sig.broadcast();
 
         --unblock_sig_waiters;
+    }
+
+// ----------------------------------------------------------------------------------------
+
+    template <
+        typename T
+        >
+    bool pipe<T>::
+    has_blocked_dequeues () const
+    {
+        auto_mutex M(m);
+        return pipe_size == 0 && dequeue_waiters > 0;
     }
 
 // ----------------------------------------------------------------------------------------


davisking:
There is a race condition where the training code is busy and not blocked on the pipe. So it will look like it's not training when really it is. There is a similar problem where the queue is non empty but the thread is calling deque but hasn't dequeued yet.

reunanen:
Ok. I'll close the PR. Do you think something like `try_get_net` (similar to [`try_lock`](http://en.cppreference.com/w/cpp/thread/mutex/try_lock)) would make sense instead? Or something else perhaps?

I just want to know when the training step has finished so I can copy the net, without blocking. Maybe there's already a way how it should be done, which I just didn't find?

reunanen:
I mean: Suppose you have new samples coming in all the time. You want to train them, but you also want to feed-forward the samples through a recent trained net.

I guess I could explicitly spin a separate thread just for the training; then that thread could easily block.

davisking:
I have a hard time seeing how that kind of try statement on the trainer doesn't lead to users creating race conditions in their programs. It's a lot less error prone to create a thread and let it block.    The coding pattern you are advocating has complex thread synchronization issues in it that I know for certain most dlib users won't understand or even recognize at all. I mean. What does the code do if the try fails?  I can see a lot of bad things here like busy waiting or doing things that make the program run way slower.  Or doing things that create race conditions.

reunanen:
Yeah, I agree. I'm ok with creating a thread.

It's just that I very much like the thread structure of your example programs, regarding how they load the data and train the net. I guess I was just trying to follow that style, maybe not recognizing that I should really be doing it differently in my case. No problem -- I'll just create a thread that can block.

(This PR is already closed.)

> I mean. What does the code do if the try fails?

It would just have used the previous version of the net (trained for fewer steps).

davisking:
Where does the previous version come from?  Is that why you were asking about in memory serialization in the other topic?  To create a copy?  You should be able to use the normal = operator to make copies of a net.

reunanen:
> Where does the previous version come from?

The previous version is a result of the same training process at an earlier time _t0_ (say, 9:00). It would be used to process new samples until another training iteration is completed at time _t1_ (say, 9:01). At _t1_, the version from _t0_ can be ditched, and the version from _t1_ can be used until yet another iteration is completed at _t2_ (say, 9:02), and so on.

Anyway, as discussed already, I can handle this by creating another thread for the training, and passing trained net objects (e.g. using `dlib::pipe`, which looks convenient, although I haven't used it yet) from that thread to where the net's `operator()` is called. Shouldn't be a problem, I think.

> Is that why you were asking about in memory serialization in the other topic? To create a copy? You should be able to use the normal = operator to make copies of a net.

No, the other topic is not related to this one. In that other topic, I just want to pass the trained network over the wire to another node (computer), where it will be used to process new data. Alternatively, I might want to save the serialized network to a database, or embed it in some other serialization. In all these cases I can first serialize to file (as is done in the examples), and then read the file -- and indeed that is exactly what I was doing first. But surely it is better to avoid such a file operation, when it is not really needed.

davisking:
You know the dnn_trainer stores a reference to the net given to it's constructor right?  So if you wanted to do something like this you will have to be making a copy of the net object and passing that copy to another thread.  Additionally, you have to think about how this maps onto your hardware resources.  I'm not sure it makes sense in most cases.  The training process will usually saturate your hardware, so it's not like you can really be running a copy of the net while training a separate copy, unless you have multiple GPUs and you are taking care to synchronize the memory (i.e. the model parameters) of this pair of network objects between the two GPUs.  


Ok, I was just making sure you weren't using serialize as a wacky operator=.  You would be surprised how often I've heard of people doing that.  For future readers of this post, **don't use serialize() when you could be using operator=.** :)

reunanen:
Yes, I'm aware of this. However, thanks for all the warnings -- much appreciated!

davisking:
Ok, well, if you know what you are doing then go for it :)

reunanen:
For what it's worth, I fixed this on the call site as follows:

First start training as usual, but create a `future` on the result:

    using namespace std:
    future<net_type> f = async(launch::async, [&my_trainer]() { return my_trainer.get_net(); });

Then later check if the training is ready as follows (and if not, then just keep using an earlier `my_processing_net`):

    if (f.wait_for(chrono::seconds(0)) == future_status::ready) {
        my_processing_net = f.get();
    }

(Note that `my_trainer` doesn't train `my_processing_net` directly, but another net, which is not shown above.)

The `wait_for(seconds(0)) == ready` is essentially the `!is_training()` that I was looking for.

I think this pretty much does what I wanted in the first place: I never need to block where `my_processing_net` is used, yet I get the new training result in use there without unnecessary delay. Moreover, there's not too much thread-related boilerplate: my main loop just receives new data, initiates a new training iteration if enough new data has been accumulated for a minibatch or so (the training of course goes to a separate thread, which dlib makes very clean), checks if there's a new training result available and updates `my_processing_net` if so (this is the second snippet above), and finally uses `my_processing_net` to process the new data (not shown above).

PR #692, opened 2017-07-08
Add test for serializing DNNs

reunanen:
I wanted to do as follows:

    net_type net, net2;
    // ...
    std::ostringstream out;
    serialize(out) << net;
    const std::string serialized = out.str();
    // ...
    std::istringstream in(serialized);
    deserialize(in) >> net2;

(In other words, I wanted to serialize DNNs, but in memory and not via files.)

I couldn't figure out a way to do this without adding this code somewhere: otherwise, i.e. with `serialize(net, out);` and `deserialize(net2, in);`, MSVC++ would just complain that it can't match `net` and `net2` (didn't try other compilers). First added the new code on the call site, but as it apparently has to be in the `dlib` namespace, thought it'd be better to have it here.

Perhaps I was doing it all wrong to begin with, but at least this addition did let me do what I wanted to do, so thought I'd share it.

### DIFF
--- a/dlib/test/dnn.cpp
+++ b/dlib/test/dnn.cpp
@@ -2486,6 +2486,32 @@ namespace
         }
     }
 
+    void test_serialization()
+    {
+        print_spinner();
+
+        using net_type = loss_mean_squared<fc<1, input<matrix<double>>>>;
+        net_type net, net2;
+
+#if 0
+        // At least on MSVC, this fails with compiler error:
+        // "C2665: 'dlib::serialize': none of the 46 overloads could convert all the argument types"
+        {
+            std::ostringstream out;
+            dlib::serialize(net, out);
+            const std::string serialized = out.str();
+            std::istringstream in(serialized);
+            dlib::deserialize(net2, in);
+        }
+#endif
+
+        // This at least compiles (with #692):
+        std::ostringstream out;
+        dlib::serialize(out) << net;
+        const std::string serialized = out.str();
+        std::istringstream in(serialized);
+        dlib::deserialize(in) >> net2;
+    }
 
 // ----------------------------------------------------------------------------------------
 
@@ -2565,6 +2591,7 @@ namespace
             test_loss_multiclass_per_pixel_outputs_on_trivial_task();
             test_loss_multiclass_per_pixel_with_noise_and_pixels_to_ignore();
             test_loss_multiclass_per_pixel_weighted();
+            test_serialization();
         }
 
         void perform_test()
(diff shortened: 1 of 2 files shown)


reunanen:
As usual, can add a unit test and documentation, if this makes sense in general.

davisking:
You don't need this code. Or shouldn't anyway. I bet you just forgot to put dlib:: in front of the serialize calls?

reunanen:
No, I don't think I forgot that.

Ok, I'll create a test case for this? I guess it could be another call site problem.

davisking:
Well. You should be able to call serialize like normal. If not then that's a huge bug in visual studio or dlib.

reunanen:
I guess I'm doing something wrong then. Will be back shortly.

dlib/test/dnn.cpp
```diff
@@ -2486,6 +2486,32 @@ namespace
...
+
+        using net_type = loss_mean_squared<fc<1, input<matrix<double>>>>;
+        net_type net, net2;
+
+#if 0
+        // At least on MSVC, this fails with compiler error:
+        // "C2665: 'dlib::serialize': none of the 46 overloads could convert all the argument types"
+        {
+            std::ostringstream out;
+            dlib::serialize(net, out);
```
reunanen: I guess this is somehow wrong then?

reunanen:
I was able to reproduce the problem as a minimal test case, which I have now added to the PR.

dlib/test/dnn.cpp
davisking: That code should compile. It's probably a bug in visual studio. There are often name lookup bugs in that compiler of this sort. The thing to do is to try and figure out why it's refusing to call serialize rather than add some proxy. 

You should check in gcc and see what happens. I would do it now but I only have my phone :/

reunanen: It seems that the gcc on Travis does not compile this line either:
https://travis-ci.org/davisking/dlib/jobs/251527618#L639

reunanen:
Ok, I do admit that your fix is more elegant than mine. :) ... And yes, now it works -- thanks a bunch!

I'm leaving the PR open; it now contains only the test that would have caught this earlier, in case you want to merge it. If not, I'll close the PR.

davisking:
No problem, thanks for reporting this bug :)

Good idea, I'll merge the PR once the integration tests finish.

reunanen:
It's a linker error for undefined reference -- as if a cpp file was missing from the build -- but the serialization should be all in the headers, right? Works locally on my Visual Studio, let's see what AppVeyor says.

davisking:
Ha, well, visual studio does the sensible thing and gcc doesn't for once :)

Anyway, I just pushed a workaround to git that should fix it.  I also included the test code so we don't need to merge this PR anymore.  I also included one call that uses dlib:: and one without to make sure the qualified name lookup works as well as the ADL based lookup.

PR #695, opened 2017-07-09
cuDNN convolution algorithms shared workspace

e-fominov:
This PR makes convolution layers in GPU mode to share workspace memory buffers. This leads to significant memory usage decrease with fast convolution algorithms

Existing dlib implementation of large networks (Resnet34) is using "prefer_smallest" cuDNN convolution algorithms to make it process large batches with same GPU RAM.
With this PR we are now able to use fast convolution algorithm with near the same RAM usage (about 100 Mb additional RAM required)

I have tested it with GTX TitanX (Maxwell/12GB) and GTX 1080 Ti (Pascal/11GB). TitanX is now able to train imagenet with batch size = 160 with fast algorithms (same batch size as smallest algorithms), and 1080 is able to train batch_size = 155

Training time:   
  * TitanX, smallest - 1.6 second per batch of 160 images
  * TitanX, fast  - 0.9 second per batch of 160 images
  * 1080Ti, smallest - 1.4 second per batch of 155 images
  * 1080Ti, fast - 0.7 second per batch of 155 images

All tests with Cudnn 6 and Cuda 8

So overall picture seems like 2x training performance improvement for large networks

Accuracy not changed - tested with MNIST example with multi-gpu mode

I am sure, someone can find more elegant solution, but this should be good as a starting point

Reading CUDNN documentation I have found that RNN networks have "reserve size" that should be kept unchanged between forward and backward passes, but workspace for all agorithms seems to be safely reusable

Also we should know, that backward algorithms require more RAM than forward. So forward-only (production) networks can be optimized even more for RAM usage

This PR changes very internal core things in Dlib, so it should be reviewed carefully before merging

### DIFF
--- a/dlib/dnn/cuda_data_ptr.h
+++ b/dlib/dnn/cuda_data_ptr.h
@@ -149,6 +149,32 @@ namespace dlib
 
     // ------------------------------------------------------------------------------------
 
+        class resizable_buffer
+        {
+            /*!
+                WHAT THIS OBJECT REPRESENTS
+                    This is a block of memory on a CUDA device that will be automatically
+                    resized if requestes size is larger than allocated
+            !*/
+        public:
+            cuda_data_void_ptr get(size_t size)
+            /*!
+                ensures
+                    - This object will return the buffer of requested size of larger
+                    - buffer.size() >= size
+            !*/
+            {
+                if (buffer.size() < size)
+                {
+                    buffer.reset();
+                    buffer = cuda_data_void_ptr(size);
+                }
+                return buffer;
+            }
+        private:
+            cuda_data_void_ptr buffer;
+        };
+
     }
 }
 
--- a/dlib/dnn/cuda_data_ptr.h
+++ b/dlib/dnn/cuda_data_ptr.h
@@ -149,7 +149,7 @@ namespace dlib
 
     // ------------------------------------------------------------------------------------
 
-        class resizable_buffer
+        class resizable_cuda_buffer
         {
             /*!
                 WHAT THIS OBJECT REPRESENTS
--- a/dlib/dnn/cudnn_dlibapi.cpp
+++ b/dlib/dnn/cudnn_dlibapi.cpp
@@ -117,17 +117,54 @@ namespace dlib
             thread_local cudnn_context c;
             return c.get_handle();
         }
+    // ------------------------------------------------------------------------------------
 
-        static std::shared_ptr<resizable_buffer> device_global_buffer()
+        class cudnn_device_buffer
         {
-            thread_local std::weak_ptr<resizable_buffer> weak_buff;
-            std::shared_ptr<resizable_buffer> buff = weak_buff.lock();
-            if (!buff)
+        public:
+            // not copyable
+            cudnn_device_buffer(const cudnn_context&) = delete;
+            cudnn_device_buffer& operator=(const cudnn_context&) = delete;
+
+            cudnn_device_buffer()
             {
-                buff = std::make_shared<resizable_buffer>();
-                weak_buff = buff;
+                buffers.resize(16);
             }
-            return buff;
+            ~cudnn_device_buffer()
+            {
+            }
+
+            std::shared_ptr<resizable_cuda_buffer> get_buffer (
+            )
+            {
+                int new_device_id;
+                CHECK_CUDA(cudaGetDevice(&new_device_id));
+                // make room for more devices if needed
+                if (new_device_id >= (long)buffers.size())
+                    buffers.resize(new_device_id+16);
+
+                // If we don't have a buffer already for this device then make one
+                std::shared_ptr<resizable_cuda_buffer> buff = buffers[new_device_id].lock();
+                if (!buff)
+                {
+                    buff = std::make_shared<resizable_cuda_buffer>();
+                    buffers[new_device_id] = buff;
+                }
+
+                // Finally, return the buffer for the current device
+                return buff;
+            }
+
+        private:
+
+            std::vector<std::weak_ptr<resizable_cuda_buffer>> buffers;
+        };
+
+
+        static std::shared_ptr<resizable_cuda_buffer> device_global_buffer()
+        {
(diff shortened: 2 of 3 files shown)


davisking:
Nice!  That's a good idea.  Now that you point it out I obviously should have implemented it this way in the first place :/

Anyway, I looked over the code and I think I see a problem.  The memory buffers need to be more than just thread_local, since you can switch the active cuda device in a thread.  If you just grab the thread local buffer you might end up with a buffer on the wrong GPU device.  You should be able to deal with this problem with the pattern I use for the cudnn context object: https://github.com/davisking/dlib/blob/master/dlib/dnn/cudnn_dlibapi.cpp#L115

In particular, imagine there was a class like this:
```
class resizable_buffer
{  public:
   cuda_data_void_ptr get(size_t size);
   /* returns a pointer to at least size bytes, will reallocate the underlying buffer if necessary.  
      Importantly, cuda_data_void_ptr tracks if anyone is still using memory from a previous 
      call to get(), so when/if we do a reallocation we don't free any memory buffer being used. 
 
      The idea here is that you don't store the output of get() anywhere other than locally in
      a function.  That is, you call get() each time you need the buffer for a call to cuDNN. */
  private: 
    cuda_data_void_ptr buffer; // get() just returns this, if it's big enough, otherwise it reallocates.
};

static std::shared_ptr<resizable_buffer> device_global_buffer(); 
```

`device_global_buffer` could be implemented similarly to how the cudnn context() method I mentioned works.  You call device_global_buffer to get the global resizable_buffer appropriate for your thread and device.  You also store the shared_ptr<resizable_buffer> as a private member of the tensor_conv so that once all the tensor_conv's are destroyed the resizable_buffer is freed (this assumes you use weak_ptrs inside the device_global_buffer implementation). 

As an aside, you can use the dlib::cuda::cuda_data_void_ptr to allocate GPU buffers, rather than dealing with the CUDA C API directly.

e-fominov:
One more thing I want to make too - remove memory fragmentation. Now I am allocating memory at first layer, than reallocating it at second (if larger)...

Is it possible to estimate workspace size of all layers, before running first convolution? May be dnn_trainer can call "setup" of all layers?

And about forward-only networks. May be better have two workspaces? One for forward running and if there will be "backward" call, it will be reallocated into larger one?

davisking:
Maybe.  You should benchmark and see if there is any significant effect of memory fragmentation.  I assume the GPUs have a TLB similar to CPUs which would mostly mitigate any effect of fragmentation, at least from what we are talking about.  But if you can show a difference then maybe. But it's hard to see how  to do this without intrusive changes to the API.  

Yeah, I would say to reuse the same workspace buffer for forward and backward.  Again, not sure why I didn't put just one workspace buffer in the code already :/

e-fominov:
Updated the code, not it is more simple with using your recommendations

dlib/dnn/cuda_data_ptr.h
```diff
@@ -149,6 +149,32 @@ namespace dlib
 
     // ------------------------------------------------------------------------------------
 
+        class resizable_buffer
```
davisking: On second thought, maybe resizable_cuda_buffer is a better name.

dlib/dnn/cudnn_dlibapi.cpp
```diff
@@ -118,6 +118,17 @@ namespace dlib
             return c.get_handle();
         }
 
+        static std::shared_ptr<resizable_buffer> device_global_buffer()
+        {
+            thread_local std::weak_ptr<resizable_buffer> weak_buff;
```
davisking: You need to add a call to cudaGetDevice() here (like on line 97 above) to make sure you are returning a resizable_buffer that's active on the currently selected cuda device.  Otherwise there is still the problem of someone switching the active thread's cuda device and then trying to use dlib, which would cause a crash or other huge problem with this code as it is.

dlib/dnn/cuda_data_ptr.h
davisking: Well, it's in the cuda namespace.  Nevermind :)

davisking: Well, but cuda_data_void_ptr has cuda in the name.  Maybe for API consistency it should be resizable_cuda_buffer.

e-fominov:
Code is now updated to support multiple devices from the same thread. Also renamed to "resizable_cuda_buffer".  

It is now more safe to initialize tensor_conv from other thread/device. 

But I think there is till possible problem of switching device/thread because tensor_conv holds several device-specific objects - workspace (this PR), workspace size (will be different for different device types), algorithm number. If we will try to run pre-initialized tensor_conv with other device, it can throw exception

I think we should add tensor_conv - wide check for current device id. if device id is changed - search new algorithm and re-estimate workspace size.

davisking:
Sweet, thanks for another nice PR.  I think it's ready to merge, unless you have objections.

I'm not sure we can do much about the issue of tensor_conv owning GPU specific buffers.  Mainly because the layers in general have parameter buffers which are all bound to a specific GPU.  Or is there something to this PR that is worse than this that I'm not seeing?

e-fominov:
I think it should work good now and can be merged. I have tried it with machine of two different GPUs (one is Maxwell other is Pascal) and training with both GPUs, it seems to be working fine (imagenet training example)

davisking:
Great.  Thanks for a sweet PR. :)

davisking:
It just occurred to me that the tensor_conv operations are asynchronous (obviously), and that the way the code was written didn't ensure the cuda_data_void_ptrs work buffer handles existed for the entire duration that cuDNN might be using them.  Meaning that, for instance, if a tensor_conv was running, and then another tensor_conv started to run but needed bigger work buffers then it would deallocate the buffer right under the previous, but still running, tensor_conv.  

I'm not sure why this wasn't happening during any of the tests.  I guess layers near the input probably use bigger workspaces most of the time so we were getting lucky.  Anyway, I just pushed a fix for this.  Let me know if you think I missed something important about it or it's not right.  It's passing my tests though and seems like a straightforward fix.

PR #702, opened 2017-07-11
Update webcam_face_pose_ex.cpp

MambaWong:
Test on a given video like this 
`cv::VideoCapture cap("Sample.avi")`
may be broken when the video frames are not enough before the main window is closed by the user.


davisking:
Good idea, thanks for the PR :)

PR #714, opened 2017-07-18
Make http_client compile in C++11 and above

Morosko:


dlib/http_client/http_client.cpp
```diff
@@ -454,9 +454,11 @@ namespace dlib
                 for (stringmap::const_iterator si = ci->second.begin(); si != ci->second.end(); ++si)
                 {
                     std::ifstream in(si->second.c_str());
+					std::string file_content((std::istreambuf_iterator<char>(in)),
+                                    std::istreambuf_iterator<char>());
                     postBody << "--" << mime_boundary << "\r\n"
                         "Content-Disposition: form-data; name=\"" << ci->first << "\"; filename=\"" << get_basename(si->second) << "\"\r\n\r\n"
-                        << in << "\r\n";
```
davisking: Doesn't putting in.rdbuf() here work?  That should be a more efficient solution.  Also, please fix the messed up tabbing.

davisking:
Cool, thanks for the PR :)

Morosko:
Thanks for the fast reply and the hint to make it more efficient! I changed it now.

davisking:
No problem. Thanks for the PR :)

PR #717, opened 2017-07-19
Report confidence of loss_multiclass_log_per_pixel_weighted

reunanen:
Make `loss_multiclass_log_per_pixel_weighted_::to_label` return not only the predicted class label, but also an associated confidence score.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1724,8 +1724,26 @@ namespace dlib
             float weight = 1.f;
         };
 
+        struct scored_label
+        {
+            scored_label()
+            {}
+
+            scored_label(uint16_t label, float score = std::numeric_limits<float>::quiet_NaN())
+                : label(label), score(score)
+            {}
+
+            operator uint16_t () const
+            {
+                return label;
+            }
+
+            uint16_t label = std::numeric_limits<uint16_t>::max();
+            float score = std::numeric_limits<float>::quiet_NaN();
+        };
+
         typedef matrix<weighted_label> training_label_type;
-        typedef matrix<uint16_t> output_label_type;
+        typedef matrix<scored_label> output_label_type;
 
         template <
             typename SUB_TYPE,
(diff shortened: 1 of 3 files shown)


dlib/dnn/loss.h
```diff
@@ -1737,7 +1755,45 @@ namespace dlib
...
-            loss_multiclass_log_per_pixel_::to_label(input_tensor, sub, iter);
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+
+            const tensor& output_tensor = sub.get_output();
+
+            DLIB_CASSERT(output_tensor.k() >= 1); // Note that output_tensor.k() should match the number of labels.
+            DLIB_CASSERT(output_tensor.k() < std::numeric_limits<uint16_t>::max());
+            DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+
+            const float* const out_data = output_tensor.host();
```
reunanen: In fact I would want to softmax the output tensor, before collecting the scores below. Is introducing a temporary tensor here too wasteful? Is there a better way?

reunanen: Hmm... Earlier [I made this `to_label` function static](https://github.com/davisking/dlib/pull/685/commits/d5794c55fe274a5e7aa7ff247d7eb7443567990c), but I guess there's no reason for that any longer. So perhaps the easiest way (smallest clean change) is to make the function non-static, and add a private member variable for the softmaxed outputs. (?)

reunanen: Hmm in fact it seems the right way to do this is to add a softmax _layer_.

reunanen: Yep, best handled with a softmax layer. Closing the PR.

reunanen:
It seems that this whole requirement is best handled by adding a `dlib::softmax` layer, like [here](https://github.com/davisking/dlib/blob/master/examples/dnn_imagenet_train_ex.cpp#L311-L316). Therefore, closing the PR.

PR #732, opened 2017-07-28
Fix strange dash in Box-Muller

elelel:
The non-standard dash character confuses Microsoft compiler and produces warnings like 'The file contains a character that cannot be represented in the current code page (949). Save the file in Unicode format to prevent data loss' when compiled on non-Latin system locale codepages


davisking:
Cool, thanks for the PR :)

PR #734, opened 2017-07-29
 differential std::clamp from c++17 with dlib:: prefix to clamp()

namar0x0309:
I need to test it out before the merge request. Just to get the ball rolling on code review.


davisking:
Sweet, thanks for the PR :)

PR #753, opened 2017-08-17
Add a python wrapper for the cnn face detector

ageitgey:
This PR adds a simple wrapper for calling the cnn face detector from python. With this, you can now use the cnn face detector in python just like the face HOG detector:

```python
import dlib

cnn_face_detection_model = dlib.cnn_face_detection_model_v1("mmod_human_face_detector.dat")

dets = cnn_face_detection_model.cnn_face_detector(img, 1)

for i, d in enumerate(dets):
  print("Detection {}: Left: {} Top: {} Right: {} Bottom: {}".format(i, d.left(), d.top(), d.right(), d.bottom()))
```

It seems to work fine, but I haven't written C++ code in a long time and I'm not very experienced with C++11 and the dlib code base. So all feedback is very welcome (including "this sucks, try again").

Thanks!

### DIFF
--- a/tools/python/CMakeLists.txt
+++ b/tools/python/CMakeLists.txt
@@ -26,6 +26,7 @@ set(python_srcs
    src/shape_predictor.cpp
    src/correlation_tracker.cpp
    src/face_recognition.cpp
+   src/cnn_face_detector.cpp
 )
 
 # Only add the GUI module if requested
@@ -35,6 +36,6 @@ endif(NOT ${DLIB_NO_GUI_SUPPORT})
 
 add_python_module(dlib ${python_srcs})
 
-# When you run "make install" we will copy the compiled dlib.so (or dlib.pyd) 
+# When you run "make install" we will copy the compiled dlib.so (or dlib.pyd)
 # library file to the python_examples folder.
 install_dlib_to(../../python_examples)
--- a/tools/python/src/dlib.cpp
+++ b/tools/python/src/dlib.cpp
@@ -19,6 +19,7 @@ void bind_object_detection();
 void bind_shape_predictors();
 void bind_correlation_tracker();
 void bind_face_recognition();
+void bind_cnn_face_detection();
 
 #ifndef DLIB_NO_GUI_SUPPORT
 void bind_gui();
@@ -51,8 +52,8 @@ BOOST_PYTHON_MODULE(dlib)
     bind_shape_predictors();
     bind_correlation_tracker();
     bind_face_recognition();
+    bind_cnn_face_detection();
 #ifndef DLIB_NO_GUI_SUPPORT
     bind_gui();
 #endif
 }
-
--- a/python_examples/cnn_face_detector.py
+++ b/python_examples/cnn_face_detector.py
@@ -1,0 +1,72 @@
+#!/usr/bin/python
+# The contents of this file are in the public domain. See LICENSE_FOR_EXAMPLE_PROGRAMS.txt
+#
+#   This example shows how to run a CNN based face detector using dlib.  The
+#   example loads a pretrained model and uses it to find faces in images.  The
+#   CNN model is much more accurate than the HOG based model shown in the
+#   face_detector.py example, but takes much more computational power to
+#   run, and is meant to be executed on a GPU to attain reasonable speed.
+#
+#   You can download the pre-trained model from:
+#       http://dlib.net/files/mmod_human_face_detector.dat.bz2
+#
+#   The examples/faces folder contains some jpg images of people.  You can run
+#   this program on them and see the detections by executing the
+#   following command:
+#       ./cnn_face_detector.py mmod_human_face_detector.dat ../examples/faces/*.jpg
+#
+#
+# COMPILING/INSTALLING THE DLIB PYTHON INTERFACE
+#   You can install dlib using the command:
+#       pip install dlib
+#
+#   Alternatively, if you want to compile dlib yourself then go into the dlib
+#   root folder and run:
+#       python setup.py install
+#   or
+#       python setup.py install --yes USE_AVX_INSTRUCTIONS --yes DLIB_USE_CUDA
+#   if you have a CPU that supports AVX instructions, you have an Nvidia GPU
+#   and you have CUDA installed since this makes things run *much* faster.
+#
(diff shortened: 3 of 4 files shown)


python_examples/mmod_face_detector.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/python
+# The contents of this file are in the public domain. See LICENSE_FOR_EXAMPLE_PROGRAMS.txt
+#
+#   This example shows how to run a CNN based face detector using dlib.  The
```
davisking: Looks good.  You should change the name of the example to cnn_face_detector.py though since the HOG model is also a MMOD based method.  It just uses HOG features instead of CNN features.  But it's otherwise the same thing.

tools/python/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(python_srcs
    src/shape_predictor.cpp
    src/correlation_tracker.cpp
    src/face_recognition.cpp
+   src/face_detection.cpp
```
davisking: This should be renamed to cnn_face_detector.cpp.  Otherwise, people looking for the HOG face detector code will probably be confused that it's not in this file.

tools/python/src/dlib.cpp
```diff
@@ -19,6 +19,7 @@ void bind_object_detection();
 void bind_shape_predictors();
 void bind_correlation_tracker();
 void bind_face_recognition();
+void bind_face_detection();
```
davisking: Rename to have cnn in name.

tools/python/src/face_detection.cpp
```diff
@@ -0,0 +1,98 @@
...
+            scale *= 2;
+            pyramid_up(image);
+        }
+
+        auto dets = net(image);
+        std::vector<rectangle> rects;
+
+        // Scale the detection locations back to the original image size
+        // if the image was upscaled.
+        for (auto&& d : dets) {
```
davisking: This should use the built in rect_up() functions.  You make a pyramid_down<2> object and then you can call rect_up() on it and it will do the right thing.  All the pyramid stuff has these things and they take care to deal with any border padding and other vagaries of this coordinate transform.

tools/python/src/face_detection.cpp
```diff
@@ -0,0 +1,98 @@
...
+};
+
+
+// ----------------------------------------------------------------------------------------
+
+void bind_face_detection()
+{
+    using boost::python::arg;
+    {
+    class_<face_detection_model_v1>("face_detection_model_v1", "This object detects human faces in an image.  The constructor loads the face detection model from a file.", init<std::string>())
```
davisking: Oh and this documentation should probably mention that it's designed to work with the model file http://dlib.net/files/mmod_human_face_detector.dat.bz2.

ageitgey:
Thanks! Those changes all make sense. I'll fix it up tomorrow most likely.

python_examples/mmod_face_detector.py
ageitgey: very good point 👍

tools/python/CMakeLists.txt
ageitgey: makes sense 👍

tools/python/src/dlib.cpp
ageitgey: 👍

ageitgey:
@davisking I made those updates. Let me know if you see anything else that needs changing. Thanks!

tools/python/src/face_detection.cpp
davisking: Ha, oops, rect_down(), but you figured that out :)

davisking:
Nice, thanks for the PR :)

People have been asking for this a lot.  It should be popular.

ageitgey:
Thanks for the help! Yeah, I finally did it because I got tired of people asking over at [ageitgey/face_recognition](https://github.com/ageitgey/face_recognition) :)

If you have a rough idea yet of when you next plan to release dlib, let me know. I've got changes to [ageitgey/face_recognition_models](https://github.com/ageitgey/face_recognition_models) and [ageitgey/face_recognition](https://github.com/ageitgey/face_recognition) to push that will use this but I'll wait until you have this change out in the next dlib version.

davisking:
I'm hoping to make the next release soon.  I've been improving some aspects of the cnn detector and I'm training a vehicle detection model as part of the new example program.  It's working pretty well but I'm still playing with it.  Once I'm done I'll make the next official dlib release.  

The damn detector keeps getting slightly better and better as I tweak the training data and retrain.  Each cycle takes about a day so it's going so.  I should probably stop since it's plenty good enough for an example program, but it's hard to stop when it's still getting better :)

Anyway, it shouldn't be more than a few weeks.

ageitgey:
Cool. Sounds like a great example model. I know what you mean about it being hard to stop when it keeps improving. I'll keep my eyes out for the release over the next few weeks.

smajida:
@ageitgey thanks for your contribution. I hope you did not get so tired of those people asking you for it :+1: 
@davisking sounds an interesting example. Can´t wait to give it a try. I have also been working on a pedestrian detector.

davisking:
No worries.  I'm not working on gaze estimation, but it would be cool if someone did it and made a PR :)

headupinclouds:
> I was wondering wether there is a plan for gaze position estimation example using traditional and dnn methods by chance

@smajida : I'm working towards a real-time gaze (i.e., viewing angle) estimation module for mobile devices in [drishti](https://github.com/elucideye/drishti) with help from dlib components.  I currently have a functional real-time pipeline for a kind of 2D eye tracker (not quite gaze) and am planning to use this as a building block to experiment with gaze estimation.  I think the current model fitting might be of some use to you.  I expect an appearance based gaze module (via dlib DNN) will ultimately be required if this is even doable for any practical use case on a mobile screen, but the gradient boosting sure is fast.  Sadly, I'm already a couple months behind on another dlib DNN issue I had planned to help with 🙁, but I do intend to try it eventually.  I've been thinking through the somewhat daunting task of data collection for this application, but this cool [SimGAN](https://github.com/wayaai/SimGAN) gift of a project could be repurposed for this with some work (Unity + GAN).

PR #772, opened 2017-08-24
Clang support for Windows

e-fominov:
This PR makes pussible compilation with Clang on Windows platform. I have tried Clang 5.0, it has full support of C++11, it is compatible with all MSVC libraries (MKL) and it supports CUDA and compiled code is faster than original MSVC-generated code

So I think, users of Dlib who need CUDA on Windows platform for DNN can easily do it with Clang

To use Clang after installation, you need to pass this arguments to CMAKE:
```
cmake ..  -G"Visual Studio 14 2015 Win64" -T"LLVM-vs2014"
```

There is one annoying warning during compilation
```
 warning : argument unused during compilation: '/Gm-' [-Wunused-command-line-argument]
```

But I can't find way to disable it. -Wno-unused-command-line-argument does not help, -Qunused-argument does not help too

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -523,7 +523,7 @@ if (NOT TARGET dlib)
             # -std=c++11 option if you let it propagate it to nvcc in some
             # cases.  So instead we disable this and manually include
             # things from CMAKE_CXX_FLAGS in the CUDA_NVCC_FLAGS list below.
-            if (APPLE)
+            if (APPLE OR MSVC)
                set(CUDA_PROPAGATE_HOST_FLAGS OFF)
                # Grab all the -D flags from CMAKE_CXX_FLAGS so we can pass them
                # to nvcc.
--- a/dlib/cmake_utils/use_cpp_11.cmake
+++ b/dlib/cmake_utils/use_cpp_11.cmake
@@ -58,6 +58,11 @@ if (CMAKE_VERSION VERSION_LESS "3.1.2")
       # Since we don't know what compiler this is just try to build a c++11 project and see if it compiles.
       test_compiler_for_cpp11()
    endif()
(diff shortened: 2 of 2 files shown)


davisking:
Sweet.  I didn't realize clang could be used with cuda.  How did you install clang? Using the option in visual studio's installer?  Is there any other setup other than to check the box in the installer?

e-fominov:
For using Clang, I downloaded version 5 from here: http://prereleases.llvm.org/win-snapshots/

davisking:
Oh, nice.  I thought it was the thing that's part of visual studio.  I forget what it said in the installer, but there were options to use clang but it didn't seem very functional when I tried it.  MS installers are also kinda horrendous.

davisking:
I got this compiler error when trying to build the dnn_mmod_ex.cpp example:

```
3>dlib.lib(dlib_generated_cuda_dlib.cu.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in dnn_mmod_ex.obj
3>dlib.lib(dlib_generated_cusolver_dlibapi.cu.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in dnn_mmod_ex.obj
3>msvcprt.lib(MSVCP140.dll) : error LNK2005: "public: char __cdecl std::basic_ios<char,struct std::char_traits<char> >::widen(char)const " (?widen@?$basic_ios@DU?$char_traits@D@std@@@std@@QEBADD@Z) already defined in dlib.lib(dlib_generated_cuda_dlib.cu.obj)
```

It seems like some of the needed compiler switches aren't being propagated to nvcc.  Did this work for you when you tried it?

e-fominov:
just checked. yes, i have same issue. will fix

PR #773, opened 2017-08-24
hotfix static runtime in cuda with clang and windows

e-fominov:
This PR fixes the compilation of examples with Clang + CUDA + Windows. There will be a lot of warnings, but compilation of dnn_mmod_exx is successul


davisking:
That fixes the compiler error but when I ran dnn_mmod_ex it just crashes.  Is that what happens for you?  I compiled it with cuda.  Were you able to run cuda programs compiled with clang?

PR #777, opened 2017-08-25
Include drectangle.h to point_transforms.h

denizevrenci:
point_transforms.h is not self compiling without including drectangle.h


davisking:
Cool, thanks for the PR :)

PR #778, opened 2017-08-25
C++11 features

denizevrenci:
Now that C++11 is required to build dlib, I would like to add some features to help with better type traits.


davisking:
Looks good, thanks for the PR.

PR #780, opened 2017-08-26
improvements to cnn face detection python interface

guigzzz:
Small improvements to cnn face detection python interface

changes:
- can use the detector in two modes:
  * single image mode - returns a mmod_rectangles object
  * list of images - returns a batch_mmod_rectangles object
    - this mode can be used for speed when CUDA is activated
    - if CUDA is deactivated then there is no real speed advantage when comparing the two

- mmod_rect object: contains a rectangle object and a confidence score for the detection
- mmod_rectangles: represents all detections in an image, can be iterated over as a list
- batch_mmod_rectangles: represents all detections over a list of images, can be iterated over as a 2d list

### DIFF
--- a/dlib/image_processing/full_object_detection.h
+++ b/dlib/image_processing/full_object_detection.h
@@ -140,6 +140,10 @@ namespace dlib
         bool ignore = false;
 
         operator rectangle() const { return rect; }
+        bool operator == (const mmod_rect& rhs) const
+        { 
+            return rect == rhs.rect; 
+        }
     };
 
     inline mmod_rect ignored_mmod_rect(const rectangle& r)
--- a/dlib/image_processing/full_object_detection.h
+++ b/dlib/image_processing/full_object_detection.h
@@ -142,7 +142,9 @@ namespace dlib
         operator rectangle() const { return rect; }
         bool operator == (const mmod_rect& rhs) const
         { 
-            return rect == rhs.rect; 
+            return rect == rhs.rect 
+                   && detection_confidence == rhs.detection_confidence
+                   && ignore == rhs.ignore;
         }
     };
 
--- a/python_examples/cnn_face_detector.py
+++ b/python_examples/cnn_face_detector.py
@@ -39,7 +39,7 @@
 #   Or downloaded from http://scikit-image.org/download.html.
 
 import sys
-
+sys.path.append('../tools/python/build')
 import dlib
 from skimage import io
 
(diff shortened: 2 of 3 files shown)


dlib/image_processing/full_object_detection.h
```diff
@@ -140,6 +140,10 @@ namespace dlib
         bool ignore = false;
 
         operator rectangle() const { return rect; }
+        bool operator == (const mmod_rect& rhs) const
```
davisking: This is a little bit funny.  If you are going to implement operator== then it needs to compare for full equality.  So checking all the fields is needed, not just rect.

python_examples/cnn_face_detector.py
```diff
@@ -39,7 +39,7 @@
 #   Or downloaded from http://scikit-image.org/download.html.
 
 import sys
-
+sys.path.append('../tools/python/build')
```
davisking: Remove this.

tools/python/src/cnn_face_detector.cpp
```diff
@@ -59,12 +53,57 @@ class cnn_face_detection_model_v1
...
+                throw dlib::error("Unsupported image type, must be 8bit gray or RGB image.");
+
+            for(auto i = 0; i < upsample_num_times; i++)
+            {
+                pyramid_up(image);
+            }
+            dimgs.push_back(image);
+        }
+
+        auto dets = net(dimgs);
```
davisking: This requires all the images to have the same size.  There is also an argument of the network that determines the mini-batch size.  For instance, if dimgs contained a million images, it wouldn't try to stuff 1 million images onto the GPU, it would be chopped up into batches of 128 (by default) and run.  The 128 value is the optional 2nd argument to net() and should also be an optional argument here, otherwise it will be impossible to run this function with large images since you will run out of GPU ram, but setting the batch size to something smaller would work fine. 

So this function needs two things.  A way to set the batch size, and a check that the user gave images that are all of the same size.  This check that all images are the same size is performed internally in the network as well, but the error message from there will probably confuse python users.

tools/python/src/cnn_face_detector.cpp
```diff
@@ -78,18 +117,42 @@ class cnn_face_detection_model_v1
...
     }
+    {
+    typedef mmod_rect type;
+    class_<type>("mmod_rect", "Wrapper around a rectangle object and a detection confidence score.")
+        .def_readwrite("rect",   &type::rect)
+        .def_readwrite("confidence", &type::detection_confidence);
+    }
+    {
+    typedef std::vector<mmod_rect> type;
+    class_<type>("mmod_rectangles", "An array of mmod rectangle objects.")
```
davisking: There is a general pattern in the dlib python api that goes like,  object, objects (for vector<object>) and objectss (for vector<vector<object>>).  This code should follow the same pattern so the API is consistent.  So these things should be called, mmod_rect, mmod_rects, and mmod_rectss.

guigzzz:
ok thanks mate, the changes make sense (especially the equality one, donno why I didnt do that), I'll get this done asap

davisking:

PR #805, opened 2017-09-05
Added face clustering example to Python API

visionworkz:
Exposed more functionality from the C++ API in the Python API:
1. Exposed the save_image_chip and save_image_chips to save a 150x150 px file with 0.25 padding which is suitable to be loaded and jittered in the dnn_metric_learning_on_images_ex.cpp
2. Added a cluster method which accepts a list of descriptors and returns a list containing integers identifying the cluster label for each descriptor.
3. Added a python example which can be pointed at a folder containing images of a common person photographed along with other different people. The example then performs clustering and saves the extracted faces (assuming the common face will form the biggest cluster) to a specified folder.

### DIFF
--- a/python_examples/face_clustering.py
+++ b/python_examples/face_clustering.py
@@ -100,7 +100,7 @@ for f in glob.glob(os.path.join(faces_folder_path, "*.jpg")):
         # middle value, such as 10, which is only 10x slower but still gets an
         # LFW accuracy of 99.3%.
 
-labels = facerec.cluster(descriptors)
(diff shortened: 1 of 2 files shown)


davisking:
Sweet, thanks for the PR :)

I'll look it over in a few days.

visionworkz:
Added threshold to the call so that it is easier to decrease the threshold if needed.

davisking:
Thanks for the PR.  I cleaned up the comments and example a bit.  You can see the changes in the git logs.

PR #809, opened 2017-09-09
Add support for non-scale-invariant MMOD

reunanen:
**What**

I want to train MMOD based detectors that are _not_ scale-invariant. So I am not even going to use an image pyramid.

How come? Think of detecting vehicles in satellite images, for example. There is essentially no perspective, and all cars and trucks are pretty much the same size. Now it would be silly to report a sedan that is 50 cm or 50 m long, for example – even if there is something in the image that, when scale is ignored, looks a lot like a sedan.

In fact it would be silly to even look for a sedan that is 50 cm or 50 m long. So bypassing the image pyramid should also make the detector run faster, when invariance to scale is not needed.

However, I still want my detector to understand different object sizes. (For example, to have separate detector windows for small trucks and large trucks.)

(In fact I'm not really processing satellite images – just wanted to give an example.)

**How**

This PR adds a new `mmod_options` constructor for this purpose. The new constructor doesn't find covering aspect ratios, but covering rectangles instead. In some cases this means that there will be
many more detector windows, which may at least partially offset the speed gains from bypassing the image pyramid. (But regardless of the speed, I still don't want to report sedans that are 50 cm or 50 m long.)

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -360,6 +360,12 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 // ----------------------------------------------------------------------------------------
 
+    enum class use_image_pyramid : uint8_t
+    {
+        no,
+        yes
+    };
+
     struct mmod_options
     {
     public:
@@ -405,12 +411,6 @@ namespace dlib
         double truth_match_iou_threshold = 0.5;
         test_box_overlap overlaps_nms = test_box_overlap(0.4);
         test_box_overlap overlaps_ignore;
-
-        enum class use_image_pyramid : uint8_t
-        {
-            no,
-            yes
-        };
 
         use_image_pyramid assume_image_pyramid = use_image_pyramid::yes;
 
@@ -685,7 +685,7 @@ namespace dlib
         {
             uint8_t assume_image_pyramid = 0;
             deserialize(assume_image_pyramid, in);
-            item.assume_image_pyramid = static_cast<mmod_options::use_image_pyramid>(assume_image_pyramid);
+            item.assume_image_pyramid = static_cast<use_image_pyramid>(assume_image_pyramid);
         }
     }
 
@@ -1052,10 +1052,10 @@ namespace dlib
         size_t find_best_detection_window (
             rectangle rect,
             const std::string& label,
-            mmod_options::use_image_pyramid assume_image_pyramid
+            use_image_pyramid assume_image_pyramid
         ) const
         {
-            if (assume_image_pyramid == mmod_options::use_image_pyramid::yes)
+            if (assume_image_pyramid == use_image_pyramid::yes)
             {
                 rect = move_rect(set_rect_area(rect, 400*400), point(0,0));
             }
@@ -1075,7 +1075,7 @@ namespace dlib
 
                 rectangle det_window;
                 
-                if (options.assume_image_pyramid == mmod_options::use_image_pyramid::yes)
+                if (options.assume_image_pyramid == use_image_pyramid::yes)
                 {
                     det_window = centered_rect(point(0,0), options.detector_windows[i].width, options.detector_windows[i].height);
                     det_window = move_rect(set_rect_area(det_window, 400*400), point(0,0));
@@ -1103,7 +1103,7 @@ namespace dlib
             const std::string& label,
             const net_type& net,
             size_t& det_idx,
-            mmod_options::use_image_pyramid assume_image_pyramid
+            use_image_pyramid assume_image_pyramid
         ) const 
         {
             using namespace std;
@@ -1118,7 +1118,7 @@ namespace dlib
             det_idx = find_best_detection_window(rect,label,assume_image_pyramid);
 
             double scale = 1.0;
-            if (options.assume_image_pyramid == mmod_options::use_image_pyramid::yes)
+            if (options.assume_image_pyramid == use_image_pyramid::yes)
             {
                 // Compute the scale we need to be at to get from rect to our detection window.
                 // Note that we compute the scale as the max of two numbers.  It doesn't
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -405,6 +405,7 @@ namespace dlib
         double truth_match_iou_threshold = 0.5;
         test_box_overlap overlaps_nms = test_box_overlap(0.4);
         test_box_overlap overlaps_ignore;
+        bool scale_invariant = true;
 
         mmod_options (
             const std::vector<std::vector<mmod_rect>>& boxes,
@@ -451,8 +452,35 @@ namespace dlib
 
             DLIB_CASSERT(detector_windows.size() != 0, "You can't call mmod_options's constructor with a set of boxes that is empty (or only contains ignored boxes).");
 
+            set_overlap_nms(boxes);
+        }
 
+        mmod_options(
+            const std::vector<std::vector<mmod_rect>>& boxes,
+            const double min_detector_window_overlap_iou = 0.75
+        )
+            : scale_invariant(false)
+        {
+            DLIB_CASSERT(0.5 < min_detector_window_overlap_iou && min_detector_window_overlap_iou < 1);
 
+            // Figure out what detector windows we will need.
+            for (auto& label : get_labels(boxes))
+            {
+                for (auto rectangle : find_covering_rectangles(boxes, test_box_overlap(min_detector_window_overlap_iou), label))
+                {
+                    detector_windows.push_back(detector_window_details(rectangle.width(), rectangle.height(), label));
+                }
+            }
+
+            DLIB_CASSERT(detector_windows.size() != 0, "You can't call mmod_options's constructor with a set of boxes that is empty (or only contains ignored boxes).");
+
+            set_overlap_nms(boxes);
+        }
+
+    private:
+
+        void set_overlap_nms(const std::vector<std::vector<mmod_rect>>& boxes)
+        {
             // Convert from mmod_rect to rectangle so we can call
             // find_tight_overlap_tester().
             std::vector<std::vector<rectangle>> temp;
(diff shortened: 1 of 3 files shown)


davisking:
Sweet, thanks for the PR.

dlib/dnn/loss.h
```diff
@@ -405,6 +405,7 @@ namespace dlib
         double truth_match_iou_threshold = 0.5;
         test_box_overlap overlaps_nms = test_box_overlap(0.4);
         test_box_overlap overlaps_ignore;
+        bool scale_invariant = true;
```
davisking: This would probably be better named something like `assume_input_layer_is_image_pyramid`

dlib/dnn/loss.h
```diff
@@ -451,8 +452,35 @@ namespace dlib
 
             DLIB_CASSERT(detector_windows.size() != 0, "You can't call mmod_options's constructor with a set of boxes that is empty (or only contains ignored boxes).");
 
+            set_overlap_nms(boxes);
+        }
+
+        mmod_options(
+            const std::vector<std::vector<mmod_rect>>& boxes,
+            const double min_detector_window_overlap_iou = 0.75
```
davisking: This kind of overloading bothers me.  I'm certain there will be confused users who call this constructor when they don't mean to.  So it would be better if the constructor had to be invoked like this:

```
mmod_options options(NO_IMAGE_PYRAMID, boxes);
```

This way it's really obvious what's happening.  You could use one of C++11's new strongly typed enum to define the NO_IMAGE_PYRAMID constant to allow safe overloading of the constructors.

dlib/dnn/loss.h
```diff
@@ -479,9 +507,6 @@ namespace dlib
             overlaps_nms = test_box_overlap(iou_thresh, percent_covered_thresh);
         }
 
-
-    private:
```
davisking: Why is this private removed?

dlib/dnn/loss.h
davisking: Actually, make this a strongly typed enum.  You will also need to update the serialization code to include this field.  That's missing from this PR at the moment.

dlib/dnn/loss.h
```diff
@@ -990,21 +1035,34 @@ namespace dlib
...
             double best_ratio_diff = -std::numeric_limits<double>::infinity();
             for (size_t i = 0; i < options.detector_windows.size(); ++i)
             {
                 if (options.detector_windows[i].label != label)
                     continue;
-                rectangle det_window = centered_rect(point(0,0), options.detector_windows[i].width, options.detector_windows[i].height);
-                det_window = move_rect(set_rect_area(det_window, 400*400), point(0,0));
+
+                rectangle det_window(options.detector_windows[i].width, options.detector_windows[i].height);
+
```
davisking: This code is a little confusing.  Put it in an if/else rather than a default assignment and an if that overrides.

dlib/dnn/loss.h
```diff
@@ -1035,14 +1094,19 @@ namespace dlib
                 throw impossible_labeling_error(sout.str());
             }
 
-            det_idx = find_best_detection_window(rect,label); 
+            det_idx = find_best_detection_window(rect,label,scale_invariant);
+            const double scale = scale_invariant
```
davisking: Use an if/else.  This bit of code looks too crazy otherwise.  I thought it was just a tabbing error at first.

dlib/dnn/loss_abstract.h
```diff
@@ -431,6 +435,9 @@ namespace dlib
                 - 0 < min_target_size <= target_size
                 - 0.5 < min_detector_window_overlap_iou < 1
             ensures
+                - scale_invariant = true
+                - This function should be used when scale-invariance is desired, and an
```
davisking: Here we should explicitly reference the input layers we expect the user to use.  So explicitly say they should be using the `input_rgb_image_pyramid`.

dlib/dnn/loss_abstract.h
```diff
@@ -461,6 +468,32 @@ namespace dlib
...
+                - scale_invariant = false
+                - This function should be used when scale-invariance is not desired, and
+                  there is no intention to apply an image pyramid.
+                - This function tries to automatically set the MMOD options to reasonable
+                  values, assuming you have a training dataset of boxes.size() images, where
+                  the ith image contains objects boxes[i] you want to detect.
+                - The most important thing this function does is decide what detector
+                  windows should be used.  This is done by finding a set of detector
+                  windows that are sized such that:
+                    - When slid over an image pyramid, each box in boxes will have an
```
davisking: Fix to say you **aren't** using an image pyramid.

davisking:
Cool.  Thanks for the PR.  This is something I knew was needed but hadn't got around to yet :)

Also, it looks right, but did you test it?  Does the dnn_mmod_ex.cpp run correctly to completion and learn to find faces for instance?

dlib/dnn/loss.h
reunanen: It isn't. I wanted to extract the `set_overlap_nms` function, rather than leave identical code in the two constructors. I made this new function private, so the diff just looks a little confusing here. 

In other words, it's still there – [right here](https://github.com/davisking/dlib/pull/809/files/90da2bbc0b2f42a4b0420c6cefd34437e209b165#diff-32a043d312608d62a296b05ce1bc096eR480), to be specific. :)

On the plus side, the diff doesn't show the function code itself as being deleted from one place and added into another place.

dlib/dnn/loss.h
reunanen: Fixed. Didn't test serialization, though.

dlib/dnn/loss.h
reunanen: Agreed, and fixed.

dlib/dnn/loss.h

dlib/dnn/loss.h
reunanen: Sorry. :)

I guess I just wanted to keep the constness.

Fixed.

dlib/dnn/loss_abstract.h
reunanen: Fixed.

dlib/dnn/loss_abstract.h
reunanen: Fixed.

dlib/dnn/loss.h
davisking: Ah, right, looks good :)

I wish there was a github button that uncollapsed all the folds without needing to click on the expand button a billion times :/

dlib/dnn/loss.h
davisking: Cool, thanks.

But won't people need to type.  `mmod_options options(mmod_options::assumed_input_layer_type_t::NO_IMAGE_PYRAMID, boxes);`?  We aren't writing java here man :)

Can you make it like `mmod_options options(use_image_pyramid::yes, boxes);` ?

dlib/dnn/loss.h
```diff
@@ -1049,20 +1065,24 @@ namespace dlib
...
-
-                if (scale_invariant) {
+                rectangle det_window;
+                
+                if (options.assumed_input_layer_type == mmod_options::assumed_input_layer_type_t::IMAGE_PYRAMID)
+                {
                     det_window = centered_rect(point(0, 0), options.detector_windows[i].width, options.detector_windows[i].height);
                     det_window = move_rect(set_rect_area(det_window, 400 * 400), point(0, 0));
                 }
+                else {
```
davisking: funny {

reunanen:
> Also, it looks right, but did you test it?

I have tested the new path (no scale-invariance) in a separate application, and it seems to be ok. Unfortunately I can't release the dataset, or the code.

The old path (image pyramid) should be unchanged, but I didn't test it.

> Does the dnn_mmod_ex.cpp run correctly to completion and learn to find faces for instance?

You mean with the new path (no image pyramid), or the old path?

I guess both could be done. So far, I haven't tried either.

davisking:
Yeah, test it both ways since they should both work.

dlib/dnn/loss.h
reunanen: Oh, ok. I thought you wanted an `enum` so that more types could be added in the future. But now I understand that the point was readability – I completely agree that `mmod_options options(true, boxes);` wouldn't be clear enough.

Fixed now.

reunanen:
Ok, I will test... With cars probably, because I have the dataset already loaded. :)

dlib/dnn/loss.h
```diff
@@ -406,6 +406,14 @@ namespace dlib
         test_box_overlap overlaps_nms = test_box_overlap(0.4);
         test_box_overlap overlaps_ignore;
 
+        enum class use_image_pyramid : uint8_t
```
davisking: Can you move it outside mmod_options?  Otherwise it's still a lot of typing.

davisking:
Oh, don't do the cars one for this test.  That thing takes a whole day to train.  The dnn_mmod_ex faces example will finish in like an hour.  Although it's possible that the faces example doesn't have enough data to make a good test.  So yeah, maybe do cars.  It's going to take a while though.

dlib/dnn/loss.h
reunanen: Right – fixed.

davisking:
Sweet, it all looks good. Let me know when the tests finish and I'll merge it :)

reunanen:
It didn't crash for me... But ok, I'll leave it alone.

dlib/dnn/loss.h
reunanen: In fact, when you pointed out the overloading confusion issue, the first solution I was thinking of was the [named constructor idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Named_Constructor). Which, I think, would fix the problem also. So when you mentioned enum, I started to think of something else (=possibility to add more modes in the future).

Anyway, I'm perfectly fine with the present approach. Just wanted to write this for completeness.

davisking: Right, that's what I was thinking about too.  C++11 enums seem like a better approach here though.

reunanen:
Works fine on the old path (image pyramid enabled):

image

However, doesn't seem to work out of the box without the image pyramid. Still working on it...

davisking:
Looks good at this point.  There is one minor problem in the deserialize() routine, but I'll fix that in a moment.  Thanks for the PR :)

PR #812, opened 2017-09-10
Let custom loss values affect MMOD gradient

reunanen:
**Problem**: It looked like `loss_per_missed_target` and `loss_per_false_alarm` only affected the calculated loss, and not the gradient. Therefore, it also looked like these values did not affect the training result.

**Solution**: Make it so that these values enter the calculation of the gradient. Add test cases as well.

---

Also, while at it, fix even loss from ignored rects when using `loss_per_missed_target != 1`.
EDIT: this fix moved from here to #813.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -763,7 +763,7 @@ namespace dlib
                         const size_t idx = (k*output_tensor.nr() + p.y())*output_tensor.nc() + p.x();
                         loss -= out_data[idx];
                         // compute gradient
-                        g[idx] = -scale;
+                        g[idx] = -scale * options.loss_per_missed_target;
                         truth_idxs.push_back(idx);
                     }
                     else
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -887,7 +887,7 @@ namespace dlib
                 for (auto&& x : final_dets)
                 {
                     loss += out_data[x.tensor_offset];
-                    g[x.tensor_offset] += scale;
+                    g[x.tensor_offset] += scale * options.loss_per_false_alarm;
                 }
 
                 ++truth;
--- a/dlib/test/dnn.cpp
+++ b/dlib/test/dnn.cpp
@@ -2875,6 +2875,44 @@ namespace
 
 // ----------------------------------------------------------------------------------------
 
+    void test_mmod_set_custom_loss_per_missed_target()
+    {
+        print_spinner();
+
+        using net_type = loss_mmod<con<1,6,6,1,1,input_rgb_image>>;
+
+        std::vector<matrix<rgb_pixel>> input(1);
+        std::vector<std::vector<mmod_rect>> truth(1);
+
+        input[0].set_size(50, 50);
+        input[0] = dlib::rgb_pixel(127, 127, 127);
+
+        truth[0].push_back(mmod_rect(rectangle(10, 10, 20, 20)));
+
+        const mmod_options options1(truth, 8, 8);
+
+        mmod_options options2 = options1;
+        options2.loss_per_missed_target *= 100;
+
+        const auto train_and_detect = [&](const mmod_options& options) {
+            print_spinner();
+            net_type net(options);
+            dnn_trainer<net_type> trainer(net);
+            trainer.set_max_num_epochs(100);
+            trainer.train(input, truth);
+            trainer.get_net();
+            return net(input[0]);
+        };
+
+        const auto detections1 = train_and_detect(options1);
+        const auto detections2 = train_and_detect(options2);
+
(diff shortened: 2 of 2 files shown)


dlib/test/dnn.cpp
```diff
@@ -2875,6 +2875,131 @@ namespace
 
 // ----------------------------------------------------------------------------------------
```
reunanen: Note that the new tests would fail with the previous `loss.h`.

I guess it is a matter of specification whether the tests themselves make sense. But at least in my world, they do. :)

dlib/dnn/loss.h
```diff
@@ -757,19 +757,19 @@ namespace dlib
...
                             // Ignore boxes that can't be detected by the CNN.
-                            loss -= 1;
+                            loss -= options.loss_per_missed_target;
                             continue;
                         }
                         const size_t idx = (k*output_tensor.nr() + p.y())*output_tensor.nc() + p.x();
                         loss -= out_data[idx];
                         // compute gradient
-                        g[idx] = -scale;
+                        g[idx] = -scale * options.loss_per_missed_target;
```
davisking: I'm very confident this isn't right.  If you work out the gradient you should find that it's 1.

davisking: Also, the loss factors do affect the gradient because they affect `truth_score_hits` which determines which parts of the gradient tensor get modified.

davisking: Anyway, if you look at the MMOD paper and do the math I'm pretty sure the gradient is +/- 1.  That isn't to say that changing the gradient to this is necessarily a bad idea.  It might actually be a better way to tweak the preference for more detections vs. fewer false alarms.  But some experiments are needed to decide if it's actually better.

davisking:
Thanks for another PR.  This might be a good idea, but you will need to run some experiments to see if it is.   This is because this changes the objective function, since I'm pretty sure the gradients were correct already.  But it might well be that this kind of loss scaling is more effective in adjusting the balance between more detections vs. fewer false alarms.

Honestly, the original MMOD objective function is pretty insensitive to the setting of these two loss parameters.  The only reason it was constructed that way was due to the limitations of the cutting plane solvers I was using way back then, but SGD probably doesn't care.  

Anyway,  do some experiments to see if this works in a sensible way :).  Like use some cloud GPUs to train a few different versions of the new vehicle detector using http://dlib.net/dnn_mmod_train_find_cars_ex.cpp.html and see what kind of detectors it learns for different loss settings.

dlib/dnn/loss.h
reunanen: Oh, ok. I admit that I was working mainly based on trial and error. But even if I made `loss_per_missed_target` really high, it didn't seem to affect the results. Guess I just need to try and err more. Thanks for your prompt reply.

davisking: Right, making it really high doesn't do much.  That's a limitation of how the MMOD loss is formulated.  So honestly, this might be better.  See my other comments about it in this PR :)

reunanen:
Davis, you are absolutely right: there are convergence issues here, most probably because the gradient does not quite match the objective function – exactly as you say. The issues just don't show up in a quick test (which, of course, is not unexpected).

So wondering if the loss function could be changed, for example to be quadratic (which might lead to a gradient that looks like the one in this PR).

At any rate, I still need something that makes my objects not go undetected.

I guess I could also change `adjust_threshold`, but it seems a less principled way: If `loss_per_missed_target` is, say, 3 times `loss_per_false_alarm`, it seems to be quite clear what is meant. But not so sure about `adjust_threshold`. What does it mean if it's -1, or -2?

davisking:
The only thing there is is the gradient.  Those loss values are just for display basically and don't feed into SGD.  So you can't make the gradient not match the objective function.  It's just impossible given how the code is structured.  

Maybe the loss can be changed.  But it's not obvious and requires a lot of experimentation to find something better.  

If you want to get more detections you should just change the detection threshold.  Try values and see what happens.  Or plot a ROC curve and pick a threshold by looking at that.

reunanen:
Oh, ok. Yes, my loss values started to increase after running for quite a while. So anyway the loss should be changed, if the gradient is changed.

davisking:
Right, the loss should be changed too for completeness, but that's not why
it's not working.   You should try a smaller learning rate.  Maybe if it
was smaller it would have worked.

reunanen:
Well. Maybe it _was_ indeed working. I just noticed that the loss was increasing, so I figured there must be something wrong.

I'll see if I can figure something out.

reunanen:
I will close this PR for now, because for sure this should not be merged as it is at the moment.

Maybe I'll suggest something else, or maybe I'll change `adjust_threshold`. We'll see.

PR #813, opened 2017-09-10
MMOD: fix loss from ignored rects when using loss_per_missed_target != 1

reunanen:
Extract this fix from #812, to be handled separately.


davisking:
Thanks for another PR, this is a good catch :)

PR #824, opened 2017-09-17
Added SARSA + Qlearning RL algorithms

adrianjav:
Hi Davis, I was planning to upload on my own repository some RL algorithms I implemented months ago and when I checked they weren't in dlib.

So I've added Q-learning and SARSA algorithms to dlib. My implementation was very basic and it used the whole state space, so I've tried to adapt them to use the feature_extractor structure properly. Their code is quite similar to the one in lspi.h and I've been thinking of using a template superclass to gather their common parts but I haven't managed to do it yet in a way I like.

I'm planning to write a proper example that uses openai to test them but for now the tests based on the one in lspi.cpp work fine.

I'm not used to contribute in open source projects, any suggestion to improve the code is well received :)


davisking:
Cool, thanks for the PR.  I looked it over quickly and it looks quite reasonable.  There are a few places that contain copy and paste errors, like DLIB_ASSERT statements with incorrect comments.  Also, why is the chain_model replicated in these test files? It looks identical to the one in lspi.cpp.  It would be better to avoid the code duplication and put all these tests in the same testing file.  Renaming lspi.cpp to something like reinforcement_learning.cpp is reasonable in that case.

adrianjav:
Damn, you're right! Those asserts were initially in the constructor and I decided to follow your approach with the setters.
I'll fix everything right now. I still want to merge all their common code, but having the template parameter in train makes everything uglier.

PS: Where can I see how the python binding works? Never saw it before

davisking:
Sounds good.

There is no python binding for this stuff.

davisking:
Let me know when this is ready for review.

adrianjav:
Ok I've been checking it and it should be ready to go.

dlib/control/sarsa_abstract.h
```diff
@@ -0,0 +1,156 @@
...
+        policy<feature_extractor> train(
+            const vector_type& samples
+        ) const;
+        /*!
+            requires
+                - samples.size() > 1
+                - samples is something with an interface that looks like
+                  std::vector<process_sample<feature_extractor>>.  That is, it should
+                  be some kind of array of process_sample objects.
+                - feature_extractor::force_last_weight_to_1 == false // TODO
```
davisking: Remove TODO statements like this from the documentation.

dlib/control/sarsa_abstract.h
```diff
@@ -0,0 +1,156 @@
...
+            ensures
+                - Trains a policy based on the given data and returns the results. The
+                  idea is to find a policy that will obtain the largest possible reward
+                  when run on the process that generated the samples. In particular,
+                  if the returned policy is P then:
+                    - P(S) == the best action to take when in state S.
+            notice
+                - Note that SARSA needs the actual next action taken so it doesn't process
```
davisking: This should be clarified.  In particular, it matters that the training samples are given in the order they happened in the environment.  That should be very clearly explained.  Other things, like lspi don't have this requirement.  For lspi, all the samples could be randomly sampled and it would work fine.   

Really for this you might be better off with a train that takes a vector of vector of samples, where each inner vector is one trial.

travek:
@adrianjav
Could you also add an example (cpp) of usage of both algorithms. And further publish them @ Examples: C++ page of the dlib site.

adrianjav:
There was something in the code that I didn't like. I've taken your review tips and reworked the code, I think it's much better now.
I'd like to add some helpers classes to easily do the training in an online fashion from a interactive environment (like the ones in openai's gym with step, end, .. methods).  The only concern I have rn is that for doing that I have to keep InputIterator non-const (or having mutables in the helper class which I don't like).

@travek fine by me, I'll write some examples when I consider the code solid enough

dlib/control/approximate_linear_models.h
```diff
@@ -56,6 +58,50 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 
     template <
+            typename feature_extractor
+            >
+    struct extended_process_sample : public process_sample<feature_extractor>
```
davisking: Please don't use inheritance here.  Only use inheritance when it makes sense to treat an object polymorphically.  But you would never want to pass an extended process sample to something expecting a process sample, or at least it would surprise me and I doubt it's what a user would really want to do.  

The idea of having an extended process sample is good though.  But make it a separate object from process_sample.

dlib/control/qlearning_abstract.h
```diff
@@ -136,9 +137,34 @@ namespace dlib
...
+                typename InputIterator,
+                typename EndIterator
+                >
+        policy<feature_extractor> train(
+            InputIterator iterator, const EndIterator& end_iterator
+        ) const;
+        /*!
+            requires
+                - iterator is an input iterator (according to the definition from the stl)
+                and it makes reference to something like std::vector<process_sample<feature_extractor>>>,
```
davisking: Please fix the formatting.

davisking:
Cool.  Looking pretty good.  Just a few more minor things :)

dlib/control/approximate_linear_models.h
adrianjav: Here I used inheritance with the intention of being able of using the three algorithms with the same code. You can see an example in the test code where I create extended_processes and by slicing they also work on qlearning & lspi.

I think whether it makes sense to use inheritance or not comes to whether people likely want to use those algorithms at the same time. Your right tho, in this case it might be better to not use it.

adrianjav:
Great, I'll fix them asap. This design is pretty much done.

However trying to make the training online I've found some design flaws and I've been thinking of adopting a design alike [openAI's gym](https://gym.openai.com/docs/), but I have some doubts I'd like to discuss before writing anything (some instant messagging app maybe?).

dlib/control/approximate_linear_models.h
davisking: Well, maybe you are right.  I'm on the fence about it after thinking a bit more.  If you want to leave it this way then that's cool with me :)

davisking:
Right, you probably want to think about the training termination conditions.  Right now it's setup to do one pass over the training data, but that's often going to be not enough.  It's worth waiting to merge this PR until you have run a real example though the code and found out if it's really convenient and practical.  Overall I think the code looks good from a software engineering point of view.  But getting the numeric solver aspect of this kind of thing right can be tricky.  In particular, the number of iterations, learning rate schedule, and termination conditions are often tough.

If you want to see a more complex example of SGD with an automatic learning rate schedule you can look at the dnn_trainer object which uses the running_gradient (http://dlib.net/algorithms.html#running_gradient) object to estimate how many iterations have transpired without substantive progress.  It's a very robust statistical test and lets the dnn_trainer know when to reduce the learning rate.  I use it for all the DNN training I do and it works pretty flawlessly.  That doesn't mean it's the right thing here, but if you are looking for ideas it's a good start.

adrianjav:
Sorry for the delay. I've been busy these days and two days ago I started the rework as I said. The code is completely different from the one at this PR but so far I find the new one quite nice.

Just for clarity I'll close this PR and open a new one in a few days when I feel like the other is ready to review.

travek:
@adrianjav 
Do you still plan to deliver SARSA and Q-learning to DLib library ? If yes, what planning do you concidertoday for this ?

adrianjav:
@travek Yep, I'm still planning to implement SARSA and Q-learning.

I'm sorry for the absence, but since my last post I've been involved in some stuff that doesn't let me any spare time. Actually, I was half-way through but I had to focus on other things.
Hopefully this Christmas I can get some time to finish and publish it.

about that, @davisking what are your thoughts about using C++11/14 and dlib's backward compability? I remember having trouble trying to write a new modern C++ interface that respects the old reinforcement learning code. I haven't found anything [here](http://dlib.net/howto_contribute.html)

davisking:
C++11 is good.  But don't use anything newer than that since there isn't
wide enough compiler support.

adrianjav:
Great, I'll keep in mind when I get into it. Thanks for the quick reply!

Issue #121, opened 2016-05-23
Python dlib => no extract_image_chips() and get_face_chip_details()


joeking11829:
Hi All,

I want to use dlib face alignment in my python program.
but, i cannot find method extract_image_chips() and get_face_chip_details() .

if i want to get alignment face using python dlib.
Should I implement it by myself or i missed something !?

Any comment !?

davisking:
Right, that's not in the pyhton API

AbdealiJK:
@davisking Any particular reason for that ?
Could you point in the direction to how someone can add it to the API ?

davisking:
There is no reason, just haven't gotten around to it.

All the code for the pyhton API is in the tools/python folder.

e-fominov:
Anybody working add this functions into python interface?

vinnitu:
I need it too.
Anybody have python wrapper yet?

PR #827, opened 2017-09-18
Image alignment and saving using Python

visionworkz:
Thank you for accepting my earlier code and adding it to dlib.

Here I have just added the size and padding as optional parameters for save_face_chip

### DIFF
--- a/python_examples/face_alignment.py
+++ b/python_examples/face_alignment.py
@@ -27,31 +27,26 @@
 #   Or downloaded from http://opencv.org/releases.html
 
 import sys
-import os
+
 import dlib
-import glob
 import cv2
 import numpy as np
 
-if len(sys.argv) != 4:
+if len(sys.argv) != 3:
     print(
         "Call this program like this:\n"
-        "   ./face_alignment.py shape_predictor_5_face_landmarks.dat dlib_face_recognition_resnet_model_v1.dat ../examples/faces/bald_guys.jpg\n"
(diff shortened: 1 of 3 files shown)


visionworkz:
Getting face chips aids in alignment as mentioned in this issue:
https://github.com/davisking/dlib/issues/121

davisking:
Thanks for another PR.  It generally looks good.  But you should clean up the example program.  For instance, why does it include the face recognition stuff when it isn't used?

visionworkz:
Thanks for your feedback @davisking. I have cleaned up the file.

visionworkz:
One thing I noticed is that calling save_face_chips frequently from Python (for example, once every time a face is detected, and also from different processes) results in some corrupt JPEG files though no error is thrown.

davisking:
Maybe you are writing to the same file?

visionworkz:
No. The file prefix I'm passing down is:
`datetime.datetime.now().isoformat()`

So for every batch of faces to be written to disk, the file prefix is quite unique.

I will try adding a UUID to it as well.

I looked at the code for save_jpeg, at first I thought it might be a memory problem if it is async. But looking at the code, it seems to be a sync call.

davisking:
I highly doubt it's something in libpng or dlib.  Python has horrible
support for multi-threading.  It's probably just a bug in how you are using
python.

visionworkz:
I'm using Python mainly to crawl datasets and build live (anonymous) ones while people pass by a camera.

davisking:

PR #839, opened 2017-09-23
imglab: add parts markers rapid fire

lorenzob:
Hi, it works like this. 

- Start imglab in "parts" mode. 
- Select a rectangle
- Now holding shift left click to add part markers, as many as you want. The counter will auto advance following the parts sequence
- use the context menu' as before to add parts and to reset the auto_add starting point right after that part

I think it's a huge improvement over the context menu. I'm not a c++ programmer so my code may need a good review.

What do you think?

### DIFF
--- a/dlib/gui_widgets/widgets.cpp
+++ b/dlib/gui_widgets/widgets.cpp
@@ -18,9 +18,6 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 // ----------------------------------------------------------------------------------------
 
-    // trz
-    int auto_part_num = 0;
-
     void toggle_button::
     set_size (
         unsigned long width,
@@ -6036,6 +6033,7 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+
     void image_display::
     on_part_add (
         const std::string& part_name
@@ -6062,14 +6060,16 @@ namespace dlib
         overlay_rects[selected_rect].parts[part_name] = c1;
         parent.invalidate_rectangle(rect); 
 
-        // trz
         // I do this here so that it advances when parts
         // are added from the context menu too
         // This also allows to restart the sequence from
         // a random point using the context menu
         int part_idx = std::distance(part_names.begin(), part_names.find(part_name));
-        //printf("part_idx: %d\n", part_idx);
-        auto_part_num = part_idx+1;
+        printf("part_idx: %d\n", part_idx);
+        if (part_idx < part_names.size())
+            auto_part_num = part_idx+1;
+        else
+            auto_part_num = 0;
 
         if (event_handler.is_set())
             event_handler();
@@ -6209,10 +6209,8 @@ namespace dlib
     {
         auto_mutex M(m);
 
-        // trz
         // reset it when we move to another image
         auto_part_num = 0;
-        //printf("reset_auto_part_num\n");
 
         overlay_rects.clear();
         overlay_lines.clear();
@@ -6510,14 +6508,12 @@ namespace dlib
     {
         scrollable_region::on_mouse_down(btn, state, x, y, is_double_click);
 
-        //trz
         /*
           Shift + click adds parts in sequence.
           I'm not sure if here is the best place to add this, but it looks like
           everything else is still working.
         */
         if (btn == base_window::LEFT && (state&base_window::SHIFT) && rect_is_selected) {
-            //printf("ADD PART\n");
 
             if (auto_part_num == part_names.size()) {
                 // I think it's safer to do nothing rather that to loop over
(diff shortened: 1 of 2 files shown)


dlib/gui_widgets/widgets.cpp
```diff
@@ -18,6 +18,9 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 // ----------------------------------------------------------------------------------------
 
+    // trz
+    int auto_part_num = 0;
```
davisking: Never ever use a global variable.  This should be a class member.

dlib/gui_widgets/widgets.cpp
```diff
@@ -6059,6 +6062,15 @@ namespace dlib
         overlay_rects[selected_rect].parts[part_name] = c1;
         parent.invalidate_rectangle(rect); 
 
+        // trz
+        // I do this here so that it advances when parts
+        // are added from the context menu too
+        // This also allows to restart the sequence from
+        // a random point using the context menu
+        int part_idx = std::distance(part_names.begin(), part_names.find(part_name));
```
davisking: This is wonky.  It would be clearer to save the string part name and use that to know what part is next.  Otherwise there are all kinds of problems, like what happens when parts are removed from the list and now the indexes don't make sense.

dlib/gui_widgets/widgets.cpp
```diff
@@ -6059,6 +6062,15 @@ namespace dlib
...
+        //printf("part_idx: %d\n", part_idx);
```
davisking: Don't leave commented out code

dlib/gui_widgets/widgets.cpp
```diff
@@ -6196,6 +6208,12 @@ namespace dlib
     )
     {
         auto_mutex M(m);
+
+        // trz
```
davisking: Don't put // trz, what does this tell the reader?

dlib/gui_widgets/widgets.cpp
```diff
@@ -6196,6 +6208,12 @@ namespace dlib
...
+        // reset it when we move to another image
+        auto_part_num = 0;
```
davisking: You need to reset the position in a lot of cases.  For instance, when someone clicks on a different rectangle in the same image.

dlib/gui_widgets/widgets.cpp
lorenzob: If they start to edit parts then everything is wonky. I didn't know it was possible, is it?

Saving the part name does not help as I may end up adding a removed part or I should check if the part is still there. And which is the part after the one that is no longer in the list?
Here I'm "iterating" over a list so I think it's clearer to use an index rather than an element and do a lookup to get back the index every time I need to move to the next. The semantic is that of an index, I wouldn't mask it to handle(?) some borderline cases.

What I can do is to reset the index to zero if the index is out of bounds. Or I could reset the index to zero every time the part list is modified.

But I cannot find the parts edit code, is there or maybe we are worrying about something that may never happen?

dlib/gui_widgets/widgets.cpp
lorenzob: I've added the reset on rectangle selection too. I'm not sure if there are any others, I couldn't find any by looking at the code and playing with the program.

dlib/gui_widgets/widgets.cpp
lorenzob: Fixed

dlib/gui_widgets/widgets.cpp
lorenzob: I've added the reset in the rectangle selection code. Are there any other cases? I couldn't find any after looking at the code and playing with the program.

dlib/gui_widgets/widgets.cpp
davisking: Yes, it is.  Not though imglab, but you have to look at the interface for the objects you are editing.  Those objects allow this, you can't just do things in some functions that will cause subsequent valid uses of the object to explode.  This is all the more important since dlib is a widely used library and who knows how many uses of these objects exist in other applications.

Yes, you have to check for everything that might happen.  Or to say it another way, you can never put the object into an invalid state.  Also, the entire API for all the objects in dlib is documented in great detail (see also http://dlib.net/howto_contribute.html#1).  These functions will say things like "Calling this function removes a part.".  It will never say something like "calling this function probably removes a part, or makes your program crash depending on how the users has been moving the mouse." because that would be a terrible contract for the function.  Anyway, my point is that you can read the documented interface for these objects to find out exactly what constitutes valid behavior and "valid vs. invalid state".  

As for the index, if the parts list changes then an integer index is meaningless.  Another thing to consider is that any time the parts list changes you should probably seek back to the start of the list.  In any case, an integer index is not inherently bad, but the code you have here is constantly dealing in the string names anyway.  You might as well skip the integer part and just use the string since the code would be simpler and detection of unexpected changes more durable.

davisking:
I just pushed a change that adds a version of this feature into the imglab tool.  Thanks for suggesting the idea :)

PR #841, opened 2017-09-25
Added an Eulerian Video Magnification example.

AronRubin:
Added an example that uses Eulerian Video Magnification in combination with dlib face tracking in-order to estimate the pulse rate of the face in the scene. It also highlights some areas for improvement in the dlib correlation tracker to provide a stable id for tracked objects.


davisking:
Thanks for the PR.

The example programs are supposed to be tutorials that teach how to use specific aspects of dlib.  This means they should essentially be essays that happen to be executable programs.  So a bunch of uncommented code doesn't fit that description, even if it does something cool.  A better thing to do is to create a new dlib class that encapsulates the important tool in this program and turn it into a documented and easily used tool.  Then you can use that tool in a short tutorial program that illustrates its use.  Also, please read the how to contribute guidelines for more detailed instructions:  http://dlib.net/howto_contribute.html

PR #845, opened 2017-09-27
cont layer: allow to set the number of filters at runtime

reunanen:
(implementation essentially copied from con layer)


davisking:
Thanks for another PR.  Don't forget to update the abstract.h file's documentation :)

reunanen:
Oops! Fixed.

davisking:
Sweet.  Thanks for the PR :)

PR #851, opened 2017-09-29
Fix warning

hwchiu:


tools/python/src/face_recognition.cpp
```diff
@@ -166,7 +166,6 @@ boost::python::list chinese_whispers_clustering(boost::python::list descriptors,
                 edges.push_back(sample_pair(i,j));
         }
     }
-    const auto num_clusters = chinese_whispers(edges, labels);
```
davisking: ?????

davisking:
You need to run the code to make sure it works before submitting a PR.  I'm saying this because this change obviously breaks the code.

tools/python/src/face_recognition.cpp
hwchiu: Sorry for my mistake. I will fix it.
And I want to know if we don't care the return value of `chinese_whispers`, do we need this variable ?

davisking: Variables that aren't used aren't needed.  But the function call is obviously hugely important.

hwchiu: Thanks, I have add it back.

davisking:

PR #857, opened 2017-10-06
Use banded Cholesky factorization if possible

seanwarren:
Hello, I have been using dlib for nonlinear optimisation with large numbers of variables. This works nicely - thank you! However in this case, the Cholesky factorization of the Hessian dominates the computation time as it scales as O(n^3). 

If the Hessian has a banded structure then it is possible to compute the factorisation in O(n\*n\*b) where b is the bandwidth. Since b << n, this can provide a considerable performance boost (\~300x in my case where n\~1500 and b\~3). 

This PR adds code to detect such banded Hessians and, if detected, uses a specialised Cholesky factorization routine. This detection step obviously imposes a small cost for non-banded matrices, however it should be negligible compared to the subsequent factorisation.

### DIFF
--- a/dlib/matrix/lapack/pbtrf.h
+++ b/dlib/matrix/lapack/pbtrf.h
@@ -1,0 +1,178 @@
+// Copyright (C) 2010  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_LAPACk_BDC_Hh_
+#define DLIB_LAPACk_BDC_Hh_
+
+#include "fortran_id.h"
+#include "../matrix.h"
+namespace dlib
+{
+    namespace lapack
+    {
+        namespace binding
+        {
+            extern "C"
+            {
+                void DLIB_FORTRAN_ID(dpbtrf) (const char* uplo, const integer* n, const integer* kd,
+                                              double* ab, const integer* ldab, integer* info);
+
+                void DLIB_FORTRAN_ID(spbtrf) (const char* uplo, const integer* n, const integer* kd,
+                                              float* ab, const integer* ldab, integer* info);
+
+            }
+
+            inline integer pbtrf (const char uplo, const integer n, const integer kd,
+                                  double* ab, const integer ldab)
+            {
+                integer info = 0;
+                DLIB_FORTRAN_ID(dpbtrf)(&uplo, &n, &kd, ab, &ldab, &info);
+                return info;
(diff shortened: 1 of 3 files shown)


davisking:
Cool, thanks for the PR.  This seems like a good idea. This change disables LAPACK usage for banded matrices, which could be bad in some cases.  Especially for matrices that are only slightly banded.  Can you update the PR to call a banded chol in LAPACK when it's available?

davisking:
Oh and definitely add a unit test for this.

seanwarren:
I've added a banded chol from LAPACK and a unit test with a simple banded matrix.

dlib/matrix/lapack/pbtrf.h
```diff
@@ -0,0 +1,178 @@
...
+            >
+        int pbtrf (
+            char uplo, matrix<T,NR1,NC1,MM,row_major_layout>& ab
+        )
+        {
+            const long ldab = ab.nr();
+            const long n = ab.nc();
+            const long kd = ldab - 1; // assume fully packed 
+
+            matrix<T,NC1,NR1,MM,row_major_layout> abt = trans(ab);
```
davisking: Can't you avoid this transpose by setting uplo to the appropriate value?

dlib/test/matrix.cpp
```diff
@@ -23,8 +24,27 @@ namespace
...
     using namespace std;
 
+    dlib::rand rnd;
+
     logger dlog("test.matrix");
 
+    template <typename type>
+    const matrix<type> rand_sp_banded(long n, long bw)
+    {
+        matrix<type> m = 10 * identity_matrix<type>(n);
```
davisking: The tests look good.  Make sure to add one for column major layout as well so both code paths are covered though.

dlib/matrix/lapack/pbtrf.h
seanwarren: Unfortunately not; storing the matrix in packed form (as required for input) breaks it's transpositional symmetry.  I've modified the chol function to ensure that the packed matrix is stored in a column major layout in practice this code path won't be used.

dlib/test/matrix.cpp
seanwarren: I have now added a column major layout test.

davisking:
Sweet, looks good.  Thanks for the nice PR :)

seanwarren:
Great, thank you.

PR #874, opened 2017-10-15
Add setter and getter for previous_loss_values_dump_amount

reunanen:
Wanted to increase `previous_loss_values_dump_amount` in my application.

### DIFF
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -574,6 +574,20 @@ namespace dlib
             return test_one_step_calls;
         }
 
+        void set_previous_loss_values_dump_amount (
+            int amount
+        )
+        {
+            DLIB_CASSERT(amount > 0);
+            previous_loss_values_dump_amount = amount;
+        }
+
+        int get_previous_loss_values_dump_amount (
+        ) const
+        {
+            return previous_loss_values_dump_amount;
+        }
+
     private:
 
         void record_test_loss(double loss)
--- a/dlib/dnn/trainer_abstract.h
+++ b/dlib/dnn/trainer_abstract.h
@@ -730,6 +730,29 @@ namespace dlib
                     - returns 0
         !*/
 
+        void set_previous_loss_values_dump_amount (
+            int amount
+        );
+        /*!
+            requires
+                - amount > 0
+            ensures
+                - Whenever the training routine decides to shrink the learning rate, it drops some
+                  previous loss values so that steps_without_progress will decrease below
+                  iter_without_progress_thresh. The number of previous loss values to drop is
+                  the parameter given to this function.
+        !*/
+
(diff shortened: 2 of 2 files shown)


davisking:
Hmm, was it terminating too early?  I've done a whole lot of testing and the current settings seemed very stable.  Although you never know.  But maybe there is some other problem here.

reunanen:
> Hmm, was it terminating too early?

Yes. Once the learning rate dropped from the initial 0.1, it quickly dropped all the way to some 1e-6. Cranking the number up to 2000 appeared to fix the problem.

> But maybe there is some other problem here.

Could be. I was using quite small mini-batches (size 10 or so); maybe that has something to do with it.

davisking:
Huh.  And when you set it to 2000 there was further decrease in the loss eventually right?  It didn't basically just grind at the same loss?

reunanen:
Correct. I bet 2000 was overkill, though.

davisking:
Hmm.  Damn.  I was hoping to have that parameter set to a value that never needed to be fiddled with.  All the things I've tested so far have worked as expected with the default value.  Can you run your code a few times with different values and see what happens?  How far off from 400 is needed?  Is it really because of the mini-batch size?

reunanen:
Ok – I'll try to. It'll take me a while to run these experiments, though.

davisking:
Thanks.  I almost don't want to have this option if it doesn't work all the time.  It does make training a lot faster in some cases, but that's not so helpful if it can't be relied upon.

reunanen:
Based on quick tests, it does look like: the smaller the mini-batch size, the more prone the learning process is to this problem (given the default value 400).

Let's assume you used a mini-batch size of around 80 most of the time. So for you, 400 steps * 80 samples per mini-batch = 32k samples. So, how about:

`previous_loss_values_dump_amount` = 32000 / mini-batch size

Just a random thought, really...

davisking:
Yeah, let me think about it.  We want to know the answer to "how many to keep so that the probability of false alarm stays below X?".  I'll see if I can figure out the right answer to that.  Maybe it's divide by N or sqrt(N) or something more complex.

davisking:
I just pushed a change that hopefully fixes this and removes the need for a user settable variable.  Can you pull the code and see if it fixes your issue?

reunanen:
> I just pushed a change that hopefully fixes this and removes the need for a user settable variable.  Can you pull the code and see if it fixes your issue?

Thanks, I'll try it at some point.

reunanen:
Looks like your fix is doing essentially the same thing as my application code, i.e. increasing the dump-amount relative to iterations-without-progress-threshold.

davisking:
Yep, seems to give good results in my testing.  Thanks for reporting this issue.

Issue #869, opened 2017-10-12
Reset or shrink momentum vector when shrinking learning rate


reunanen:
Not sure at all, but to me it looks like any momentum is preserved after learning rate is shrunk.

If so, then I guess as a result the optimizer might end up going in an unfavorable direction for quite a while, which may take a very long time to backtrack, given the new (smaller-than-before) learning rate.

I think it would make sense to reset or shrink also the momentum vector, whenever the learning rate is shrunk.

@davisking, can you comment if the above makes any sense? Just an observation from a single training run – didn't look in a lot of detail yet. If it already works like this, then I'm happy to close the issue right away.

reunanen:
I mean, even getting the momentum to zero may take quite a while if `learning_rate_shrink` is small.

davisking:
Do other frameworks do this?  In general I think they don't and many SGD
algorithms definitely don't do that kind of thing.  You can try it and see
what happens.  But I expect it's not better.

reunanen:
I have no idea if other frameworks do this. Just thought it'd make sense. Will perhaps try at some point.

reunanen:
I guess it makes less difference if learning rate is decreased gradually and not quickly by a factor of 10 or so.

davisking:
Maybe.

reunanen:
In fact, I observed the following sequence of events:

1. Loss goes up for a while.
2. As a consequence, learning rate is shrunk.
3. Loss keeps going up, possibly due to momentum.
4. Learning rate is shrunk even more, until it's really really small.
5. Loss keeps growing at a speed that can't be explained by the new, very small learning rate. Guessing it had to be momentum.

davisking:
It could be, but the momentum term gets small fast. With a momentum of 0.9
it takes about 22 steps to shrink it by a factor of 10, which is not very
many steps.   The "steps without progress" metric is going to be set to
look at way more than 20 steps when it's deciding if it should shrink the
learning rate.  So this shouldn't matter.

reunanen:
Right. Maybe it was something else then. Closing the issue, at least for now.

reunanen:
Indeed the behaviour is (qualitatively) the same even with momentum = 0. So it must be something else.

davisking:
There is also backtracking built into dnn_trainer.  If it sees that the loss suddenly goes up it should reload the solver state from disk and try again.  That should happen unless you have disabled it, assuming the loss is really going up.  Unless it's going up really slowly.

[...]

PR #875, opened 2017-10-15
Add get_net parameter that allows to call the function without forcing flush to disk

reunanen:
(see the discussion in #869)


davisking:
Looks good, thanks for another good PR :)

PR #880, opened 2017-10-16
disjoint_subsets: add get_number_of_sets and get_size_of_set functions

facug91:
The function get_number_of_sets returns the current number of sets in the UFDS.
The function get_size_of_set(i) returns the number of elements in the set represented by i.
It also add noexcept (C++11) directives to size and clear functions, to make them faster.


davisking:
Thanks for the PR.  This is a very nice PR.  Hardly anyone does it right on the first time :)  I like the documentation and the tests especially.  

Anyway, there is one issue.  This change has a performance impact for existing use cases.  For example, disjoint_subsets is the main computational component of the segment_image() routine (https://github.com/davisking/dlib/blob/master/dlib/image_transforms/segment_image.h) and this change will have a significant runtime impact.  If you look at the code you will see that the segment_image() is already doing this same bookkeeping to keep track of the set sizes, so one possible solution is to update segment_image() to use this new bookkeeping inside disjoint_subets.  However, there are other uses that don't require this bookkeeping at all, and more importantly, there are other dlib users out there who use this tool and won't be updating their code just to avoid the set size duplication of computation.

So, a better approach would be to make a `disjoint_subsets_sized` or `disjoint_subsets_with_set_sizes` class that contains an instance of disjoint_subsets or inherits from it and provides this additional functionality.  This way existing uses of the class don't pay the runtime overhead for this, but new uses can use it if appropriate.

facug91:
> Thanks for the PR. This is a very nice PR. Hardly anyone does it right on the first time :) I like the documentation and the tests especially.

Thanks for this Davis :)

> Anyway, there is one issue. This change has a performance impact for existing use cases. For example, disjoint_subsets is the main computational component of the segment_image() routine (https://github.com/davisking/dlib/blob/master/dlib/image_transforms/segment_image.h) and this change will have a significant runtime impact. If you look at the code you will see that the segment_image() is already doing this same bookkeeping to keep track of the set sizes, so one possible solution is to update segment_image() to use this new bookkeeping inside disjoint_subets. However, there are other uses that don't require this bookkeeping at all, and more importantly, there are other dlib users out there who use this tool and won't be updating their code just to avoid the set size duplication of computation.
> So, a better approach would be to make a disjoint_subsets_sized or disjoint_subsets_with_set_sizes class that contains an instance of disjoint_subsets or inherits from it and provides this additional functionality. This way existing uses of the class don't pay the runtime overhead for this, but new uses can use it if appropriate.

I've made a sub-class disjoint_subsets_sized, as you said. Anyway, it could have a little performance issue, because now methods in disjoint_subsets are virtual, but it's so little that I don't know if you care about it. If this is an issue, I can try to think something else.

davisking:
Thanks for the update.  

Don't make it virtual, that will impact the performance of some uses.  It's better to use containment in this case rather than inheritance.  Also, please use spaces instead of tabs.

facug91:
Sorry about the tabs, my IDE added them automatically.
Now disjoint_subsets_sized is implemented in terms of disjoint_subsets, instead of inherit from it.

davisking:
Looks good, thanks for the PR  :)

Issue #881, opened 2017-10-17
SVM threaded cross validation trainer return type inconsistent and fails to compile


jpblackburn:
The return value of cross_validate_trainer is 
`const matrix<double, 1, 2, typename trainer_type::mem_manager_type>`

The return value of cross_validate_trainer_threaded is
`const matrix<typename trainer_type::scalar_type, 1, 2, typename trainer_type::mem_manager_type>`

Note that the value type of the matrix switched from `double` to `trainer_type::scalar_type`.  This is confusing as I expected the threaded version to be a drop-in replacement for the unthreaded version with the exception of the number of threads argument.  

Even worse, the threaded version fails to compile if `trainer_type::scalar_type` is not `double`.  I do not mind the return type being double even if `trainer_type::scalar_type` is something else (like `float`), but it should compile.  The error is a compile time assert from matrix/matrix.h line 1474 that the assignment has incompatible types.  The error message references svm/svm_threaded.h line 60.  

I think the issues is that `cvtti_helpers::task::operator()` is instantiated with `matrix_type` = `matrix<trainer_type::scalar_type, 1, 2, trainer_type::mem_manager_type>`.  See svm/svm_threaded.h lines 55, 140, and 211.  Unfortunately, the return type of `test_binary_decision_function` is `const matrix<double,1,2> test_binary_decision_function`.  This causes an assignment  at svm/svm_threaded.h line 60 from a `matrix<double,1,2>` to a `matrix<trainer_type::scalar_type,1,2>`.  

It would see like changing the return type of cross_validate_trainer_threaded to `matrix<double,1,2>` and making the corresponding changes to cvtti_helpers::task would solve both issues.  

https://github.com/davisking/dlib/blob/38a2846ddad5a81fab9905b76d061ad473001eba/dlib/svm/svm.h#L536-L547

https://github.com/davisking/dlib/blob/38a2846ddad5a81fab9905b76d061ad473001eba/dlib/svm/svm_threaded.h#L227-L239

https://github.com/davisking/dlib/blob/38a2846ddad5a81fab9905b76d061ad473001eba/dlib/matrix/matrix.h#L1472-L1475

https://github.com/davisking/dlib/blob/38a2846ddad5a81fab9905b76d061ad473001eba/dlib/svm/svm_threaded.h#L48-L60

https://github.com/davisking/dlib/blob/38a2846ddad5a81fab9905b76d061ad473001eba/dlib/svm/svm_threaded.h#L140

[...]

PR #883, opened 2017-10-17
Allow cross_validate_trainer_threaded to use non-double data

jpblackburn:
This fixes issue #881


davisking:
Looks good, thanks for the PR.

PR #897, opened 2017-10-20
Replace push_back to emplace_back to perfect forwarding the arguments

michal915:


davisking:
Thanks for the PR.  Did you see speed improvements with these?  This looks like a blanket transformation was applied to push_back.  I don't really want to do this unless there is some benefit since this code works.  Making stylistic changes with no upside and the possibility of adding bugs is not great.

PR #901, opened 2017-10-21
Missing interfaces in add_prev for compatibility with mmod loss.

grochefort:
Hi,

I should be wrong, but I got some errors when trying to use an add_prev layer with a mmod loss.
I fixed it and hope it could be usefull for someone else.

Regards,
Gilles Rochefort.

PS:
Following is a small snippet of the errors I got
> In file included from 3rdparty/dlib/dlib-src/dlib/../dlib/dnn/layers.h:15:
> 3rdparty/dlib/dlib-src/dlib/../dlib/dnn/utilities.h:218:41: error: no member named 'map_output_to_input'
>  in 'dlib::add_prev_<tag1>'
>                 p = net.layer_details().map_output_to_input(p);
>                     ~~~~~~~~~~~~~~~~~~~ ^

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2078,6 +2078,9 @@ namespace dlib
         const tensor& get_layer_params() const { return params; }
         tensor& get_layer_params() { return params; }
 
+        inline dpoint map_input_to_output (const dpoint& p) const { return p; }
+        inline dpoint map_output_to_input (const dpoint& p) const { return p; }
+
         friend void serialize(const add_prev_& , std::ostream& out)
         {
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  This is a good idea.  I was initially hesitant to add it to this particular layer because I was worried about users getting into trouble since it's possible to make network architectures where the mapping is weird (e.g. two layers with different ways of mapping to the input image are merged, then which mapping do you pick?).  But that doesn't seem to be likely given the experience I've been having.  All the other layers that do this kind of merging have been updated to allow this, I just forgot about add_prev.  So this is a good change.

PR #906, opened 2017-10-23
pedantic: fix "extra `;`" warnings when compiling with -Wpedantic

fenollp:
Doesn't add anything, just makes my compilation way calmer :)


davisking:
Cool, thanks for the PR :)

PR #913, opened 2017-10-27
Win lapack

seanwarren:
On windows, it seems like the CMake build only looks for the Intel MKL versions of LAPACK and BLAS. This PR uses find_package to try and find LAPACK and BLAS. This is useful if you have installed these libraries via vcpkg, for example.


davisking:
Good idea, thanks for the PR.

PR #919, opened 2017-11-04
Remove unused variable

grochefort:
Just removed the offending unused variable. Not sure it is the way to fix it but it suppress some warning.


davisking:

PR #934, opened 2017-11-09
Determine lapack fortran linking convention in CMake

seanwarren:
LAPACK functions are often decorated in different ways by the compiler. Currently, dlib attempts to determine the correct decoration by testing the compiler flags in fortran_id.h. However, this fails when using CLAPACK on windows as this package uses decorated (trailing underscore) function names whereas the platform FORTRAN compiler does not. 

This commit explicitly looks for decorated and undecorated versions of the function in CMAKE and, if detected, sets a preprocessor definition which is detected in fortran_id.h. The old detection approach is still used if this search fails. This definition is propagated using target_compile_definitions(dlib INTERFACE ...)


dlib/CMakeLists.txt
```diff
@@ -502,6 +502,11 @@ if (NOT TARGET dlib)
          if (DLIB_USE_LAPACK)
             if (lapack_found)
                set (dlib_needed_libraries ${dlib_needed_libraries} ${lapack_libraries})
+               if (lapack_with_underscore)
+                  set (dlib_needed_definitions ${dlib_needed_definitions} -DLAPACK_FORCE_UNDERSCORE)
```
davisking: You should add these things to the `active_preprocessor_switches` variable using `list(APPEND active_preprocessor_switches "-DWHATEVER")` rather than introducing a new variable.

seanwarren:
Thanks for the comments - I've now modified the code to add the preprocessor switches correctly and added them to config.h.in

dlib/cmake_utils/cmake_find_blas.txt
```diff
@@ -346,3 +348,22 @@ if (NOT lapack_found)
     endif()
 endif()
 
+# If using lapack, determine whether to mangle functions
+if (lapack_found)
+   include(CheckFunctionExists)
+   include(CheckFortranFunctionExists)
+   set(CMAKE_REQUIRED_LIBRARIES ${LAPACK_LIBRARIES})
```
davisking: I just tested this and it doesn't work unless `LAPACK_LIBRARIES` is all lowercase since cmake is case sensitive.  For instance, I've got the Intel MLK installed and this bit of code doesn't find it because all the other variables in this file use the lowercase version of this variable name.  Can it be changed to lowercase or is there a reason you did it this way?

seanwarren: Apologies - this was just an oversight. I've now changed the variable to the correct capitalization

davisking:

PR #935, opened 2017-11-09
Remove explicit specification of library path in dlibConfig.cmake

seanwarren:
Currently, dlib hardcodes dlib_LIBRARIES in dlibConfig.cmake. This is not required as the library and dependencies will be picked up in a configuration dependent manner by dlib-{configuration}.cmake if dlib_LIBRARIES is set to dlib::dlib. The current approach means that different configurations (Release/Debug/...) cannot easily be installed side-by-side on Windows. 

This PR removes this hardcoded library path. I've tested that CMake successfully detects the library and dependencies on Windows 7 and macOS.

### DIFF
--- a/dlib/cmake_utils/dlibConfig.cmake.in
+++ b/dlib/cmake_utils/dlibConfig.cmake.in
@@ -29,6 +29,7 @@ if(NOT TARGET dlib-shared AND NOT dlib_BINARY_DIR)
 endif()
 
 set(dlib_LIBRARIES dlib::dlib)
+set(dlib_LIBS      dlib::dlib)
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR, seems good. You should leave the `dlib_LIBS` variable in there though so existing users aren't broken by it vanishing.  There is also a deprecation warning in there when people use these variables.  So hopefully they will fall from use soon anyway.  But in the meantime we shouldn't break anything.

seanwarren:
Good point, I've restored the dlib_LIBS variable

davisking:
Cool, thanks again for the PR :)

PR #937, opened 2017-11-10
Keep recent losses and current step counts when restoring previous state from disk

reunanen:
Problem: It happened to me that the training state was saved to disk when the previous loss values were quite low. After this, the loss generally started to go up. At the same time, I had `iterations_without_progress_threshold` set to a relatively high value, at least when compared to `time_between_syncs`, so the threshold was not reached before another sync-to-disk-call was due. Therefore, the learning rate wasn't shrunk, but instead the old state was restored from disk. Consequently, the learning rate never went down, and the process just kept restoring the old state from disk. (I guess it would have gone down at some point, with a probability of 1, but hey – life is short...)

Solution: When loading a previously saved training state from disk, do not load the previous loss values, but rather keep the values currently in memory. This way, we should eventually reach `iterations_without_progress_threshold`, shrink the learning rate, and get back on track.

To avoid confusion, also keep the step counts (not sure if this part is necessary / useful).

### DIFF
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -1012,28 +1012,9 @@ namespace dlib
                 // previously saved state in the hopes that the problem won't reoccur.
                 if (loss_increased_since_last_disk_sync()) 
                 {
-                    // keep the loss values, because otherwise we might end up never shrinking the learning rate
-                    const auto previous_loss_values_to_keep = std::move(previous_loss_values);
-                    const auto test_previous_loss_values_to_keep = std::move(test_previous_loss_values);
-
-                    // keep also the step counts, in order to avoid confusion
-                    const auto train_one_step_calls_to_keep = train_one_step_calls;
-                    const auto test_one_step_calls_to_keep = test_one_step_calls;
-                    const unsigned long steps_without_progress_to_keep = steps_without_progress;
-                    const unsigned long test_steps_without_progress_to_keep = test_steps_without_progress;
-
                     std::ifstream fin(newest_syncfile(), std::ios::binary);
                     deserialize(*this, fin);
                     sync_file_reloaded = true;
-
-                    // restore the loss and the step count values
-                    previous_loss_values = std::move(previous_loss_values_to_keep);
-                    test_previous_loss_values = std::move(test_previous_loss_values_to_keep);
-                    train_one_step_calls = train_one_step_calls_to_keep;
-                    test_one_step_calls = test_one_step_calls_to_keep;
-                    steps_without_progress = steps_without_progress_to_keep;
-                    test_steps_without_progress = test_steps_without_progress_to_keep;
-
                     if (verbose)
                         std::cout << "Loss has been increasing, reloading saved state from " << newest_syncfile() << std::endl;
                 }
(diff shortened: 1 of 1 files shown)


dlib/dnn/trainer.h
```diff
@@ -1012,9 +1012,28 @@ namespace dlib
...
+                    // keep the loss values, because otherwise we might end up never shrinking the learning rate
+                    const auto previous_loss_values_to_keep = std::move(previous_loss_values);
+                    const auto test_previous_loss_values_to_keep = std::move(test_previous_loss_values);
+
+                    // keep also the step counts, in order to avoid confusion
+                    const auto train_one_step_calls_to_keep = train_one_step_calls;
+                    const auto test_one_step_calls_to_keep = test_one_step_calls;
+                    const unsigned long steps_without_progress_to_keep = steps_without_progress;
+                    const unsigned long test_steps_without_progress_to_keep = test_steps_without_progress;
+
```
reunanen: I guess these 6 lines of code could be moved to a function that returns a struct. Then, after deserializing, the struct could be passed to another function.

At only 6 lines (*), not sure if doing so would make things any cleaner. But if there will be more variables that should be handled like this, then it'd make sense, I suppose.

(*): Or even less? Still not sure about the step counts...

davisking:
Well, the whole point of reloading is to fix a problem where the trainer drops the loss too early because of a sudden spike in loss.  It also helps that it reloads old network parameters that might be less busted by the loss spike, but it's important that the loss values get thrown away too to prevent early termination. 

The logic in loss_increased_since_last_disk_sync() should only allow it to reload a few times before it gives up.  I forget how many times, it does this exponential decay thing, but it shouldn't let it reload more than like 8 times before it gives up.

reunanen:
> Well, the whole point of reloading is to fix a problem where the trainer drops the loss too early because of a sudden spike in loss.

Hmm, ok. I was reading [this](https://github.com/davisking/dlib/blob/master/dlib/dnn/trainer.h#L1015-L1018) a little differently.

I'm now thinking of either closing this PR, or adding your above clarification to the code as a comment.

> The logic in loss_increased_since_last_disk_sync() should only allow it to reload a few times before it gives up. I forget how many times, it does this exponential decay thing, but it shouldn't let it reload more than like 8 times before it gives up.

Ok, thanks for the explanation. I guess I just didn't wait long enough, then.

davisking:
That change to the comment is confusing.  Now I don't know what it's saying :)

I'm just saying, before I had this logic in there you would have this problem of early termination.  I would come look at my stuff and it would have suddenly terminated with a crappy solution, when if the optimizer had just kept running it would have been fine.  I'm pretty sure reloading the network parameters helps too, but that's secondary, at least most of the time it seems less important.  

Also, if it keeps reloading from a save then something has probably gone terribly wrong in the optimization.  Like maybe it's just generally unstable for some reason.  Maybe the learning rate is too high to the network is too hard to learn.

reunanen:
> That change to the comment is confusing. Now I don't know what it's saying :)

Ok, I'll just close this PR then.

> Also, if it keeps reloading from a save then something has probably gone terribly wrong in the optimization. Like maybe it's just generally unstable for some reason. Maybe the learning rate is too high to the network is too hard to learn.

I think you are right. I believe my input to `std::log` in calculating logarithmic loss sometimes goes too small, even 0, which creates stability issues. I think I'll create another PR where instead of `std::log(g[idx])`, something like `std::log(std::max(g[idx], 1e-6))` is calculated (in log-loss).

davisking:
Cool.  Test it a bunch and be careful :)

It's really easy to mess up an optimizer by adding clipping or fudge
factors like that.  It's tricky to get right.

PR #938, opened 2017-11-10
Problem: log loss may become infinite, if g[idx] goes zero

reunanen:
Solution: limit the input of the log function to 1e-6 (or more)

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -239,10 +239,10 @@ namespace dlib
     };
 
     template <typename T>
-    T safe_log(T input)
+    T safe_log(T input, T epsilon = 1e-10)
     {
         // Prevent trying to calculate the logarithm of a very small number (let alone zero)
-        return std::log(std::max(input, static_cast<T>(1e-6)));
+        return std::log(std::max(input, epsilon));
     }
 
     template <typename SUBNET>
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -238,6 +238,13 @@ namespace dlib
 
     };
 
+    template <typename T>
+    T safe_log(T input)
+    {
+        // Prevent trying to calculate the logarithm of a very small number (let alone zero)
+        return std::log(std::max(input, static_cast<T>(1e-6)));
+    }
+
     template <typename SUBNET>
     using loss_binary_log = add_loss_layer<loss_binary_log_, SUBNET>;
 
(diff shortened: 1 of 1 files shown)


davisking:
This might be a good idea.  Do some testing first though on the problem that was causing trouble, and also on some other problems to make sure this doesn't have bad side effects.   For instance, maybe 1e-6 is way too big.

reunanen:
At least [here](https://community.drivendata.org/t/0s-and-logarithmic-loss-metric/51/2), values like 1e-5 and 1e-15 are suggested.

But yes, I'll test.

davisking:
I know, but you need to test it.  I write a lot of optimizers.  These
little details sometimes cause me a lot of grief, especially considering
how simple they seem.  Without testing you don't see the effects.

reunanen:
Yes, I understand. But at least this change affects only the loss and not the objective function, right? Although surely the loss affects other stuff (as discussed in #937).

reunanen:
But as said, yes – I'll test. :)

davisking:
Yeah.  I think this is probably the right thing to do :) 

But if I had a dollar for every time I was in situations like this and thought something was the right thing, but upon detailed testing found out that there was some subtle issue, I would have a lot of dollars :)

reunanen:
At the very least, this does fix the problem that led me to do #937. (The problem was relatively easy to reproduce, so I can already say this with some certainty.)

I'll just keep this change included in everything I do, and see if any new issues arise. I'm doing a lot of log-loss based semantic segmentation for relatively large amounts of image data, so this change will get tested extensively.

davisking:
Sounds good.  Let me know when it's ready to merge :)

reunanen:
Sure. Thanks again for your help!

reunanen:
Results on the VOC 2012 data:

```
step#: 117761  learning rate: 0.0001  average loss: 0.0340257    steps without apparent progress: 2954
step#: 117847  learning rate: 0.0001  average loss: 0.0346309    steps without apparent progress: 4908
step#: 117933  learning rate: 0.0001  average loss: 0.0330298    steps without apparent progress: 3227
step#: 118019  learning rate: 0.0001  average loss: 0.0339596    steps without apparent progress: 4983
Saved state to pascal_voc2012_trainer_state_file.dat_
saving network
Testing the network...
train accuracy  :  0.965768
val accuracy    :  0.835585
```

So not really much better (than what was reported in #943), but not worse either.

The change in this PR makes a bigger difference with some other datasets, but for the better in my experience. So I would say it's now ready to merge.

davisking:
Wasn't this the output before?

```
train accuracy  :  0.990882
val accuracy    :  0.849609
```

reunanen:
Only with batch normalization right before the loss layer.

davisking:
Right, duah.  Cool, looks good then :)

Thanks for another PR.

PR #953, opened 2017-11-14
Update input.h

MambaWong:
I think this may makes it easier to understand.


davisking:
Thanks for the PR, but this doesn't seem any simpler to read to me.  I also don't want to spend time testing something that's a purely cosmetic code change.

PR #964, opened 2017-11-16
imglab: select next/previous image with 's' and 'w'

hannometer:
Less hand movement from <TAB> position to arrow keys - this helps me annotate faster.

### DIFF
--- a/tools/imglab/src/metadata_editor.cpp
+++ b/tools/imglab/src/metadata_editor.cpp
@@ -354,6 +354,14 @@ on_keydown (
             select_image(image_pos);
         }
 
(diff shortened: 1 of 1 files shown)


tools/imglab/src/metadata_editor.cpp
```diff
@@ -354,6 +354,14 @@ on_keydown (
             select_image(image_pos);
         }
 
+        if (key == 'w' && !overlay_label.has_input_focus())
```
davisking: If you want to add more keys you should add the check as part of the `if (key == base_window::KEY_whatever)` if.  That way all the other keyboard mods like shift and control will work for a and s as well as up and own.

davisking:

PR #966, opened 2017-11-17
Update dnn_mmod_find_cars2_ex.cpp

amincheloh:


davisking:
Good catch, thanks for the PR :)

alexandreofbh:
Why in this network, despite having two outputs, front and rear of the vehicle, is the network defined with only one output?
using net_type = loss_mmod<con<1,9,9,1,1,rcon5<rcon5<rcon5<downsampler<input_rgb_image_pyramid<pyramid_down<6>>>>>>>>;

davisking:
It's not actually 1 channel of output, see http://dlib.net/dnn_mmod_train_find_cars_ex.cpp.html.  In particular this line
```
    // The final layer of the network must be a con layer that contains 
    // options.detector_windows.size() filters.  This is because these final filters are
    // what perform the final "sliding window" detection in the network.  For the dlib
    // vehicle dataset, there will be 3 sliding window detectors, so we will be setting
    // num_filters to 3 here.
    net.subnet().layer_details().set_num_filters(options.detector_windows.size());
```

alexandreofbh:
Oh, I see, I was actually thinking about that, but I got confused and thought that filters and outputs were different things.
Thank you very much!

alexandreofbh:
Hello Davis King,

I am trying to create a classifier for uppercase characters and numbers in photos (2..9, A..Z). In total, there are 34 different labels, as I am considering the number 0 as the letter 'O' and the number 1 as the letter 'I'.

I am trying to use the structure from the example file 'dnn_mmod_train_find_cars_ex.cpp' for training.

In cases where there are blurred characters or characters that are too small to identify, I mark a box without any label as an ignored box for these uncertain characters, such as the noisy 'O' and 'Q'.

To avoid confusion between '1' and 'I', '0' and 'O', whenever the number 0 or 1 appears, I label them as the letter 'O' and 'I' respectively.

So far, I haven't had any success using the 'net_type' structure from 'dnn_mmod_train_find_cars_ex.cpp'. Even when I adjusted the 'nc' and 'nr' parameters, as well as the 'stride_x' and 'stride_y' in the network, to reduce the downsampling, because the boxes are smaller than the fronts and rears of the cars.

With your experience, I would like to know if you think that using this 'net_type' and 'dnn_mmod_train_find_cars_ex.cpp' approach, I have a chance of succeeding with this working, or if I should consider exploring another path?

I also made the following adjustments:
#define LONGER_SIDE_BOX 12
#define SHORTER_SIDE_BOX 10
mmod_options options(boxes_train, LONGER_SIDE_BOX, SHORTER_SIDE_BOX);
cropper.set_randomly_flip(false);
cropper.set_min_object_size(LONGER_SIDE_BOX-1, SHORTER_SIDE_BOX-2);

I would greatly appreciate your suggestions, and thank you for always responding to us!

davisking:
Those are probably too small for that network.  Take note of the comments right above `mmod_options options(boxes_train, 70, 30);` in that example.  The stride is 8 which is pretty big for objects that are 10ish pixels wide.   You could upsample the image and try that, which would work but becomes very slow.  You probably want to use a network specifically designed for finding small letters.

PR #971, opened 2017-11-20
generic image all the way - colormaps.h

kino-dome:
Still hunting down pre-generic_image implementations


kino-dome:
Hope this is the last bit :)

kino-dome:
BTW Davis if I want to publish my bridge-to-dlib code, which dictation you'd like better? "Dlib" or "dlib" ? I know it's a very specific question but sometimes people actually have preference of one over another, so I figured I'll ask :) Cheers!

davisking:
Cool, thanks for another PR :)

I generally prefer dlib.

PR #980, opened 2017-11-25
Exposed jitter_image in Python and added an example

visionworkz:
Exposed jitter_image function in Python API along with an example.

This is useful while building datasets using the Python API.


davisking:
Thanks for the PR.  I don't see anything wrong with the code but I do have a question.  What are you doing with the jittered images?  The DNN training API isn't exposed through python.  Are you planning on exposing part of that as well?

visionworkz:
My thought was that jittering could be useful while building a dataset:
1. Get images for class
2. Align
3. Cluster
4. Jitter
5. Store

Another use, while training a classifier:
Let training data consist of _**m**_ images with labels
1. For (image, label) in raw image training data
2. Jitter image **_n_** times
3. Get **_n_** representations from model (for the **_n_** jittered images)
4. Add to training data (descriptor_1, label), (descriptor_2, label), ...., (descriptor_n, label)
5. Repeat until all **_m_** images are done
6. Now training data has **_n x m_** (descriptor, label) entries
7. Train a classifier on the **_n x m_** dataset 

I could consider porting the training part as well though the current C++ example serves me well.

davisking:
You definitely don't want to store jittered images to disk.  Jittering is
much more efficiently done one the fly.

It makes sense for training models though.  My only reservation is that
this python interface to jittering is going to be really slow because of
all the list building.  So I would never use it myself and I'm generally
hesitant to put things in dlib that I wouldn't be willing to use.  Isn't
there a way to make returning an image to python more efficient?

visionworkz:
I guess we could pass matrix<rgb_pixel> to Python. But at the end of the day, would still need to convert from that data structure to numpy array for OpenCV stuff.

I did think about the list construction, but at the end of the day, since rgb_pixel is a class, I am not sure we can do some kind of memory map.

It seems to me once would need to iterate through the pixels either in C++ like it is now or in Python to get to a Numpy image structure.

Do you have any better ideas? I could look into improving it.

davisking:
Python list construction is much slower than looping over pixels in C++
since the python list involves a lot f pointer indirection, memory
allocations (this means mutex locks and system calls in most cases), page
faults, and so on.

I would either avoid using opencv entirely, thus allowing you to use a
matrix<rgb_pixel> directly, or figure out how to construct numpy objects
from C++ so you can copy the image into one of them in C++ code, which will
be thousands of times faster than doing all that list stuff.

visionworkz:
I am able to return numpy arrays with the following code snippet:
```
#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
#include <numpy/ndarrayobject.h>

// ... previous code

boost::python::list get_jitter_images(object img, size_t num_jitters = 1, bool disturb_colors = false)
{
    if (!is_rgb_python_image(img))
        throw dlib::error("Unsupported image type, must be RGB image.");

    // Convert the image to matrix<rgb_pixel> for processing
    matrix<rgb_pixel> img_mat;
    assign_image(img_mat, numpy_rgb_image(img));

    // The top level list (containing 1 or more images) to return to python
    boost::python::list jitter_list;

    // Size of the numpy array
    npy_intp dims[3] = { num_rows(img_mat), num_columns(img_mat), 3};

    for (int i = 0; i < num_jitters; ++i) {
        // Get a jittered crop
        matrix<rgb_pixel> crop = dlib::jitter_image(img_mat, rnd_jitter);
        // If required disturb colors of the image
        if(disturb_colors)
            dlib::disturb_colors(crop, rnd_jitter);

        PyObject *pyObj = PyArray_SimpleNewFromData(3, dims, NPY_UINT8, image_data(crop));
        boost::python::handle<> handle(pyObj);
        // Append image to jittered image list
        jitter_list.append(object(handle));
    }
           
    return jitter_list;
}

BOOST_PYTHON_FUNCTION_OVERLOADS(get_jitter_images_with_defaults, get_jitter_images, 1, 3)

void bind_image_classes()
{
    using boost::python::arg;
    import_array();

    {
    class_<rgb_pixel>("rgb_pixel")
        .def(init<unsigned char,unsigned char,unsigned char>( (arg("red"),arg("green"),arg("blue")) ))
        .def("__str__", &print_rgb_pixel_str)
        .def("__repr__", &print_rgb_pixel_repr)
        .add_property("red", &rgb_pixel::red)
        .add_property("green", &rgb_pixel::green)
        .add_property("blue", &rgb_pixel::blue);
    }

    def("jitter_image", &get_jitter_images, get_jitter_images_with_defaults(
    "Takes an image and returns a list of jittered images."
    "The returned list contains num_jitters images (default is 1)."
    "If disturb_colors is set to True, the colors of the image are disturbed (default is False)", 
    (arg("img"), arg("num_jitters"), arg("disturb_colors"))
    ));
}

```

I also changed the cmake file (`tools/python/CMakeLists.txt`) to check for presence of numpy:
```

CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)


set(USE_SSE4_INSTRUCTIONS ON CACHE BOOL "Use SSE4 instructions")
# Make DLIB_ASSERT statements not abort the python interpreter, but just return an error.
add_definitions(-DDLIB_NO_ABORT_ON_2ND_FATAL_ERROR)

include(../../dlib/cmake_utils/add_python_module)

# Test for numpy
FIND_PACKAGE(PythonInterp)
IF(PYTHONINTERP_FOUND)
    EXECUTE_PROCESS( COMMAND ${PYTHON_EXECUTABLE} -c "import numpy" OUTPUT_QUIET ERROR_QUIET RESULT_VARIABLE NUMPYRC)
    IF(NUMPYRC EQUAL 1)
        MESSAGE(FATAL_ERROR "Python interpreter found but 'numpy' package not available")
    ELSE(NUMPYRC EQUAL 1)
        MESSAGE(STATUS "Found Python with installed numpy package")
        EXECUTE_PROCESS( COMMAND ${PYTHON_EXECUTABLE} -c "import sys; from numpy import get_include; sys.stdout.write(get_include())" OUTPUT_VARIABLE NUMPY_INCLUDE_PATH)
        MESSAGE(STATUS "Numpy include path '${NUMPY_INCLUDE_PATH}'")
        include_directories(${NUMPY_INCLUDE_PATH})
    ENDIF(NUMPYRC EQUAL 1)
ELSE(PYTHONINTERP_FOUND)
    MESSAGE(FATAL_ERROR "Could not find Python interpreter")
ENDIF(PYTHONINTERP_FOUND)

#include_directories($(python -c "from numpy import get_include; print(get_include())"))

add_definitions(-DDLIB_VERSION=${DLIB_VERSION})

# Tell cmake to compile all these cpp files into a dlib python module.
set(python_srcs
   src/dlib.cpp
   src/matrix.cpp
   src/vector.cpp
   src/svm_c_trainer.cpp
   src/svm_rank_trainer.cpp
   src/decision_functions.cpp
   src/other.cpp
   src/basic.cpp
   src/cca.cpp
   src/sequence_segmenter.cpp
   src/svm_struct.cpp
   src/image.cpp
   src/rectangles.cpp
   src/object_detection.cpp
   src/shape_predictor.cpp
   src/correlation_tracker.cpp
   src/face_recognition.cpp
   src/cnn_face_detector.cpp
)

# Only add the GUI module if requested
if(NOT ${DLIB_NO_GUI_SUPPORT})
   list(APPEND python_srcs src/gui.cpp)
endif(NOT ${DLIB_NO_GUI_SUPPORT})

add_python_module(dlib ${python_srcs})

# When you run "make install" we will copy the compiled dlib.so (or dlib.pyd)
# library file to the python_examples folder.
install_dlib_to(../../python_examples)
```

As long as the numpy module is installed via pip, `tools/python' can be built.

But I am facing a problem of the images not being correctly returned to Python. The returned object is an ndarray with the right dimensions but in certain instances only the blue channel is present and all other channel values are 0, resulting in a blue image.

Could you help spot the issue?

davisking:
You aren't using `PyArray_SimpleNewFromData` right.  I googled that function and the manual tells you what's wrong with what you did.  I'm also not super excited about adding a numpy dependency.  Isn't it possible to return an appropriate buffer to python without making the dlib compile dependent on numpy headers?

visionworkz:
@davisking thanks for pointing that out. Here is my modified function:

```
boost::python::list get_jitter_images(object img, size_t num_jitters = 1, bool disturb_colors = false)
{
    if (!is_rgb_python_image(img))
        throw dlib::error("Unsupported image type, must be RGB image.");

    // Convert the image to matrix<rgb_pixel> for processing
    matrix<rgb_pixel> img_mat;
    assign_image(img_mat, numpy_rgb_image(img));

    // The top level list (containing 1 or more images) to return to python
    boost::python::list jitter_list;

    // Size of the numpy array
    npy_intp dims[3] = { num_rows(img_mat), num_columns(img_mat), 3};

    for (int i = 0; i < num_jitters; ++i) {
        // Get a jittered crop
        matrix<rgb_pixel> crop = dlib::jitter_image(img_mat, rnd_jitter);
        // If required disturb colors of the image
        if(disturb_colors)
            dlib::disturb_colors(crop, rnd_jitter);

        void* img_data_copy = malloc(img_mat.nr() * width_step(crop));
        memcpy(img_data_copy, image_data(crop), img_mat.nr() * width_step(crop));
        PyObject *pyObj = PyArray_SimpleNewFromData(3, dims, NPY_UINT8, img_data_copy);
        PyArray_ENABLEFLAGS((PyArrayObject*)pyObj, NPY_ARRAY_OWNDATA);
        boost::python::handle<> handle(pyObj);
        // Append image to jittered image list
        jitter_list.append(object(handle));
    }
           
    return jitter_list;
}
```

There are many questions regarding returning a Numpy array from dlib including this old question:
https://sourceforge.net/p/dclib/discussion/442518/thread/533998a7/

I myself was hoping to return a Numpy array straightaway from the c++ code.

However, if you think adding Numpy dependency is undesirable, you can simply close this pull request.

tools/python/src/image.cpp
```diff
@@ -32,44 +37,32 @@ boost::python::list get_jitter_images(object img, size_t num_jitters = 1, bool d
...
-            }
-            img_as_list.append(row_list);
-        }
-
-        jitter_list.append(img_as_list);
+            dlib::disturb_colors(crop, rnd_jitter);
+
+        void* img_data_copy = malloc(img_mat.nr() * width_step(crop));
+        memcpy(img_data_copy, image_data(crop), img_mat.nr() * width_step(crop));
+        PyObject *pyObj = PyArray_SimpleNewFromData(3, dims, NPY_UINT8, img_data_copy);
```
davisking: Now you have a memory leak.

davisking:
Maybe adding a numpy dependency is fine.  It's not like you get very far in any reasonable python use without numpy anyway.  I have to think about it.

tools/python/src/image.cpp
visionworkz: From my understanding:
```
PyArray_ENABLEFLAGS((PyArrayObject*)pyObj, NPY_ARRAY_OWNDATA);
```
Indicated the PyArrayObject owns the data. I will reconfirm.

davisking: Hmm, maybe that's the case.  However, it seems like you should use PyArray_SimpleNew instead.

visionworkz: As per your suggestion, I am now directly allocating memory via `PyArray_SimpleNew`

The function is below:
```
boost::python::list get_jitter_images(object img, size_t num_jitters = 1, bool disturb_colors = false)
{
    if (!is_rgb_python_image(img))
        throw dlib::error("Unsupported image type, must be RGB image.");

    // Convert the image to matrix<rgb_pixel> for processing
    matrix<rgb_pixel> img_mat;
    assign_image(img_mat, numpy_rgb_image(img));

    // The top level list (containing 1 or more images) to return to python
    boost::python::list jitter_list;

    size_t rows = num_rows(img_mat);
    size_t cols = num_columns(img_mat);

    // Size of the numpy array
    npy_intp dims[3] = { num_rows(img_mat), num_columns(img_mat), 3};

    for (int i = 0; i < num_jitters; ++i) {
        // Get a jittered crop
        matrix<rgb_pixel> crop = dlib::jitter_image(img_mat, rnd_jitter);
        // If required disturb colors of the image
        if(disturb_colors)
            dlib::disturb_colors(crop, rnd_jitter);
        
        PyObject *arr = PyArray_SimpleNew(3, dims, NPY_UINT8);
        npy_uint8 *outdata = (npy_uint8 *) PyArray_DATA((PyArrayObject*) arr);

        size_t index = 0;
        for(int i=0; i<rows; i++) {
            for(int j=0; j<cols; j++) {
                rgb_pixel pixel = crop(i, j);
                outdata[index++] = pixel.red;
                outdata[index++] = pixel.green;
                outdata[index++] = pixel.blue;
            }
        }
                
        boost::python::handle<> handle(arr);
        // Append image to jittered image list
        jitter_list.append(object(handle));
    }
           
    return jitter_list;
}
```

However, do you see any problem with the previous malloc approach? `memcpy` should be faster than the `for` loop here(?)

davisking:
You can still use memcpy here, which yes, is probably faster.

visionworkz:
@davisking I earlier tried a `for` loop (which worked), but since we get a pointer to the buffer, I changed to `memcpy` instead. So now we get a list of numpy arrays directly returned by this function.

The only concern I can think of is if someone is using a really old (incompatible) version of Numpy and they want to upgrade dlib.

visionworkz:
@davisking I've made presence of Numpy during compilation optional. Also made sure that if Numpy is present, it can be v1.5.1 or later (which was released on Nov 18, 2010 - before the first release of dlib on Github)

visionworkz:
@davisking what are your thoughts here concerning inclusion of Numpy? I can accordingly change the PR.

davisking:
Sorry, been busy the last few days.  

I think the PR is good.  Making it disable and continue to build successfully if numpy's headers aren't available is a good way to do it.  I'll look it over more carefully and merge it soon.

davisking:
Thanks, look good :)

PR #1016, opened 2017-12-15
little corrections in global optimization

ernestotapiar:
Dear Davis,

this PR makes two little corrections to the code of the global optimization module.

Best regards,
Ernesto.

### DIFF
--- a/dlib/global_optimization/upper_bound_function.h
+++ b/dlib/global_optimization/upper_bound_function.h
@@ -141,8 +141,6 @@ namespace dlib
             const long dims = points[0].x.size();
 
             using sample_type = std::vector<std::pair<size_t,double>>;
-
-            using sample_type = std::vector<std::pair<size_t,double>>;
             using kernel_type = sparse_linear_kernel<sample_type>;
             std::vector<sample_type> x;
             std::vector<double> y;
--- a/dlib/global_optimization/upper_bound_function_abstract.h
+++ b/dlib/global_optimization/upper_bound_function_abstract.h
@@ -53,6 +53,7 @@ namespace dlib
                             double local_bound = p.y + sqrt(noise_terms[i] + trans(p.x-x)*M*(p.x-x))
                             min_ub = min(min_ub, local_bound)
                         }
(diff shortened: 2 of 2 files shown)


davisking:
Sweet, thanks for the PR.  I was just going over the documentation for all this stuff myself and trying to get it all fully fleshed out :)

Have you used this function yet for anything?  So far I've been very pleased with it in all my own testing.

ernestotapiar:
No, I haven't yet. I was just looking for algorithms for global optimization and discovered you recently introduced  one in dlib. I read the references you metion in the documentation and was curious  to know your variant of the algorithm by Malherbe et al. 

The application I have in mind should use a surrogate function beacuse the actual function to be globally optimized  is extremely expensive to compute.

davisking:
Cool.  You should give it a shot.  I've found that using a LIPO style method is much better than pure random search when the problem has many local optima.  For problems with only one or a small number of optima it's not super different from random search, but if there are a lot then it's way better.  Then when combined with a local trust region solver it works very well.

ernestotapiar:
Tank you for sharing this information on LIFO methods. One of the first references I found was the tutorial by Brochu et al. (https://arxiv.org/abs/1012.2599) who describe Bayesian optimization using Gaussian (Student-t) processes. These methods also assume Lipschitz-continuity on the objective function. 

The Bayesian framework seems to me very attractive due to its theroretical basis, but it suffers, in my personal opinion, from the choice of meta parameters and functions (covariance, aquisition) they require beforehand. The solution you offer overcome nicely these drawbacks. I just wonder how both methods could be teoretically related apart from Lipschitz-continuity.

ernestotapiar:
I totally agree with you. In some cases, the Bayesian approach would require conducting many experiments to find the most suited function space that models best the original function. This also makes no sense for functions expensive to compute.

More than a concrete idea, I had just thought it would be nice that your approach could be a limit (sort of deterministic) case of a Bayesian model.

I will use for sure your method in my application, however, not in the short term, because the availability of the needed data, unfortunately, doesn't depend only on me.

PR #1024, opened 2017-12-20
Make sure intel architecture is properly detected

pguyot:
On systems such as FreeBSD, `uname -p` returns amd64 (in lowercase) and `-DUSE_AVX_INSTRUCTIONS=1` does not work (the option is ignored).


davisking:

PR #1027, opened 2017-12-24
>2x performance increase if DLIB_DO_NOT_USE_SIMD is defined

andriy-gerasika:
avoid excessive copying of simd4f structs by making low() & high() methods of simd8f struct return "const simd4f&"

### DIFF
--- a/dlib/simd/simd8f.h
+++ b/dlib/simd/simd8f.h
@@ -125,8 +125,8 @@ namespace dlib
                 return _high[idx-4];
         }
 
-        inline simd4f low() const { return _low; }
-        inline simd4f high() const { return _high; }
+        inline const simd4f& low() const { return _low; }
+        inline const simd4f& high() const { return _high; }
 
     private:
         simd4f _low, _high;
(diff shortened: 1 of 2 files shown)


davisking:
Thanks for the PR.

Huh, that's odd.  What compiler and what compiler switches does this speedup appear under?

andriy-gerasika:
Ubuntu 16.04, gcc5, -DDLIB_DO_NOT_USE_SIMD 
I am trying to profile dlib using callgrind, it does not support sse/avx instructions, so I had to turn off sse/avx and noticed this stuff.

davisking:
Did you give -O3 or any other optimizations?  I'm asking because the change you made shouldn't effect the speed, unless you haven't enabled compiler optimizations.

andriy-gerasika:
yes, I gave -O3 and -ffast-math.
The change affects speed only if DLIB_DO_NOT_USE_SIMD is defined, which is not going to happen in production.

However, check screenshot of callgrind before the change:
http://www.gerixsoft.com/tmp/dlib-simd-perf0.png
and after the change:
http://www.gerixsoft.com/tmp/dlib-simd-perf1.png

i.e. the change makes profiling easier

davisking:
Yeah, that's surprising.  But cool, thanks for the PR :)

PR #1028, opened 2017-12-25
fix typo in dnn mmod train find cars example

ducthienbui97:
I found a typo in dnn mmod train find cars example:
```"is there a care here?"``` which should be ```"is there a car here?"```


davisking:
Sweet, thanks for the PR :)

Issue #293, opened 2016-10-14
Switch Python Binding to pybind11 Instead of Boost.Python


davisking:
I've recently found out about [pybind11](https://github.com/pybind/pybind11), which is essentially a header only reimplementation of Boost.Python with C++11, and without any dependency on boost.  It also has superior documentation compared to Boost.Python, and while it's API is very similar to Boost.Python it does seem a little cleaner.  

Over the last year, it seems like the vast majority of people complaining about difficulties building dlib have actually been complaining about difficulties installing/building boost.  So it would be great to get rid of Boost.Python and use pybind11.  It doesn't look like too much work since the pybind11 API is very similar to Boost.Python.

edmBernard:
I would be glad to help you for this. 
Have you begin this switch or we need to start from scratch ?

davisking:
That would be sweet.  Needing to install boost is the number 1 problem new dlib users have.  So this would be a very well received update :)

I've just looked at pybind11 but haven't done anything yet.

davisking:
@edmBernard Are you still working on this?  We still get almost daily questions from people having trouble installing boost.python (e.g. #473) so switching to pybind11 would be pretty sweet :)

edmBernard:
sorry I don't already got time to moving forward on this task but that's still on my project.

davisking:
Yeah I saw that was missing.  Guess it's not a simple find/replace.  Still totally worth it though :)

edmBernard:
yeah I thinks too.

edmBernard:
to help in the switch I find this lib that already switch : 
before : (with boost::python) : [here](https://bitbucket.org/glotzer/hoomd-blue/src/9714d3421c291abe9f333f28b663c2a34674b565/hoomd/md/MuellerPlatheFlow.cc?at=fix_getar_writejson_mpi&fileviewer=file-view-default)
after : (with partialy pybind11) : [here](https://bitbucket.org/glotzer/hoomd-blue/src/8bdadec6bd43a6b1076d69cee3744a953d858c7d/hoomd/md/MuellerPlatheFlow.cc?at=fix_getar_writejson_mpi&fileviewer=file-view-default)

edmBernard:
with the diff : [here](https://bitbucket.org/glotzer/hoomd-blue/diff/hoomd/md/MuellerPlatheFlow.cc?diff2=8bdadec6bd43&at=fix_getar_writejson_mpi)

edmBernard:
[...]

PR #1040, opened 2017-12-29
Replace boost::python with pybind11

supervacuus:
This is the initial pull request for #293 

* tested with python 3.6.3, gcc 7.2.1 and  pybind11 (more recent than 2.2)
* it depends on pybind11 to be found by find_package()
* all python examples run
* i've done a few review rounds to ease your review
* no detailed testing of sub-components has been done. i'll send you a message when the pr is ready for an in-depth review.
* if you @davisking have fly-over peek and find something disgusting, let me know ;-)


davisking:
Thanks for the PR.  This is going to be a popular change to dlib and I'll probably create v19.9 just to push this out once it's been merged and sufficiently tested.  I looked it over a bit and I don't see any obvious issues, so you seem to be on the right track.  Let me know when it's ready for a full review.

supervacuus:
* now also tested with python 2.7.14
* removed external find_package dependency and packaged pybind11 v2.3.dev0 in dlib/external
* as a consequence, i was able to get both python CI builds running
* further bug fixes

Rikorose:
Why is dlib not using git submodules? This would also make this PR clearer.

davisking:
Because I don't want to get confused posts from users.  I'm sure I would get lots of issues posted like "I pulled the code from github, but I get a pybind11 not found error!" since git clone doesn't automatically clone submodules.  You should look at the issues and questions I get on various dlib forums. There are a lot of people very new to programming.

davisking:
Although for the purpose of this PR we can use submodules to make it easier to review.  I would be fine with that.  I have a hard time believing sticking with submodules doesn't create user confusing though considering the level of skill I see in posters every day.

supervacuus:
i can also move the packaged pybind11 external folder into a separate commit, then you can view only the commit with the binding changes while the PR as a whole is still buildable without any changes.

davisking:
I'm fine either way.  We can give submodules a try or not.  It's easy enough already to ignore the pybind11 folder when reviewing the PR.

supervacuus:
I think we can start with the review and the PR testing round. I added around 50 test-cases for the basic data-structures and fixed a few minor issues. This covers the areas with the biggest changes.

I think bugs/regressions will mostly pop-up due to non-obvious semantic differences between pybind11 and boost::python like the use of [bind_vector](http://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#binding-stl-containers) vs vector_indexing_suite  as well as the use of [opaque types](http://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#making-opaque-types).

supervacuus:
sorry, premature comment-submit...

open issues, that should not conflict with a review or testing-round and could be added in a separate PR:

- the build currently combines the build-parameter initialization from pybind11 with your CMake macro add_python_module. We can try to merge this by going full pybind11 or setting a few parameters in from pybind11 from add_python_module. The latter seems like more maintenance in the long-term, but might still be the way to go. Anyway, i think we should do this together, or at least a short chat session would be helpful before i can work on this.
- i have not written a test for cca. Since there is no complexity in the binding code for cca, i think it would make sense to simply replicate the test you did for C++ as a pytest setup. Will do next week.
- i did not attempt to write tests for the use-case classes, because they would be integration tests anyway and the corresponding modules are unit-tested in C++ on detail-level that i cannot replicate with the current python interfaces anyway.
- i have not integrated the pytests in the build-process yet (i simply copy the shared library to the test-directory and invoke pytest there). I will work on this next week.

Late happy new year.

davisking:
Sounds good.  As for cmake, I would assume we can use pybind11 the way their documentation shows, e.g.

```
add_subdirectory(pybind11)
pybind11_add_module(example example.cpp)
```

Does that not work the way you would expect?  It seems like we can delete dlib's add_python_module file entirely.  Or is it not as straightforward as it seems?

supervacuus:
It is not that bad, but

- `pybind11_add_module` cannot create "dlib", because another target with the same name exists (which is the static dlib).  You bypassed this in `add_python_module` by suffixing the shared target with `_` but keeping the output name. I cannot do that with `pybind11_add_module` AFAIK. Maybe pybind11 has infra for this scenario, but i have not seen it yet.
- there are also a few things in the `add_python_module` file that you do besides defining the macro, that i haven't dived into enough and compared with the pybind11_add_module to have clear idea if we can just drop it.

Have a look at https://github.com/supervacuus/dlib/tree/merge_pybind_build this is my current state with module name `dlib_.so` and no dependency on `add_python_module`.

davisking:
That looks reasonable.  I would assume you could use
set_target_properties() on dlib_ to set the output name to not include the
_ just like add_python_module did.  Does that not work because of something
funny in pybind11_add_module()?

Also, we need something like the PYTHON3 option to
set PYBIND11_PYTHON_VERSION to the right version of python to build
against.  Or we could remove the PYTHON3 option in favor of some message to
set PYBIND11_PYTHON_VERSION.  For instance, when PYBIND11_PYTHON_VERSION
isn't set maybe it prints a big warning message saying that you should set
it to the version you want to use.  setup.py would have to be updated as
well.  In any case, I'm open to whatever is simplest and works reliably.

davidishere:
i just have the problem "Boost python library not found." when i install dlib in Debian Jessia, can i use the new code now ?

supervacuus:
@davisking regarding set_target_properties: yes, of course *facepalm*

I would leave the automatic assignment of `PYBIND11_PYTHON_VERSION` as it is defined in the following condition: https://github.com/supervacuus/dlib/blob/merge_pybind_build/tools/python/CMakeLists.txt#L9

Additionally i would check whether `PYBIND11_PYTHON_VERSION` has been defined already and only if not use the automatic assignment.

Since one main driver for doing all of this is helping people with the setup, i would not require setting a python version explicitly. This way, someone who knows what he is doing can override, while all others have sane defaults.

davisking:
Yeah, that's what I meant.  Let the default PYBIND11_PYTHON_VERSION do it's thing and only mess with it if the user specifically requests it.  But even that may be excessive.  It's really just these two things that should work:

- `python setup.py build` should always work and build against whatever python interpreter that ran setup.py
- Using cmake to build against any installed python interpreter should be easy.  Manually setting PYBIND11_PYTHON_VERSION via a -D switch is fine.  How to do that just needs to be documented somewhere people will see it. It's probably a good idea to make it a message in the cmake output like "PYBIND11_PYTHON_VERSION not set, using default python interpreter.  To change this do something like: -DPYBIND11_PYTHON_VERSION=3.4".  However, other than printing a message when it's not set we probably shouldn't do anything.

davisking:
@davidishere Yes, give it a try and let us know if you have any issues.

soundreactor:
i just tried building the merge branch with VS 2017 and something went wrong.
but i have no clue what. 
i tried with -T 140 and without

  P:\000_ML\nununu\dlib-merge_pybind_build\tools\python\src\dlib.cpp(41): error
C2059: syntax error: ';' [P:\000_ML\nununu\dlib-merge_pybind_build\tools\python\
build\dlib_.vcxproj]
    4 Warning(s)
    1 Error(s) 

in the cmake error log 
 fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory 

just yesterday i compiled the normal master branch with -T v140 no errors.
but when i tried import in python it had link errors. (i suspected boost)
my only hope is this pybind branch.

do i really have to install VS 2015? :(

supervacuus:
@soundreactor can you please use the version from the PR which is `supervacuus:pybind11`. My `merge_pybind_build` branch is experimental and used only to merge the build approaches. It is quite possible that some build-aspect is not working there (like in your case the DLIB_VERSION not being picked up correctly from the CMakeLists.txt). The version in the pybind11 branch definitely builds in VS2017

soundreactor:
@supervacuus ok i did just that and it compiled and installed but when i import it in python i get:

  File "<stdin>", line 1, in <module>
  File "C:\Users\KElian\AppData\Local\Programs\Python\Python36\lib\site-packages
\dlib-19.8.99-py3.6-win-amd64.egg\dlib\__init__.py", line 1, in <module>
    from .dlib import *
ImportError: DLL load failed: A dynamic link library (DLL) initialisation routin
e failed.

i tried using "Process Monitor " to understand what the problem is but i don't really know how to read it.
the pcmon recording:
https://docs.google.com/spreadsheets/d/1hM6SAVo5QSr1MfdpoK-kVfZ-wsqQ9jBtISWJJgFOH_Q/edit?usp=sharing

supervacuus:
@soundreactor i have tested this now on two different windows machines (one with 2.7 and one with 3.6), but cannot reproduce. This error is typically raised when there is CRT mismatch...

* how did you compile and install?
* what did you run?
* do you have more than one python version installed?
* are you sure that you have used that same python for building and installing as you did for executing?
* can you load the `dlib.pyd` in the `python_examples` folder (simply change to the directory, start python and `import dlib`)?

soundreactor:
@supervacuus  after renaming my 2 other python folders (2.7 and 3.5) it worked just fine with 3.6
thanks for the help !

supervacuus:
@soundreactor That is great news and valuable feedback for the PR. Let me know if any other issue arise from the current branch.

davisking:
@supervacuus Are you switching cmake to not use add_python_module?  If not then I'll try and do it later tonight.  I want to wait until that change is made before asking people to test the PR.  But once it's there I'll see if I can get a bunch of people to try it out.

supervacuus:
@davisking I'll try to get the `merge_pybind_build` branch to work during the next two to three hours. If i am succesful i will merge it to `pybind11` (no rebases anymore?). In any case, i'll send you an update.

davisking:
Sweet, sounds good.  Let me know when it's ready.

supervacuus:
@davisking I'm sorry but i could not invest as much time as i wished to. I committed and pushed my change into my `merge_pybind_build` branch. Basically I went through all the initialization you were doing before the `add_python_module` definition and checked whether `pybind11_add_module` did that already. If not (most cases), i copied it into the `CMakeLists.txt` of the python module.

I also left the `PYTHON3` option in the meantime, but only for the executable lookup. The selection-setup of python versions is left to the pybind11 build process. The executable lookup is something that probably also should be changed... if most people use `setup.py` as their build-process anyway then simply passing `sys.executable` to the cmake-build is much more robust than trying to guess this yourself in cmake. For instance if you create a virtualenv with python3 in order to setup dlib there, by default the python executable on the path will be `python` and not `python3`.

I tested it with both `setup.py` and individual cmake builds for python 2.7 and 3.6 under linux... successfully. Under windows i still get an empty `DLIB_VERSION` which is curious but did not have the time to investigate further. If you can wait another day, i will have a go at that.

davisking:
No problem.  Thanks for all your work on this :)

It's no rush, it can definitely wait until tomorrow.

davidishere:
just ‘python3 setup.py install‘ successfully, very fast and easy, thanks

supervacuus:
@davisking Okay, add_python_module is no more. Tested on linux (2.7.14, 3.6.3) and windows10 (VS2017, 3.6.4). There are a few things that could be further optimized in the python build, but i consider those orthogonal to this PR.

davisking:
Awesome.  I tried it in a few places and all worked without issue.

ageitgey:
I tried building this on macOS 10.12.6. I ran into the following error:

```
[ 90%] Building CXX object CMakeFiles/dlib_.dir/src/svm_struct.cpp.o
/Users/ageitgey/projects/dlib/tools/python/src/svm_struct.cpp:41:91: error: expected '(' for function-style cast or type construction
        psi = problem.attr("get_truth_joint_feature_vector")(idx).cast<feature_vector_type&>();
                                                                       ~~~~~~~~~~~~~~~~~~~^
/Users/ageitgey/projects/dlib/tools/python/src/svm_struct.cpp:41:92: error: expected expression
        psi = problem.attr("get_truth_joint_feature_vector")(idx).cast<feature_vector_type&>();
                                                                                           ^
/Users/ageitgey/projects/dlib/tools/python/src/svm_struct.cpp:41:94: error: expected expression
        psi = problem.attr("get_truth_joint_feature_vector")(idx).cast<feature_vector_type&>();
                                                                                             ^
3 errors generated.
make[2]: *** [CMakeFiles/dlib_.dir/src/svm_struct.cpp.o] Error 1
make[1]: *** [CMakeFiles/dlib_.dir/all] Error 2
make: *** [all] Error 2
```

I didn't have a ton of time tonight, so I didn't try to debug it yet. I just thought I'd share in case it was something obvious. I can take another look tomorrow.

The full build log is here: https://gist.github.com/ageitgey/1d3db1f85c821c91ac2db3f001d57769

The compiler is `Apple LLVM version 9.0.0 (clang-900.0.39.2)`.

darugar:
I'm getting the following error trying to compile with Python on OS X (c++ examples compile fine):

OS X 10.11.6
Python 3.6.4
git pull origin pull/1040/head

python setup.py install --yes USE_AVX_INSTRUCTIONS

[ 90%] Building CXX object CMakeFiles/dlib_.dir/src/svm_struct.cpp.o
dlib.git/tools/python/src/svm_struct.cpp:41:91: error: expected '(' for function-style cast or type construction
        psi = problem.attr("get_truth_joint_feature_vector")(idx).cast<feature_vector_type&>();
                                                                       ~~~~~~~~~~~~~~~~~~~^
dlib.git/tools/python/src/svm_struct.cpp:41:92: error: expected expression
        psi = problem.attr("get_truth_joint_feature_vector")(idx).cast<feature_vector_type&>();
                                                                                           ^
dlib.git/tools/python/src/svm_struct.cpp:41:94: error: expected expression
        psi = problem.attr("get_truth_joint_feature_vector")(idx).cast<feature_vector_type&>();

supervacuus:
Thanks for the report @darugar and @ageitgey . The cast is a type-dependent expression that can require the `template` keyword  before invoking the cast function (compilers obviously do disagree where it is needed). I pushed a fix.

ageitgey:
@supervacuus Thanks for the quick fix! 2af8c0e solved the compilation issue. I'll move on to testing the functionality now.

ageitgey:
I ran all the unit tests and examples from the face_recognition lib and everything worked perfectly. So surprisingly, I wasn't able to find any issues at all. However I didn't test every function of dlib, just those relating to face stuff.

Also, the computer I have available at the moment doesn't have an nvidia card, so I only tested non-CUDA builds on Mac (2.9 GHz Intel Core i7):

```
python3 setup.py install --yes USE_AVX_INSTRUCTIONS --no DLIB_USE_CUDA
```

I also wanted to verify that there weren't any major changes to performance, so I ran [a very simple benchmark](https://github.com/ageitgey/face_recognition/blob/master/examples/benchmark.py). Here are the results across dlib versions:

|                                    | dlib 19.7            | dlib 19.8.99 master  | dlib 19.8.99 no-boost |
|------------------------------------|----------------------|----------------------|-----------------------|
| Face locations (240p)              | 0.0554s (18.05 fps)  | 0.0560s (17.86 fps)  | 0.0553s (18.08 fps)   |
| Face landmarks (240p)              | 0.0022s (455.91 fps) | 0.0023s (436.39 fps) | 0.0024s (416.37 fps)  |
| Encode face inc. landmarks (240p)  | 0.0185s (54.19 fps)  | 0.0189s (52.86 fps)  | 0.0187s (53.38 fps)   |
| End-to-end (240p)                  | 0.0742s (13.47 fps)  | 0.0746s (13.40 fps)  | 0.0754s (13.27 fps)   |
| Face locations (480p)              | 0.2177s (4.59 fps)   | 0.2156s (4.64 fps)   | 0.2180s (4.59 fps)    |
| Face landmarks (480p)              | 0.0022s (453.43 fps) | 0.0023s (427.95 fps) | 0.0023s (441.14 fps)  |
| Encode face inc. landmarks (480p   | 0.0187s (53.56 fps)  | 0.0190s (52.52 fps)  | 0.0191s (52.26 fps)   |
| End-to-end (480p)                  | 0.2328s (4.30 fps)   | 0.2336s (4.28 fps)   | 0.2339s (4.28 fps)    |
| Face locations (720p)              | 0.4888s (2.05 fps)   | 0.4949s (2.02 fps)   | 0.5033s (1.99 fps     |
| Face landmarks (720p)              | 0.0022s (454.01 fps) | 0.0023s (436.43 fps) | 0.0023s (435.91 fps)  |
| Encode face inc. landmarks (720p)  | 0.0186s (53.67 fps)  | 0.0190s (52.66 fps)  | 0.0190s (52.52 fps)   |
| End-to-end (720p)                  | 0.5149s (1.94 fps)   | 0.5161s (1.94 fps)   | 0.5231s (1.91 fps)    |
| Face locations (1080p)             | 1.1388s (0.88 fps)   | 1.1428s (0.88 fps)   | 1.1406s (0.88 fps)    |
| Face landmarks (1080p)             | 0.0023s (436.72 fps) | 0.0025s (397.40 fps) | 0.0024s (414.53 fps)  |
| Encode face inc. landmarks (1080p) | 0.0199s (50.15 fps)  | 0.0196s (50.95 fps)  | 0.0200s (49.92 fps)   |
| End-to-end (1080p)                 | 1.1616s (0.86 fps)   | 1.1634s (0.86 fps)   | 1.1640s (0.86 fps)    |

So performance appears basically identical in this dumb benchmark. Any tiny fluctuations in numbers here is meaningless. I just wanted to make sure it wasn't suddenly 2x slower or anything and it seems fine. 👍

supervacuus:
@ageitgey thanks for testing! great to see that no major performance regression were found in your test-setup.

lextoumbourou:
Tested and working on the following:

   * Python 3.6.3 on Ubuntu 16.4
   * Python 3.6.1 on Amazon Linux AMI release 2017.03

This is a massive improvement!

***

Side note, I originally built Dlib with AVX instructions (`python3 setup.py install --yes USE_AVX_INSTRUCTIONS --no DLIB_USE_CUDA`) and got this error when I try to import:

```
In [1]: import dlib
Illegal instruction (core dumped)
```

I realised that my machine doesn't support AVX instructions and I was able to recompile with the correct flags, but the error message isn't very helpful. I wonder if Dlib could detect "bad" CPU extensions and display a helpful error message like what Tensorflow does?

```
In [1]:  import tensorflow
F tensorflow/core/platform/cpu_feature_guard.cc:35] The TensorFlow library was compiled to use AVX instructions, but these aren't available on your machine.
Aborted
```

davisking:
Thanks for testing this.

Yeah, that's a good idea about the error message.  I'll add that to dlib.

davisking:
@supervacuus, I'm good with merging this PR if you think it's ready.  Is there anything else you want to do before the merge?

supervacuus:
@davisking If the PR is ready for you, then it is ready for me. I have written down a few things that are not directly related with the switch to pybind11 that i would add in the future when i have more time if that is ok with you.

davisking:
Sounds good.  Thanks for the great PR and definitely submit more if you have the time :)

PR #1100, opened 2018-01-31
Add Python rvm_trainer and init functions

ksemb:
This PR exposes the `rvm_trainer` class to Python. Additionally, all trainers are provided with an `init()` function.

I am not sure if the `rvm_trainer<linear_kernel<sample_type> >` and `rvm_trainer<sparse_linear_kernel<sparse_vect> >` classes are correct, since there does not exist an rvm counterpart of the `svm_c_linear_trainer` class.

### DIFF
--- a/tools/python/src/svm_c_trainer.cpp
+++ b/tools/python/src/svm_c_trainer.cpp
@@ -72,56 +72,51 @@ template <typename trainer_type>
 double get_c_class2 ( const trainer_type& trainer) { return trainer.get_c_class2(); }
 
 template <typename trainer_type>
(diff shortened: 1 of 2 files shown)


davisking:
Thanks for the PR :)

Can you add some tests for these things to the python/tests folder?  You could do something as simple as using the training data from python_examples/svm_binary_classifier.py to test that it trains and gets the right answers.

ksemb:
I just added a test case.

davisking:
Nice, looks good. Thanks for the PR :)

PR #1135, opened 2018-02-15
Enable multi-process builds with MSVC, to massively speed up the build.

kohei101:
Also suppress warning about macro define redefinitions.

### DIFF
--- a/dlib/cmake_utils/set_compiler_specific_options.cmake
+++ b/dlib/cmake_utils/set_compiler_specific_options.cmake
@@ -126,6 +126,22 @@ if (MSVC)
       # Clang can compile all Dlib's code at Windows platform. Tested with Clang 5
       list(APPEND active_compile_opts "-Xclang -fcxx-exceptions")
    endif()
+
+   if(${CMAKE_VERSION} VERSION_GREATER 3.0.1)
+      # See if we can enable multi-process build.
+      cmake_host_system_information(RESULT num_logical_cores QUERY NUMBER_OF_LOGICAL_CORES)
+      cmake_host_system_information(RESULT available_ram_mb QUERY AVAILABLE_PHYSICAL_MEMORY)
+      math(EXPR required_ram_mb "${num_logical_cores} * 768")  # 768MB per process.
+      message(STATUS "Available physical memory: ${available_ram}MB")
+      message(STATUS "Logical core count: ${num_logical_cores}")
+      message(STATUS "Minimum memory required for multi-process build: ${required_ram_mb}MB")
+      if(${available_ram_mb} GREATER_EQUAL ${required_ram_mb})
+         message(STATUS "Enabling multi-process build")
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR, but I don't think dlib's cmake files should be forcing /MP on users. If a user wants to run a parallel build they can enable it for their project or not. Some people build on computers with a small amount of ram and /MP might cause their build to crash.

davisking:
Actually, you could use this https://cmake.org/cmake/help/v3.0/command/cmake_host_system_information.html to query the system to see if they have a reasonable amount of RAM. Like check if the amount of RAM available is 4gb * number of cores or greater or something like that and then enable the option if there seems to be enough RAM.  

Also, if you do this be sure to include appropriate checks against the cmake version since dlib supports cmake 2.8.12 or newer but this host info stuff might not have been available back then.  You should also set any compiler switches in here https://github.com/davisking/dlib/blob/master/dlib/cmake_utils/set_compiler_specific_options.cmake rather than editing CMAKE_CXX_FLAGS.

kohei101:
Hi Davis,

Thanks for your quick response.  Let me look into the issues you've raised.  I'll get back to you.

kohei101:
I've made updates to my previous change, to hopefully address your comments.  I need to note that, while you've suggested 4 GB per process as required available memory, I've lowered that to 768 MB per process.  The reason for that is, my build machine has a 4 physical core (w/ 8 logical cores) and 16 GB of system RAM, and normally I have 8 GB available when nothing else is running.  If we set the required RAM per process to be 4 GB, then it would not pass on my environment.  In my observation the dlib build process, with multi-process build enable, typically uses roughly 0.5 GB or less on my machine. So I believe it should be sufficient to set the per-process requirement to 768 MB.

If you still want the memory requirement set higher, please let me know, and I'll change it.

Thanks!

davisking:
This seems sensible.  However, your goal is to enable parallel builds only for dlib right?  I was initially thinking that /MP is set project wide (i.e. including not just dlib .cpp files but all files in the whole visual studio solution) and that is why the memory requirement would have to be set very large.  But if you want to do it only for dlib's files, which is probably the right thing to do here, then you need to do this a little differently.  This is because things in `active_compile_opts` are propagated to all dependent projects and will, for instance, cause a project that uses dlib to use /MP as well.

Can you change that code you added into a cmake function that adds a `PRIVATE` target compile option of /MP to a target. Then you can call that function in dlib/CMakeLists.txt to add it to the dlib target only without impacting anything other than dlib's own .cpp files.

kohei101:
Sorry for not replying sooner.  I just got caught up with other things, but I'd like to come back to it at some point and finish this up hopefully sooner rather than later.

kohei101:
Actually I don't need this piece anymore. Let's close it.

davisking:
I think it's basically covered by recent changes to dlib's cmake scripts.  Thanks for the PR in any case though :)

Issue #1137, opened 2018-02-16
CMakeLists.txt: find_package(dlib) fails on Fedora/Red Hat


kimwalisch:
Hi,

When installing ```dlib-devel``` using the package manager of Fedora or Red Hat ```find_package(dlib)``` fails when I try to use dlib in my ```CMakeLists.txt```:

```CMake
cmake_minimum_required(VERSION 3.1)
find_package(dlib REQUIRED)

add_executable(example example.cpp)
target_link_libraries(example dlib::dlib_shared)
```

Here is the error message:

```
cmake .
CMake Error at /usr/local/lib/cmake/dlib/dlib.cmake:85 (message):
  The imported target "dlib::dlib" references the file

     "/usr/local/lib/libdlib.a"

  but this file does not exist.  Possible reasons include...
```

The error is caused by a Fedora/Red Hat policy that static libraries must be put into its own package (e.g. ```dlib-static```), hence ```dlib-devel``` contains only the shared dlib library. The ```dlib-devel``` package also installs the CMake config files which export the ```dlib::dlib``` target for the static dlib library which is missing in the ```dlib-devel``` package. Because of this ```find_package(dlib)``` fails. 

In the ```dlib.spec``` file which is used to build the official dlib RPM packages the static dlib library is simply deleted after the build has finished:

```spec
%install
pushd build
%make_install
popd
rm -f %{buildroot}/%{_libdir}/*.a
```

There is no easy workaround I know that could be used in the ```dlib.spec``` to fix the official dlib RPM packages (The only workaround I know is to include dlib.a in the ```dlib-devel``` package which would fix the issue but this is not allowed by Fedora/Red Hat).

davisking:
I wasn't involved or even knew about the creation of the dlib RPM package.
The package maintainers need to test their packages before they publish
them to make sure they haven't somehow broken the package. I don't know
what you want me to do about this. You should complain to the package
maintainer. Maybe there is some change to the dlib CMake files that might
make the package maintainer's life easier, but we have to begin from a
point where the maintainer is at least testing their packages before
pushing them out.

davisking:
[...]

Issue #923, opened 2017-11-06
Simplify shared/static linking


SoapGentoo:
Hi @davisking 
one problem in distributions we face is the fact that CMake wasn't designed to concurrently handle building shared and static libraries without adding two targets etc. This leads to every CMake-based buildsystem having a different way of handling enabling/disabling static and/or shared libraries. I would like to recommend the following plan, which is what Kitware recommend:

Only define one library. The type of library can be switched by the native CMake variable `BUILD_SHARED_LIBS`. I recommend building static libraries by default (as most users who manually build dlib will prefer this). If users want shared libraries, they just pass `-DBUILD_SHARED_LIBS=ON`. This makes building dlib a lot easier for us, as for instance when building the test suite we don't build the static libs which we throw away in the end. Furthermore, because `BUILD_SHARED_LIBS` is a standard CMake library, it integrates well with how distributions can globally only build shared libraries. Furthermore, this would simplify the build system a lot for you too by removing lots of redundant definitions and conditional code.

Would you be willing to accept a PR for this?

davisking:
What about when someone installs dlib on a unix system by?

```
mkdir build
cd build
cmake ..
make install
```

That's now only going to install a static library, or a shared library depending on the switches.  

Also, what do you mean by "This makes building dlib a lot easier for us, as for instance when building the test suite we don't build the static libs which we throw away in the end".  What is made easier?  As an aside, I want the tests to build dlib in multiple modes to make sure the build works.  They are tests after all.

SoapGentoo:
@davisking correct. The proper CMake way is to configure it twice out of tree, and install that twice. In general for users this doesn't matter, as people who compile dlib don't usually want both (and in practice prefer static libs, as that doesn't require any `RPATH` or `LD_LIBRARY_PATH` mucking). The current solution (producing the shared out of the static library) has another disadvantage: it requires the same compiling the same source code twice (once with `-fPIC` and once without `-fPIC`), which given that as a distro we generally only install the shared library, as a waste of CPU cycles.

I don't see how the current system tests dlib in different modes? Currently, all tests link to the static library. In fact, with my suggestion, the shared library is also tested.

davisking:
Yes, that's true.  I mean it tests the build. 

Anyway, if you want to submit a PR that makes this change I'm fine with it so long as it doesn't impact existing users.

davisking:
What happens when you configure it two times, and get two different dlib/config.h files?

xsacha:
[...]

PR #1138, opened 2018-02-16
Only define one dlib target in CMake and shared vs. static is determined by BUILD_SHARED_LIBS variable.

davisking:
This PR removes the dlib-shared CMake target and also causes the dlib CMake target to be either a shared or static library depending on the state of CMake's BUILD_SHARED_LIBS variable. Currently, the dlib target is always a static library.

There has been some discussion about this already: https://github.com/davisking/dlib/issues/923, @SoapGentoo, @xsacha, https://github.com/davisking/dlib/issues/1137, @kimwalisch.

This PR addresses some of the issues.  However, it isn't perfect. Here is one problematic case: Someone builds their own library via cmake and uses `add_subdirectory(dlib)` to add dlib as a dependency. They then set `BUILD_SHARED_LIBS=1` so that their library (and now dlib as well) are built as shared libraries. They then do `make install`, but since dlib's cmake files don't have an install target when invoked from a parent project via `add_subdirectory(dlib)` this user's install will be broken since it will be missing the dlib shared library files.

The obvious fix is to make dlib's CMake files always define install targets.  However, I am certain that would confuse and/or irritate many users who use `add_subdirectory(dlib)`. For example, people who don't set `BUILD_SHARED_LIBS=1` have no need to have any dlib files installed anywhere then they run `make install` on their own projects. To use a specific example, the dlib/tools/imglab program depends on dlib via `add_subdirectory(dlib)` and it would be inappropriate for `make install` of imglab to try to install dlib as well. It should install only imglab. In addition to this being just weird and irritating for experienced C++ users, I'm very confident I would get lots of questions from inexperienced users who are simply confused and assume they are supposed to do something with the static dlib library files and headers that cmake would be trying to install in this situation. 

Another option is to cause any project that depends on dlib via `add_subdirectory(dlib)` to build dlib as static library regardless of the state of BUILD_SHARED_LIBS. I think this is the best option.  However, it has the downside that a top level cmake file that merely aggregates many cmake projects together so that they can all be `make installed` with one command will not work as expected. There doesn't seem to be a way in cmake to distinguish between the two cases (1) someone using `add_subdirectory(dlib)` to get the `dlib::dlib` target to link to vs (2) someone using `add_subdirectory(dlib)` for meere cmake project aggregation. 

I'm not sure we care about project aggregation. Is there a reason to care about this case?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,14 +2,35 @@ cmake_minimum_required(VERSION 2.8.12)
 
 
 
+
+#############################################################################
+#                                                                           #
+#  READ examples/CMakeLists.txt TO SEE HOW TO USE DLIB FROM C++ WITH CMAKE  #
+#                                                                           #
+#############################################################################
+
+
+
+
+
 get_directory_property(has_parent PARENT_DIRECTORY)
 if(NOT has_parent)
-   # Set this so that the dlib subfolder will build both a static and shared
-   # library, since the only reason to build this CMakeLists.txt by itself is
-   # if you want to install dlib.  It should be noted however that installing
-   # dlib is not necessary.  A simpler approach is to use it the way shown in
-   # examples/CMakeLists.txt 
+   # When you call add_subdirectory(dlib) from a parent CMake project dlib's
+   # CMake scripts will assume you want to statically compile dlib into
+   # whatever you are building rather than create a standalone copy of dlib.
+   # This means CMake will build dlib as a static library, disable dlib's
+   # install targets so they don't clutter your project, and adjust a few other
+   # minor things that are convenient when statically building dlib as part of
+   # your own projects.
+   #
+   # On the other hand, if there is no parent CMake project or if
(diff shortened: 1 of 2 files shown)


xsacha:
I care about this case primarily as I try to integrate this package in to Hunter (C++ package manager) and every other package follows this convention. The main changes I have to maintain are install targets and single static/shared target.

davisking:
@xsacha and @SoapGentoo, can you try out the code in this PR and see if it behaves in a way that works for you? There is just the dlib target now and it builds based on BUILD_SHARED_LIBS, unless dlib is being built by calling `add_subdirectory(dlib)`, then it will build statically (with -fPIC if BUILD_SHARED_LIBS==true).  This behavior can be overridden by setting `DLIB_IN_PROJECT_BUILD==false` in addition to setting `BUILD_SHARED_LIBS`.

xsacha:
Thanks. Changes look good.

PR #1141, opened 2018-02-19
Added support for variadic Python functions in find_max_global.

mortehu:
Example calls that are made to work with this PR:

* `find_max_global(lambda *args: sum(args), [0, 0, 0], [1, 1, 1], 10)`
* `find_max_global(lambda a, b, *args: a+b+sum(args), [0, 0, 0], [1, 1, 1], 10)`

This call is still disallowed, because the function being optimized requires too many arguments:

* ~~`find_max_global(lambda a, b, c, d, *args: a+b+c+d+sum(args), [0, 0, 0], [1, 1, 1], 10)`~~

### DIFF
--- a/tools/python/src/global_optimization.cpp
+++ b/tools/python/src/global_optimization.cpp
@@ -45,17 +45,18 @@ py::list mat_to_list (
     return l;
 }
 
-size_t num_function_arguments(py::object f)
+size_t num_function_arguments(py::object f, size_t expected_num)
 {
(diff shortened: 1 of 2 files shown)


davisking:
Awesome, thanks for the PR :)

Can you add some tests to tools/python/test that make sure this works as expected: your example works, and errors are still thrown for giving functions with the wrong number of arguments.

davisking:
Great, thanks for the nice PR :)

PR #1148, opened 2018-02-23
Problem: integer overflow when calculating sizes (may happen e.g. with very large images)

reunanen:
Consider an innocent-looking 2D-array whose size is, say, 70,000 * 70,000. When using `unsigned long` all the way, there is an overflow in calculating the size. Similar for tensors, although there it may happen even more easier, if `n > 1` and/or `k > 1`.

Solution: change some types from `unsigned long` to `size_t`.

Note: certainly didn't think everything through yet – or even test almost anything (only my own case where overflow actually did happen). Really just opening the PR to start a discussion as to whether this kind of a direction makes sense. (Personally I think it's more future-proof, but I guess there may also be arguments against this approach.)

### DIFF
--- a/dlib/array/array_kernel.h
+++ b/dlib/array/array_kernel.h
@@ -245,11 +245,12 @@ namespace dlib
     {
         try
         {
-            // This is probably the least likely value for a max_size from an old version.
+            // This is probably the least likely value for a max_size from an old version,
+            // from among numbers that are guaranteed to be representable using unsigned long.
             // TODO: start version numbering from 1, when it can be reasonably expected that
             //       there aren't too many objects around any longer serialized without any
             //       version number (but directly a max_size).
-            const unsigned long version = std::numeric_limits<unsigned long>::max();
+            const unsigned long version = 4294967295;
 
             serialize(version,out);
 
@@ -279,7 +280,7 @@ namespace dlib
             unsigned long version;
             deserialize(version, in);
 
-            if (version != std::numeric_limits<unsigned long>::max()) {
+            if (version != 4294967295) {
                 // old version - the "version" field is actually max_size, serialized as unsigned long
                 //             - also size is serialized as unsigned long
                 // TODO: remove this branch completely, when backward compatibility is no longer needed
--- a/dlib/array/array_kernel.h
+++ b/dlib/array/array_kernel.h
@@ -245,15 +245,6 @@ namespace dlib
     {
         try
         {
-            // This is probably the least likely value for a max_size from an old version,
-            // from among numbers that are guaranteed to be representable using unsigned long.
-            // TODO: start version numbering from 1, when it can be reasonably expected that
-            //       there aren't too many objects around any longer serialized without any
-            //       version number (but directly a max_size).
-            const unsigned long version = 4294967295;
-
-            serialize(version,out);
-
             serialize(item.max_size(),out);
             serialize(item.size(),out);
 
@@ -277,29 +268,13 @@ namespace dlib
     {
         try
         {
-            unsigned long version;
-            deserialize(version, in);
-
-            if (version != 4294967295) {
-                // old version - the "version" field is actually max_size, serialized as unsigned long
-                //             - also size is serialized as unsigned long
-                // TODO: remove this branch completely, when backward compatibility is no longer needed
-                unsigned long max_size = version, size;
-                deserialize(size, in);
-                item.set_max_size(max_size);
-                item.set_size(size);
-                for (unsigned long i = 0; i < size; ++i)
-                    deserialize(item[i], in);
-            }
-            else {
-                size_t max_size, size;
-                deserialize(max_size, in);
-                deserialize(size, in);
-                item.set_max_size(max_size);
-                item.set_size(size);
-                for (size_t i = 0; i < size; ++i)
-                    deserialize(item[i], in);
-            }
+            size_t max_size, size;
+            deserialize(max_size,in);
+            deserialize(size,in);
+            item.set_max_size(max_size);
+            item.set_size(size);
+            for (size_t i = 0; i < size; ++i)
+                deserialize(item[i],in);
         }
         catch (serialization_error e)
         { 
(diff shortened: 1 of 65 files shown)


davisking:
This is probably a good idea on the whole, but some things aren't right in the details.

dlib/array/array_kernel.h
```diff
@@ -245,10 +245,18 @@ namespace dlib
     {
         try
         {
+            // This is probably the least likely value for a max_size from an old version.
+            // TODO: start version numbering from 1, when it can be reasonably expected that
+            //       there aren't too many objects around any longer serialized without any
+            //       version number (but directly a max_size).
+            const unsigned long version = std::numeric_limits<unsigned long>::max();
+
+            serialize(version,out);
```
davisking: This breaks backwards compatibility.  Why do you think you need to change the serialization code?

dlib/array/array_kernel.h
```diff
@@ -268,13 +276,29 @@ namespace dlib
...
-            deserialize(max_size,in);
-            deserialize(size,in);
-            item.set_max_size(max_size);
-            item.set_size(size);
-            for (unsigned long i = 0; i < size; ++i)
-                deserialize(item[i],in);
+            unsigned long version;
+            deserialize(version, in);
+
+            if (version != std::numeric_limits<unsigned long>::max()) {
```
davisking: The max unsigned long value isn't a specific number.  It could be any number of possible values depending on how unsigned long is represented.

More to the point here though, I don't see why you need to edit the serialization code at all.

dlib/dnn/tensor.h
```diff
@@ -205,7 +205,7 @@ namespace dlib
         long m_k;
         long m_nr;
         long m_nc;
-        long m_size; // always equal to m_n*m_k*m_nr*m_nc
```
davisking: Changing anything from signed to unsigned is going to create hard to find bugs.  It would be better to use long long here.

dlib/array/array_kernel.h
reunanen: For backward compatibiity, I thought. At least when deserializing data from an old version to a new version. But if it's not needed, then that's even better of course. Didn't really have a look at how dlib's serialization works under the hood – if it readily handles the different types, then that's really great.

dlib/array/array_kernel.h
reunanen: Good point. Changed to 4294967295, but then removed the whole serialization change.

dlib/dnn/tensor.h
reunanen: Fixed.

davisking: You mean to use long long right?

davisking: It's probably fine to just change them all to long long. Probably even preferable since there are many places where people will multiply these things together.

reunanen: Changed all to long long.

faboules:
Thank you for show fixed

reunanen:
The tests now pass on the build servers, but not locally on my laptop: `test_timer` fails non-deterministically, although should be pretty much unrelated?

That said, I'm sure there are still places in documentation where `unsigned long` should be changed to `size_t`, and also in individual classes some functionality that will fail for sizes larger than 2^32. But that damage should largely be limited to each individual class (and its users), I think.

reunanen:
Yeah, `test_timer` fails the same way in `master` too.

davisking:
When was test_timer failing? All the failed tests I saw were compiler errors.

reunanen:
It failed only locally, on my laptop.

davisking:
Was it under heavy load?  There are tests that check if the timer waits for the appropriate amount of time and the only time I've had problems with them is when the system doing the testing was under enough load so as to cause large execution delays.

reunanen:
No, it wasn't under heavy load. But in fact I was using the laptop over a Remote Desktop connection from a desktop. Maybe that affects somehow. Can try without the RDP connection.

Anyway, I think this issue is not related to the PR, especially when this test failed the same way also in the master branch.

davisking:
Right, it's definitely unrated.  I was just curious.

davisking:
Anyway, can you update the abstract.h files as well? I looked over the PR and there are a bunch of cases where they are now inconsistent with the code.  

There are also a few bits that deal with returning `dlib::matrix` or `matrix_exp` objects, but to change those really we would need to change the `matrix_exp` and related code to use `long long`. That should probably be done in its own PR since that's a bigger thing.

Alternatively, you might want to make a smaller PR that impacts only the tensor and related code that caused you the trouble.

reunanen:
Interestingly, eliminating RDP from the equation did help!

Yes, I know that the *abstract.h files need to be changed also. Will look into it at some point.

And yes, I started from the tensor code... But then there was some dependency that forced me to dig a little deeper, I thought. But I will double-check if that was really necessary or not.

reunanen:
In fact it seems that many of the `*_abstract.h` files do not mention the `size()` function even currently... so not sure what to do.

dlib/dnn/tensor_abstract.h
```diff
@@ -467,7 +467,7 @@ namespace dlib
         !*/
 
         explicit resizable_tensor(
-            long long n_, long long g k_ = 1, long long nr_ = 1, long long nc_ = 1
+            long long n_, long long k_ = 1, long long nr_ = 1, long long nc_ = 1
```
reunanen: I guess regardless of what is done to #1148, at least this change should go to `master`. (Sorry, my bad...)

davisking:
A lot of the abstract.h files don't mention size because those objects inherit from enumerable which is what defines their size().

reunanen:
Yeah, that's what I thought... But it also means that I didn't find a lot of abstract.h files that would now be inconsistent.

davisking:
Cool.  Let me know when you think the PR is ready.

reunanen:
Basically, now. There is still [this error](https://travis-ci.org/davisking/dlib/jobs/346296797), but it looks unrelated to me.

Not saying that there isn't a lot of other places that could/should be changed at the same time, though.

davisking:
Cool, looks good. Thanks for another good PR :)

I used to purposefully avoid using size_t because there were old compilers that would give an error like 'what is this size_t type?'. Standards support is way better than it used to be :)

PR #1151, opened 2018-02-25
Change types of tensor's size-related members to prevent integer overflow

reunanen:
Like #1148, but for `tensor` only.

### DIFF
--- a/dlib/dnn/tensor.h
+++ b/dlib/dnn/tensor.h
@@ -387,7 +387,7 @@ namespace dlib
             m_nr = nr_;
             m_nc = nc_;
             m_size = n_*k_*nr_*nc_;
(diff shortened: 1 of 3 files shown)


dlib/dnn/tensor.h
```diff
@@ -377,7 +377,7 @@ namespace dlib
         }
 
         void set_size(
-            long n_, long k_ = 1, long nr_ = 1, long nc_ = 1
+            long long n_, long long k_ = 1, long long nr_ = 1, long long nc_ = 1
```
davisking: You missed a (long) cast a few lines down.

reunanen: Oops, fixed.

davisking:
Sweet, looks good. Thanks for another PR :)

PR #1187, opened 2018-03-09
Clear flags

jjaloszy2:
Set istream flags to 0 in old_deserialize_floating_point.  This fixes [[nodiscard]] warning.

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -576,7 +576,7 @@ namespace dlib
     )
     {
         std::ios::fmtflags oldflags = in.flags();  
-        in.flags(); 
+        in.flags(0); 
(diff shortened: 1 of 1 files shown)


davisking:
This seems right.  Did you close it because there was some problem?

jjaloszy2:
The CI server gave a build error for clang.   It compiled fine with the MS compiler.  I guessed 0 wasn’t guaranteed to be implicitly convertible to fmtflags.  But looking into it more, it looks like a 0 can be used to initialize BitmaskType.
________________________________
LECO Corporation Notice: This communication may contain confidential information intended for the named recipient(s) only. If you received this by mistake, please destroy it and notify us of the error. Thank you.

PR #1189, opened 2018-03-09
Allow to set arbitrary cmake opts from setup.py

### LINKED ISSUE #1188: Not possible to pass CUDA_HOST_COMPILER when compile Python extension
## Expected Behavior

`setup.py` should allow to set options like `-DCUDA_HOST_COMPILER` or `-DCUDA_NVCC_FLAGS` for cmake in order to properly compile CUDA code in case if e.g. `CMAKE_C_COMPILER` != `CUDA_HOST_COMPILER`.

## Current Behavior

Currently only `--yes` and `--no` type of options are supported.

---

* **Version**: 19.9
* **Where did you get dlib**: github repo
* **Platform**: Gentoo x86-64
* **Compiler**: gcc 7.3.0 / gcc 6.4.0 for CUDA

Kagami:
Fixes #1188

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,6 +26,7 @@ Additional options:
     --clean: delete any previous build folders and rebuild.  You should do this if you change any build options
              by setting --compiler-flags or --yes or --no since last time you ran a build to make sure the changes
              take effect.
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the PR :)

PR #1198, opened 2018-03-18
Add python example for concurrent global optimization

tsoernes:
I'm not sure how to use `solver_epsilon` when using the global optimizer this way.
Also, it should be noted that an iteration last for as long as the longest of the concurrent simulation runs

### DIFF
--- a/python_examples/concurrent_global_optimization.py
+++ b/python_examples/concurrent_global_optimization.py
@@ -1,32 +1,42 @@
 import logging
 from functools import partial
-from multiprocessing import Pool
+from multiprocessing import Process, Queue
+from time import sleep
 
 import dlib
+import numpy as np
 
 logger = logging.getLogger('')
 
 
 class Simulation:
     def __init__(self, alpha, beta, epsilon):
-        pass
+        self.alpha = alpha
+        self.epsilon = epsilon
 
     def simulate(self):
-        return 1
+        # Run simulation, return result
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.

Example programs should be tutorials with some kind of narrative structure.  So there should be a comment explaining what the example teaches, then someone should be able to read the example top to bottom and learn something. This means comments that explain what it's doing along the way and why it's doing it.  

I'm not sure what you mean about `solver_epsilon`. I don't see it being referenced in the example.

Also, an example like this needs to teach best practices. I'm not familiar with python's threading tools, but certainly it should not process the function evaluations in waves that all wait for the slowest evaluation. That doesn't take advantage of the solver very well.

tsoernes:
Added independent optimizations. I can't do any guide for at least a couple of months when I finish my thesis.

tsoernes:
So `dlib.find_min_global` takes a `solver_epsilon` parameter. As far as I can tell neither `function_spec` nor `global_function_search` allows it. I don't know if its possible to use it when using the two latter functions instead of the first.

davisking:
All the member functions of `global_function_search` are exposed. So you can call `.set_solver_epsilon()`.

tsoernes:
It ended up as its own module instead https://github.com/tsoernes/gfsopt , so I won't finish this PR

davisking:
Sweet :)

PR #1212, opened 2018-03-28
Downsample layer

davemers0160:
This is my first attempt at trying to do a pull request.  Hopefully I got it right.  I added the downsample_ code to the layers.h and the description to the layers_abstract.h file.  I also added a test to the dnn.cpp file under dlib/test/dnn.cpp.  The test ran successfully, but not sure if I implemented it correctly.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -741,6 +741,112 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    class downsample_
+    {
+    public:
+
+        explicit downsample_(
+            double scale_x_ = 0.5,
+            double scale_y_ = 0.5
+        ) : scale_x(scale_x_), scale_y(scale_y_)
+        {
+            DLIB_CASSERT(0 < scale_x && scale_x < 1, "downsampling scale_x should be between 0.0 and 1.0");
+            DLIB_CASSERT(0 < scale_y && scale_y < 1, "downsampling scale_y should be between 0.0 and 1.0");
+      
+        }
+
+        double get_scale_x() const { return scale_x; }
+        double get_scale_y() const { return scale_y; }
+        
+        template <typename SUBNET>
+        void setup (const SUBNET& /*sub*/)
+        {
+        }
+    
+        template <typename SUBNET>
+        void forward(const SUBNET& sub, resizable_tensor& output)
+        {
+            output.set_size(
+                sub.get_output().num_samples(),
+                sub.get_output().k(),
+                scale_y*sub.get_output().nr(),
+                scale_x*sub.get_output().nc());
+            tt::resize_bilinear(output, sub.get_output());
+        } 
+        
+        template <typename SUBNET>
+        void backward(const tensor& gradient_input, SUBNET& sub, tensor& /*params_grad*/)
+        {
+            tt::resize_bilinear_gradient(sub.get_gradient_input(), gradient_input);
+        }
+        
+        inline dpoint map_input_to_output (dpoint p) const 
+        { 
+            p.x() = p.x()/scale_x;
+            p.y() = p.y()/scale_y;
+            return p; 
+        }
+
+        inline dpoint map_output_to_input (dpoint p) const 
+        { 
+            p.x() = p.x()*scale_x;
+            p.y() = p.y()*scale_y;
+            return p; 
+        }
+        
+        const tensor& get_layer_params() const { return params; }
+        tensor& get_layer_params() { return params; }
+        
+        friend void serialize(const downsample_& item, std::ostream& out)
+        {
+            serialize("downsample_", out);
+            serialize(item.scale_x, out);
+            serialize(item.scale_y, out);
+        }
+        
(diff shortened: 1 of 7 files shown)


davisking:
Cool, thanks for the PR :)

I think this would be more useful if the layer resized a tensor to a target shape rather than downsampling by a factor.  I realize that ultimately both layer types are equivalent, but my impression is that being able to say `resize_to<340,280,subnet>` is more convenient for the sorts of uses this kind of layer helps with.  Or is that not a good understanding of the target uses?

davemers0160:
Are you thinking about replacing both the upsample and downsample with a single layer than could do both?  I think that you are right on the input.  I can look at reworking the inputs and having the scale calculated inside the layer instantiation.

davisking:
You don't need a scale if you have at a target size.

davemers0160:
OK, I see where I can directly put in the new (smaller) nr, nc.  What do you recommend for a check of the inputs to make sure that they don't exceed the input tensor nr & nc values.  I can put a DLIB_CASSERT in the setup routine or is there a better place to put it?

davisking:
You can put an assert in the setup(). However, I don't understand why you
would do that.  The point of the layer is to compel the output to be a
certain size right? Let the layer make the output whatever the uses says.
If it's bigger it's bigger if it's smaller it's smaller.  Why would it be a
problem if the output is bigger?

davemers0160:
I was thinking that if the template inputs were bigger than the subnet inputs you would be getting a fractional upsample.  Also I'm not sure but I think that I would have to add some conditional checks in the two input/output mapping functions to correctly map up in scale or down in scale (I'd have to think about the best way to handle that).  If I don't limit the input size is resize_to_ (resize_to) taken anywhere else in dlib?

davisking:
I'm not sure I follow. We are talking about making a layer that takes an
input tensor and bilinearly resizes it so that it has a user specified
number of rows and columns right?

davemers0160:
You're right, I was more stuck on the downsample side.  With the change to allowing the user to specify the output tensor size (nr,nc) which could be larger than the input tensor, calling it a downsample layer is a little misleading.  I like the resize_to name, but I'll defer to you on that.

davisking:
Cool.  resize_to is good I think.

davemers0160:
OK, I think that I've got al the changes made.

dlib/dnn/layers.h
```diff
@@ -739,6 +739,119 @@ namespace dlib
...
+        static_assert(NC_ >= 1, "NC resize parameter can't be less than 1.");
+        
+        resize_to_()
+        {
+        }
+        
+        template <typename SUBNET>
+        void setup (const SUBNET& sub)
+        {
+            scale_x = (double)NC_/(double)sub.get_output().nc();
```
davisking: This should go in the forward() function since the size of the network output could change from one invocation to the next.  Some layers require the input size to be constant, like the fc_ layer and they make more heavy use of setup() as a result.  But layers that make sense for tensors of varying sizes should support that.  So here that means these scale parameters should be calculated each time at the start of forward().

dlib/dnn/layers_abstract.h
```diff
@@ -1224,6 +1224,64 @@ namespace dlib
...
+
+            WHAT THIS OBJECT REPRESENTS
+                This is an implementation of the EXAMPLE_COMPUTATIONAL_LAYER_ interface
+                defined above.  In particular, it allows you to resize a layer using
+                bilinear interpolation.  To be very specific, it resizes each of the
+                channels in an input tensor.  Therefore, if IN is the input tensor to this
+                layer and OUT the output tensor, then we will have:
+                    - OUT.num_samples() == IN.num_samples()
+                    - OUT.k()  == IN.k() 
+                    - OUT.nr() == IN.nr()*scale_y
```
davisking: Fix documentation.

davemers0160:
I've made the requested changes.  Let me know if I missed anything.  Thanks,
Dave

dlib/dnn/layers_abstract.h
```diff
@@ -1224,6 +1224,64 @@ namespace dlib
...
+                This is an implementation of the EXAMPLE_COMPUTATIONAL_LAYER_ interface
+                defined above.  In particular, it allows you to resize a layer using
+                bilinear interpolation.  To be very specific, it resizes each of the
+                channels in an input tensor.  Therefore, if IN is the input tensor to this
+                layer and OUT the output tensor, then we will have:
+                    - OUT.num_samples() == IN.num_samples()
+                    - OUT.k()  == IN.k() 
+                    - OUT.nr() == NR_
+                    - OUT.nc() == NC_
+                    - scale_y == NR_/IN.nr()
```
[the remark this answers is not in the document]
davisking: Remove them.  They don't make sense in context.  Nothing else in the documentation here mentions them so it's not clear what they mean or refer to.  Someone should be able to read the documentation and understand it all on its own.

davisking:

davemers0160:
Thanks, This kind of a fun process.  Thanks for the patience

davisking:
No problem.  Thanks for the PR.

PR #1216, opened 2018-03-30
prevent C++ compilation flags leaking to CUDA

jasjuang:
CMakeLists.txt

```
cmake_minimum_required(VERSION 3.10)

project(Example LANGUAGES CXX CUDA)

set(PROJECT_SRCS ${PROJECT_SOURCE_DIR}/main.cu)

find_package( dlib REQUIRED )

add_executable(${PROJECT_NAME} ${PROJECT_SRCS})

target_link_libraries(${PROJECT_NAME} dlib::dlib)
```

main.cu
```
int main(){
return 0;
}
```

Will cause error because NVCC doesn't recognize flags like `-msse4`. This PR fixes it.


davisking:
Thanks for the PR, but it looks like this breaks the build.

jasjuang:
I tried it in vcpkg and it works. I suspect the cmake in appveyor is outdated. What version of CMake is it?

jasjuang:
But I think it should be encouraged. If we really want to be backwards compatible to a certain degree, I suppose we can try to add some sort of checks on the cmake version to decide this fix should be applied or not.

davisking:
That’s fine. Figure out what version of cmake first supported this and add an if.

jasjuang:
@davisking I added the check as you requested

davisking:
Can you rebase to the current dlib and see if the tests pass?

jasjuang:
I merged with the latest master. It's not "going to work with future cmake", it works on my visual studio with the latest official cmake 3.11 release.

davisking:
Cool. I thought you were only running gcc. 

Tests are still failing though :)

jasjuang:
It failed on "Variant: old-cmake". Are you sure you still want to support cmake 2.8...? If you really do I suppose we can try to figure out what exact version of cmake start supports the `COMPILER_LANGUAGE` clause in linux.

davisking:
Yes. Lots of people still use it.

jasjuang:
I figured out. It's introduced in cmake 3.3. Now this should be ready to merge.

davisking:

Issue #1220, opened 2018-03-30
Memory leak in Python get_face_chip


visionworkz:
Calling get_face_chip repeatedly, causes memory leak

## Expected Behavior
Calling  get_face_chip in a for loop should free memory and not cause more and more memory to be used.

## Current Behavior
System memory is constantly reduced as more and more files are processed with get_face_chip

## Steps to Reproduce
```
import time

import cv2
import dlib

__DETECTOR = dlib.get_frontal_face_detector()
__PREDICTOR =  dlib.shape_predictor("ktrain/image/shape_predictor_5_face_landmarks.dat")


def load_hundred():
    images = []
    for _ in range(100):
        img = cv2.imread("face2.jpg")
        rgb_img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
        dets = __DETECTOR(rgb_img)
        shape = __PREDICTOR(rgb_img, dets[0])
        face_chip = dlib.get_face_chip(rgb_img, shape)
        images.append(face_chip)
    return images

def main():
    while True:
        images = load_hundred()
        time.sleep(0.25)

if __name__ == "__main__":
    main()
```

```
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          32065        1994       14741         112       15328       28438
Swap:         32667           0       32667
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          32065        1995       14740         112       15328       28437
Swap:         32667           0       32667
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          32065        1998       14738         112       15328       28435
Swap:         32667           0       32667
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          32065        2070       14688          88       15305       28385
Swap:         32667           0       32667
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          32065        2109       14649          88       15305       28347
Swap:         32667           0       32667
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          32065        2138       14615          93       15311       28312
Swap:         32667           0       32667
```

Please also see:
https://github.com/pybind/pybind11/issues/27

* **Version**: 19.10
* **Where did you get dlib**: Github
* **Platform**: Ubuntu
* **Compiler**: gcc

davisking:
Thanks for reporting the bug.  Can you submit a PR that fixes it?

visionworkz:
[...]

PR #1222, opened 2018-03-31
Fixed reference count issue

visionworkz:
Fix for https://github.com/davisking/dlib/issues/1220
@davisking @supervacuus


davisking:
Cool, thanks for the PR.  At some point we should switch to a C++ rather than C interface so that these kinds of errors are impossible, like maybe py::array or write something ourselves if pybind11's numpy interfaces aren't sufficient (although they probably are).

@visionworkz Does this PR fix the problems you were observing?

davisking:
Er, duah, you are @visionworkz.  Thought @supervacuus submitted the PR for some reason :/

visionworkz:
@davisking yeah I fixed it first since it was just a ref count issue. I agree switching to pybind interfaces may be a better solution in the long run. 

I have tested it though. The leak is fixed.

PR #1223, opened 2018-03-31
Remove unnecessary semicolon

ariaaan:
That's pretty much it, there was an unnecessary semicolon in the `train_object_detector.py` python example.


davisking:
Cool, thanks for the PR :)

PR #1246, opened 2018-04-10
Fixed refcount issue in Python dlib.jitter_image and dlib.get_face_chips

visionworkz:
Pybind was incrementing reference count when adding handle to a list. This has been fixed now. Apologies for not catching it in the last commit.

Test code:
```
import sys

import dlib
import cv2
import numpy as np

# Load all the models we need: a detector to find the faces, a shape predictor
# to find face landmarks so we can precisely localize the face
detector = dlib.get_frontal_face_detector()
sp = dlib.shape_predictor("shape_predictor_5_face_landmarks.dat")

# Load the image using OpenCV
bgr_img = cv2.imread("tom.jpg")

# Convert to RGB since dlib uses RGB images
img = cv2.cvtColor(bgr_img, cv2.COLOR_BGR2RGB)

# Ask the detector to find the bounding boxes of each face.
dets = detector(img)
face = sp(img, dets[0])

# Get the aligned face image
image = dlib.get_face_chip(img, face, size=320)
count = sys.getrefcount(image)
assert count == 2, "Memory leak detected in get_face_chip"

# Get 5 jittered images without data augmentation
jittered_images = dlib.jitter_image(image, num_jitters=5)

# Ref count check for jitter_image
count = sys.getrefcount(jittered_images[0])
assert count == 2, "Memory leak detected in jitter_image"

# Ref count check for  get_face_chips
faces = dlib.full_object_detections()
faces.append(face)
images = dlib.get_face_chips(img, faces, size=320)
count = sys.getrefcount(images[0])
assert count == 2, "Memory leak detected in get_face_chips"

print("No memory leak detected")
```


davisking:
Cool, thanks for catching this.

PR #1249, opened 2018-04-10
Remove python C-API refs from numpy_returns.cpp

supervacuus:
Hi @davisking and @visionworkz, this PR would be my approach to remove the Python C-API from the numpy_returns wrapper. This could be used in other places as well, but i'd like to review and test this first in isolation.

I have written a very basic regression-test for #1220 , but marked the test-cases as "skip" because i have not invested time for the CI setup but used it continuously during my local development (@davisking i've seen you typically solve test-data with a base64-encoded string, i could give it a try here too).

Tell me what you think and how to proceed.


davisking:
Awesome, I was also bothered by the C API.  You can save some test files in the test folder. Don't worry about base64 encoding it or anything funny like that, just stash them in python/test and access them as appropriate. 

Also make sure the automated tests pass.  It seems like they are failing because opencv isn't installed on the test servers. It would be better if the tests didn't require opencv to be installed.

tools/python/test/test_numpy_returns.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_test_face_chip():
+    rgb_img, shape = get_test_image_and_shape()
+    return dlib.get_face_chip(rgb_img, shape)
+ 
+@pytest.mark.skip
+def test_get_face_chip():
+    face_chip = get_test_face_chip()
+    expected = numpy.load("test_face_chip.npy")
+    assert numpy.array_equal(face_chip, expected)
```
visionworkz: I would suggest that there is also an assertion on the ref count:
```
import sys
count = sys.getrefcount(face_chip)
assert count == 2, "Memory leak detected in get_face_chip"
```

tools/python/test/test_numpy_returns.py
```diff
@@ -0,0 +1,60 @@
...
+
+@pytest.mark.skip
+def test_get_face_chips():
+    face_chips = get_test_face_chips()
+    expected = numpy.load("test_face_chip.npy")
+    assert numpy.array_equal(face_chips[0], expected)
```
visionworkz: Similarly here:
```
count = sys.getrefcount(face_chips)
assert count == 2, "Memory leak detected in get_face_chips"
count = sys.getrefcount(face_chips[0])
assert count == 2, "Memory leak detected in get_face_chips"
```

visionworkz:
Thanks @supervacuus!

visionworkz:
Shall we introduce a load/save image method in the dlib library? This will remove any dependencies on OpenCV or skimage. Sample code is given below:
```
#include <dlib/image_io.h>
py::object load_rgb_image (const std::string &path)
{
    matrix<rgb_pixel> img;
    load_image(img, path);
    return create_array_from_rgb_image(img);
}

bool has_ending (std::string const full_string, std::string const &ending) {
    if(full_string.length() >= ending.length()) {
        return (0 == full_string.compare(full_string.length() - ending.length(), ending.length(), ending));
    } else {
        return false;
    }
}

void save_rgb_image(py::object img, const std::string &path)
{
    if (!is_rgb_python_image(img))
        throw dlib::error("Unsupported image type, must be RGB image.");

    std::string lowered_path = path;
    std::transform(lowered_path.begin(), lowered_path.end(), lowered_path.begin(), ::tolower);

    if(has_ending(lowered_path, ".bmp")) {
        save_bmp(numpy_rgb_image(img), path);
    } else if(has_ending(lowered_path, ".dng")) {
        save_dng(numpy_rgb_image(img), path);
    } else if(has_ending(lowered_path, ".png")) {
        save_png(numpy_rgb_image(img), path);
    } else if(has_ending(lowered_path, ".jpg") || has_ending(lowered_path, ".jpeg")) {
        save_jpeg(numpy_rgb_image(img), path);
    } else {
        throw dlib::error("Unsupported image type, image path must end with one of [.bmp, .png, .dng, .jpg, .jpeg]");
    }
    return;
}
```
and
```
    m.def("load_rgb_image", &load_rgb_image, 
	"Takes a path and returns a numpy array (RGB) containing the image",
	py::arg("path")
    );

    m.def("save_rgb_image", &save_rgb_image, 
	"Saves the given (RGB) image to the specified path. Determines the file type from the file extension specified in the path",
	py::arg("img"), py::arg("path")
    );
```

supervacuus:
@visionworkz From my pov it would be great not to depend on external libraries for basic image io. I can integrate your code in this PR. We could use this for some of the examples too (in another PR).

@davisking In this case I was specifically talking about the trained model for the shape predictor. It has ~9MB and i think there was a reason why you did not include it in the repository. If you can provide a lighter weight model to feed the test predictor that would be great.

visionworkz:
@supervacuus from my pov as well, it will be very useful to have image io exposed from Python. I also suggest the inclusion of a show image function:
```
#include <dlib/gui_widgets.h>
void show_rgb_image(py::object img, const std::string &window_name) {
    if (!is_rgb_python_image(img))
        throw dlib::error("Unsupported image type, must be RGB image.");
    image_window the_window(numpy_rgb_image(img), window_name);
    the_window.wait_until_closed();
}
```
and
```
    m.def("show_rgb_image", &show_rgb_image, 
	"Shows the image in a window",
	py::arg("img"), py::arg("window_name")="Image"
    );
```

I am adding the code samples here instead of in a PR as they depend on your `create_array_from_rgb_image` function included in this PR.

With these changes, we can remove skimage and OpenCV dependencies from all Python examples.

@davisking thoughts?

davisking:
@supervacuus Ah, if you want a shape predictor test then just run the trainer on the example test data and then load the resulting model and test that it does what it's supposed to.  You could train it on one image, using really mild settings that train very quickly and then test that the model works well on the image it's trained on.

@visionworkz Yeah, removing the external dependencies in the examples is great.  I don't think you need show_rgb_image though.  Just using the dlib.image_window is good right? I guess it needs to be updated to deal with numpy arrays but that's not a big deal, just another overload of set_image().

Also, can't a `def_buffer` statement be added to the `dlib::matrix` classes so that the image loader can just return a `matrix<rgb_pixel>` directly without doing a copy? That would run faster. See http://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html.  I haven't used that pybind11 feature but it looks good.

visionworkz:
@davisking the following code works perfectly:
```
import dlib
img = dlib.load_rgb_image("someimage.jpg")
window = dlib.image_window()
window.set_image(img)
```
I think we should leave one Python example with OpenCV as it demonstrates that images returned from dlib can be used with OpenCV too after swapping Red and Blue channels.

davisking:
Yeah. Leaving one opencv example is a good idea.

visionworkz:
@davisking, @supervacuus  one last consideration:
If we return an object via `def_buffer`, on the Python side, we would have to use something like:
```
np.array(matrix_instance, copy = False)
```
This does introduce more complexity on the Python side.

With the memory copy approach, it is already a Numpy array and can be used directly. 

For me personally, I prefer the convenience of just getting a Numpy array without additional complexity.

Ideally, if we can find an approach which can return a Numpy array without the memory copying, that would be best in my opinion.

davisking:
Yeah, I agree.  We shouldn't require python users to write weird code like
that.  It seems like there should be a way to make it transparent, but
maybe that's not the case.

I'm not adverse to adding a method to dlib::matrix like "steal_memory()" or
something that just returns the pointer to the matrix's memory block and
causes the matrix to go into an empty matrix state since it no longer has
any contents (since the memory was taken). Then that memory block could be
stuffed into something else.

supervacuus:
So to summarize:

- introduce basic image io interface to the python module
- rewrite the tests in terms of this interface to drop dependencies
- trains simple predictor model for the test to drop dependency on the heavy *.dat

with that we can close the PR.

further steps:
- evaluate options in pybind to bind array_t to matrix without a copy
- adapt examples to use basic image io but leave/create one example to show possible interoperability with opencv, skimage or pillow.

I just have to add, that i won't be able to work on this before the weekend. So in case you want to hijack this in the meantime @visionworkz feel free.

visionworkz:
@supervacuus I never intended to hijack this thread :-) I know you were the person who wrote the pybind11 stuff and that's why I mentioned you, so that you could eyeball the new stuff better. I appreciate your contributions!

@davisking I did some further research and here is what I found:
How to bind and return `matrix<rgb_pixel>` directly:
```
#include <pybind11/numpy.h>
matrix<rgb_pixel> get_face_chip (
    py::object img,
    const full_object_detection& face,
    size_t size = 150,
    float padding = 0.25
)
{
    if (!is_rgb_python_image(img))
        throw dlib::error("Unsupported image type, must be RGB image.");

    matrix<rgb_pixel> chip;
    extract_image_chip(numpy_rgb_image(img), get_face_chip_details(face, size, padding), chip);
    return chip;
}
```
and
```
    typedef matrix<rgb_pixel> type;
    py::class_<type>(m, "image", py::buffer_protocol())
   .def_buffer([](type &m) -> py::buffer_info {
        return py::buffer_info(
            image_data(m),                                      /* Pointer to buffer */
            sizeof(npy_uint8),                                  /* Size of one scalar */
            py::format_descriptor<npy_uint8>::format(),         /* Python struct-style format descriptor */
            3,                                                  /* Number of dimensions */
            { num_rows(m), num_columns(m), 3UL },               /* Buffer dimensions */
            { sizeof(npy_uint8) * num_columns(m) * 3,           /* Strides (in bytes) for each index */
              sizeof(npy_uint8) * 3,
              sizeof(npy_uint8) }
        );
    });
```

This works, and in Python, you would show the image as follows:
```
# Get the aligned face image
dlib_image = dlib.get_face_chip(img, face, size=320)
count = sys.getrefcount(dlib_image)
assert count == 2, "Memory leak detected in get_face_chip"
image = np.array(dlib_image, copy=False) # This step is necessary as dlib_image is of type dlib.image
count = sys.getrefcount(dlib_image)
assert count == 3, "Unexpected ref count for dlib image"
count = sys.getrefcount(image)
assert count == 2, "Unexpected ref count for numpy array"
window = dlib.image_window()
window.set_image(image)
window.wait_until_closed()
```

We **have** to initialize the array with `np.array(dlib.get_face_chip(img, face, size=320), copy=False)` which is non-ideal. So I prefer @supervacuus approach, unless we can find something else.

One more thing. Since, we are returning a numpy array, we know the dimensions will be H x W x 3. I did not get why we need to use `sizeof(rgb_pixel)`. Shouldn't the code be:
```
pybind11::array_t<uint8_t> create_array_from_rgb_image(const matrix<rgb_pixel> &rgb_image)
{
    const size_t dtype_size = sizeof(uint8_t);
    const auto rows = static_cast<const size_t>(num_rows(rgb_image));
    const auto cols = static_cast<const size_t>(num_columns(rgb_image));
    const size_t channels = 3;
    const size_t image_size = dtype_size * rows * cols * channels;

    auto arr = new uint8_t[image_size];
    memcpy(arr, image_data(rgb_image), image_size);

    return pybind11::array_t<uint8_t>(
        {rows, cols, channels},                                                     // shape
        {dtype_size * cols * channels, dtype_size * channels, dtype_size},          // strides
        arr,                                                                        // pointer
        pybind11::capsule{
            arr, [](void *arr_p) {
                delete[] reinterpret_cast<uint8_t *>(arr_p);
            }
        }
    );
}
```

With this, the Python code becomes dead simple:
```
# Get the aligned face image
image = dlib.get_face_chip(img, face, size=320)
count = sys.getrefcount(image)
assert count == 2, "Memory leak detected in get_face_chip"
window = dlib.image_window()
window.set_image(image)
window.wait_until_closed()
```

supervacuus:
@visionworkz "hijacking" was meant jokingly... just wanted to make sure that you know that i am not married to the code :-)

visionworkz:
@davisking and @supervacuus. I have consolidated the changes discussed here in PR https://github.com/davisking/dlib/pull/1258

supervacuus:
@davisking I think you can close this then?

davisking:
Yeah, that makes sense.  Thanks for working on this with @visionworkz. The overall API is now a lot nicer :)

Issue #1252, opened 2018-04-11
dlib Python module references save_jpeg() even if built with DLIB_JPEG_SUPPORT=0


aldanor:
When building dlib with `-DDLIB_JPEG_SUPPORT=0`, Python bindings shouldn't unconditionally refer to `save_jpeg()` (in face_recognition module), otherwise dlib becomes unimportable like so:

```
ImportError: dlib.cpython-36m-x86_64-linux-gnu.so: undefined symbol: 
_ZN4dlib9save_jpegERKNS_7array2dINS_9rgb_pixelENS_33memory_manager_stateless_kernel_1IcEEEERKSsi
```

davisking:
Why would you compile the python bindings without jpeg support in the first
place?

aldanor:
In my particular case? Since I only use dlib for its global opt module, nothing to do with image-related stuff - and because of this build issue all Python environments that use dlib have to carry a ljbjpeg around.

davisking:
Are there any problems this causes? What bad things happen because of this?

I'm not excited about adding more switches because users find them
confusing.  I don't want to get questions from people like, "I compiled
with DLIB_JPEG_SUPPORT=0 and then got an error that I couldn't load a jpeg
file. I am confused."  I get lots of such questions already. :(

aldanor:
My point of view, as a user, may be different from yours, of course :) Ditto re: not being excited about adding more switches, but here goes:

- Most 'normal' users of the Python side of dlib would just do a `pip install` or `conda install` a pre-built tarball, not having to mess with cmake compile flags and the like. This gives them the 'whole thing' if possible so, thus no questions.
- Those who disable things on purpose (e.g. jpeg support) probably do so because they know what they're doing and why they're doing it - in which case it should be no surprise to them that this or that feature ends up being disabled; I would be very surprised if someone purposefully disabled jpeg support and then opened a ticket to complain about, as you've noted above.
- There's parts of dlib (e.g. global optimization) that don't deal with any image related stuff; making libjpeg.so a hard dependency of the Python package is pretty brutal, since now you either have to ensure the boxes you deploy your Python package to have libjpeg installed, or haul it around in every conda environment your package gets installed to, depending on how you're structuring/packaging your Python code. If it wasn't for libjpeg, the only dependency is `numpy` - which makes total sense.
[...]

PR #1258, opened 2018-04-17
Add basic image io and remove python C-API refs from numpy_returns.cpp

visionworkz:
This is a consolidation of the discussion in https://github.com/davisking/dlib/pull/1249

For your review @davisking and @supervacuus 

@davisking I tried to fix the build but it cannot load the image in the test folder. I verified that the image exists. Maybe an issue related to DLIB_JPEG_SUPPORT flag?

There is further impact due to https://github.com/davisking/dlib/issues/1252

@aldanor if this PR is merged before your PR, you might wish to `ifdef` some portions of the code in this PR as well.


aldanor:
@visionworkz I started a local branch (with ifdef stuff) this weekend but haven't finished it yet. I'll wait for this batch of work to get merged in first then.

davisking:
@visionworkz, all the file paths when running `python setup.py test` should be relative to the root of the repository.  The ones in the new tests are relative to the python/test folder, I'm pretty sure that's why you are getting the file load errors.

tools/python/test/test_numpy_returns.py
```diff
@@ -24,17 +34,26 @@ def get_test_face_chips():
 def get_test_face_chip():
     rgb_img, shape = get_test_image_and_shape()
     return dlib.get_face_chip(rgb_img, shape)
- 
+
+# The tests below will be skipped for Python 2.7 as there is a pickle issue
```
davisking: This should work in python 2.7.  Can't you pickle using an earlier pickling version, like 2? It's certainly not the case that pybind11 pickling doesn't work in python 2.7.

davisking: If you use cPickle and a call like .dumps(variable,2) it works.  See http://pybind11.readthedocs.io/en/stable/advanced/classes.html#pickling-support

davisking:
Looks awesome, thanks for the nice PR :)

PR #1265, opened 2018-04-19
Renamed BOOST_JOIN to DLIB_BOOST_JOIN 

Zyrin:
This prevents name clashes when working with boost.

This happed to us after updateing from boost 1.58 to 1.67.


davisking:
Cool, thanks for the PR :)

PR #1276, opened 2018-04-26
New constructor for dnn_trainer

MrLIk:
This PR provide to create dnn_trainer on GPU that has specific device_id.
It is useful sometimes to run separate tasks on different GPUs. In current realization it is impossible not to run training process on 0 device. The input in one of the existing constructors `const std::vector<int>& cuda_extra_device` gives ability only to launch learning on additional GPUs and on GPU with id = 0 .


davisking:
It's not impossible. You just call set_device() in your code before you make the trainer.  Also, this behavior is misleading since you can make a network and then create a trainer using some device other than the one that is currently selected, but that would cause a cuda error. Or is there something I'm missing here?

MrLIk:
My fault, I just try to call `dlib::cuda::set_device()` before dnn_trainer creation and it is what I need. I thought for some reasons that setting device should be inside dnn_trainer object. I close PR, sorry for false alarm :)

davisking:
No worries :)

PR #1277, opened 2018-04-26
Supporting bottom-up DIBs

pjuzeliunas:
Rows in bitmap image file can be arranged either top-down, or bottom-up. Current implementation of BMP file reader in dlib supports only top-down row arrangement.

Fix: negative `biHeight` value (from DIB) means that we have a bottom-up arrangement. Therefore rows are assigned to an image in a reverse order.

### Literature
 - [Top-Down vs. Bottom-Up DIBs](https://msdn.microsoft.com/en-us/library/windows/desktop/dd407212(v=vs.85).aspx) by Microsoft

### DIFF
--- a/dlib/image_loader/image_loader.h
+++ b/dlib/image_loader/image_loader.h
@@ -111,8 +111,8 @@ namespace dlib
             a = buf[i]; b = buf[i+1]; c = buf[i+2]; d = buf[i+3];
             biHeight = a | (b<<8) | (c<<16) | (d<<24);
             
-            bool topDown = biHeight < 0;
-            if (topDown)
+            bool bottomUp = biHeight < 0;
+            if (bottomUp)
                 biHeight = 0 - biHeight;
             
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the PR.

dlib/image_loader/image_loader.h
```diff
@@ -110,7 +110,11 @@ namespace dlib
             i += 4;
             a = buf[i]; b = buf[i+1]; c = buf[i+2]; d = buf[i+3];
             biHeight = a | (b<<8) | (c<<16) | (d<<24);
-
+            
+            bool topDown = biHeight < 0;
```
davisking: Shouldn't this variable be called bottomUp?

pjuzeliunas: Yes, you are right. Fixed.

davisking:
Awesome.

PR #1281, opened 2018-04-30
parallelize image rotations

Vermeille:
Hello Davis,

This PR is about parallelizing `add_image_rotations` which takes an unreasonable amount of time on my machine, while there are a lot of idling CPUs.

I'm not sure about some low level arbitrary choices I've made and implementation details, so please, get back to me. I'll remove the debug prints and all that while treating your feedback.

Many thanks!

### DIFF
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -3,6 +3,9 @@
 #ifndef DLIB_INTERPOlATIONh_
 #define DLIB_INTERPOlATIONh_ 
 
+#include <thread>
+#include <algorithm>
+
 #include "interpolation_abstract.h"
 #include "../pixel.h"
 #include "../matrix.h"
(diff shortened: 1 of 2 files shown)


davisking:
Thanks for the PR. 

This should probably use dlib::parallel_for().  And yes, please remove unrelated things like print statements and other stuff.  A PR should be focused and clean.

Vermeille:
Hello again,

I think that fixes what you asked for :)

Hope it helps!

dlib/image_transforms/interpolation.h
```diff
@@ -1543,33 +1546,32 @@ namespace dlib
...
+        using namespace impl;
 
-        std::vector<T> objtemp;
-        std::vector<U> objtemp2;
-        typename image_array_type::value_type temp;
-        for (long i = 0; i < angles.size(); ++i)
-        {
-            for (unsigned long j = 0; j < images.size(); ++j)
+        image_array_type new_images(images.size() * angles.size());
+        Objects new_objects(images.size() * angles.size());
```
davisking: Don't make the Objects typedef. You only use it once.  So adding it is just more code and more confusing indirection for someone reading it.  Just say std::vector<std::vector<T>>.  That's much clearer.

dlib/image_transforms/interpolation.h
```diff
@@ -1543,33 +1546,32 @@ namespace dlib
...
+        Objects2 new_objects2(images.size() * angles.size());
+
+        dlib::parallel_for(0, images.size(), [&](int j) {
+            typename image_array_type::value_type temp;
+
+            int dst_base = j * angles.size();
+            for (long i = 0; i < angles.size(); ++i)
             {
+                int dst = dst_base + i;
```
davisking: Be consistent in index type.  In particular, these ints should be longs.

dlib/image_transforms/interpolation.h
Vermeille: Agreed. Fixed.

dlib/image_transforms/interpolation.h
Vermeille: I guess it's right, now?

davisking:

PR #1293, opened 2018-05-08
Introduced new rgb_normalized_pixel as a pixel type

visionworkz:
When dealing with images for deep learning, it is often useful to normalise them first.

This PR introduces a new pixel type rgb_normalized_pixel which makes it easy to deal with colour values centered at 0 (+/- 0.5)

Loading an image which is normalised is as easy as:
```
matrix<rgb_normalized_pixel> image;
load_image(image, "test.png");
// And this can be directly saved too
save_png(image, "rgb-norm.png");
save_bmp(image, "rgb-norm.bmp");
save_jpeg(image, "rgb-norm.jpg");
// save_dng(image, "rgb-norm.dng"); <= This is broken currently
```

@davisking I have tried to add some tests but this PR will benefit from your more thorough look through. Currently save_dng is broken. This will also be useful later when/if we want to expose normalised images to Python in a queue or some other mechanism.

### DIFF
--- a/dlib/image_saver/save_png.h
+++ b/dlib/image_saver/save_png.h
@@ -80,7 +80,7 @@ namespace dlib
 
             impl::impl_save_png(file_name, row_pointers, img.nc(), impl::png_type_rgb_alpha, 8);
         }
-        else if (pixel_traits<pixel_type>::lab || pixel_traits<pixel_type>::hsi || pixel_traits<pixel_type>::rgb)
+        else if (pixel_traits<pixel_type>::lab || pixel_traits<pixel_type>::hsi || pixel_traits<pixel_type>::rgb || pixel_traits<pixel_type>::rgb_normalized)
         {
             // convert from Lab or HSI to RGB (Or potentially RGB pixels that aren't laid out as R G B)
             array2d<rgb_pixel> temp_img;
--- a/dlib/matrix/matrix_utilities.h
+++ b/dlib/matrix/matrix_utilities.h
@@ -2779,7 +2779,8 @@ namespace dlib
             pixel_traits<P>::rgb,
             pixel_traits<P>::hsi,
             pixel_traits<P>::rgb_alpha,
-            pixel_traits<P>::lab
+            pixel_traits<P>::lab,
+            pixel_traits<P>::rgb_normalized
             >::value
         >
     struct pixel_to_vector_helper;
@@ -2858,6 +2859,21 @@ namespace dlib
         }
     };
 
+    template <typename P>
+    struct pixel_to_vector_helper<P,6>
+    {
+        template <typename M>
+        static void assign (
+                M& m,
+                const P& pixel
+        )
+        {
+            m(0) = static_cast<typename M::type>(pixel.red);
+            m(1) = static_cast<typename M::type>(pixel.green);
+            m(2) = static_cast<typename M::type>(pixel.blue);
+        }
+    };
+
 
     template <
         typename T,
@@ -2882,7 +2898,8 @@ namespace dlib
             pixel_traits<P>::rgb,
             pixel_traits<P>::hsi,
             pixel_traits<P>::rgb_alpha,
-            pixel_traits<P>::lab
+            pixel_traits<P>::lab,
+            pixel_traits<P>::rgb_normalized
             >::value
         >
     struct vector_to_pixel_helper;
@@ -2958,6 +2975,21 @@ namespace dlib
             pixel.l = static_cast<unsigned char>(m(0));
             pixel.a = static_cast<unsigned char>(m(1));
             pixel.b = static_cast<unsigned char>(m(2));
+        }
+    };
+
+    template <typename P>
+    struct vector_to_pixel_helper<P,6>
+    {
+        template <typename M>
+        static void assign (
+                P& pixel,
+                const M& m
+        )
+        {
+            pixel.red = static_cast<float>(m(0));
+            pixel.green = static_cast<float>(m(1));
+            pixel.blue = static_cast<float>(m(2));
         }
     };
 
--- a/dlib/pixel.h
+++ b/dlib/pixel.h
@@ -34,6 +34,7 @@ namespace dlib
         This traits class will define the following public static members:
             - bool grayscale
             - bool rgb
+            - bool rgb_normalized
             - bool rgb_alpha
             - bool hsi
             - bool lab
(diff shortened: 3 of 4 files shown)


davisking:
Thanks for the PR.

It's not totally clear to me what the use case is for this. What does this allow you to do that was difficult to do before?

visionworkz:
Hi @davisking,

This is basically syntactic sugar. Instead of writing a function with a `for` loop that does the conversion, it can be done directly by implementing this type.

My end goal was to introduce a multi-threaded queue for batch sampling without replacement on the C++ side and then exposing it to Python as well. The queue would encapsulate the mechanism used [here](https://github.com/davisking/dlib/blob/master/examples/dnn_metric_learning_on_images_ex.cpp) (for example).

I wanted the queue to return normalised pixel values to avoid processing on the Python side while images are being read from the queue.

Since the batch will then typically be copied to the GPU, I wanted to return the batch as a Numpy array of the form (batch_size, height, width, 3).

To that end, I have the following generic code:
```
template <
    typename pixel_type
    >
    void assert_is_image_array(const std::vector<matrix<pixel_type>>& imgs) {
        // We expect every image to be the same size
        size_t rows_from_images = -1;
        size_t cols_from_images = -1;

        for(auto image : imgs) {
            if(rows_from_images == -1)
                rows_from_images = num_rows(image);
            if(cols_from_images == -1) {
                cols_from_images = num_columns(image);
            }
            size_t current_image_rows = num_rows(image);
            size_t current_image_cols = num_columns(image);
            if(current_image_rows != rows_from_images || current_image_cols != cols_from_images) {
                throw dlib::error("Cannot convert from std::vector<matrix<pixel_type>> to Numpy as images in vector have different sizes");
            }
        }
    }

template <
    typename pixel_type
    >
class numpy_images : public py::array_t<typename pixel_traits<pixel_type>::basic_pixel_type>
{
    /*!
        REQUIREMENTS ON pixel_type
            - is a dlib pixel type, this just means that dlib::pixel_traits<pixel_type>
                is defined.
        WHAT THIS OBJECT REPRESENTS
            This is collection of N images of the same size which can be created/assigned from 
            an std::vector containing matrix<pixel_type> objects. This object helps return a 
            Numpy array of the shape (N, image_rows, image_cols, 3)
    !*/
public:

    numpy_images() = default;

    numpy_images(std::vector<matrix<pixel_type>>&& rhs) {
        *this = convert_to_numpy(std::move(rhs));
    }

    numpy_images& operator= (
        std::vector<matrix<pixel_type>>&& rhs
    )
    {
        *this = convert_to_numpy(std::move(rhs));
        return *this;
    }

    numpy_images& operator= (
        const py::array_t<typename pixel_traits<pixel_type>::basic_pixel_type>& rhs
    )
    {
        py::array_t<typename pixel_traits<pixel_type>::basic_pixel_type>::operator=(rhs);
        return *this;
    }

private:
    static py::array_t<typename pixel_traits<pixel_type>::basic_pixel_type> convert_to_numpy(std::vector<matrix<pixel_type>>&& imgs)
    {
        assert_is_image_array(imgs);

        using basic_pixel_type = typename pixel_traits<pixel_type>::basic_pixel_type;
        const size_t dtype_size = sizeof(basic_pixel_type);
        const auto batch = static_cast<const size_t>(imgs.size());
        const auto rows = batch != 0 ? static_cast<const size_t>(num_rows(imgs[0])) : 0;
        const auto cols = batch != 0 ? static_cast<const size_t>(num_columns(imgs[0])) : 0;
        const size_t channels = pixel_traits<pixel_type>::num;
        const size_t block_size = dtype_size * batch * rows * cols * channels;
        const size_t image_ptr_counter_step = rows * cols * channels;
        const size_t image_size = dtype_size * rows * cols * channels;
        size_t image_ptr_counter = 0;
        basic_pixel_type* memory = (basic_pixel_type*) malloc(block_size);

        int image_num = 1;
        for(matrix<pixel_type> img : imgs) {
            std::unique_ptr<pixel_type[]> arr_ptr = img.steal_memory();
            basic_pixel_type* arr = (basic_pixel_type *) arr_ptr.release();
            basic_pixel_type* destination = memory + image_ptr_counter;
            memcpy (destination, arr, image_size);
            delete[] arr;
            image_ptr_counter += image_ptr_counter_step;
        }

        return pybind11::template array_t<basic_pixel_type>(
            {batch, rows, cols, channels},                                                  // shape
            {
                dtype_size * rows * cols * channels, 
                dtype_size * cols * channels, 
                dtype_size * channels, 
                dtype_size
            },                                                                              // strides
            memory,                                                                         // pointer
            pybind11::capsule{ memory, [](void *arr_p) { delete[] reinterpret_cast<basic_pixel_type*>(arr_p); } }
        );
    }
};
```

As you can see, introducing the rgb_normalized_pixel, the above code would work without any special handling for normalisation.

Instead of submitting one big PR, I decided to introduce the changes in small chunks and add tests as we go along so that the functionality can be well tested.

davisking:
I think you should do the scaling inside `convert_to_numpy`.  That's a much simpler way to code it.  I also don't want to enshrine a particular normalization strategy in a pixel type, especially since this one is peculiar to some kinds of neural networks and not generally what people doing image processing on images that happen to have multiple floating point bands want.  

I would be open to adding a multi-band float pixel type that behaved the way the existing single band float pixel types do, which is to say no scaling. They simply contain whatever the underlying value was and when being converted to other pixel types they perform a simple saturation at the max or min values of the target type. 

Also, the code you posted is a little more complex and error prone than it needs to be.  There is manual memory management which should not be there.  It also is destructive of the source image array when it doesn't need to be since it's not like you pass ownership of the source image memory blocks to python, they are simply copied and then destroyed.  You shouldn't be stealing the memory block at all here since you just immediately destroy it. In general you should try and not have raw owning pointers since it creates the opportunity for a memory leak.

Also don't call malloc. Instead you should just make a py::array_t of the appropriate size using the py::array_t's constructor.  Let it allocate it's own memory.  Then just use a for loop to copy the data into the py::array_t you made.

visionworkz:
@davisking thanks for your comments and inputs.

Yes, I agree that a generic multi-band `float` and `unsigned char` type is a good idea, using min and max values for scaling.

In particular, it could kill two birds with one stone by also solving this request:
https://github.com/davisking/dlib/pull/518

Would you like to comment on the template type for this and if you think we should implement it using pixel_traits?

As for the memory, I can see why stealing is not necessary in this case.

visionworkz:
Ok. I think we can close this PR then. I will just work with a matrix of float and appropriate override.

PR #1296, opened 2018-05-09
allow to override the initial shape of the shape predictor

Vermeille:
Hi,

I did as you suggested, but I'm not sure you're actually interested in the change. Feel free to discard it.

Have a nice day :)

### DIFF
--- a/dlib/image_processing/shape_predictor_trainer.h
+++ b/dlib/image_processing/shape_predictor_trainer.h
@@ -184,6 +184,19 @@ namespace dlib
         padding_mode_t get_padding_mode (
         ) const { return _padding_mode; }
 
+        void set_initial_shape (
+            const matrix<float, 0, 1>& mat
+        )
+        {
+            _initial_shape = mat;
+        }
+
+        const matrix<float, 0, 1>& get_initial_shape (
+        )
+        {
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  Did it improve the results?  Also, how are you getting the initial shape in the first place.  There doesn't seem to be a way to get the shape used by a simpler dataset in the PR.

Vermeille:
The results are better but still not satisfying unfortunately. I guess I'll just train a regression forest to predict an angle of rotation so that I put the face straight up before alignment.

To get the the initial shape I trained a model without face rotations, and deserialized manually the model to keep only the initial shape part of it. A bit hacky, but that works.

davisking:
Ok, I'm not too surprised.  I think the right thing to do here is to do some more experiments to see what really would improve the tool.  I suspect that a better initial shape could be found by aligning all the shapes together with a similarity transform prior to averaging as that would fix the underlying problem.  There is still the issue about why what you did didn't work.  It might just be that you need a deeper cascade.  Who knows.  Anyway, I'm going to close this PR since, like you said, this is a bit hacky and the right thing to do is more work.  If you want to do the more complex thing that would be cool, but you certainly aren't obligated :)

PR #1302, opened 2018-05-11
Added option to link against mkl_sequential

Zyrin:


davisking:
Thanks for the PR.  However, users should use the MKL_NUM_THREADS environment variable to control this rather than via linker settings. 

Future readers can also see the discussion here: https://github.com/davisking/dlib/issues/1303

Zyrin:
In my opinion, for small applications/projects an environement variable might be ok, but not for huge production applications (especially with a lot of libraries). My problem with them is that you either will need to add startup scripts for every appliation, or you will need to set the variable programatically. In both cases this is done in complete unrelated parts of the application. That makes it hard to find again later and it is error prone.

davisking:
Why not call the MKL’s API to set it then? The MKL exposes all this stuff via an API you can call from your app and tell it to do whatever you want.

Zyrin:
Its the same problem as with environment variables.
Let me give you a concrete example from the company I work at, maybe that will make it clearer why I want to be able to linking against mk_sequential. We do realtime imageprocessing and use dlibs neural nets for some classification problems. Our applications are already heavily parallelized, so adding more threads that just waste cpu cycles (dlibs dnns in combination with multithreaded mkl) is a huge no go.

Now if I were to use mkls multithreaded mkl library for dlib I would have to paste mkl_set_num_threads in quite a lot of places, that in themselves have nothing to do with mkl.

In addition, other pealpe that want to use dlibs networks for new agorithms would would have to know and remeber about setting mkls threadcount to one. And it is almost garanteed that one way ore another, someone would not do that, thus wastig recouces.

I just want to link dlib against mkl_sequential, so that everything regarding dlib just works without any waste and no one else has a chance of doing it wrong.

davisking:
How do you link to dlib now?  Using cmake and add_subdirectory()?

Zyrin:
No, via targets. dlib is prebuild once on a buildserver and then fetched in other projects via a find_package

davisking:
Alright, I relent.  Did you test this PR on both windows and linux?

Zyrin:
Well my company uses dlib with this patch included and builds windows as well as linux projects with it. So yes. But just to be shure I can test the current Head at home in a fiew ours.

Zyrin:
I compiled dtest on ubuntu 16.04 and ran the tests successfully.

davisking:
Sweet.  Thanks for the PR :)

PR #1307, opened 2018-05-15
Modified README.txt at imglab module with README.md

daVincere:
This pull request
1. Modified the README file for imglab module for better readability.
2. Added a section regarding adding parts to objects as I found many having the same issue when training their own model.


davisking:
Thanks for the PR.  

This seems to be basically duplicative of the documentation in about->help.  I don't want to just duplicate it since it just becomes one more thing to update or get out of sync.  However, maybe switching to markdown and putting a huge bold message that says to read about->help is a good idea since the lack of people reading that information is really the issue.  

Also noting that imglab does part annotations in the first paragraph is probably not a bad idea though.

daVincere:
Thanks for the prompt response, Davis.
I committed a similar mistake of not reading the about->help documentation hence wanted others to not go through it. However, I'll keep exploring the library and hopefully, eventually contribute in some substantial way.

Issue #1299, opened 2018-05-10
Shape Predictor -- better/different error metrics


akalra:
Hi --

First off -- thank you for your excellent work. We love dlib, use it extensively, and point to it's thoughtful design and documentation as our guide posts for code we write.

For one of our own projects, we're discussing extending dlib to evaluate errors on a per-landmark basis, and perhaps providing more statistics around them. Specifically, we'll:

1. compute error per landmark across all images
2. return more error statistics (variance, min, max, etc, along with the existing average)
3. potentially log error on a per-image basis in some low-priority way to aid in finding "difficult" images that are throwing off the overall metric

For our project, some landmarks are more error prone than others, and our relatively 'noisy' input dataset may mean that there's images that are hurting more than helping overall. We're hoping these changes will better help us evaluate the accuracy of the models we train.

If you think this is an interesting extension to dlib proper, I'd be happy to extend test_shape_predictor to conditionally return 'detailed' errors if asked to -- it should be straightforward to do it in a backwards compatible way. running_stats already computes many of these things.

If not, we can continue with our plans to do it outside dlib. Thoughts?

Thanks!

Ankur

PR #1308, opened 2018-05-15
WIP: Shape Predictor metrics

akalra:
This is a work-in-progress, but I wanted to get your feedback on my approach to #1299 before I spent more time fleshing it out. Think of this as an initial rough sketch ;)

The goal is to provide more nuanced error metrics while testing a shape_predictor. We now return a running_statistics object for both the overall error and for each individual landmark, which you can interrogate for min, max, average, etc.

Things about this I'm not thrilled about:

 - to maintain backwards compatibility, I had to basically double the number of relevant function signatures, and chain the old implementations to the new ones. This feels a little awkward, since a further orthogonal change will double or triple the number of signatures again. All these signatures may make it hard for developers using the library to find the right version, and will also need to dealt with on the python binding layer. It seems like there should be a better pattern to signal optional behavior with less overhead?

 - I'm unsure about the return type -- I thought about creating a new class, or hand-picking specific values from running_statistics, but didn't have a strong perspective either way. I went with the simplest thing for now, happy to revisit.

 - I mimicked the existing test case, but was unsure if I should do something stronger. It felt like verifying the mathematical properties (i.e. total min is min of all errors, etc) was going beyond testing the actual implementation of how errors were computed and returned.

Things I still need to do:

 - wire this up to the python layer
 - extend the example code to demonstrate how to use this

Very open to feedback. I'm comfortable in C++ but don't live in it, so also open to stylistic / idiomatic suggestions.

Thoughts?

Ankur

### DIFF
--- a/dlib/image_processing/shape_predictor.h
+++ b/dlib/image_processing/shape_predictor.h
@@ -9,7 +9,6 @@
 #include "../matrix.h"
 #include "../geometry.h"
 #include "../pixel.h"
-#include "../tuple.h"
 #include "../statistics.h"
 #include <utility>
 
@@ -436,11 +435,23 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 // ----------------------------------------------------------------------------------------
 
+    struct shape_predictor_statistics
+    {
+        shape_predictor_statistics( running_stats<double> error_across_landmarks,
+                                    std::vector< running_stats<double> > error_by_landmark)
+                                        : error_across_landmarks(error_across_landmarks),
+                                          error_by_landmark(error_by_landmark)
+        {}
+
+        running_stats<double> error_across_landmarks;
+        std::vector< running_stats<double> > error_by_landmark;
+    };
+
+
     template <
         typename image_array
         >
-    dlib::tuple<running_stats<double>, std::vector<running_stats<double>>>
-        test_shape_predictor_with_detailed_statistics (
+    shape_predictor_statistics test_shape_predictor_with_detailed_statistics (
             const shape_predictor& sp,
             const image_array& images,
             const std::vector<std::vector<full_object_detection> >& objects,
@@ -504,11 +515,7 @@ namespace dlib
             }
         }
 
-        dlib::tuple<running_stats<double>, std::vector<running_stats<double>>> result;
-        result.get<0>() = rs;
-        result.get<1>() = rs_by_landmark;
-
-        return result;
+        return shape_predictor_statistics(rs, rs_by_landmark);
     }
 
 // ----------------------------------------------------------------------------------------
@@ -516,8 +523,7 @@ namespace dlib
     template <
         typename image_array
         >
-    dlib::tuple<running_stats<double>, std::vector<running_stats<double>>>
-        test_shape_predictor_with_detailed_statistics (
+    shape_predictor_statistics test_shape_predictor_with_detailed_statistics (
         const shape_predictor& sp,
         const image_array& images,
         const std::vector<std::vector<full_object_detection> >& objects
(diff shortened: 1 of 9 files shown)


davisking:
Thanks for the PR.  It looks generally fine, except don't return a tuple.  Instead create a struct with named fields that are suggestive of their meaning/content. Also, if the returned variable is a struct then it's easy to add more statistics later without disrupting the interface.

Also yeah, add a few more tests.  Don't go crazy but a little more coverage of the new behavior would be good.

akalra:
Thanks, I incorporated that feedback.

I'm working on the Python bindings now, but this will take a bit longer than I anticipated since running_stats is not currently used at the Python layer. I'll have to investigate how to pipe that through cleanly, it (on its own) is a fair bit more involved than my change.

Incidentally, I noticed that the Python-to-C++ implementation for test_shape_predictor was distributed between ./tools/python/src/shape_predictor.cpp and ./tools/python/src/shape_predictor.h in a non-obvious (to me) way. 

In shape_predictor.cpp:
 - test_shape_predictor_with_images_py (actual Python-to-C++ code)
 - test_shape_predictor_with_images_no_scales_py (actual Python-to-C++ code)
 - .. (actual pybind11 bindings)

In shape_predictor.h:
 - test_shape_predictor_py (actual Python-to-C++ code)
 - test_shape_predictor_with_images (this is post-translation, pure C++)

If that's intentional, can you help me understand how you think about what goes in which file? Just want to make sure I put my stuff in the right places.

Thanks!

Ankur

davisking:
Yeah, a lot of people have edited that code so it's a little wonky and
could be cleaned up. So don't read anything into the organization and
anything you want to cleanup is fine to cleanup :)

akalra:
PIped everything up to the python layer, and updated the documentation.

Now python has full access to running_stats and shape_predictor_statistics, and the Python API exposes a regular 'test_shape_predictor' (which returns mean as before) and a 'test_shape_predictor_with_stats' (which returns a shape_predictor_statistics).

Let me know if you have any outstanding concerns. I think I've incorporated all your feedback so far, but happy to make other changes if necessary.

Thanks!

Ankur

Note: I think I accidentally rebased on upstream changes incorrectly while trying to resolve conflicts, so there's a messy git history in the PR. I can submit a new PR to clean it up if you think that's important, but I'm guessing you squash this all into a single commit anyway?

davisking:
Ha, oh yeah, definitely make a new PR or rebase or something.  The PR is pretty wonky now :)

davisking:
If only mercurial had won out over git the world would be a better place :)

akalra:
Cleaned this up :) 

Sorry for the churn, I should know better than to try to do anything complicated with git late at night. Coffee helps with the incantations.

Ankur

akalra:
I think we had all the changes in, I just had to sync to the latest version of master and make sure there were no conflicts. Will try to verify that and sync to tip of master in the next week.

Sorry I dropped the ball on this. Our project moved past this stage of the work, and I got caught up in other things.

Ankur

davisking:
No problem

RalphMasson:
Hello, I am interested in using the function dlib.test_shape_predictor_with_stats but it doesn't seem to exist anymore. Is there anything close to this available in dlib? thank you

davisking:
No.  This PR never got finished up.  It seemed like a decent idea though :)

akalra:
I'm not working on this actively anymore, but it was a useful piece of
functionality when I needed it. If there's still interest, I'm happy to
take a crack at it again or pair with someone on it.

Pity, in re-reading this thread, it sounds like we got super close to
finishing it up. I think I may have gotten the last round of feedback in a
particularly busy time and then never quite got back to it.

Thanks!

Ankur

PR #1319, opened 2018-05-18
Fix compile without dependencies

xsacha:
When compiled without dependencies (such as JPEG/PNG) and installed with the created Config, the find_package would fail


davisking:
Thanks for the PR.

I don't understand what this does or the problem it fixes though.  For instance, I just built dlib with the dependencies off and find_package seemed to work fine.

xsacha:
Sorry, I'll close this for now. There's more to this PR.

davisking:
No worries.

PR #1330, opened 2018-05-23
Add a default cpack "package" target.

JosiahKane:
Add a default cpack "package" target when dlib is directly built, as an alternative to building for immediate installation.

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -24,6 +24,7 @@ endif()
 include(cmake_utils/release_build_by_default)
 include(cmake_utils/use_cpp_11.cmake)
 
+set(CPACK_PACKAGE_NAME "dlib")
 
 set(CPACK_PACKAGE_VERSION_MAJOR "19")
 set(CPACK_PACKAGE_VERSION_MINOR "12")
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.

I've never used cpack.  Can you say what does this change enable/why is it useful?

JosiahKane:
It enables building an archive which neatly packages up the built library. 

This PR can be used by running
`cmake --build . --target package --config Release`
where you would otherwise run
`cmake --build . --config Release`

The result on my mac is an archive called dlib-19.12.99-Darwin.tar.gz, containing the headers under include/dlib and libdlib.a under lib. There are also pkgconfig and cmake configuration files. As usual passing BUILD_SHARED_LIBS determines whether that build library is shared or static; I would instead get libdlib.19.12.99.dylib under lib with that enabled.

Our particular use case is to isolate building dlib from building the rest of our source code, while still having something we can easily depend on with cmake. We want the isolation primarily because we'll want to build dlib in release mode for speed even while building our code in debug mode. It also allows for slightly faster rebuilds of our code, because we do not usually want to rebuild our stable dependencies.

davisking:
Cool, sounds good :)

Can the `include(CPack)` statement be put with the other install and configure directives? Or does it need to be where it is?  It would be slightly cleaner if co-located with the other install instructions.

JosiahKane:
Sure. I've moved it down.

davisking:
Awesome, thanks for the PR :)

PR #1335, opened 2018-05-28
Allow batched face recognition for greater performances

Vermeille:
Hi Davis!

As always, feel free to voice any criticism, opinion, or suggestion for improvement, or to ask any question you might have.

Have a great day!

### DIFF
--- a/tools/python/src/face_recognition.cpp
+++ b/tools/python/src/face_recognition.cpp
@@ -11,6 +11,7 @@
 #include <dlib/image_io.h>
 #include <dlib/clustering.h>
 #include <pybind11/stl_bind.h>
+#include <pybind11/stl.h>
 
 
 using namespace dlib;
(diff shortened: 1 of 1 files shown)


tools/python/src/face_recognition.cpp
```diff
@@ -47,33 +48,85 @@ class face_recognition_model_v1
...
         face_descriptors.reserve(face_chips.size());
 
         if (num_jitters <= 1)
         {
+            int img_i = 0;
+            int i = 0;
             // extract descriptors and convert from float vectors to double vectors
-            for (auto& d : net(face_chips,16))
-                face_descriptors.push_back(matrix_cast<double>(d));
+            for (auto& d : net(face_chips,16)) {
```
davisking: This loop is kinda bizarre.  Rewrite it so it's clearer.  Probably allocate the size of face_descriptors in it's constructor so you can just index and push_back would be much clearer.

davisking:
Thanks for the PR.  It looks good, except for those funny loops I noted.  You should also make the {} style match the rest of the code.  Right now in the PR it's all mismatched.

Vermeille:
It should be good now :)

tools/python/src/face_recognition.cpp
```diff
@@ -47,33 +48,82 @@ class face_recognition_model_v1
...
         {
+            int img_i = 0;
+            int i = 0;
             // extract descriptors and convert from float vectors to double vectors
             for (auto& d : net(face_chips,16))
-                face_descriptors.push_back(matrix_cast<double>(d));
+            {
+                face_descriptors[img_i].push_back(matrix_cast<double>(d));
+                ++i;
+                while (i == batch_faces[img_i].size())
```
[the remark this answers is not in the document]
Vermeille: I'm out of town for a while and don't have much time to review this thoroughly, though, in order not to let you hanging with this: I'm `while`-ing that condition, because:

1. we've computed the good number of faces in that image, so the condition is true.
2. That means we're starting over to count the number of faces in that said image, `i = 0`.
3. _We're looping while there is no face in the `img_i`th pic_. If we have a pic with no face in it, we can go out of sync and have the arrays unmatching

But I'm not super super formal about this and I'm actually questioning the correctness of this if the _first_ image has actually zero faces in it.

As I said, I'm out of town currently, but I'm not forgetting this PR on which I depend. In the meantime, if you want to check that for yourself, you're obviously really welcome.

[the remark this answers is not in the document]
davisking: So long as the code is simplified it's fine :)

But right now it's way too squirrely.

Vermeille:
Should be fine :)

tools/python/src/face_recognition.cpp
```diff
@@ -47,33 +48,93 @@ class face_recognition_model_v1
...
 
+        std::vector<std::vector<matrix<double,0,1>>> face_descriptors(batch_imgs.size());
         if (num_jitters <= 1)
         {
+            int img_i = 0;
+            int i = 0;
             // extract descriptors and convert from float vectors to double vectors
             for (auto& d : net(face_chips,16))
-                face_descriptors.push_back(matrix_cast<double>(d));
+            {
```
davisking: Sorry, but these loops are still freaking me out.  I think you should be able to use this instead:

```
auto descriptors = net(face_chips,16);
auto next = descriptors.begin();
for (auto& img : face_descriptors)
{
    for (auto& face : img)
        face = matrix_cast<double>(*next++);
}
```

Does that work or am I missing something?  All you need to do is initialize face_descriptors appropriately.

Vermeille: I'm sorry, I don't understand your example. What is to be done with `face` then?

davisking:
I'm not sure I understand what you mean.  The face variable in the code I
posted?

davisking:
Yeah, that's fine too.

Vermeille:
I guess this one's good, for real.

Thanks for insisting on making the code easier. I didn't realize this could have been written more elegantly, and it's now a new pattern in my mental bank I'll certainly use again.

tools/python/src/face_recognition.cpp
```diff
@@ -47,33 +48,78 @@ class face_recognition_model_v1
...
         {
-            for (auto& fimg : face_chips)
-                face_descriptors.push_back(matrix_cast<double>(mean(mat(net(jitter_image(fimg,num_jitters),16)))));
+            // extract descriptors and convert from float vectors to double vectors
+            auto fimg = std::begin(face_chips);
+            for (int i = 0; i < batch_faces.size(); ++i)
+            {
+                for (int j = 0; j < batch_faces[i].size(); ++j)
+                {
+                    auto& r = matrix_cast<double>(mean(mat(net(
```
davisking: You have `matrix_cast<double>` two times here.  It would also probably be faster if you did a std::move() on the insertion to the face_descriptors array.

davisking:
Yeah, no problem.  Thanks for the PR :)

Did you test both the jittered and non jittered code paths to make sure they match the non-batch outputs?  If so then I think it's basically ready to merge.

Vermeille:
I just did. The element-wise difference is almost exactly zero (when not zero) in the non jittered version, and about 1e-4 / 1e-5 in the jittered version

davisking:
Cool, thanks for the PR :)

davisking:
Maybe I'm being an idiot, but this doesn't seem to work.  How do you call this from python?

davisking:
Oh, nevermind, you need to use dlib.full_object_detections.  The pybind11 docstrings aren't super great for this method :/

Vermeille:
Do you need me to change anything?

I have to say I'm using it through https://github.com/ageitgey/face_recognition/pull/547 , so maybe the API is not the best possible, dlib-wise

davisking:
No, it's fine.  I pushed a change that fixed the docstrings.

davisking:
It's just that you need to define types, like dlib.full_object_detections via a call to def() before you use them or pybind11 messes up the docstrings.

davisking:
Although it might be a good idea to allow a pyhton list of lists of shapes to be given, which is what I was doing.  If you wanted to add another overload that worked with that that would probably be a good idea.

Vermeille:
I might need it at some point. When and if that happens, I'll do :)

PR #1341, opened 2018-06-01
Fix bad_alloc exception when loading bmp images with negative height

Palmitoxico:
The 'long' type can be 32 bits or 64 bits depending on what compiler you are using (LLP64 vs LP64). If the long type is 64 bits the biHeight variable will receive a very big number when the bmp image height is negative, causing a std::bad_alloc exception when calling the 'image.set_size' method.

I think that would be a good idea to replace all numeric types with the ones declared in \<cstdint\>(C++11) and get rid of "../uintn.h".

### DIFF
--- a/dlib/image_loader/image_loader.h
+++ b/dlib/image_loader/image_loader.h
@@ -44,7 +44,7 @@ namespace dlib
             unsigned long bfReserved;
             unsigned long biSize;
             unsigned long biWidth;
-            long biHeight;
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the PR :)

Yes, using cstdint would be better.  I don't want to get rid of uintn.h though since it's just going to break people's old code and it's not doing any harm.

PR #1343, opened 2018-06-02
Ignore bmp reserved area

Palmitoxico:
The 4 bytes starting from the 0x06 offset of a bmp file are marked as 'Reserved' but it doesn't guarantee that these bytes are zeroed, some applications uses this area for its own purposes and the output file
still a valid bmp file.

### DIFF
--- a/dlib/image_loader/image_loader.h
+++ b/dlib/image_loader/image_loader.h
@@ -41,7 +41,6 @@ namespace dlib
             unsigned long bytes_read_so_far = 0;
             unsigned long bfSize;
             unsigned long bfOffBits;
-            unsigned long bfReserved;
             unsigned long biSize;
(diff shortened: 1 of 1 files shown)


Palmitoxico:
It failed the Variant=Test job with gcc on travis-ci 'test_timer'. I've no idea how my changes would influence that.

davisking:
That might have been due to some funny loading in travis-ci.  It's testing that timer statements pause for the amount of time they are supposed to.  In rare cases under a lot of system load the tests can fail since the timer pauses for more time than it's supposed to.  I told the test to rerun.  What you changed should definitely have no impact.  So it's probably fine.

Palmitoxico:
That makes sense. Thanks!

davisking:

PR #1377, opened 2018-06-25
Add related works section

c9s:
Hi Davis,

I am working on a gRPC server wrapper that runs dlib as the model serving service, therefore I added the related work section in the readme.  Hope you like it.

Thanks,
c9s


davisking:
Thanks for the PR, however, this isn't the right place to put this kind of content.  You can add this project o the "who uses dlib?" wiki instead: https://sourceforge.net/p/dclib/wiki/Known_users/

PR #1405, opened 2018-07-09
Added static_cast and additional braces to fix compilation warnings.

YuriiOlenych:


davisking:
Thanks for the PR.  

The static_casts are fine, but those extra {} are bizarre.  What compiler generates a warning for code like array<int> a = {1,2,3}; and requires you to double up the {} to make the warning go away?  What is the warning?

YuriiOlenych:
Warning: "suggest braces around initialization of subobject [-Wmissing-braces]".
Please take a look at the example in https://en.cppreference.com/w/cpp/container/array.

davisking:
That page shows = using just {}, not {{}}.

YuriiOlenych:
`std::array<int, 3> a1{ {1, 2, 3} }; // double-braces required in C++11 (not in C++14)`

davisking:
Yes, but that's not what's in the PR.

davisking:
Yes.  It's most definitely legal to say things like `std::array<int, 3> a = {1,2,3};`.

davisking:
This honestly seems like a compiler bug.

YuriiOlenych:
Yeah, probably. Found this: https://stackoverflow.com/questions/31555584/why-is-clang-warning-suggest-braces-around-initialization-of-subobject-wmis
But still, in our project, we are treating all warning as errors.

davisking:
Yeah, what can you do :)

PR #1430, opened 2018-07-26
add check against null pointer

360CodeSafe:
Hi all,
This is Qihoo360 CodeSafe Team, we found some bugs in dlib:

in some situation, such as memory exhausted, 'new LibpngData' will return a null pointer. Then it could cause a derefrence of null pointer in statement 'ld_->png_ptr_ = ...'.
we may need to check whether the return pointer of expression 'new LibpngData' is null or not.


360CodeSafe:
if the return type is not a refrence type, an unnecessary constructor may be called, especially for some basic class.

PR #1432, opened 2018-07-26
add check against null pointer

360CodeSafe:
Hi all,
This is Qihoo360 CodeSafe Team, we found some bugs in dlib:

in some situation, such as memory exhausted, 'new LibpngData' will return a null pointer. Then it could cause a derefrence of null pointer in statement 'ld_->png_ptr_ = ...'.
we may need to check whether the return pointer of expression 'new LibpngData' is null or not.

### DIFF
--- a/dlib/image_loader/png_loader.cpp
+++ b/dlib/image_loader/png_loader.cpp
@@ -110,7 +110,12 @@ namespace dlib
 
     void png_loader::read_image( const char* filename )
     {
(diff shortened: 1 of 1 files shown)


davisking:
This is wrong, new throws std::bad_alloc upon failure.  It doesn't return NULL.  

This also seems like a PR generated by a bot, so if some human is paying attention to this, please don't submit auto-generated pull requests.

PR #1442, opened 2018-08-07
Added `OPENBLAS_LIBS` detection to provide custom OpenBLAS library pa…

RedFox20:
Added `OPENBLAS_LIBS` detection to provide custom OpenBLAS library paths. This enables BLAS cross-compilation with Android and Raspberry.


davisking:
Why can't you do this by setting CMAKE_LIBRARY_PATH or similar CMake environment variables for controlling discovery?  Typically, that is how this kind of thing is done.

RedFox20:
Hi, thank you for taking the time to review this. We ran into a peculiar issue where configuring CMake library path just didn't work and we spent hours trying to get it detected.
Whereas not as ideal, this is a very straightforward parameter for dlib CMakeLists.txt, which will most likely save at least some headache of people less versed with CMake internals :)

teleger:
If you want to uncomment the #define DLIB_USE_BLAS of the config.h file, you can use openblas. I'm going to comment this out, and it reports an error.                                                                               
 error::   matrix_blas_bindings.h:216: error: undefined reference to 'cblas_dgemm'.                                           What can be done to solve this problem?

davisking:
@RedFox20 But how is someone going to know to set OPENBLAS_LIBS?  Anyone using cmake should know about CMAKE_PREFIX_PATH, CMAKE_LIBRARY_PATH, and the like, or can easily discover it via google search.  But I don't see how other people would come to know about this OPENBLAS_LIBS variable.  It's basically just an openblas specific variable that replicates the behavior of CMAKE_LIBRARY_PATH and CMAKE_PREFIX_PATH.

The only way I can see that using the built in cmake variables could fail is if the openblas library wasn't called openblas.  Is that what's happening?

RedFox20:
@davisking I was not able to get CMAKE_LIBRARY_PATH and CMAKE_PREFIX_PATH working for static libs.
However, after testing dlib performance with (BLAS=TRUE, LAPACK=FALSE), I found that it's substantially slower than just relying on dlib SIMD. For example, `dlib_face_recongition_resnet_model_v1` went from 400ms to 2000ms, so 5x slower on Windows desktop, latest MSVC.

Are there any platforms where OpenBLAS provides a substantial speedup?

davisking:
Please answer my question. What is the name of the openblas file? 

Using openblas should be faster, but maybe you compiled it in some non-performant mode.  I don't know.  I would use the IntelMLK on windows.

RedFox20:
Alright, here are the configuration details for OpenBLAS:
```
'NOFORTRAN=TRUE', 'BUILD_SHARED_LIBS=FALSE'
```

And this is the output from dlib CMakeLists.txt. The OPENBLAS_LIBS variable is injected by a **custom** build system. The name for the variable on windows is `openblas.lib` and on Linux it is `libopenblas.a`.

```
    def configure(self):
        self.add_cmake_options(
            'DLIB_NO_GUI_SUPPORT=TRUE',
            'DLIB_ENABLE_ASSERTS=OFF',
            'DLIB_PNG_SUPPORT=OFF',
            'DLIB_JPEG_SUPPORT=OFF',
            'DLIB_GIF_SUPPORT=OFF',
            'DLIB_LINK_WITH_SQLITE3=OFF',
            'DLIB_USE_CUDA=OFF',
            'DLIB_USE_LAPACK=OFF')
        if not self.windows:
            self.add_cl_flags('-Wno-tautological-constant-compare')
        if self.openblas:
            self.add_cmake_options('DLIB_USE_BLAS=ON')
            self.inject_products('dlib', 'OpenBLAS', 'OPENBLAS_INCLUDE', 'OPENBLAS_LIBS')
```

```
-- Found OpenBLAS library
-- Found BLAS: {omitted}/lib/RELWITHDEBINFO/openblas.lib;{omitted}/driver/level2/driver_level2.dir/RelWithDebInfo/driver_level2.lib;{omitted}/driver/level3/driver_level3.dir/RelWithDebInfo/driver_level3.lib;{omitted}/driver/others/driver_others.dir/RelWithDebInfo/driver_others.lib;{omitted}/interface/interface.dir/RelWithDebInfo/interface.lib;{omitted}/kernel/kernel.dir/RelWithDebInfo/kernel.lib
```

davisking:
Ok, so why doesn't cmake's find_package() find openblas.lib?  If you really set CMAKE_PREFIX_PATH to contain the folder with openblas.lib it should find it. You should figure out what's wrong with your setup and use the normal cmake method for finding a library, rather than adding something that's duplicative and less general than find_package().  

So I'm not going to merge this PR unless you can demonstrate that there is some bug in find_package() or show some use case where find_package() fails.  If there really is a bug in cmake and this works around it then that's fine, but I need to see proof of that first.

RedFox20:
Wow ok, never mind then. :)

davisking:
Well, seriously.  You aren't supposed to be using variables like this in modern cmake.

RedFox20:
Adding `CMAKE_LIBRARY_PATH=/home/jorma/Projects/FaceOne/build/OpenBLAS/linux/lib` works and OpenBLAS detects `libopenblas.a`, however it fails to detect `cblas_ddot`
```
-- find_library Found OpenBLAS library: /home/jorma/Projects/FaceOne/build/OpenBLAS/linux/lib/libopenblas.a
-- Looking for sgetrf_
-- Looking for sgetrf_ - not found
-- Looking for cblas_ddot
-- Looking for cblas_ddot - not found
-- BLAS library does not have cblas symbols, so dlib will not use BLAS or LAPACK
```

Even though these symbols exist:
```
nm -C /home/jorma/Projects/FaceOne/build/OpenBLAS/linux/lib/libopenblas.a | grep cblas_ddot
cblas_ddot.c.o:
0000000000000000 T cblas_ddot
```
There is only partial match for `sgetrf_`:
```
nm -C /home/jorma/Projects/FaceOne/build/OpenBLAS/linux/lib/libopenblas.a | grep sgetrf_
0000000000000000 T sgetrf_single
0000000000000000 T sgetrf_parallel
```

So BLAS is still disabled on Linux :/
I moved OPENBLAS_LIBS detection to the end of `find_blas.cmake`:
```
# When all else fails use CMake's built in functions to find BLAS and LAPACK
if (NOT blas_found)
  if (OPENBLAS_LIBS)
    set(cblas_lib ${OPENBLAS_LIBS})
    set(blas_libraries ${cblas_lib})
    set(blas_found 1)
    message(STATUS "OPENBLAS_LIBS Found OpenBLAS library: ${blas_libraries}")
    set(CMAKE_REQUIRED_LIBRARIES ${blas_libraries})
  else()
```
Which forcefully enables BLAS.

Here are some performance results on Linux:
```
// BLAS enabled
$ FaceDetection elapsed: 142.682ms
$ LandmarkDetection elapsed: 80.412ms
$ ResNet elapsed: 89.004ms
```
```
// BLAS disabled
$ FaceDetection elapsed: 147.942ms
$ LandmarkDetection elapsed: 16.689ms
$ ResNet elapsed: 178.217ms
```
Overall, only 30ms speed gain with BLAS enabled on Linux. Do note that I have LAPACK disabled.

And here's the same code running on Windows:
```
// BLAS enabled
$ FaceDetection elapsed: 3.778ms
$ LandmarkDetection elapsed: 3.120ms
$ ResNet elapsed: 2216.690ms
```
```
// BLAS disabled
$ FaceDetection elapsed: 3.826ms
$ LandmarkDetection elapsed: 3.202ms
$ ResNet elapsed: 331.316ms
```
For windows, ResNet really suffers especially with BLAS enabled, while FaceDetection and LandmarkDetection are almost instant. :|
I did notice that letting dlib automatically detect `Accelerate.framework` on `macOS` and `iOS` gives a much more noticeable performance improvement. In the order of 42 seconds --> 1 second for RGB face detection (but I changed that to grayscale now).

So, currently there is no extra motivation from me to merge this pull request, since BLAS speedup is somehow inconsistent.

teleger:
@RedFox20  Thank you for working on the dlib version of android later, which I don't know much about. The use of dlib is just a little bit familiar, I'm just a novice

davisking:
@RedFox20 Thanks for debugging this.   So I just ran some tests myself on Ubuntu 16.04, I used the version of openblas from apt (libopenblas-dev), I also had libblas-dev installed which I think just uses openblas, but I'm not sure since it seems to perform better than whatever libopenblas-dev installs.  

But in any case, both of these versions of blas are much faster than not using blas.  In particular, libblas-dev is about 25 times faster at running the resnet model (the one in dnn_imagenet_ex) than not using blas at all. Using libopenblas-dev it's 12.6x faster than not using blas.  And using the Intel MLK is 37x faster than not using blas.  Moreover, I can tell you with certanty that the dlib blas code that you get if you don't link to a blas library is not going to be faster.  I wrote it and it's not really optimized at all.  It's just there so you get something rather than nothing if you don't link to a blas library.  

So the only way I can see that you could get the numbers you are seeing is if you compiled openblas yourself and accidentally did so without optimizations.  Like if you built it in some unoptimized debug mode.

Also, those checks in cmake for things like cblas_ddot is because sometimes people somehow get copies of blas libraries onto their machines that are missing symbols.  What used to happen without those checks is they would just get linker errors when they tried to use them since those symbols were missing.  These cmake checks prevent this.  I don't know why they are apparently falsely saying your blas is missing symbols.  Maybe it really is missing them, it might be that the code you are compiling doesn't use those symbols.  I've never seen cmake mess up the check for "does this symbol exist".  So I don't know.

RedFox20:
@davisking Thanks for double checking it. Then the current OpenBLAS configuration is definitely wrong. I will investigate it again tonight. I will also take a look at Intel MLK, since you reported better performance with that.

It might be related to OpenBLAS CMake support being partial or incomplete. Or it might be related to having LAPACK disabled.

Regardless, I'll do some more investigation on this and perhaps at least contribute a BLAS readme for dlib at some point.

davisking:
Do you mean you compiled openblas with cmake?  Many cmake scripts have a
default setting of no optimizations enabled.

RedFox20:
@davisking We use a custom build and package management tool which sets RelWithDebInfo by default. I need to dig into OpenBLAS CMakeLists.txt to see what might be going wrong. Perhaps it's due to LAPACK disabled. Did you run with LAPACK enabled? Can you try disabling LAPACK for dlib and see what the results are?

davisking:
It's not LAPACK.  The resnet models don't use LAPACK at all.

PR #1453, opened 2018-08-23
Remove redundant line

xsacha:
With C++17 enabled on VS15.6+, this triggered a nodiscard warning which led to failed builds when warnings as errors is enabled.


davisking:
Cool, thanks for the PR :)

PR #1467, opened 2018-08-31
add python bindings for update_noscale method of the correlation tracker

grynko:
Hi,

I've updated python bindings for correlation tracker a little bit. Now there are two new functions: `update_noscale` and `update_noscale_guess`.

### DIFF
--- a/tools/python/src/correlation_tracker.cpp
+++ b/tools/python/src/correlation_tracker.cpp
@@ -82,6 +82,45 @@ double update_guess (
     }
 }
 
+double update_noscale (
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the PR.  

Can you update the doc strings you used? They don't correctly reflect the behavior of those functions.  I assume you copy and pasted them from the other methods already in the python API.  But these are different functions, they have different behavior.  Right now the doc strings are wrong.  

Also please put some input validation on those functions to catch precondition violation.  I.e. get_position().is_empty() == false

PR #1488, opened 2018-09-20
Python Image and Rectangle Scaling

samuarl:
Added a python function `scale_image` that uses dlib `resize_image` to do scaling. Not sure if this is OK as its own function or rather should just be another signature of `resize_image` binding?

I often have to down scale large images in order to fit into graphics memory for cnn face detection but then want to use the original full size image for pose prediction. I have added `scale_rect` for rectangle and drectangle which accepts a scale factor so that the detected face rect can be rescaled back to fit the face in the original full size image.

### DIFF
--- a/tools/python/src/image2.cpp
+++ b/tools/python/src/image2.cpp
@@ -29,6 +29,20 @@ numpy_image<T> py_resize_image (
 // ----------------------------------------------------------------------------------------
 
 template <typename T>
+numpy_image<T> py_scale_image (
+    const numpy_image<T>& img,
+    double scale
+)
+{
+    DLIB_CASSERT(scale > 0, "Scale factor must be greater than 0");
+    numpy_image<T> out = img;
+    resize_image(scale, out);
+    return out;
+}
+
+// ----------------------------------------------------------------------------------------
+
+template <typename T>
 numpy_image<T> py_equalize_histogram (
     const numpy_image<T>& img
 )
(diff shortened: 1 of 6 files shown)


davisking:
Also be sure to be precise in describing the behavior of functions in the documentation.  Someone should know exactly what something does just by reading the docs.  So saying "scales the image" isn't clear because there are multiple ways to interpret scale.

tools/python/src/image2.cpp
```diff
@@ -28,6 +28,20 @@ numpy_image<T> py_resize_image (
...
 // ----------------------------------------------------------------------------------------
 
+template <typename T>
+numpy_image<T> py_scale_image (
+    const numpy_image<T>& img,
+    double scale
+)
+{
+    DLIB_CASSERT(scale > 0, "Scale factor must be greater than 0");
+    numpy_image<T> out = img;
```
davisking: I'm pretty sure this will result in the input image being modified.  Isn't that what happens?

You need to call the version of resize_image() that takes two images, an in and an out.

davisking:
Looks pretty good! :)

Except for the part I just pointed out.

tools/python/src/image2.cpp
samuarl: When I test in python it works as expected and is not modifying the input.

As I understand it and my intention here is that `out` is a copy of `img` instantiated via the implicit copy constructor of `numpy_image`. The copy `out` is modified when passed to `scale_image` and returned without modifying the original input `img`.

I have little experience with c++ so was hoping to implement the functionality whilst touching as few lines as possible - for good or bad. Happy to give it another shot if you would still prefer it implemented as suggested.

davisking:
Hmm, I guess it works out in this case. That = doesn’t make a deep copy though. It’s just holding a pointer to the original image and that = casts away the constness. 

In this case it ends up being fine since resize_image has to allocate memory for the resized image and so ends up not modifying the input image’s pixels.

davisking:

PR #1494, opened 2018-09-24
build: compilation fix with OpenCV 4

ayounes-nviso:


davisking:
Did you test that this doesn’t break when used with older versions of opencv?

ayounes-nviso:
> Did you test that this doesn’t break when used with older versions of opencv?

No, and it won't work probably because of missing cvIplImage(), I will add some #ifdef

ayounes-nviso:
cvIplImage() was added only in 3.4.3

davisking:
Thanks for the PR.  Is there any documentation somewhere that explains what CV__DISABLE_C_API_CTORS does?  I just googled a bit and the first hit I got was your github issue filed here. There were only 5 hits in google and none very informative.

dlib/CMakeLists.txt
```diff
@@ -784,6 +784,8 @@ if (NOT TARGET dlib)
 
    endif ()  ##### end of if NOT DLIB_ISO_CPP_ONLY ##########################################################
 
+   # OpenCV4 is cleaning up the old C API compat.
+   target_compile_definitions(dlib PRIVATE -DCV__DISABLE_C_API_CTORS)
```
davisking: There are no .cpp files in dlib that #include any opencv code.  So it seems like this cmake statement should have no effect.  Or is there something I'm overlooking?

davisking:
I just tried using this patch with opencv 2.4.9 and got a compiler error.  It appears that the opencv version numbers go EPOCH.MAJOR.MINOR.

ayounes-nviso:
> Thanks for the PR. Is there any documentation somewhere that explains what CV__DISABLE_C_API_CTORS does? I just googled a bit and the first hit I got was your github issue filed here. There were only 5 hits in google and none very informative.

OpenCV 4 will cleanup the C API and probably remove remove all legacy API like IplImage. CV__DISABLE_C_API_CTORS is I guess here to proceed step by step but as I found out the latest code  will not compile without it.

I think my patch is not a proper fix; instead you should get rid of IplImage and use cv::Mat directly. This I guess will break compatibility with OpenCV 2.X

davisking:
I still don't follow.  There are no such thing as destructors in C.  Can you remove that cmake statement since it seems to be a noop?

Also, please adjust the #ifdefs to check the correct opencv version so that users with older setups are not impacted.  I wager you just need to change the word MAJOR to EPOCH. If the solution for maintaining backwards compatibility was ultra hard then that would be one thing, but this seems very simple.

dlib/CMakeLists.txt
davisking: Then in this case it's a noop.  Please remove it from the PR :)

ayounes-nviso:
> I still don't follow. There are no such thing as destructors in C. Can you remove that cmake statement since it seems to be a noop?

The macro is necessary otherwise <opencv2/core/types_c.h> won't compile. I can add it prior to its inclusion.

> Also, please adjust the #ifdefs to check the correct opencv version so that users with older setups are not impacted. I wager you just need to change the word MAJOR to EPOCH. If the solution for maintaining backwards compatibility was ultra hard then that would be one thing, but this seems very simple.

Well, CV_VERSION_EPOCH has disappeared since 3.X so the check shall be something like that:

#if (defined(CV_VERSION_EPOCH) && CV_VERSION_EPOCH == 2) || (CV_VERSION_MAJOR < 4)

ayounes-nviso:
I pulled latest OpenCV master and it turns out that this patch is no more necessary... for now.

jasjuang:
I had to pull the latest OpenCV because there's a critical fix for my application, but it breaks dlib and it was quite frustrating but I stumbled upon this PR that fixes my problem. Thank you so much. IMO I won't be the last one that stumbles into this problem, so I think it's worth ironing out the details on how to make it compatible with older OpenCV.

davisking:
Please file your own issue and fill out the issue template

davisking:
For reference, I just downloaded the latest opencv-4.1.2 and built dlib's webcam example program that uses opencv with it and it built fine.

jasjuang:
@davisking to be clear I pulled the lastest commit on OpenCV instead of the latest release, I think it makes sense to just support the releases, I say we fix it if this problem is still there on the next release.

PR #1505, opened 2018-09-29
add finetuning example

sandsmark:
seems like people request this a lot, so I just quickly hacked together something.

it compiles and seem to work (it should be fast enough to train even on CPU, but I've only tested on my 1080ti), but I'm not sure if this is sane (and there's probably some bugs here, I haven't really used the dlib dnn APIs for anything before), so if you have any feedback it'd be appreciated.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -140,6 +140,7 @@ if (NOT USING_OLD_VISUAL_STUDIO_COMPILER)
    add_example(dnn_imagenet_train_ex)
    add_example(dnn_semantic_segmentation_train_ex)
    add_example(dnn_metric_learning_on_images_ex)
+   add_example(dnn_finetune_ex)
 endif()
 
 
--- a/examples/dnn_finetune_ex.cpp
+++ b/examples/dnn_finetune_ex.cpp
@@ -1,0 +1,329 @@
+// The contents of this file are in the public domain. See LICENSE_FOR_EXAMPLE_PROGRAMS.txt
+/*
+    This example shows how to retrain an existing model.
+
+    Specifically we're going to first run a number of images through the lower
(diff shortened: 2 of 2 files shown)


sandsmark:
the same API/library concepts are covered by existing examples, and everything is in the documentation (otherwise I wouldn't have been able to figure out how to hack together this quickly :-).

but apart from being a compile&run-thing for people that just want to finetune, it also shows how to do custom input layers and chopping up and putting together different networks.

davisking:
Also, another issue is that the example code is very complex.  A good strategy when making examples it to make them as simple as possible.  So to illustrate how to, for instance, replace the last layer in a network with something else you would be better off using a simpler setup, like training on mnist or some other dataset where the labels are less complex and so the related code is less distracting to the reader.

Perhaps a good example would be to stick a multiclass classifier on the end of the face recognition network so that it can predict the identity of the people in the examples/johns folder.  I would also point out that training a linear classifier doesn't need to be done with SGD, and in fact probably shouldn't since SGD is a terrible and slow optimizer.  For learning linear classifiers there are much better algorithms available and I would always encourage someone to use one of them.  Examples are dlib::svm_multiclass_linear_trainer and dlib::svm_c_linear_dcd_trainer.  Those should be the tools of choice for retraining the final layer of a network.  The svm_c_linear_dcd_trainer in particular is blindingly fast and accurate compared to SGD, which allows you to then use a hyper parameter optimizer along with it to automatically find the best setting of the final layer (e.g. http://dlib.net/model_selection_ex.cpp.html)

So another option is to make an example that is more along those lines.  Or even a new tool in dlib that wrapped that code up into a convenient utility function and did it all automatically, accompanied with an example that explained its use.  

As a side note, you don't need to make that new input layer.  The regular input<matrix> layer already does this.

PR #1514, opened 2018-10-05
Apparently fix #1513 by avoiding the cudaStreamSynchronize call

### LINKED ISSUE #1513: Training gets stuck on GeForce RTX 2080 Ti
I'm training semantic-segmentation networks, along the lines of #288.

The code works great on GeForce GTX products, but with new RTX hardware training gets stuck randomly. Looks like a race condition or similar, because the freeze happens after multiple iterations and usually after different number of iterations (from run to run).

Using CUDA 10 and cuDNN 7.3 on 64-bit Windows (but same issue with CUDA 8 and cuDNN 5). Latest `master` from GitHub. MSVC debugger shows the code is waiting [on this line](https://github.com/davisking/dlib/blob/master/dlib/cuda/gpu_data.cpp#L130).

Because steps to reproduce include acquisition of RTX hardware, I'll rather spend time trying to debug the issue than writing a complete set of steps – at least at this point.

reunanen:
resolves #1513

Candidate solution: busy-loop until `cudaStreamQuery` returns `cudaSuccess`

### DIFF
--- a/dlib/cuda/gpu_data.cpp
+++ b/dlib/cuda/gpu_data.cpp
@@ -118,8 +118,10 @@ namespace dlib
         }
     }
 
+#ifdef WIN32
     // This should be pretty much the same as cudaStreamSynchronize, which for some
-    // reason makes training freeze on GeForce RTX 2080 Ti hardware.
+    // reason makes training freeze on some Windows machines.
+    // (see https://github.com/davisking/dlib/issues/1513)
     void synchronize_stream(cudaStream_t stream)
     {
         while (true)
@@ -133,6 +135,7 @@ namespace dlib
             }
         }
     }
+#endif // WIN32
 
     void gpu_data::
     async_copy_to_device() const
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  This is probably fine.  A lot of cuda routines do busy waiting internally anyway, so it's probably what cudaStreamSynchronize() is doing.  

I'll wait a bit to merge it just to make sure you don't run into any more issues.  Let me know after you have use it for a while.  If it fixes the problem and no other issues surface we can merge it :)

reunanen:
@davisking Made it a Windows-only thing.

davisking:
Ok, so it's been a while.  Should I still merge this or have some funny behaviors been discovered?  If nothing new has surfaced I'm going to merge this PR :)

reunanen:
I haven't seen any issues that I would have attributed to this change.

What about you, @davemers0160?

davisking:
That's good enough for me :)

davemers0160:
@reunanen , @davisking I have not seen any negative affects of the change.

xsacha:
The reports (and my experience) only show this issue occuring during training.

Has anyone seen an issue during inferencing caused by this issue or know why it wouldn't be caused?

PR #1537, opened 2018-11-03
Added loss_binary_log_weighted

Raincode:
Hello,

I found the need to train with imbalanced data using a binary loss layer. I found a very easy way to add weights (using the labels), and slightly modified loss_binary_log. Since it worked well for my purposes, I thought perhaps you would like to add it.

My IDE sort of cleaned up the whitespace, I hope that's not a problem (was not really intentional).

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -19,7 +19,7 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-    class loss_binary_hinge_ 
+    class loss_binary_hinge_
     {
     public:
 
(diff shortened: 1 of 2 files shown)


davisking:
Cool, thanks for the PR.  This is a good idea.

There don't need to be two separate classes though. That is, it would be better to just make the change to loss_binary_log as it doesn't break any existing uses.  Also please don't include whitespace changes in the PR.

Raincode:
I'm sorry, I will have to reopen this pull-request. I'm not really good with git yet...

davisking:
No worries :)

PR #1538, opened 2018-11-04
Added weighted labels to loss_binary_log layer.

Raincode:
[Previous PR](https://github.com/davisking/dlib/pull/1537)


davisking:
Sweet, thanks for the PR :)

Raincode:
I like the way you wrote the doc, it's easier to understand.

On a sidenote, it actually worked quite well giving the data which is very common a label less than 1 (e. g. 0.2) in Addition to increasing the weight of the other label.  It further increased Performance in my case, where simply increasing the other label e. g. from -30 to -50 would not improve Performance, if anything make it worse.

PR #1544, opened 2018-11-08
Fixed random_forest_regression.h warnings

Lukas-Buricin:


davisking:
Thanks for the PR.  Something on Appveyor is just broken.  I keeps failing with `The 'pytest' distribution was not found and is required by the application` for Python 3.5 tests.  In any case, it's nothing to do with this PR :)

PR #1557, opened 2018-11-19
Fix #787: find MKL FFT with CMake on Windows

### LINKED ISSUE #787: Found Intel MKL BLAS/LAPACK library but DLIB_USE_MKL_FFT off
Hi,

I try to compile dlib with Intel MKL on Windows but DLIB_USE_MKL_FFT is off. I think there is a problem in [dlib/cmake_utils/cmake_find_blas.txt](https://github.com/davisking/dlib/blob/master/dlib/cmake_utils/cmake_find_blas.txt) file but not sure how to deal with it.

capture

facug91:
Issue #787 was closed, because no PR was made, but the bug in the CMake remained.  
The problem was that on Windows, CMake script was not looking up for mkl headers path. I've just added quite the same code used on UNIX, with different hint paths.

I need this for the tracker module, and there's where I've tried it. I've also compiled and run all tests.

Anyway, there was a test that failed, but it doesn't seem to have anything to do with this change. This is the exact error message:  
```
Error occurred at line 108.  
Error occurred in file c:\dlib\dlib\test\is_same_object.cpp.  
Failing expression was is_same_object(a,b) == false.  
```

### DIFF
--- a/dlib/cmake_utils/find_blas.cmake
+++ b/dlib/cmake_utils/find_blas.cmake
@@ -312,6 +312,20 @@ elseif(WIN32 AND NOT MINGW)
 
    INCLUDE (CheckFunctionExists)
 
+   # Get mkl_include_dir
+   set(mkl_include_search_path
+      "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/mkl/include"
+      "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/compiler/include"
+      "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/compiler/include"
+      "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/mkl/include"
+      "C:/Program Files (x86)/Intel/Composer XE/mkl/include"
+      "C:/Program Files (x86)/Intel/Composer XE/compiler/include"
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  It looks good to me, but did you mean that the `is_same_object` test passed before this change but failed after?  I don't see how that could happen, since it should be unrelated, but if that's what happened then we need to figure out why.

facug91:
Sorry, I was not clear in my message. That error in that test was happening before my change, and I think it has no relation to this problem. That is what I meant.  
Anyway, I supposed there's another problem I might be facing, if I have that test failing.

davisking:
Cool, looks good then. 

That other error is odd though. Are you using visual studio?

facug91:
Yes, VS 2017 version 15.9.2.  
It seems as the compiler is breaking the unit test by optimizing the code.  
I've added two couts to print the address of 'a' and 'b', and then the test pass ok, but breaks in:  
 ```DLIB_TEST( is_same_object(sb,sd) == false) ;```  
If I do the same for this two object, all tests pass correctly.

davisking:
That's weird.  The dlib unit tests are regularly run on VC2017 and the error is not encountered.  Even weirder, that test is just checking that `int a, b; &a != &b;`.  If two different `int` variables somehow have the same address then the compiler is deeply broken.

PR #1561, opened 2018-11-22
Add MKL with TBB support to CMake

facug91:
MKL can be used with TBB for multithreading, instead of OpenMP.  
In this PR I add an option to CMake to link to these libraries, by default in OFF to maintain current behavior.  
  
I've performed a few tests by my own, but I haven't used a powerful machine, where the difference can be higher. I've trained some DNNs (a specific number of epochs only, because it takes too long to finish training on CPU), and on Ubuntu 18.04 seems to be ~8% slower using TBB, while on Windows 10 seems to be the opposite, ~14% faster or even more.  
I wanted to ask you Davis if there're some kind of test you think would be better to benchmark this.

### DIFF
--- a/dlib/cmake_utils/find_blas.cmake
+++ b/dlib/cmake_utils/find_blas.cmake
@@ -334,16 +334,12 @@ elseif(WIN32 AND NOT MINGW)
    # Get mkl_include_dir
    set(mkl_include_search_path
       "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/mkl/include"
-      "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/tbb/include"
       "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/compiler/include"
       "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/mkl/include"
-      "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/tbb/include"
       "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/compiler/include"
       "C:/Program Files (x86)/Intel/Composer XE/mkl/include"
-      "C:/Program Files (x86)/Intel/Composer XE/tbb/include"
       "C:/Program Files (x86)/Intel/Composer XE/compiler/include"
       "C:/Program Files/Intel/Composer XE/mkl/include"
-      "C:/Program Files/Intel/Composer XE/tbb/include"
       "C:/Program Files/Intel/Composer XE/compiler/include"
       )
    find_path(mkl_include_dir mkl_version.h ${mkl_include_search_path})
--- a/dlib/cmake_utils/find_blas.cmake
+++ b/dlib/cmake_utils/find_blas.cmake
@@ -3,17 +3,17 @@
 # information about it at http://www.cmake.org
 #
 #
-# This cmake file tries to find installed BLAS and LAPACK libraries.  
+# This cmake file tries to find installed BLAS and LAPACK libraries.
 # It looks for an installed copy of the Intel MKL library first and then
-# attempts to find some other BLAS and LAPACK libraries if you don't have 
+# attempts to find some other BLAS and LAPACK libraries if you don't have
 # the Intel MKL.
 #
 #  blas_found               - True if BLAS is available
 #  lapack_found             - True if LAPACK is available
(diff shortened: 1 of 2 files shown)


facug91:
Sorry for the whitespaces, some of my IDEs must have added them.  

No, MKLDNN is not what I'm talking about. MKL can use OpenMP (libiomp5md.lib) or TBB (tbb.lib) for thread management. According to Intel, using TBB should increasing performance comparing to OpenMP, that´s why I wanted to try it on dlib.

facug91:
I've performed another test on a desktop computer with an i5-7400 processor and Windows 10, and the execution time was the same, but the usage of CPU was notably lower with TBB. With OpenMP was always at ~99%, and with TBB ~67%.   


> Also, what is the significance of using the MKL while also linking to TBB? Are you talking about the MKLDNN library, as opposed to the MKL?   

If you know different, or consider it is not important, I can stop working on it. But I thought that if we can improve efficiency of dlib in CPU by just changing the link in CMake, it'd be great.

facug91:
As the CPU usage with tbb was low, I've performed the same test, but running two at the same time with each library, and with TBB was ~25% faster.

davisking:
Sounds good.  

Please remove the whitespace changes from the PR though.

dlib/cmake_utils/find_blas.cmake
```diff
@@ -85,6 +85,9 @@ if (UNIX OR MINGW)
          /opt/intel/mkl/lib/intel64
          /opt/intel/lib/intel64
          /opt/intel/mkl/lib
+         /opt/intel/tbb/*/lib/em64t/gcc4.7
+         /opt/intel/tbb/lib/intel64/gcc4.7
```
davisking: Can't this be done in a way that doesn't hard code 4.7?

dlib/cmake_utils/find_blas.cmake
```diff
@@ -94,6 +97,8 @@ if (UNIX OR MINGW)
          /opt/intel/mkl/*/lib/32
          /opt/intel/mkl/lib/ia32
          /opt/intel/lib/ia32
+         /opt/intel/tbb/*/lib/32/gcc4.7
+         /opt/intel/tbb/lib/ia32/gcc4.7
```
davisking: Same here

dlib/cmake_utils/find_blas.cmake
```diff
@@ -287,24 +298,32 @@ elseif(WIN32 AND NOT MINGW)
    if (SIZE_OF_VOID_PTR EQUAL 8)
       set( mkl_search_path
          "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/mkl/lib/intel64" 
+         "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/tbb/lib/intel64/vc14" 
          "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries_*/windows/compiler/lib/intel64" 
-         "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/compiler/lib/intel64" 
          "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/mkl/lib/intel64"
+         "C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/tbb/lib/intel64/vc14"
```
davisking: Is there a way to avoid hard coding versions here too?

facug91:
I thought about that when I made the PR, I've should have added that to the message.  
The folders used by TBB are fixed for some compilers by Intel, as for gcc, they are gcc4.1, gcc4.4, and gcc4.7. Every FindTBB file I've read, add all possible paths to hints, or check the compiler used to see which one to use.  
We could do that, but as dlib need a compiler with C++11 support, I thought it was not necessary, and we could use directly gcc4.7 and vc14. Although in the future, if Intel add other ones, they'll must be added too, by hand.

davisking:
Hmm, ok.  I guess that's fine.  Thanks for the PR :)

dlib/cmake_utils/find_blas.cmake
amincheloh: Not sure about this path, but on my installation is `C:/Program Files (x86)/IntelSWTools/compilers_and_libraries/windows/tbb/lib/intel64/vc_mt`

facug91: You should have both paths, but you must be compiling with /MT flag, am I right?

Issue #1579, opened 2018-12-05
max_pointwise between matrix and scalar


facug91:
I was trying to do that, max_pointwise between a matrix and a scalar, and couldn't find the proper function. I wanted to know if there's a reason for that, or if a PR with the functionality would be ok.

davisking:
There is no function, I would do like `max(max(mymatrix),5.0)`.

davisking:
`lowerbound(mymatrix, 0.5)`

facug91:
That's what I was looking for, thanks!

facug91:
Davis, one more thing. What about making pointwise matrix division? Is there a function to do that, which I'm not seeing?  
By the time, I was able to do it with pointwise_multiple(M1, 1.0 / M2), but I've tried implementing a quick pointwise_divide function, and it seems to perform ~10% faster.

davisking:
Yeah, that sounds cool to me :)

PR #1586, opened 2018-12-08
Add pointwise_divide function

facug91:
As discussed in issue #1579, there was no function on dlib to perform pointwise division of matrices.  
This PR adds that function and corresponding unit tests.  

Here are some performance tests I've ran, all of them performing a pointwise division. The number at the end is the number of arguments used for each test.  
  
With a matrix of size 2016x1512
pointwise_multiply2           17 ms  
pointwise_multiply3           20 ms   
pointwise_multiply4           23 ms   
pointwise_division2           15 ms   
pointwise_division3           18 ms   
pointwise_division4           22 ms   

With a matrix of size 4032x3024
pointwise_multiply2           70 ms   
pointwise_multiply3           81 ms   
pointwise_multiply4           92 ms   
pointwise_division2           62 ms   
pointwise_division3           72 ms   
pointwise_division4           89 ms   

We can see an improvement in performance between ~6% and ~12%.  

Also, I've found a place were pointwise_multiple was used to actually perform a division, in the affine layer, so I've replaced it.

### DIFF
--- a/dlib/matrix/matrix_utilities.h
+++ b/dlib/matrix/matrix_utilities.h
@@ -2772,164 +2772,164 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-	template <typename M1, typename M2>
-	struct op_pointwise_divide : basic_op_mm<M1,M2>
-	{
-		op_pointwise_divide( const M1& m1_,const M2& m2_) : basic_op_mm<M1,M2>(m1_,m2_) {}
+    template <typename M1, typename M2>
+    struct op_pointwise_divide : basic_op_mm<M1,M2>
+    {
+        op_pointwise_divide( const M1& m1_,const M2& m2_) : basic_op_mm<M1,M2>(m1_,m2_) {}
 
-		typedef typename impl::compatible<typename M1::type, typename M2::type>::type type;
-		typedef const type const_ret_type;
-		const static long cost = M1::cost + M2::cost + 1;
+        typedef typename impl::compatible<typename M1::type, typename M2::type>::type type;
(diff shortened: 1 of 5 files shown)


dlib/matrix/matrix_utilities.h
```diff
@@ -2772,6 +2772,165 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+	template <typename M1, typename M2>
+	struct op_pointwise_divide : basic_op_mm<M1,M2>
+	{
+		op_pointwise_divide( const M1& m1_,const M2& m2_) : basic_op_mm<M1,M2>(m1_,m2_) {}
```
davisking: Use spaces instead of tabs.

dlib/test/matrix2.cpp
```diff
@@ -134,6 +134,9 @@ namespace
 
         DLIB_TEST(squared(m4) == pointwise_multiply(m4,m4));
         DLIB_TEST(cubed(m4) == pointwise_multiply(m4,m4,m4));
+        DLIB_TEST(m4 == pointwise_divide(squared(m4),m4));
+        DLIB_TEST(m4 == pointwise_divide(cubed(m4),m4,m4));
```
davisking: Test the overload for 4 arguments as well.

facug91:
I've changed identation to be exactly as in pointwise_multiply, and replace all tabs I've found. Sorry about that.  
Anyway, have you considered to add some kind of code formatter to the project? For example, I've heard of some like clang-format, but I've never used any of them. Maybe adding something like that to dlib would be another good PR? :)

PR #1603, opened 2018-12-21
Find MKL where it actually is based on environment variables

mboisson:


davisking:
This PR has merge conflicts.  You need to fix that before I can be merged into dlib.

PR #1612, opened 2019-01-05
remove register keyword from jdhuff for C++17

tdp2110:
the register storage specifier was deprecated in C++11 and removed in C++17.
The presence of `register` prevents `dlib` from compiling against the C++17 standard.


davisking:
Nice. Thanks for the PR :)

PR #1625, opened 2019-01-18
fixed check for excessive detections in loss_mmod_

davemers0160:
recently ran into the problem where dets.size() was equal to max_num_initial_dets which then throws a subscript out of range error when accesing: dets[max_num_initial_dets].detection_confidence


davemers0160:
Not sure why this is failing the Python 35 checks.

davisking:
Yeah, I’m sure the PR is good :)

Appveyor is just flaky. I reran the test but it’s still failing. I’ll look into it further. But in any case, thanks for the PR.

davisking:
I've been getting lots of funny failures with python 3.5 on appveyor (e.g. missing python packages that are part of setuptools or other stuff that shouldn't be missing).  I think it's just broken on appveyor.  It's also weirdly non-deterministic.  Only seems to happen on python 3.5 though so I'm just removing that test on appveyor.

Anyway, PR looks good.  Thanks :)

PR #1627, opened 2019-01-18
GPU-Ready Dockerfile

enric1994:
Dockerfile with Dlib installed + CUDA and cuDNN. Ready to use on the GPU with Nvidia-Docker

It can dramatically reduce the number of issues regarding the usage on the GPU.

Full example: https://github.com/enric1994/face_recognition_docker_gpu

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM nvidia/cuda:8.0-cudnn5-devel
+
+RUN apt update -y; apt install -y \
+git \
+cmake \
+libsm6 \
+libxext6 \
+libxrender-dev \
+gcc-4.8 \
+g++-4.8 \
+python3 \
+python3-pip
+
+RUN pip3 install \
+scikit-build
+
+RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 50
+RUN update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 50
+
+RUN git clone https://github.com/davisking/dlib.git
+RUN mkdir -p /dlib/build
+
+RUN cmake -H/dlib -B/dlib/build -DDLIB_USE_CUDA=1 -DUSE_AVX_INSTRUCTIONS=1
+RUN cmake --build /dlib/build
+
+RUN cd /dlib; python3 /dlib/setup.py install --yes USE_AVX_INSTRUCTIONS --yes DLIB_USE_CUDA


davisking:
I'm not sure this makes things easier.  A user would have to install nvidia-docker, docker, and understand how to use docker.  It's really easy to install nvidia drivers in ubuntu.

When would someone want to use this docker container and how would they use it?

enric1994:
Docker can very useful in the following situations (even if you don't have GPU or nvidia-docker installed):

1. Try out dlib without the overhead of installing it and their dependencies on your system
2. You have a version of CUDA/cuDNN which is not compatible with dlib
3. Reproduce errors or features in a constrained environment
4. Deploy it on a production environment

Two basic ways to use it:
* If you don't have GPU: 
```
docker build -t dlib .
docker run -it dlib bash
```
* If you do have GPU:
```
docker build -t dlib .
nvidia-docker run -it dlib bash
```
This will create a ready-to-use container with dlib installed, and start a Bash session.

davisking:
I don't understand how this has less overhead installing it.  You still install all the same stuff.  Moreover, dlib without cuda has no dependencies, aside from a C++ compiler. What version of cuda/cudnn isn't compatible with dlib?

My point though is I don't understand who is going to use this docker file. I have a hard time imaging that the sort of person who finds dlib hard to install, considering it's very simple, somehow finds using docker more straightforward. Conversely, I don't know anyone who is familiar with docker who would find the presence of this dockerfile useful in setting up a docker container that happened to use dlib.

dazcona:
I did have some issues installing dlib with CUDA support on Google Cloud Platform. Dlib is an awesome library (thanks very much @davisking), however we spent a lot of time just to get it working. Docker would solve that by virtualizing the installation of dlib that can also be used for deployments in production. I have been using Docker for a few months now and I find myself using it more for collaboration and version control in a team just to work on the same environment 🤖🙂 so thanks both for your work and contributions

davisking:
Yes, I am also a fan of docker. I’m just not sure this fits in the main dlib repo. A more appropriate place might be in a dlib FAQ like “what’s an example of using dlib in docker?”

enric1994:
You are right, all CUDA versions are compatible. The dependency issues are more related with the compiler.

I am seeing quite a number of people that do not find dlib easy to install, specially with CUDA:
There are 400+ issues in this repo related with "install dlib" and in StackOverflow.

In my opinion, considering the amount of issues with the installation, it would be useful to add this to the FAQ but also in the main repo and README.

davisking:
I suspect you can find even more hits on the internet for “how do I install docker?”. 

I also get a lot of questions like “I don’t want to use cmake or a normal compiler. I want to do some silly thing”. Most people won’t want to use docker.  They want to integrate dlib into some other application. 

For example, a lot of people are confused about how to use visual studio and they post questions about dlib that are actually general questions about visual studio. Most of these questions are of that kind, that is, “how do I do this generic programming task? Oh and I happen to be using dlib so I’m asking in a dlib forum”. 

Like honestly, the dlib docs show exactly what to type to get dlib working and it’s not complex. But you need to install a compiler. Docker is one tool among many that lets someone install a compiler.

I don’t want there to be a docker file in the dlib repo because many users will then assume you have to use docker to use dlib, which is crazy. I don’t want to start getting generic “how do I use docker?” questions in addition to all the other generic “how do I use a computer/program” questions I already get. 

A well written FAQ that walks a user though setting up docker and runs some dlib examples in a docker container would be cool though and I’m sure would help some users.

PR #1640, opened 2019-01-26
Make sure all embeddings are compared in dnn_metric_learning_on_images example

ksachdeva:


davisking:
Thanks for the PR, but I'm not sure I understand the changes.  This appears to cause each pair of images to be counted two times.

PR #1654, opened 2019-02-13
Added scikit-build as required by setup

ccoulombe:
Fix for 
```
...
ModuleNotFoundError: No module named 'skbuild'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -232,6 +232,7 @@ setup(
     zip_safe=False,
     tests_require=['pytest==3.8'],
     install_requires=['cmake'],
+    setup_requires=['scikit-build'],
     packages=['dlib'],
     keywords=['dlib', 'Computer Vision', 'Machine Learning'],
     classifiers=[


davisking:
I don't think dlib requires scikit-build.  For example, I don't have scikit-build installed and never have this issue. Why do you think dlib requires this?

ccoulombe:
Sorry, I should have detailed more the issue and where it is coming from.

In `setup.py`, `cmake` is an install dependency in order to install `dlib`: https://github.com/davisking/dlib/blob/41a87e5926935c2328e9057a683c6ba6f9214cc9/setup.py#L234 
from this commit 027e7faf4e92e1e49aeeca34ffecd2280f441030.

This yield:
```
Processing /home/coulombc/dlib
Collecting cmake (from dlib==19.16.99)
  Using cached https://files.pythonhosted.org/packages/72/7b/8d3d83f884283ddfc86ca407d10152d66b84dff5baf5ed4537c127f988cb/cmake-3.13.3.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-wheel-xu8_03mf/cmake/setup.py", line 7, in <module>
        from skbuild import setup
    ModuleNotFoundError: No module named 'skbuild'
```
when you `python setup.py install` or `pip wheel .` from the dlib directory.

`skbuild` comes from `cmake` from PyPI that is imported here: https://github.com/scikit-build/cmake-python-distributions/blob/32412e1c655909af680a79b56964eb9af67b495e/setup.py#L7

But the another issue (the main one actually) is that `cmake.setup.py` has no install requirements and the setup does not require `skbuild` but import it.

davisking:
Huh.  But I still don't understand.  I just ran `python setup.py install` on a machine that doesn't have skbuild installed and it worked. When does it fail?

andrewplesniak:
I ran into the same error when I ran `python setup.py install`.  It was the very end of the compiling process that it crashed. I was able to fix it by doing `pip install scikit-build` and then re-rerunning `python setup.py install`. `Dlib` then able to finish compiling a few seconds later. 

I believe what he is saying is that the current version of `cmake` available via `sudo apt-get install cmake` requires `scikit-build` but it is not listed as a dependency on their end. It appears I am running `cmake version (3.7.2-1)` on my raspberry pi where I got the error. I do not know a whole lot about this stuff as I am still a newbie, but maybe some of the information will help.

davisking:
Sure, I get what is being said.  What is surprising to me is that I don't experience this error when try to install dlib.  This PR also breaks on python 2.7, seemingly, according to the travis ci tests.  If not for that I would probably be fine with merging it.  Although the error on travis-ci doesn't mention scikit-build.  So I'm rerunning it to see if it was some transient travis failure. 

Also, @andrewplesniak, did you have cmake already installed before trying to install dlib?  The reason I added the explicit cmake dependency was to hopefully get pip to automatically pull the dependency.  It does this, but I found that by default, on many, but not all systems, it won't put cmake into the PATH and so the cmake commands run by setup.py fail anyway.  So the other option is to just remove the cmake dependency listed in setup.py since the pip version of cmake is currently sort of busted anyway.

andrewplesniak:
Yes I did have cmake installed on my system because right before compiling dlib I complied OpenCV. However the process I followed to do that was I installed cmake via `sudo apt-get install cmake` which gave me an older version (3.7.2-1). I am thinking that calling setup.py in my virtual environment used pip to install the busted version because I appear to have version (3.13.3) only in my virtual environment where I was trying to install dlib.

davisking:
After thinking about this and playing around a bit more I decided to just remove the cmake dependency all together.  So this puts us back to how dlib was in the last release.  It's up to the user to install cmake.  This seems like the most sensible thing to do given the general flakiness of the current pip version of cmake.  When things improve we can reconsider.

To summarize the current state though, sometimes having pip pull in cmake made installing dlib nice and seamless.  Other times it does nothing, making no attempt to even download cmake, and so still required the user to manually install cmake.  A few times I ran into confusing issues where the pip installed cmake destructively interfered with the system cmake and basically broke my ability to use cmake at all on my system until I went and cleaned it up.  It's totally unacceptable that pip installing dlib might wreck your system, so for now this dependency is removed.

I was also eventually able to replicate the errors in this thread.  I have two Ubuntu 16.04 machines, seemingly identically configured, and one exhibits the error about skbuild being missing and the other does not.  Neither has skbuild installed.  I have no idea why this happens.  My general experiences with pip's dependency handling have always been bad and this is just how it goes I guess :(

Anyway, thanks for bringing this to my attention.  I'm going to close the PR since it's no longer needed, as the cmake dependency is gone for the time being.

PR #1675, opened 2019-03-01
optimization: Add search strategy parameter accessors

jschueller:
in progress, did not run any tests

### DIFF
--- a/dlib/optimization/optimization_search_strategies_abstract.h
+++ b/dlib/optimization/optimization_search_strategies_abstract.h
@@ -51,6 +51,15 @@ namespace dlib
                   this search strategy is used with the line_search() function.
         !*/
 
+        void set_wolfe_rho (
+          double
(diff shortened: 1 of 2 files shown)


dlib/optimization/optimization_search_strategies_abstract.h
```diff
@@ -51,6 +51,15 @@ namespace dlib
                   this search strategy is used with the line_search() function.
         !*/
 
+        void set_wolfe_rho (
+          double
```
davisking: Give arguments names.

davisking: Also, in the docs it's better to say `#get_wolfe_rho() == new_value` rather than to replicate the docs from `get_wolfe_rho`.  This way it's very clear what the correspondence is and avoids documentation string replication.

PR #1704, opened 2019-03-27
Change CUDA version check for cudaSync from 10000 to 10010

facug91:
When upgrading to CUDA 10.1, I've started to have the same problem fixed in #1514.  
This is because the CUDA_VERSION wasn't change in gpu_data.cpp file.  
This pull request simply upgrades version check from 10000 to 10010 (CUDA_VERSION in CUDA 10.1).


davisking:
Sweet, thanks for the PR :)

PR #1717, opened 2019-04-05
Semantic-segmentation loss calculation: fix buffer usage on multi-GPU training

### LINKED ISSUE #1677: Training using data loaders and multiple GPUs
## Expected Behavior

I was trying the `dnn_semantic_segmentation_train_ex`. It works perfectly with 1 GPU, however I can't make it work on multiple GPUs.
According to `dnn_introduction2_ex`, it should be as straightforward as adding a `std::vector<int>` to the constructor of the `dnn_trainer` with the ids of the GPUs to run the training on.
So I changed:
https://github.com/davisking/dlib/blob/master/examples/dnn_semantic_segmentation_train_ex.cpp#L290
to look like:
``` cpp
dnn_trainer<bnet_type> trainer(bnet,sgd(weight_decay, momentum), {0,1});
```

## Current Behavior

I run the training with the modified line above and I get this error:

```
$ ./build/dnn_semantic_segmentation_train_ex VOCdevkit/VOC2012/



SCANNING PASCAL VOC2012 DATASET

images in dataset: 1464
mini-batch size: 23

dnn_trainer details: 
  net_type::num_layers:  239
  net size: 0.0165014MB
  net architecture hash: ca957805a9e35b40ff4481384dc5d66e
  loss: loss_multiclass_log_per_pixel
  synchronization file:                       pascal_voc2012_trainer_state_file.dat
  trainer.get_solvers()[0]:                   sgd: weight_decay=0.0001, momentum=0.9
  learning rate:                              0.1
  learning rate shrink factor:                0.1
  min learning rate:                          1e-05
  iterations without progress threshold:      5000
  test iterations without progress threshold: 500

step#: 0     learning rate: 0.1   average loss: 0            steps without apparent progress: 0
It's what string is: 
Error while calling cudaMemset(loss_cuda_work_buffer, 0, sizeof(float)) in file /home/arrufat/dlib/dlib/cuda/cuda_dlib.cu:1676. code: 11, reason: invalid argument
Aborted (core dumped)
```

I tried the training on multiple GPUs from the `dnn_introduction2_ex` and it works, so I wonder if I had to handle the threads of the data loaders somehow.

## Steps to Reproduce
I just modified that line and run the training.
I also noticed the following behaviour:
- if the ids are: `{1,2}`, GPUs 0, 1, 2 are loaded.
- if the ids are: `{0,1}`, but I launch prepending `CUDA_VISIBLE_DEVICES=1,2`, only GPUs 1 and 2 are loaded.

It's not a big issue, but is that the expected behavior? Maybe I am supposed to index always from 0 and use the `CUDA_VISIBLE_DEVICES` environment variable to set the cards, but I found that a bit odd, nonetheless.

* **Version**: latest commit from master: 02ed083c4c8f2d9c5edb5df52523326d13b9d7ae
* **Where did you get dlib**: I built it from this repo
* **Platform**: x86_64 GNU/Linux
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.4 LTS
Release:        16.04
Codename:       xenial

* **Compiler**: GCC-5.4
* **CUDA**: 9.0

Thanks in advance for all your hard work.

reunanen:
Hopefully resolves #1677.

### DIFF
--- a/dlib/cuda/cuda_dlib.h
+++ b/dlib/cuda/cuda_dlib.h
@@ -423,7 +423,6 @@ namespace dlib
             compute_loss_multiclass_log_per_pixel(
             )
             {
-                work = device_global_buffer();
             }
 
             template <
@@ -439,7 +438,7 @@ namespace dlib
                 const size_t bytes_per_plane = subnetwork_output.nr()*subnetwork_output.nc()*sizeof(uint16_t);
                 // Allocate a cuda buffer to store all the truth images and also one float
                 // for the scalar loss output.
-                cuda_data_void_ptr buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
+                cuda_data_void_ptr buf = device_global_buffer()->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
 
                 cuda_data_void_ptr loss_buf = buf;
                 buf = buf+sizeof(float);
@@ -466,8 +465,7 @@ namespace dlib
                 tensor& gradient,
                 double& loss
             );
-            
-            std::shared_ptr<resizable_cuda_buffer> work;
+
         };
 
     // ------------------------------------------------------------------------------------
--- a/dlib/cuda/cuda_dlib.h
+++ b/dlib/cuda/cuda_dlib.h
@@ -438,7 +438,11 @@ namespace dlib
                 const size_t bytes_per_plane = subnetwork_output.nr()*subnetwork_output.nc()*sizeof(uint16_t);
                 // Allocate a cuda buffer to store all the truth images and also one float
                 // for the scalar loss output.
-                cuda_data_void_ptr buf = device_global_buffer()->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
+                if (!work)
+                {
+                    work = device_global_buffer();
+                }
+                cuda_data_void_ptr buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
 
                 cuda_data_void_ptr loss_buf = buf;
                 buf = buf+sizeof(float);
@@ -465,7 +469,8 @@ namespace dlib
                 tensor& gradient,
                 double& loss
             );
-
+            
+            mutable std::shared_ptr<resizable_cuda_buffer> work;
         };
 
     // ------------------------------------------------------------------------------------


reunanen:
Why this helps (if it does):

On multiple GPUs, we need to have a separate work buffer for each.

(Note that [`device_global_buffer()`](https://github.com/davisking/dlib/blob/9433bfd6ab6a20292354237db0a8672da8eacd6f/dlib/cuda/cuda_data_ptr.cpp#L129) is essentially `thread_local`.)

dlib/cuda/cuda_dlib.h
```diff
@@ -423,7 +423,6 @@ namespace dlib
             compute_loss_multiclass_log_per_pixel(
             )
             {
-                work = device_global_buffer();
```
davisking: Oh yeah, good catch, this is definitely the issue as this pins the buffer to a specific device.

dlib/cuda/cuda_dlib.h
```diff
@@ -439,7 +438,7 @@ namespace dlib
                 const size_t bytes_per_plane = subnetwork_output.nr()*subnetwork_output.nc()*sizeof(uint16_t);
                 // Allocate a cuda buffer to store all the truth images and also one float
                 // for the scalar loss output.
-                cuda_data_void_ptr buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
```
davisking: This can't go here, since once buf goes out of scope the buffer may be deallocated and if do_work() launched any cuda kernels that are still using it things will blow up.  It happens to be the case that do_work() doesn't currently do that, it syncs at the end, but who knows, maybe it will be different in the future.  And in any case, cuda allocations take forever, so we should be careful to avoid them. 

So instead, you could keep the class member and just check if it's initialized here and allocate it if it's not initialized. That would make everything work out properly I think.

reunanen: Ok. Did think about that, but figured (incorrectly it seems) that we would then need to have a separate work buffer for each thread / device (=> a `vector` or `unordered_map` of these `shared_ptr`s, or something like that... which would then need to be protected by a `mutex` as well, if new items could be added simultaneously by multiple threads).

But looking at it again, it seems that there are as many `compute_loss_multiclass_log_per_pixel` objects as there are threads / devices. So the above shouldn't be a problem.

So... just pushed a suggested fix to this problem.

davisking: That’s essentially what device_global_buffer is, a thread and device specific buffer.  It’s held in a weak_ptr so it gets deallocated if no o e is using it in a specific device/thread combination. 

But right, any of the layer objects are only used in one device. 

This class should have a custom copy constructor too so that the work buffer never gets copied to another thread. Otherwise you could get a work buffer from the wrong thread which could be bad. The global_device_buffer method should probably return a unique_ptr to make this kind of mistake less likely now that I’m thinking about the use patterns more.

davisking: Although I guess you can’t use unique_ptr with weak_ptr. Hmm. It would be cool if somehow copying the buffer pointers to other threads was difficult. Not sure how to accomplish that though.

davisking: Need a non-copyable shared pointer 😅

davisking: Anyway, don’t worry about this part in this PR. I’ll do something about it in a bit.

davisking:
Thanks for another great PR :)

PR #1737, opened 2019-04-18
Fixed constness for lapack function declarations

dodomorandi:
This brings const correctness for lapack function parameters, according to the documentation provided by netlib.

~~I was not sure about code formatting and west const/east const, because I found some mixed styles. For the latter, I decided to keep all the changes with east const. Anyway, if you want me to format the code in a different way, it is not a problem.~~ I changed the format style using the _west const_ following the suggestion of @davisking.

P.s.: ironically, I sent a similar patch to armadillo because some lapack functions in dlib were const correct and it created some conflicts armadillo declarations. Now that the patch is in trunk and there is a release candidate, I noticed that exact same problem for dlib.

### DIFF
--- a/dlib/matrix/lapack/gees.h
+++ b/dlib/matrix/lapack/gees.h
@@ -21,15 +21,15 @@ namespace dlib
 
             extern "C"
             {
-                void DLIB_FORTRAN_ID(dgees) (char *jobvs, char *sort, L_fp select, integer *n, 
-                                             double *a, integer *lda, integer *sdim, double *wr, 
-                                             double *wi, double *vs, integer *ldvs, double *work, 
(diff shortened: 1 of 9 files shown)


davisking:
Sweet, thanks for the PR :)

The style in dlib is `const type` rather than `type const`, so change that and the PR is ready to merge.

dodomorandi:
@davisking I start wondering if github sent a notification for my update. Nevertheless, I changed the style for the const. Let me know if it is necessary to do something else :blush:

davisking:
Oh yeah, thanks for the update.  Github didn't email me :/

Anyway, looks good.  Thanks for the PR :)

Issue #354, opened 2016-12-05
input_grayscale_image_pyramid?


grisevg:
It's probably a stupid question, but could you explain why there isn't an `input_grayscale_image_pyramid` that works on `matrix<unsigned_char>` instead of `matrix<rgb_pixel>` in `input_rgb_image_pyramid`?

Does CNN behave better on coloured images than on grayscale?

davisking:
Yes, in many applications color is useful.  Although sometimes it obviously
doesn't matter.  So there isn't any reason why there couldn't be a
grayscale version of the input layer.

grisevg:
So the only practical reason to make `input_grayscale_image_pyramid` would be to safe performance?

davisking:
Right.

grisevg:
Thank you very much for clarifying that

PR #1761, opened 2019-05-09
Add input_grayscale_image_pyramid, issue #354

facug91:
Issue #354  
I'm doing some test with it, I'll be uploading a performance comparison tomorrow using it in the mmod (the training takes several hours).  
In the meantime I wanted to create the pull request, to be sure it's a good contribution, and to start reviewing the code.

### DIFF
--- a/dlib/dnn/input.h
+++ b/dlib/dnn/input.h
@@ -636,6 +636,68 @@ namespace dlib
         }
 
     protected:
+
+        template <typename forward_iterator>
+        void to_tensor_init (
+                forward_iterator ibegin,
+                forward_iterator iend,
+                resizable_tensor& data,
+                unsigned int k
+        ) const
+        {
+
+            DLIB_CASSERT(std::distance(ibegin,iend) > 0);
+            auto nr = ibegin->nr();
+            auto nc = ibegin->nc();
+            // make sure all the input matrices have the same dimensions
+            for (auto i = ibegin; i != iend; ++i)
+            {
+                DLIB_CASSERT(i->nr()==nr && i->nc()==nc,
+                             "\t input_grayscale_image_pyramid::to_tensor()"
+                                     << "\n\t All matrices given to to_tensor() must have the same dimensions."
+                                     << "\n\t nr: " << nr
+                                     << "\n\t nc: " << nc
+                                     << "\n\t i->nr(): " << i->nr()
+                                     << "\n\t i->nc(): " << i->nc()
+                );
+            }
+
+            long NR, NC;
+            pyramid_type pyr;
+            auto& rects = data.annotation().get<std::vector<rectangle>>();
+            impl::compute_tiled_image_pyramid_details(pyr, nr, nc, pyramid_padding, pyramid_outer_padding, rects, NR, NC);
+
+            // initialize data to the right size to contain the stuff in the iterator range.
+            data.set_size(std::distance(ibegin,iend), k, NR, NC);
+
+            // We need to zero the image before doing the pyramid, since the pyramid
+            // creation code doesn't write to all parts of the image.  We also take
+            // care to avoid triggering any device to hosts copies.
+            auto ptr = data.host_write_only();
+            for (size_t i = 0; i < data.size(); ++i)
+                ptr[i] = 0;
+
+        }
+
+        // now build the image pyramid into data.  This does the same thing as
+        // standard create_tiled_pyramid(), except we use the GPU if one is available.
+        void create_tiled_pyramid (
+                const std::vector<rectangle>& rects,
+                resizable_tensor& data
+        ) const
+        {
+            for (size_t i = 1; i < rects.size(); ++i) {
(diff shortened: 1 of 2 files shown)


davisking:
Thanks for the PR.  The code looks correct.  You should use spaces instead of tabs though.  And if you could figure out how to reduce the code duplication that would be good too, like making a base class that contains most of the code and is inherited from.

facug91:
Sorry about the tabs, I've fixed it already.  
About the hierarchy, what I could do is create a base class input_image_pyramid, which implement all methods except `to_tensor`, and all those related to serialization, which would go to the derived classes. Does this seem better for you?  
The problem is that `to_tensor` is a template member function, so I can not specialize it in derived classes.  
One more thing. If the above is good for you, what would be the correct way to document it in the abstract?

davisking:
Sorry for the late reply.  Yeah, that sounds fine. Don't use any virtual functions, you can just publicly inherit from the base class to get all the methods.  As for `to_tensor`, yeah, that's not really sharable.  But their are parts of it that could be trivially moved into protected helper methods in the base class and called from derived classes.

I wouldn't mention this inheritance in the documentation at all.  It's an implementation detail users don't need to know about.

facug91:
With this commit I've create a base clase `input_image_pyramid`, which has the common methods of the derived classes.

facug91:
And now I've refactored to_tensor method, to move repeated code to base class.

facug91:
Some benchmarking I've made:  

| | rgb | grayscale |
|-|------|--------------|
| 640x360 | 18.14 us | 16.05 us |
| 400x300 | 10.31 us | 9.22 us |

It is around 11% faster.

dlib/dnn/input.h
```diff
@@ -591,30 +591,13 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 
     template <typename PYRAMID_TYPE>
-    class input_rgb_image_pyramid
+    class input_image_pyramid
```
davisking: This should probably go in a `detail` namespace to discourage users from touching it.

dlib/dnn/input.h
```diff
@@ -623,8 +606,8 @@ namespace dlib
         void set_pyramid_outer_padding (unsigned long value) { pyramid_outer_padding = value; }
 
         bool image_contained_point (
-            const tensor& data,
-            const point& p
+                const tensor& data,
```
davisking: Fix tabbing here and other places.  Should be 4 spaces.

dlib/dnn/input.h
```diff
@@ -633,45 +616,49 @@ namespace dlib
...
             for (auto i = ibegin; i != iend; ++i)
             {
                 DLIB_CASSERT(i->nr()==nr && i->nc()==nc,
-                    "\t input_rgb_image_pyramid::to_tensor()"
-                    << "\n\t All matrices given to to_tensor() must have the same dimensions."
-                    << "\n\t nr: " << nr
-                    << "\n\t nc: " << nc
-                    << "\n\t i->nr(): " << i->nr()
-                    << "\n\t i->nc(): " << i->nc()
+                             "\t input_grayscale_image_pyramid::to_tensor()"
```
davisking: `input_image_pyramid`

dlib/dnn/input.h
```diff
@@ -690,11 +677,158 @@ namespace dlib
...
+        friend void to_xml(const input_grayscale_image_pyramid& item, std::ostream& out)
+        {
+            out << "<input_grayscale_image_pyramid"
+                <<"' pyramid_padding='"<<item.pyramid_padding
+                <<"' pyramid_outer_padding='"<<item.pyramid_outer_padding
+                <<"'/>";
+        }
+
+    private:
+        using input_image_pyramid<PYRAMID_TYPE>::pyramid_padding;
```
davisking: Is this needed?  Do you get an error in the friend functions without this?

dlib/dnn/input_abstract.h
```diff
@@ -271,6 +271,155 @@ namespace dlib
...
+        >
+    class input_grayscale_image_pyramid
+    {
+        /*!
+            REQUIREMENTS ON PYRAMID_TYPE
+                PYRAMID_TYPE must be an instance of the dlib::pyramid_down template.
+
+            WHAT THIS OBJECT REPRESENTS
+                This input layer works with gray scale images of type matrix<unsigned char>.
+                It is identical to input except that it outputs a tensor containing a tiled
```
davisking: Say "identical to the input layer except..."

dlib/dnn/input.h
facug91: I think it used to throw me a compilation error, but maybe the problem was somewhere else, and I thought it was that. I removed it and it's still working.

dlib/dnn/input_abstract.h
facug91: Sorry, but I don't understand what you mean.

davisking: I just mean to replace "It is identical to input except that it outputs a tensor" with "It is identical to the input layer except that it outputs a tensor"

davisking:

PR #1763, opened 2019-05-13
more pedantic silencing

arrufat:
Sorry, I didn't catch these with my previous commit (now I've rebuilt from scratch all the examples)


arrufat:
I'm not sure about this second commit, if you don't like it, I will revert it.

dlib/dnn/layers.h
```diff
@@ -2241,12 +2241,14 @@ namespace dlib
             if (version != "add_prev_")
                 throw serialization_error("Unexpected version '"+version+"' found while deserializing dlib::add_prev_.");
         }
-
+        #pragma GCC diagnostic push
+        #pragma GCC diagnostic ignored "-Wunused-parameter"
```
davisking: This is a pretty heavy solution.  How about just saying `friend std::ostream& operator<<(std::ostream& out, const add_prev_& )`?

arrufat: Yes, you're right, that works better :) thank you!

davisking:

arrufat:
my pleasure :)

Issue #1794, opened 2019-06-07
Setting "y" of a dlib.point changes "x" instead


jtoss:
Using the python API of dlib I noticed that changing the "y" property of a dlib.point changes its "x" property instead. 

Python 3.6.6 
```
dlib.__version__
'19.17.0'
```

## Steps to Reproduce
```
>>> import dlib
>>> p = dlib.point(0,0)
>>> p 
point(0, 0)
>>> p.x = 1
>>> p
point(1, 0)
>>> p.y = 2
>>> p
point(2, 0)
```

jotapem:
Could reproduce it with Python 3.5.2 and Dlib 19.16.0

PR #1795, opened 2019-06-08
 Fix setting a point's y coordinate changes x instead (Python bindings)

mchelem:
* Add test to reproduce issue #1794 
* Fix python bindings


davisking:
Thanks for reporting and fixing this :)

Issue #1828, opened 2019-07-17
load_image documentation misleading


pauldreik:
The documentation for load_image (as of 20190717):
http://dlib.net/imaging.html#load_image

says the extension of the filename is used to use the right routine to parse the file. However, the implementation ignores the filename and selects the routine by [inspecting the contents of the file.](https://github.com/davisking/dlib/blob/2704f3e26df4e7deadd92174b659e54be89bc0dd/dlib/image_loader/load_image.h#L74)

Current documentation:
This global function takes a file name, looks at its extension, and then loads it into an array2d of pixels using the appropriate image loading routine. The supported types are BMP, PNG, JPEG, GIF, and the dlib DNG file format. 

Proposed:
This global function takes a file name and loads it into an array2d of pixels using the appropriate image loading routine. The supported types are BMP, PNG, JPEG, GIF, and the dlib DNG file format.

davisking:
Indeed, that would be better. Feel free to submit this stuff as a pull request rather than an issue. You don’t have to ask permission to make a pull request :)

pauldreik:
There you go :-)

PR #1835, opened 2019-07-18
fix mismatch between documentation and implementation of load_image

pauldreik:
This should close https://github.com/davisking/dlib/issues/1828


davisking:
Looks good, thanks for fixing this :)

Issue #1842, opened 2019-07-26
pixel value assignment outside range


pauldreik:
Some pixel values cause the hue to be between -90 and 0 degrees, but it is expected to be within 0 and 360. This causes undefined behaviour when converting to the range of an unsigned int.

https://github.com/davisking/dlib/blob/48b9bf3f93f47f6173e101bc81343d55eedf242d/dlib/pixel.h#L1226

This was found through fuzzing.

pauldreik:
still not fixed

Issue #1841, opened 2019-07-26
integer overflow in load_bmp


pauldreik:
In case biHeight is INT_MIN, negating it causes a signed integer overflow which is undefined behaviour.

https://github.com/davisking/dlib/blob/48b9bf3f93f47f6173e101bc81343d55eedf242d/dlib/image_loader/image_loader.h#L107

This was found through fuzzing.

pauldreik:
prevent auto close

Issue #1840, opened 2019-07-26
load_image indirectly uses uninitialised value


pauldreik:
Running a short image file through load_image() makes the program use unitialized values, which is undefined behaviour.

I happened to see this in the source, and verified it.

The problem is that the buffer https://github.com/davisking/dlib/blob/48b9bf3f93f47f6173e101bc81343d55eedf242d/dlib/image_loader/load_image.h#L37 is filled with 0 to 8 bytes by the file reading, then the elements are referenced a bit further down.

To reproduce:
- build in debug mode
- create a short file, for instance with ```echo -e B >short.bin``` or ```cat /dev/null >empty.bin```
-invoke ```load_image("short.bin")``` through valgrind

The following is output by valgrind:
```
paul@tonfisk:~/code/delaktig/dlib/dlib/test/build_normal$ valgrind ./dtest -d  --test_image_load short.bin 
==22027== Memcheck, a memory error detector
==22027== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==22027== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==22027== Command: ./dtest -d --test_image_load short.bin
==22027== 
Running test_image_load / ==22027== Conditional jump or move depends on uninitialised value(s)
==22027==    at 0x1C5EFC: dlib::image_file_type::read_type(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (load_image.h:48)
==22027==    by 0x1C6E63: void dlib::load_image<dlib::array2d<dlib::rgb_pixel, dlib::memory_manager_stateless_kernel_1<char> > >(dlib::array2d<dlib::rgb_pixel, dlib::memory_manager_stateless_kernel_1<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (load_image.h:74)
==22027==    by 0x1C4464: (anonymous namespace)::image_load_tester::invoke_on_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (image_load.cpp:58)
==22027==    by 0x1C424C: (anonymous namespace)::image_load_tester::invoke_on_file_or_dir(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (image_load.cpp:44)
[...]

Issue #1830, opened 2019-07-17
integer overflow in array2d_kernel.h set_size


pauldreik:
I fuzzed dlib::load_image and found this with undefined behaviour sanitizer:
nr_*nc_ overflows.

https://github.com/davisking/dlib/blob/2704f3e26df4e7deadd92174b659e54be89bc0dd/dlib/array2d/array2d_kernel.h#L492

How to reproduce:
 - compile with undefined behaviour sanitizer (-fsanitize=undefined) on clang or gcc
 - feed a [bad image](https://github.com/pauldreik/dlib/blob/fuzz/dlib/test/images/problematic/integer_overflow_3d4e1dc8.txt) to image_load

You can do this by checking out the [fuzz branch in my fork](https://github.com/pauldreik/dlib/tree/fuzz), build the unit tests, and executing
```./dtest -d --test_image_load ../images/problematic/integer_overflow_3d4e1dc8.txt```

Here is an example output:
```
paul@tonfisk:~/code/delaktig/dlib/dlib/test/build$ ./dtest -d --test_image_load ../images/problematic/integer_overflow_3d4e1dc8.txt 
Running test_image_load / /home/paul/code/delaktig/dlib/dlib/../dlib/image_loader/../image_transforms/../statistics/../matrix/../array2d/array2d_kernel.h:509:47: runtime error: signed integer overflow: 4280353023 * 4280352826 cannot be represented in type 'long int'
                                                                               

Testing Finished
Total number of individual testing statements executed: 0
All tests completed successfully

```

I made a simple check for large values by summing the integer logarithms of nr and nc to verify it's reasonable before multiplying and allocating. Are you interested in that fix?

davisking:
Sure, a PR that converts this into a proper exception would be welcome. :)

pauldreik:
This is now included in https://github.com/davisking/dlib/pull/1834

pauldreik:
just fending off autoclose

PR #1843, opened 2019-07-26
fix various issues found during fuzzing

pauldreik:
The fuzzers are not going to be accepted into dlib as presented in [the pull request](https://github.com/davisking/dlib/pull/1834). The bug fixes were asked to be submitted separately (https://github.com/davisking/dlib/pull/1834#issuecomment-513504883).

To get separate issues to refer and link to I created separate issues. All these are fixed in this pull request:
- #1841 integer overflow
- #1842 assignment outside range
- #1840 use of uninitialized value
- #1830 integer overflow

There is a new unit test and directory with problematic files, to make it easier to prove the presence of bugs and that they are fixed. Not sure how to get it running in CI, however.


dlib/test/image_load.cpp
```diff
@@ -0,0 +1,77 @@
...
+            for(const auto& file: files) {
+                invoke_on_file(file);
+            }
+            }
+        }
+        void invoke_on_file(const std::string& filename) {
+            print_spinner();
+            dlog << dlib::LINFO << "attempting to parse image "<<filename;
+
+                   dlib::array2d<dlib::rgb_pixel> img;
```
davisking: Please fix tabbing

dlib/test/image_load.cpp
```diff
@@ -0,0 +1,77 @@
...
+           print_spinner();
+
+           // recursively load files
+           invoke_on_file_or_dir(m_arg);
+        }
+     private:
+        void invoke_on_file_or_dir(const std::string& dir_or_file) {
+            if(dlib::file_exists(dir_or_file)) {
+                invoke_on_file(dir_or_file);
+            } else {
```
davisking: Use dlib's brace style so code is consistent with rest of codebase.

davisking: Also make sure the indenting is correct everywhere.

dlib/test/image_load.cpp
```diff
@@ -0,0 +1,77 @@
...
+        void invoke_on_file(const std::string& filename) {
+            print_spinner();
+            dlog << dlib::LINFO << "attempting to parse image "<<filename;
+
+                   dlib::array2d<dlib::rgb_pixel> img;
+                   try {
+                       dlib::load_image(img, filename);
+                       dlog << dlib::LINFO << "successfully parsed image "<<filename;
+                   } catch(std::exception& e) {
+                       dlog << dlib::LERROR << "got an exception while loading image "<<filename<<": "<<e.what();
```
davisking: Shouldn't we be putting DLIB_TESTs here about "yes, this image should have failed or no it shouldn't have failed to load"?

dlib/test/pixel.cpp
```diff
@@ -753,6 +753,17 @@ namespace
...
 
+        {
+            // bad pixel example, found through fuzzing
+            rgb_pixel src;
+            src.red=246;
+            src.green=170;
+            src.blue=245;
+            hsi_pixel dest;
+            assign_pixel(dest,src);
+            DLIB_TEST(dest.h>=0);
```
davisking: These tests are always true, since `hsi_pixel` holds `unsigned char` typed fields.  How could this test have failed?

dlib/test/images/README.txt
```diff
@@ -0,0 +1,5 @@
+This directory contains images that are problematic to parse,
+for instance because they contain errors.
+
+They are here, to be able to prove bugs, which may have been fixed.
```
davisking: Rename `dlib/test/images/problematic/integer_overflow_3d4e1dc8.txt ` to something other than a `.txt` file

dlib/internal/integer_utils.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <cstdint>
+#include <limits>
+
+namespace dlib
+{
+namespace detail {
+long safe_multiply(long a, long b) {
+    long ret;
+    bool overflow=false;
+#ifdef DLIB_HAS_BUILTIN_MUL_OVERFLOW
```
davisking: This should be done with a `#ifdef` check for the compiler version, rather than using a macro and cmake rule.  This is to simplify the build system.  Not everyone uses cmake.  You would also have to update the cmake install scripts if we really went with `DLIB_HAS_BUILTIN_MUL_OVERFLOW` so that `config.h` is properly handled as well.  But this sort of thing could be done more simply using `#ifdef such and such gcc or clang version` style

dlib/internal/integer_utils.cpp
```diff
@@ -0,0 +1,42 @@
...
+        auto tmp=large{a}*b;
+        if(tmp<=std::numeric_limits<small>::max() && tmp>=std::numeric_limits<small>::min()) {
+            overflow=false;
+            ret=tmp;
+        } else {
+            overflow=true;
+        }
+    } else {
+       // unsupported compiler
+       throw std::logic_error("signed integer overflow check not implemented on this platform");
```
davisking: This code must be made to work everywhere. 

I think you should be able to do all this much more simply, by essentially doing:
```
result = a*b;
if (result/b != a) then there was overflow
```
You have to do it in unsigned arithmetic since signed overflow is undefined behavior in C++.  But you can accomplish this here by converting the signed integers into unsigned, doing the multiply, and converting it back to signed, multiplying by -1 if needed, and checking that the conversion to signed is equal to the unsigned value (accounting for the -1).

dlib/pixel.h
```diff
@@ -809,8 +809,12 @@ namespace dlib
                 c2.s = delta / (c2.l < 0.5 ? (2*c2.l) : (2-2*c2.l));
             c2.h = 0;
             if (delta > 0) {
-                if (themax == c1.r && themax != c1.g)
+                if (themax == c1.r && themax != c1.g) {
                     c2.h += (c1.g - c1.b) / delta;
+                    if(c1.g < c1.b) {
+                        c2.h += 6;
```
davisking: Why + 6?

Also, please use the coding style used in the rest of dlib, so { on a new line.

pauldreik:
Would it be possible that you provide a .clang-format file? Manual formatting is not fun. Or do you have some other way of keeping the format (like using eclipse builtin formatter or so)?

dlib/pixel.h
pauldreik: Because it splits 360 degrees in six parts, it is clear if you inspect the surrounding code.

dlib/test/pixel.cpp
pauldreik: It didn't, it's just there so the interesting event is actually triggered by unit test. Good for debugging, and good for keeping code coverage up and asserting what the function promises.

davisking: I'm still confused about this.  Does something different happen if `DLIB_TEST(dest.h>=0);` is removed?

davisking: Because it seems that it should be removed, as it can't fail and it suggests to the reader that somehow `dest.h` could be negative, but that is impossible.

dlib/test/pixel.cpp
```diff
@@ -753,6 +753,17 @@ namespace
...
+            DLIB_TEST(dest.h<=360);
```
reunanen: Surely an `unsigned char` can't be >360 either?

Issue #1844, opened 2019-07-26
outside bounds access during xml parsing


pauldreik:
Fuzzing dlib::xml_parser I found a crash after 177 milliseconds. It's a string access outside bounds in [this function](https://github.com/davisking/dlib/blob/48b9bf3f93f47f6173e101bc81343d55eedf242d/dlib/xml_parser/xml_parser_kernel_1.h#L1204)

I will submit a pull request with the fix (replacing str[i] with str.at(i)).

pauldreik:
The crashing input was:
```cpp
 "<?>\0"
```
but it can probably be minimized more.

davisking:
Thanks for finding this.  Don't replace it with `str.at(i)` though.  That will break the contract for this method, and for the parser as a whole.  In this case, you should return -1, as this is what the contract in the header for this method indicates should happen in such situations.

pauldreik:
I guess I will have to spend more than 30 s on the fix :-)

PR #1845, opened 2019-07-26
fix xml parser read outside bounds #1844

pauldreik:
This fix made the parser survive more than 100 million fuzzer iterations, so I think it works.

https://github.com/davisking/dlib/issues/1844


pauldreik:
still not fixed

PR #1848, opened 2019-07-29
pybind11: cmake: ignore the check between host-python and cross-compiler

notoriou1pig:
When dlib is compiling, cmake will compare python architecture and target
architecture. So in cross-compiling case, it is irrevelant because host and
target architecture often differs. The main problem come from checking python
architecture on host and not on target.

Here is an error when compiling dlib from x86_64 to arm 32-bit target :
```
Python config failure: Python is 64-bit, chosen compiler is 32-bit
```

So :
- Skipping the comparation when cross-compiling is enabled.

Signed-off-by: Romain Naour <romain.naour@smile.fr>
Signed-off-by: Alexandre PAYEN <alexandre.payen@smile.fr>


davisking:
Thanks for the PR :)

PR #1855, opened 2019-08-08
Fix #1849 by resetting the cuda_compute object when loss increases

### LINKED ISSUE #1849: Semantic Segmentation example crashes with multiple GPUs and loss increase
Hi, I'm using dlib from latest git revision and the semantic segmentation example crashes when used with multiple GPUs and the loss increases, so the trainer reloads from a previous state.
I suspect this might be related to https://github.com/davisking/dlib/issues/1677.
As with that issue, the only modification I made to the example code was adding the GPU initializer list to the `dnn_trainer`.

Here's the exact error:
```
step#: 64369  learning rate: 0.1   average loss: 0.424571     steps without apparent progress: 7989
Loss has been increasing, reloading saved state from pascal_voc2012_trainer_state_file.dat_
It's what string is: 
Error while calling cudaMemset(loss_cuda_work_buffer, 0, sizeof(float)) in file ../dlib/cuda/cuda_dlib.cu:1676. code: 11, reason: invalid argument
```

I suspect this might be hard to reproduce on other people's machines, but I have saved the state files when reloaded from makes this happen, so if you come up with a fix, I can try it easily.

Thanks in advance.

reunanen:
(= when the net is deserialized from disk)

Appears to fix the crash at least on my machine. @arrufat , want to try?

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -2581,12 +2581,17 @@ namespace dlib
             serialize("loss_multiclass_log_per_pixel_", out);
         }
 
-        friend void deserialize(loss_multiclass_log_per_pixel_& , std::istream& in)
+        friend void deserialize(loss_multiclass_log_per_pixel_& item, std::istream& in)
         {
             std::string version;
             deserialize(version, in);
             if (version != "loss_multiclass_log_per_pixel_")
                 throw serialization_error("Unexpected version found while deserializing dlib::loss_multiclass_log_per_pixel_.");
+
+#ifdef DLIB_USE_CUDA
+            // Reset the cuda_compute object to fix https://github.com/davisking/dlib/issues/1849
+            item.cuda_compute = cuda::compute_loss_multiclass_log_per_pixel();
+#endif
         }
 
         friend std::ostream& operator<<(std::ostream& out, const loss_multiclass_log_per_pixel_& )


arrufat:
I might not be able to try it in the following week. I'll try then. Thank you so much for looking into this :)

arrufat:
@reunanen, I've just tried it and it seems to be working, thank you so much! :tada:

dlib/dnn/loss.h
```diff
@@ -2581,12 +2581,17 @@ namespace dlib
...
+        friend void deserialize(loss_multiclass_log_per_pixel_& item, std::istream& in)
         {
             std::string version;
             deserialize(version, in);
             if (version != "loss_multiclass_log_per_pixel_")
                 throw serialization_error("Unexpected version found while deserializing dlib::loss_multiclass_log_per_pixel_.");
+
+#ifdef DLIB_USE_CUDA
+            // Reset the cuda_compute object to fix https://github.com/davisking/dlib/issues/1849
+            item.cuda_compute = cuda::compute_loss_multiclass_log_per_pixel();
```
davisking: This seems a little funny to me that it would solve this.  Over in the `compute_loss_multiclass_log_per_pixel` class we do `cuda_data_void_ptr buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));`, and use that `buf` and let it get destructed.  We do hold onto `work` though. However, if someone else on the same device and thread called `work->get()` they could potentially cause the buffer to reallocate, causing the thing pointed to by `buf` to be given up by `work`, and since `compute_loss_multiclass_log_per_pixel` isn't holding `buf` anymore `buf`'s memory block would get deallocated.  But if the async kernel is still using it that would be bad.  Maybe that's what's happening here.  So if you make `compute_loss_multiclass_log_per_pixel` hold onto `buf` like it holds onto `work` does that make this go away?

reunanen: You mean something like [this](https://github.com/reunanen/dlib/commit/3922729851f353c03b516d30431b2532c74c0e6c)?

No, that doesn't seem to help, at least on my machine. But maybe I misunderstood?

arrufat: @reunanen I confirm that does not work either on my machine.

davisking: No, that's not exactly what I mean.`DLIB_ASSERT` is only on if you explicitly activated it and I suspect if those were `DLIB_CASSERT` (which is always on) those asserts would be failing. 

I mean make only 2 lines of code different.  Put `cuda_data_void_ptr buf;` as a member variable.  Then change 
```
cuda_data_void_ptr buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
```
to
```
buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
```

reunanen: So like [this](https://github.com/reunanen/dlib/commit/318c02c15b237fa65ecef51ba5d0dc3ed1430bb4). (?)

Well, the result is the same as previously. At least superficially, the crash looks identical. Didn't go to any depth this time, though.

I admit that my previous change was overly complex, however I think it should have been essentially equivalent.

davisking: Yeah, that’s what I meant. Huh. I’ll look into this a little it more. But maybe this is working around a cuda bug?  I’m surprised the code here doesn’t fix it while the code in the PR does.

davisking: Ah, so I think I see the error.  Or at least one part of the error.  So network objects are supposed to be copyable in the usual way, e.g. `net1 = net2`.  This is however broken here in `compute_loss_multiclass_log_per_pixel` since its constructor does `work = device_global_buffer();` which puts a device specific buffer in the classe's state.  But when you go and copy a network to another network, as happens incidentally during deserialization when the trainer reloads all the networks, you run into the problem of all the networks using buffers for device 0.  

I think the right solution is to move the `work = device_global_buffer();` from the constructor to inside `operator()`.  There isn't anything particularly expensive about calling `device_global_buffer()`.  The only reason to save `work` (and `buf` really as well) is to ensure they have lifetimes as long as the cuda kernel launch that happens there.

davisking: So does doing that fix the bug here?  (where "that" is put `work = device_global_buffer();` in `operator()` rather than the constructor of `compute_loss_multiclass_log_per_pixel` and also make `buf` a member variable as well)

reunanen: @davisking, sorry but I don't follow. Isn't [it](https://github.com/davisking/dlib/blob/b02be8c734c55fc9de53edff8078376fea9c66fb/dlib/cuda/cuda_dlib.h#L443) already in `operator()`, and not in the constructor?

reunanen: Should there be copy constructors and assignment operators that leave `work` of the copy uninitialized?

davisking: Oops, sorry, I was looking at a branch of dlib with older code.  Yes, it's in the operator() already.  Just change 
```
if (!work)
{   
    work = device_global_buffer();
}   
```
to simply
```
work = device_global_buffer();
```

`device_global_buffer()` should be plenty fast so there isn't any need to avoid calling it here.

reunanen: Yeah, that fixes the problem too, at least on my machine. 👍 

Will create a new PR. Guessing we can abandon this one.

davisking: Sweet, thanks for debugging this :)

arrufat: Thank you so much again to both of you for looking into this :)

davisking:
No problem.  Thanks for reporting the bug :)

reunanen:
This PR was made obsolete by #1862.

PR #1862, opened 2019-08-17
Fix #1849 by calling device_global_buffer() unconditionally

### LINKED ISSUE #1849: Semantic Segmentation example crashes with multiple GPUs and loss increase

reunanen:
Also made `buf` a member variable.

### DIFF
--- a/dlib/cuda/cuda_dlib.h
+++ b/dlib/cuda/cuda_dlib.h
@@ -438,15 +438,11 @@ namespace dlib
                 const size_t bytes_per_plane = subnetwork_output.nr()*subnetwork_output.nc()*sizeof(uint16_t);
                 // Allocate a cuda buffer to store all the truth images and also one float
                 // for the scalar loss output.
-                if (!work)
-                {
-                    work = device_global_buffer();
-                }
+                work = device_global_buffer();
                 buf = work->get(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
 
                 cuda_data_void_ptr loss_buf = buf;
                 buf = buf+sizeof(float);
-
 
                 // copy the truth data into a cuda buffer.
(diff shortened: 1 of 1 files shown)


davisking:
Sweet, thanks for fixing this :)

davisking:
This whole thing has mad me realize the device global buffer API was too complex and error prone. I just pushed a diff that makes it a lot simpler.  Now you just do:
```
cuda_data_void_pointer buf = device_global_buffer(size);
```
I tested everything locally and it seems fine. Can you grab the latest dlib and use that in your testing as well?  

In retrospect, I'm not sure why I didn't write it this way in the first place :|

reunanen:
Yeah, even the latest master works on my test machine (2 * GTX 1080 (non-Ti)).
(At least in this respect. Didn't test anything else for the time being.)

davisking:
Sweet

PR #1863, opened 2019-08-19
add loss_mean_squared_per_channel

arrufat:
This pull request adds support for a loss that computes independently on each plan of the output tensor.
This loss is useful to estimate keypoints (for pose estimation, etc).
The article that motivated me to add this kind of loss was [Simple Baselines for Human Pose Estimation and Tracking](https://arxiv.org/abs/1804.06208).
I have successfully trained a network that replicates that article with this loss.

As a unit test, I added a small example that takes a matrix with 9 white dots as an input and outputs 9 channels, each one with a white dot if there was a dot at the associated position.

Sorry for the delay in making this PR... I couldn't think of a simple example (that can actually be used to learn how to use the loss)

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -2891,7 +2891,149 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-    class loss_dot_ 
+    template<long _num_channels>
+    class loss_mean_squared_per_channel_
+    {
+    public:
+
+        typedef std::array<matrix<float>, _num_channels> training_label_type;
+        typedef std::array<matrix<float>, _num_channels> output_label_type;
+
+        template <
+            typename SUB_TYPE,
+            typename label_iterator
+            >
+        void to_label (
+            const tensor& input_tensor,
+            const SUB_TYPE& sub,
+            label_iterator iter
+        ) const
+        {
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+
+            const tensor& output_tensor = sub.get_output();
+
+            DLIB_CASSERT(output_tensor.k() == _num_channels, "output k = " << output_tensor.k());
+            DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+
+            const float* out_data = output_tensor.host();
+
+            for (long i = 0; i < output_tensor.num_samples(); ++i, ++iter)
+            {
+                for (long k = 0; k < output_tensor.k(); ++k)
+                {
+                    (*iter)[k].set_size(output_tensor.nr(), output_tensor.nc());
+                    for (long r = 0; r < output_tensor.nr(); ++r)
+                    {
+                        for (long c = 0; c < output_tensor.nc(); ++c)
+                        {
+                            (*iter)[k].operator()(r, c) = out_data[tensor_index(output_tensor, i, k, r, c)];
+                        }
+                    }
+                }
+            }
+        }
+
+
+        template <
+            typename const_label_iterator,
+            typename SUBNET
+            >
+        double compute_loss_value_and_gradient (
+            const tensor& input_tensor,
+            const_label_iterator truth,
+            SUBNET& sub
+        ) const
+        {
+            const tensor& output_tensor = sub.get_output();
+            tensor& grad = sub.get_gradient_input();
+
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+            DLIB_CASSERT(input_tensor.num_samples() != 0);
+            DLIB_CASSERT(input_tensor.num_samples() % sub.sample_expansion_factor() == 0);
+            DLIB_CASSERT(input_tensor.num_samples() == grad.num_samples());
+            DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+            DLIB_CASSERT(output_tensor.k() >= 1);
+            DLIB_CASSERT(output_tensor.k() < std::numeric_limits<uint16_t>::max());
+            DLIB_CASSERT(output_tensor.nr() == grad.nr() &&
+                output_tensor.nc() == grad.nc() &&
+                output_tensor.k() == grad.k());
+            for (long idx = 0; idx < output_tensor.num_samples(); ++idx)
+            {
+                const_label_iterator truth_matrix_ptr = (truth + idx);
+                for (long k = 0; k < output_tensor.k(); ++k)
+                {
+                    DLIB_CASSERT((*truth_matrix_ptr)[k].nr() == output_tensor.nr() &&
+                        (*truth_matrix_ptr)[k].nc() == output_tensor.nc(),
+                        "truth size = " << (*truth_matrix_ptr)[k].nr() << " x " << (*truth_matrix_ptr)[k].nc() << ", "
+                        "output size = " << output_tensor.nr() << " x " << output_tensor.nc());
+                }
+            }
+
+            // The loss we output is the average loss over the mini-batch, and also over each element of the matrix output.
+            const double scale = 1.0 / (output_tensor.num_samples() * output_tensor.k() * output_tensor.nr() * output_tensor.nc());
+            double loss = 0;
+            float* const g = grad.host();
+            const float* out_data = output_tensor.host();
+            for (long i = 0; i < output_tensor.num_samples(); ++i, ++truth)
+            {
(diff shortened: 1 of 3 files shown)


dlib/dnn/loss.h
```diff
@@ -2891,7 +2891,149 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-    class loss_dot_ 
+    template<long _num_channels>
+    class loss_mean_squared_per_channel_
```
davisking: How about calling this `loss_mean_squared_per_channel_and_pixel_`?

dlib/dnn/loss.h
```diff
@@ -2891,7 +2891,149 @@ namespace dlib
...
+            const tensor& output_tensor = sub.get_output();
+            tensor& grad = sub.get_gradient_input();
+
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+            DLIB_CASSERT(input_tensor.num_samples() != 0);
+            DLIB_CASSERT(input_tensor.num_samples() % sub.sample_expansion_factor() == 0);
+            DLIB_CASSERT(input_tensor.num_samples() == grad.num_samples());
+            DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+            DLIB_CASSERT(output_tensor.k() >= 1);
+            DLIB_CASSERT(output_tensor.k() < std::numeric_limits<uint16_t>::max());
```
davisking: Shouldn't these things assert that `output_tensor.k() == _num_channels`?

dlib/dnn/loss.h
```diff
@@ -2891,7 +2891,149 @@ namespace dlib
...
+            DLIB_CASSERT(output_tensor.nr() == grad.nr() &&
+                output_tensor.nc() == grad.nc() &&
+                output_tensor.k() == grad.k());
+            for (long idx = 0; idx < output_tensor.num_samples(); ++idx)
+            {
+                const_label_iterator truth_matrix_ptr = (truth + idx);
+                for (long k = 0; k < output_tensor.k(); ++k)
```
davisking: Should assert that `truth_matrix_ptr`'s size is `_num_channels`

arrufat:
I've updated with the requested changes. Your suggested name is more self-explanatory.

arrufat:
My last commit fixed a build error on the windows checks (it built fine on my machine with GCC-9.1).
Is there anything else I need to do?

davisking:
Thanks, the code looks good.  I tried running the test and it passes, but I notice the loss only drops a little.  Is it possible to make the test so that it runs to a low loss? Otherwise it's not obvious that it's working.  For instance, you could ask it to do something simple like train 10 linear functions rather than use this more complex network in the test now.

davisking:
For instance, make up 10 random linear functions.  Then use them to make training data.  E.g. make a 10x5 random matrix `w`, a random 5x1000 random matrix `x` and set `y` to `w*x`.  Train a network with just one linear layer to map `x` to `y`.  It should be able to learn it to basically 0 error and do so very quickly.

arrufat:
Ok, I'll try that. I also noticed the loss drops only a little, and I think it's because predicting everything as 0 already gives a pretty low error (since there's only one pixel that would differ). I'll think of a more obvious test.

dlib/test/dnn.cpp
```diff
@@ -2495,6 +2495,113 @@ namespace
...
+        // the main thing to notice when using this layer is that the output of
+        // the subnetwork needs to have the same number of channels as the
+        // channels we specify
+        using net_type = loss_mean_squared_per_channel_and_pixel<num_channels,
+                            htan<con<num_channels, 1, 1, 1, 1,
+                            relu<cont<32, 2, 2, 2, 2,
+                            relu<con<32, 2, 2, 2, 2,
+                            input<matrix<float>>>>>>>>>;
+        net_type net;
+        matrix<float> background = zeros_matrix<float>(num_rows, num_cols);
```
arrufat: In order to make sure the loss decreases, I've made this small change:
``` cpp
matrix<float> background = -ones_matrix<float>(num_rows, num_cols);
```
Since the network output is a `htan` the range of its values are `[-1, 1]`. That also has the benefit of making the network have a higher loss at the beginning:
- `error_before = 1.0636`
- `error_after = 0.0370037`

If the problem was the network not reducing the loss enough, would that solve it?
Or do you still prefer your example with random linear functions?
The reason I kind of like this example, is because it shows how it can be used to learn a network that estimates landmark heatmaps.

davisking:
That’s better. But a little linear thing that dropped the loss really low would be nice. 

The unit tests also aren’t a good way to do documentation. That kind of thing should go in a fully documented example program (you don’t have to make one for this though).

arrufat:
Ok, I understand. I'll redo the tests then, probably during this weekend :)

arrufat:
I've implemented the suggested linear test. I did not fiddle much with the parameters, but I'm not very happy with the results. Maybe learning that many random functions is not easy...

- `error_before = 2.03315`
- `error_after = 0.216274`

For the moment, I haven't removed the previous test. Please let me know what you think, or if I misunderstood your test.

dlib/test/dnn.cpp
```diff
@@ -2497,6 +2497,75 @@ namespace
...
+        const auto compute_error = [&inputs, &labels, &net, num_channels]()
+        {
+            const auto out = net(inputs);
+            double error = 0.0;
+            for (size_t i = 0; i < out.size(); ++i)
+            {
+                for (size_t c = 0; c < num_channels; ++c)
+                {
+                    auto diff = out[i][c] - labels[i][c];
+                    auto sqerror = pointwise_multiply(diff, diff);
```
davisking: To get the mse between to matrices you can just do `mean(squared(a-b))`

dlib/test/dnn.cpp
```diff
@@ -2497,6 +2497,75 @@ namespace
 
 // ----------------------------------------------------------------------------------------
 
+    void test_loss_mean_squared_per_channel_and_pixel_simple()
+    {
+        print_spinner();
+
+        const int num_samples = 1000;
+        const long num_channels = 10;
+        const long dimension = 1000;
```
davisking: This all looks great.  I suspect the reason it's still not converging super fast is just because these are large.  Try like 2 channels and dimensions of 3.

dlib/test/dnn.cpp
```diff
@@ -2497,6 +2497,75 @@ namespace
...
+                }
+            }
+            return error / out.size() / num_channels;
+        };
+
+        const auto error_before = compute_error();
+
+        dnn_trainer<net_type> trainer(net);
+        trainer.set_learning_rate(0.1);
+        trainer.set_iterations_without_progress_threshold(100);
```
davisking: Also setting this to something more than 100 might be reasonable. We don't want the unit tests to run super long though, but still, going a little above 100 could be good. Maybe 500 for instance.

arrufat:
Thank you for all the suggestions, I've simplified the MSE computation and reduced the dimensions of the data, it still does not converge to 0 as fast I'd like:

- `error_before: 2.95371`
- `error_after: 0.137677`

dlib/test/dnn.cpp
arrufat: Wow, thanks! I must say I'm not very familiar with all the matrix operations, yet...

arrufat:
I'm suspecting maybe the tests are too random? Maybe I need to generate a more structured wight matrix `w`?

davisking:
Don’t call `set_max_num_epochs`.  That’s probably telling it to stop early.

arrufat:
The only reason why I put the value of 160 is because it stops at epoch 157 (where it reaches a learning rate of 1e-7). I've also tried with smaller learning rate, but I get the same results.

arrufat:
I've modified the network a little bit:

```c++
using net_type = loss_mean_squared_per_channel_and_pixel<num_channels,                                                                                                                                                               
                    extract<0, num_channels, 1, dimension,                                                                                                                                                                           
                    fc<num_outputs,                                                                                                                                                                                                  
                    relu<bn_fc<fc<500,                                                                                                                                                                                               
                    input<matrix<float>>>>>>>>;
```

Now, the errors look like this:

- `error_before: 2.25673`
- `error_after: 0.0214044`

arrufat:
Great! Thank you! I'm probably more happy than you about this PR (my first serious one).
I wanted to contribute to dlib so badly!!

PR #1865, opened 2019-08-23
don't cast away constness

tdp2110:
there was no actual harm here: the unconsting was consted up by the member (but it needn't have been had the member been pointer to non-const). 

generally speaking, `reinterpret_cast` is better than C-style casting because it cannot cast away constness.

### DIFF
--- a/dlib/image_processing/generic_image.h
+++ b/dlib/image_processing/generic_image.h
@@ -186,7 +186,7 @@ namespace dlib
         image_view(
             image_type& img
         ) : 
-            _data((char*)image_data(img)), 
+            _data(reinterpret_cast<char*>(image_data(img))), 
             _width_step(width_step(img)),
             _nr(num_rows(img)),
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks for the PR :)

tdp2110:
btw, check out clang-tidy if you haven't already. I ran it on this repo and it showed a couple of things, like potential nullptr dereference in entropy_decoder_model_kernel_5.h on `node *n = temp->child_context;`. I briefly thought about fixing it, but wasn't sure how :)

davisking:
I’m not super concerned. That code is ancient and extremely well used without any issues.

davisking:
Yeah, that might be a real bug.  It's hard to be sure.  Can you trigger it
at runtime with some known input?

tdp2110:
Unfortunately no, I only have the analyzer output.

PR #1866, opened 2019-08-24
Fixing dpoint:  y-property mutates x-coord 

supervacuus:
See #1794, but this time for dpoint.


davisking:
Thanks for the fix :)

PR #1870, opened 2019-08-30
Clear truth_idxs between samples

reunanen:
To address the possible issue mentioned [here](https://github.com/davisking/dlib/commit/29db3ee566d68b3a197e24394c3beb64513ca98e#r34900715), clear the `truth_idxs` vector between samples.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1139,7 +1139,8 @@ namespace dlib
                     det_thresh_speed_adjust = std::max(det_thresh_speed_adjust,dets[max_num_initial_dets].detection_confidence + options.loss_per_false_alarm);
                 }
 
-                std::vector<size_t> truth_idxs;  truth_idxs.reserve(truth->size());
+                std::vector<size_t> truth_idxs;
+                truth_idxs.reserve(truth->size());
 
                 // The loss will measure the number of incorrect detections.  A detection is
                 // incorrect if it doesn't hit a truth rectangle or if it is a duplicate detection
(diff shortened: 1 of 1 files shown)


dlib/dnn/loss.h
```diff
@@ -1240,7 +1240,7 @@ namespace dlib
 
                 hit_truth_table.assign(hit_truth_table.size(), false);
                 final_dets.clear();
-
+                truth_idxs.clear();
```
davisking: Just move `truth_idxs` into the loop body so the `clear()` is unneeded. I'm not sure what I was thinking.  The formatting of the `truth_idxs` declaration is even weird :|

davisking:
Nice catch, thanks for fixing this :)

reunanen:
No problem!

Actually, should we also `push_back` something (e.g., 0) to `truth_idxs` [here](https://github.com/davisking/dlib/blob/170877da88fdd887b9404aec78646e604d8368d8/dlib/dnn/loss.h#L1158)?

dlib/dnn/loss.h
reunanen: Moved. (Didn't change formatting.)

davisking:
> No problem!
> 
> Actually, should we also `push_back` something (e.g., 0) to `truth_idxs` [here](https://github.com/davisking/dlib/blob/170877da88fdd887b9404aec78646e604d8368d8/dlib/dnn/loss.h#L1158)?

Yeah, definitely. `truth_idxs` needs to end up with `truth_idxs.size() == truth->size()`.

dlib/dnn/loss.h
```diff
@@ -1140,6 +1139,7 @@ namespace dlib
                     det_thresh_speed_adjust = std::max(det_thresh_speed_adjust,dets[max_num_initial_dets].detection_confidence + options.loss_per_false_alarm);
                 }
 
+                std::vector<size_t> truth_idxs;  truth_idxs.reserve(truth->size());
```
davisking: And make this two lines.  Not sure what I was thinking with this funky formatting

davisking:
Awesome, thanks for another PR :)

PR #1896, opened 2019-10-10
Ignore truth rects that overlap too much (have same index in feature coordinates)

### LINKED ISSUE #1894: MMOD loss may go negative if detection doesn't match a truth box but does match an ignore box
## Expected Behavior
MMOD loss never goes negative for valid input.

(Note that what constitutes "valid input" may need to be discussed. In my case, there are correct truth-rects that are very close to each other. They also overlap so much that it does not look like a great idea to try to teach all of them. So I'm setting a stricter `overlaps_nms`, and setting some of the truth-rects to be ignored. However, I don't want to completely disregard any of them (they contain valuable information), so for different epochs I'm randomly selecting which rects are to be ignored. Thought this should be fine, as long as there are no non-ignored rects that overlap, as per the `overlaps_nms` that I have set.)

## Current Behavior
When there are ignore-rects very close to and overlapping with the (non-ignore) truth-rects, it looks like it's possible that `truth_match <= options.truth_match_iou_threshold`, while `overlaps_ignore_box(*truth, dets[i].rect) == true`. In other words, a candidate detection doesn't quite match the (correctly-)best-matching non-ignore rect with `> options.truth_match_iou_threshold`, but does overlap an ignore-box.

As a result, it seems that the loss may go negative. (Specifically because in this case the candidate detection isn't added to `final_dets`, and therefore the corresponding `out_data` isn't added to `loss`.)

Note that it's not just the loss that goes sour; also the training itself does not really "work", i.e. the gradients are not updated correctly it seems. (This seems to hold at least in a real-world project I'm working with, where the network after a long training run eventually detects nothing – whereas a shorter training run works ok. But for this gradient part I don't have how-to-reproduce instructions, at least just yet. However, I guess loss and gradient go hand-to-hand, so possibly fixing the negative loss will fix the gradient updates as well.)

## Workarounds
I have tried the following workarounds:
- Set `options.truth_match_iou_threshold = 0.25` (arbitrary selection).
- Change code so that the gradient is not updated for samples for which the contribution to loss would be negative: https://github.com/reunanen/dlib/commit/0d55d3cf1bd9cd082f5bd8361fe551972c92ee61

While these approaches seem to alleviate the problem to some degree at least, I can't say I'm very happy with either of them. (Think they are only band-aid at best.)

## Steps to Reproduce

Clone current `master` branch, and cherry-pick this: https://github.com/reunanen/dlib/commit/781065f62f57ccc4466aa5e110f66642e99bc438

Run the dnn tests – the newly added test should fail.

reunanen:
resolves #1894

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1156,10 +1156,17 @@ namespace dlib
                         {
                             // Ignore boxes that can't be detected by the CNN.
                             loss -= options.loss_per_missed_target;
-                            truth_idxs.push_back(0);
+                            truth_idxs.push_back(-1);
                             continue;
                         }
                         const size_t idx = (k*output_tensor.nr() + p.y())*output_tensor.nc() + p.x();
+                        if (std::find(truth_idxs.begin(), truth_idxs.end(), idx) != truth_idxs.end())
+                        {
+                            // Ignore duplicate truth box in feature coordinates.
+                            loss -= options.loss_per_missed_target;
+                            truth_idxs.push_back(-1);
+                            continue;
+                        }
                         loss -= out_data[idx];
                         // compute gradient
                         g[idx] = -scale;
@@ -1169,7 +1176,7 @@ namespace dlib
                     {
                         // This box was ignored so shouldn't have been counted in the loss.
                         loss -= options.loss_per_missed_target;
-                        truth_idxs.push_back(0);
+                        truth_idxs.push_back(-1);
                     }
                 }
 
@@ -1226,6 +1233,7 @@ namespace dlib
                         if (options.overlaps_nms(best_matching_truth_box, (*truth)[i]))
                         {
                             const size_t idx = truth_idxs[i];
+                            assert(idx != -1);
                             // We are ignoring this box so we shouldn't have counted it in the
                             // loss in the first place.  So we subtract out the loss values we
                             // added for it in the code above.
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1143,7 +1143,7 @@ namespace dlib
                 std::vector<size_t> truth_idxs;
                 truth_idxs.reserve(truth->size());
 
-                std::unordered_map<size_t, std::deque<rectangle>> truth_idx_to_truth_rects;
+                std::unordered_map<size_t, rectangle> idx_to_truth_rect;
 
                 // The loss will measure the number of incorrect detections.  A detection is
                 // incorrect if it doesn't hit a truth rectangle or if it is a duplicate detection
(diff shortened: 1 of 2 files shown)


dlib/dnn/loss.h
```diff
@@ -1160,10 +1163,18 @@ namespace dlib
                             continue;
                         }
                         const size_t idx = (k*output_tensor.nr() + p.y())*output_tensor.nc() + p.x();
+                        if (unique_truth_idxs.find(idx) != unique_truth_idxs.end())
+                        {
+                            // Ignore duplicate truth box in feature coordinates.
+                            loss -= options.loss_per_missed_target;
+                            truth_idxs.push_back(0);
```
davisking: Setting this to 0 is going to cause this to interact badly with the use of `truth_idxs` later (here https://github.com/davisking/dlib/pull/1896/files#diff-32a043d312608d62a296b05ce1bc096eR1239-R1244). That code later on is checking if truth boxes are being knocked out by NMS, this check here is essentially the same thing, so should probably be done in the same place. Effectively, we have a floor on how tight you can make the NMS because of the CNN's resolution.

So you could keep track of how many times one of these CNN output cells is hit by a truth box and if it's more than once pretend that the NMS triggered.  Then treat it the same way as the NMS check at https://github.com/davisking/dlib/pull/1896/files#diff-32a043d312608d62a296b05ce1bc096eR1239-R1244 does (with an updated warning message to say why the box is being ignored).

reunanen: Hmm those links don't appear to lead anywhere?

reunanen: But no worries, I think I get it anyway. Will push some suggestion.

reunanen: Hmm what if I just push `idx` and not `0` here?

davisking: Huh, yeah, links worked at first.  Anyway, I mean down at line 1228 where `truth_idxs` is used.

reunanen: Tried to do a minimal (yet correct) change.

dlib/dnn/loss.h
```diff
@@ -1156,20 +1159,34 @@ namespace dlib
...
-                            truth_idxs.push_back(0);
+                            truth_idxs.push_back(-1);
                             continue;
                         }
                         const size_t idx = (k*output_tensor.nr() + p.y())*output_tensor.nc() + p.x();
+                        const auto i = truth_idx_to_truth_rects.find(idx);
+                        if (i != truth_idx_to_truth_rects.end())
+                        {
+                            // Ignore duplicate truth box in feature coordinates.
+                            std::cout << "Warning, ignoring object.  We encountered a truth rectangle located at " << x.rect;
```
davisking: The logic here looks good to me.  Can you make this warning message more gramatical though?  Right now it will say `We encountered a truth rectangle located at [blah blah] because it maps to the exact same feature coordinates...`

dlib/dnn/loss.h
```diff
@@ -1142,6 +1143,8 @@ namespace dlib
                 std::vector<size_t> truth_idxs;
                 truth_idxs.reserve(truth->size());
 
+                std::unordered_map<size_t, std::deque<rectangle>> truth_idx_to_truth_rects;
```
davisking: Seems like this could be simply `std::unordered_map<size_t, rectangle>` since only one element of the deque is used.  That would also likely execute faster, which is worth it here as this code can sometimes be a bottleneck.

dlib/dnn/loss.h

dlib/dnn/loss.h
reunanen: Changed.

reunanen:
> This looks good to me. Does it end up making those training issues you mentioned in #1894 all go away?

Indirectly, yes. Based on your advice, I first [removed the NMS limit](https://github.com/reunanen/dlib/commit/27f193df34a3d69506457c42e57857cc50a0d116#diff-7d2445557c03b31a758c2ef7d310bc53), and made the network try to learn everything – even heavily overlapping truth rects.

That however caused a new issue, when the truth rects would map to the same feature coordinates. This PR appears to fix this new issue.

davisking:
Nice, thanks for another quality PR :)

PR #1897, opened 2019-10-14
Primarily match to truth rects that haven't been hit already

reunanen:
Consider two truth rects that overlap heavily (say, A and B). Also consider basically good response, i.e. two detections (say, C and D) that are pretty close to the truth rects. Now in the current `master` version it may very well happen that both detections C and D happen to be closest to A (and not B), meaning that there's one false positive (both C and D mapped to A) and also one false negative (B not found at all).

This PR modifies the matching process so that detections are primarily matched to truth rects that haven't been hit already. Therefore, detection D (previously false positive) would now be matched to truth rect B (while det C is still matched to truth A), meaning that there's no longer a false positive, and no longer a false negative (assuming of course that `truth_match_iou_threshold` is exceeded for the D-B pair).

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1204,7 +1204,7 @@ namespace dlib
 
                     const auto& det_label = options.detector_windows[dets[i].tensor_channel].label;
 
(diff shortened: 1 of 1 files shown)


davisking:
Nice, this is a good idea. :)

PR #1899, opened 2019-10-15
fix signed/unsigned warning

arrufat:
Since the `truth_idxs` might hold negative values, it should be declared as a `std::vector<int>` instead of `std::vector<size_t>`


davisking:
Good catch, thanks for the PR :)

PR #1905, opened 2019-10-24
Propagate CMake flags to try_compile (fixes #1900)

### LINKED ISSUE #1900: Seamless integration with package managers
Hi,

I tried to package dlib so that it works with Conan: it's mostly good and most of it worked on the first try, but the `try_compile` sanity checks in `find_libjpeg.cmake` and `find_libpng.cmake` make it impossible to package dlib if those libraries are only installed through the package manager and not directly in the system directories.

## Expected Behavior
The CMake support for Conan allows to download libraries in `~/.conan` and automatically changes variables such as `CMAKE_PREFIX_PATH` to point to the Conan directories, which allows `find_package` to find the required libraries in the Conan directories. If the CMake files don't do anything specific, everything works smoothly enough.

## Current Behavior
When I try to use the CMake/Conan support to package dlib, the libraries (libjpeg, libpng, etc.) downloaded by Conan are correctly found by `find_package`, but the subsequent sanity checks in `find_libjpeg.cmake` and `find_libpng.cmake` fail.

The reason is that the aforementioned tests use `try_compile` to compile the test projects in `test_for_libjpeg` and `test_for_libpng`, and those call do not propagate the CMake environments to the callees (specifically the variables that tell CMake where to find its packages), so their `find_package` calls fail if libjpeg and libpng are only installed through Conan and don't exist in the system directories.

It should be possible to forward the required variables somehow, probably through the `CMAKE_FLAGS` argument of `try_compile`.

## Steps to Reproduce
If you're familiar enough with Conan, here is a stripped down version of the recipe I used to fetch, compile and install dlib (requires Python 3.7):

```python
import os
from pathlib import Path

from conans import ConanFile, CMake, tools


class LibDlibConan(ConanFile):
    name = "dlib"
    version = "19.18"
    url = "https://github.com/davisking/dlib"
    generators = "cmake"
    settings = "os", "arch", "compiler", "build_type"

    _source_subfolder = "source_subfolder"
    _build_subfolder = "build_subfolder"

    scm = {
        "type": "git",
        "subfolder": _source_subfolder,
        "url": "https://github.com/davisking/dlib.git",
        "revision": f"v{version}"
    }

    def source(self):
        # Tell CMake about the subdirectories
        tools.replace_in_file(
            Path(self._source_subfolder) / "dlib" / "CMakeLists.txt",
            'project(dlib)',
            f'''project(dlib)
                include(${{CMAKE_BINARY_DIR}}/../conanbuildinfo.cmake)
                conan_basic_setup()'''
        )

        # Avoid try_compile issues at the cost of fewer checks 
        tools.replace_in_file(
            Path(self._source_subfolder) / "dlib" / "cmake_utils" / "find_libjpeg.cmake",
            "if(JPEG_FOUND)",
            "if(NOT JPEG_FOUND)"
        )
        tools.replace_in_file(
            Path(self._source_subfolder) / "dlib" / "cmake_utils" / "find_libpng.cmake",
            "if(PNG_FOUND)",
            "if(NOT PNG_FOUND)"
        )
        
    def requirements(self):
        self.requires("libjpeg/9c")
        self.requires("libpng/1.6.37")
        self.requires("zlib/1.2.11")

    def build(self):
        cmake = CMake(self)
        cmake.configure(build_folder=self._build_subfolder)
        cmake.build()

    def package(self):
        self.copy("*.a", "lib", keep_path=False)
        self.copy("*.lib", "lib", keep_path=False)
        self.copy("*.so*", "lib", keep_path=False)
        self.copy("*.h", "include/dlib", os.path.join(self._source_subfolder, "dlib"))
        self.copy("LICENSE.txt", "licenses", os.path.join(self._source_subfolder, "dlib"), keep_path=False)

    def package_info(self):
        self.cpp_info.libs = tools.collect_libs(self)
```

As you can see I worked around the issue by skipping the sanity checks for libjpeg and libpng entirely, but I would rather keep said checks if possible. Forwarding the appropriate CMake variable to `try_compile` instead would probably help to keep the checks while being package manager-friendly :) 

Having an installation step in CMake to simplify the process would be nice too, but that's a different issue and I seem to recall that it was already discussed somewhere else.

* **Version**: 19.18
* **Where did you get dlib**: GitHub releases
* **Platform**: Ubuntu 18.04 64-bit
* **Compiler**: irrelevant

Morwenn:
Sorry for the noise (4 commits is a bit much), I had to tinker a bit directly from the web UI, don't hesitate to squash the commits before merging.

This pull request addresses the concerns I had in #1900: it propagates `CMAKE_PREFIX_PATH`, `CMAKE_INCLUDE_PATH` and `CMAKE_LIBRARY_PATH` to `try_compile` in the files `find_libjpeg.cmake` and `find_libpng.cmake`, allowing those libraries to be found by CMake even when they live in uncommon directories (my use case being libraries installed with Conan).


davisking:
Cool, thanks for the PR :)

PR #1914, opened 2019-11-06
Openblas type fix

cchadowitz-pf:


dlib/matrix/cblas_constants.h
```diff
@@ -3,6 +3,15 @@
...
+#ifdef USING_OPENBLAS
+#include <cblas.h>
+#define CBLAS_DOUBLE_TYPE openblas_complex_double
+#define CBLAS_FLOAT_TYPE openblas_complex_float
+#else
+#define CBLAS_DOUBLE_TYPE double
+#define CBLAS_FLOAT_TYPE float
+#endif
+
 #if !(defined(__GSL_CBLAS_H__) || defined(CBLAS_H))
```
davisking: A better thing to do is probably to add to this list of "if this is defined then don't let dlib define the cblas symbols again".  There is almost certainly some `#define` set by the openblas header you are running into that you could trigger on.

cchadowitz-pf: So I've gotten a better understanding of where the issue is since then. openblas v0.2.20 (the one that ships with ubuntu 18.04) has this bug still not fixed https://github.com/xianyi/OpenBLAS/issues/1322

Since v0.2.20 (so v0.3.0+), it's been fixed with this https://github.com/xianyi/OpenBLAS/pull/1347/files

However that version of openblas only ships with later versions of ubuntu. I've found one way of resolving this is by manually installing a later version of openblas.

When I put my defines in the "if this is defined don't let dlib define the cblas symbols again", it causes errors in ubuntu 16.04. So I'm not sure what the best way is to get dlib to build under these situations in both ubuntu 16.04 and 18.04 without manually updating openblas to v0.3.0+. In the meantime, I can rely on updating openblas until the next ubuntu LTS version is released and includes a later version of openblas natively.

davisking: Ah, so maybe the right thing to do is to introduce the `CBLAS_DOUBLE_TYPE` then :|. Although that requires being able to reliably detect their presence. 

I'm also fine with just saying users shouldn't `#include` the broken version of the openblas headers too.

cchadowitz-pf: Yeah - the whole situation is a bit messy. I'd defer to your thoughts/preferences on this, though I'm glad to have a better understanding of the situation as a whole.

PR #1918, opened 2019-11-10
Instance segmentation

### LINKED ISSUE #1646: Instance segmentation: new layer (if needed), and an example program
I think I need to implement _instance segmentation_. For example, this looks like a useful approach to me: https://arxiv.org/pdf/1807.10007.pdf

If someone else implements any kind of approach before I get my head around this matter, great. If not, then I might give it a try.

Any feedback or comments are very welcome.

reunanen:
(resolves #1646)

This PR adds an _instance segmentation_ example.

In the sample images below, colors of rectangles depend on class, whereas segmentation colors are random (to show where one instance changes to another).

Some examples from the training set (PASCAL VOC 2012, but with bicycles, cars and cats only):

image
image
image
image
image
image
image
image
image

Some examples not seen during training:

image
image
image
image
image
image
image
image
image

(Yes, these is certainly overfitting, for example in the cat detector. But avoiding that is not really the point of this PR.)

The proposed implementation is (I guess) pretty much the simplest possible approach: first run standard MMOD object detector, then predict the correct segmentation mask for each found object. So object detection and segmentation are decoupled (separate networks).

An alternative approach would be a single loss layer, predicting both objects and instance masks at the same time. This is how [Mask R-CNN](https://arxiv.org/pdf/1703.06870), for example, works.
(Actually started implementing that direction also. But here it wasn't entirely straightforward (to me at least) to figure out how the MMOD side of it could run an image pyramid, without forcing the segmentation head to do that as well. But this may nevertheless be a reasonable direction for future work; the PR at hand should be seen only as a baseline example, upon which we can later improve.)

While the decoupled approach is sort-of unexciting, there are also benefits to it:
* When a common (downscaling) path is not used, distributing training to two different machines is trivial (the machines need not even be networked).
* You don't really need to find a balance between detection loss and segmentation loss, like you do in Mask R-CNN. I suppose that's mostly a trial-and-error exercise in practice (haven't really tried though). At any rate, find it nice that in the proposed approach we don't need to set another parameter whose scale perhaps is not very easily understood.

---

Note that in addition to the instance segmentation example, this PR also contains a proposed change to the trainer code itself. In particular, this trainer code change tries to fix a problem where dumping some of the previous loss values (when shrinking the learning rate) appears to cause the trainer not to load from its synchronization file, even though training loss may have increased a whole lot since the latest disk sync:

image

(This problem should be rather easy to reproduce by training instance segmentation networks _without_ the proposed trainer changes.)

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -993,6 +993,36 @@ namespace dlib
         }
     }
 
+    std::ostream& operator<<(std::ostream& out, const std::vector<mmod_options::detector_window_details>& detector_windows)
+    {
+        // write detector windows grouped by label
+        // example output: aeroplane:74x30,131x30,70x45,54x70,198x30;bicycle:70x57,32x70,70x32,51x70,128x30,30x121;car:70x36,70x60,99x30,52x70,30x83,30x114,30x200
+
+        std::map<std::string, std::deque<mmod_options::detector_window_details>> detector_windows_by_label;
+        for (const auto& detector_window : detector_windows)
+            detector_windows_by_label[detector_window.label].push_back(detector_window);
+
+        size_t label_count = 0;
+        for (const auto& i : detector_windows_by_label)
+        {
+            const auto& label = i.first;
+            const auto& detector_windows = i.second;
+
+            if (label_count++ > 0)
+                out << ";";
+            out << label << ":";
+
+            for (size_t j = 0; j < detector_windows.size(); ++j)
+            {
+                out << detector_windows[j].width << "x" << detector_windows[j].height;
+                if (j + 1 < detector_windows.size())
+                    out << ",";
+            }
+        }
+
+        return out;
+    }
+
 // ----------------------------------------------------------------------------------------
 
     class loss_mmod_ 
@@ -1395,37 +1425,10 @@ namespace dlib
         {
             out << "loss_mmod\t (";
 
-            out << "detector_windows:(";
             auto& opts = item.options;
 
-            {
-                // write detector windows grouped by label
-                // example output: detector_windows:(aeroplane:74x30,131x30,70x45,54x70,198x30;bicycle:70x57,32x70,70x32,51x70,128x30,30x121;car:70x36,70x60,99x30,52x70,30x83,30x114,30x200)
-                typedef std::deque<const dlib::mmod_options::detector_window_details*> detector_windows;
-                std::map<std::string, detector_windows> detector_windows_by_label;
-                for (const auto& detector_window : opts.detector_windows)
-                    detector_windows_by_label[detector_window.label].push_back(&detector_window);
+            out << "detector_windows:(" << opts.detector_windows << ")";
 
-                size_t label_count = 0;
-                for (const auto& i : detector_windows_by_label)
-                {
-                    const auto& label = i.first;
-                    const auto& detector_windows = i.second;
-
-                    if (label_count++ > 0)
-                        out << ";";
-                    out << label << ":";
-
-                    for (size_t i = 0; i < detector_windows.size(); ++i)
-                    {
-                        out << detector_windows[i]->width << "x" << detector_windows[i]->height;
-                        if (i+1 < detector_windows.size())
-                            out << ",";
-                    }
-                }
-            }
-
-            out << ")";
             out << ", loss per FA:" << opts.loss_per_false_alarm;
             out << ", loss per miss:" << opts.loss_per_missed_target;
             out << ", truth match IOU thresh:" << opts.truth_match_iou_threshold;
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1397,12 +1397,34 @@ namespace dlib
 
             out << "detector_windows:(";
             auto& opts = item.options;
-            for (size_t i = 0; i < opts.detector_windows.size(); ++i)
+
             {
-                out << opts.detector_windows[i].width << "x" << opts.detector_windows[i].height;
-                if (i+1 < opts.detector_windows.size())
-                    out << ",";
+                // write detector windows grouped by label
+                // example output: detector_windows:(aeroplane:74x30,131x30,70x45,54x70,198x30;bicycle:70x57,32x70,70x32,51x70,128x30,30x121;car:70x36,70x60,99x30,52x70,30x83,30x114,30x200)
+                typedef std::deque<const dlib::mmod_options::detector_window_details*> detector_windows;
+                std::map<std::string, detector_windows> detector_windows_by_label;
+                for (const auto& detector_window : opts.detector_windows)
+                    detector_windows_by_label[detector_window.label].push_back(&detector_window);
+
+                size_t label_count = 0;
+                for (const auto& i : detector_windows_by_label)
+                {
+                    const auto& label = i.first;
+                    const auto& detector_windows = i.second;
+
+                    if (label_count++ > 0)
+                        out << ";";
+                    out << label << ":";
+
+                    for (size_t i = 0; i < detector_windows.size(); ++i)
+                    {
+                        out << detector_windows[i]->width << "x" << detector_windows[i]->height;
+                        if (i+1 < detector_windows.size())
+                            out << ",";
+                    }
+                }
             }
+
             out << ")";
             out << ", loss per FA:" << opts.loss_per_false_alarm;
             out << ", loss per miss:" << opts.loss_per_missed_target;
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -993,7 +993,7 @@ namespace dlib
         }
     }
 
-    std::ostream& operator<<(std::ostream& out, const std::vector<mmod_options::detector_window_details>& detector_windows)
+    inline std::ostream& operator<<(std::ostream& out, const std::vector<mmod_options::detector_window_details>& detector_windows)
     {
         // write detector windows grouped by label
         // example output: aeroplane:74x30,131x30,70x45,54x70,198x30;bicycle:70x57,32x70,70x32,51x70,128x30,30x121;car:70x36,70x60,99x30,52x70,30x83,30x114,30x200
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -697,10 +697,10 @@ namespace dlib
 
                     // Check if we should shrink the learning rate based on how the test
                     // error has been doing lately.
-                    if (learning_rate_shrink != 1 && steps_since_last_learning_rate_shrink > iter_without_progress_thresh)
+                    if (learning_rate_shrink != 1)
                     {
                         test_steps_without_progress = count_steps_without_decrease(test_previous_loss_values);
-                        if (test_steps_without_progress >= test_iter_without_progress_thresh)
+                        if (test_steps_without_progress >= test_iter_without_progress_thresh && steps_since_last_learning_rate_shrink >= test_iter_without_progress_thresh)
                         {
                             test_steps_without_progress = count_steps_without_decrease_robust(test_previous_loss_values);
                             if (test_steps_without_progress >= test_iter_without_progress_thresh)
@@ -809,13 +809,11 @@ namespace dlib
                 // have a "budget" that prevents us from calling
                 // count_steps_without_decrease() every iteration.  We do this because
                 // it can be expensive to compute when previous_loss_values is large.
-                if (gradient_check_budget > iter_without_progress_thresh
-                    && learning_rate_shrink != 1
-                    && steps_since_last_learning_rate_shrink > iter_without_progress_thresh)
+                if (gradient_check_budget > iter_without_progress_thresh && learning_rate_shrink != 1)
                 {
                     gradient_check_budget = 0;
                     steps_without_progress = count_steps_without_decrease(previous_loss_values);
-                    if (steps_without_progress >= iter_without_progress_thresh)
+                    if (steps_without_progress >= iter_without_progress_thresh && steps_since_last_learning_rate_shrink >= iter_without_progress_thresh)
                     {
                         // Double check that we aren't seeing decrease.  This second check
                         // discards the top 10% largest values and checks again.  We do
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -1047,7 +1047,10 @@ namespace dlib
                     // might be better off giving up at this learning rate, and trying a
                     // lower one instead.
                     if (prob_loss_increasing_thresh >= prob_loss_increasing_thresh_max_value)
+                    {
+                        std::cout << "(and while at it, also shrinking the learning rate)" << std::endl;
                         learning_rate = learning_rate_shrink * learning_rate;
+                    }
                 }
                 else
                 {
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -706,10 +706,10 @@ namespace dlib
                                 // optimization has flattened out, so drop the learning rate. 
                                 learning_rate = learning_rate_shrink*learning_rate;
                                 test_steps_without_progress = 0;
+
                                 // Empty out some of the previous loss values so that test_steps_without_progress 
                                 // will decrease below test_iter_without_progress_thresh.  
-                                for (unsigned long cnt = 0; cnt < test_previous_loss_values_dump_amount+test_iter_without_progress_thresh/10 && test_previous_loss_values.size() > 0; ++cnt)
-                                    test_previous_loss_values.pop_front();
+                                drop_some_test_previous_loss_values();
                             }
                         }
                     }
@@ -826,10 +826,10 @@ namespace dlib
                             // optimization has flattened out, so drop the learning rate. 
                             learning_rate = learning_rate_shrink*learning_rate;
                             steps_without_progress = 0;
+
                             // Empty out some of the previous loss values so that steps_without_progress 
                             // will decrease below iter_without_progress_thresh.  
-                            for (unsigned long cnt = 0; cnt < previous_loss_values_dump_amount+iter_without_progress_thresh/10 && previous_loss_values.size() > 0; ++cnt)
-                                previous_loss_values.pop_front();
+                            drop_some_previous_loss_values();
                         }
                     }
                 }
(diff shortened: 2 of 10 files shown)


dlib/dnn/loss.h
```diff
@@ -1397,12 +1397,34 @@ namespace dlib
...
             out << "detector_windows:(";
             auto& opts = item.options;
-            for (size_t i = 0; i < opts.detector_windows.size(); ++i)
+
             {
-                out << opts.detector_windows[i].width << "x" << opts.detector_windows[i].height;
-                if (i+1 < opts.detector_windows.size())
-                    out << ",";
+                // write detector windows grouped by label
+                // example output: detector_windows:(aeroplane:74x30,131x30,70x45,54x70,198x30;bicycle:70x57,32x70,70x32,51x70,128x30,30x121;car:70x36,70x60,99x30,52x70,30x83,30x114,30x200)
```
reunanen: This chunk only improves (?) the output, as per the comment.

dlib/dnn/trainer.h
```diff
@@ -1057,34 +1076,67 @@ namespace dlib
...
+            };
+            const auto can_drop_an_additional_previous_loss_value = [&]() {
+                return !additional_previous_loss_values_to_keep_until_disk_sync.empty()
+                    && total_previous_loss_values() > max_previous_loss_values_to_consider;
+            };
+
+            while (can_drop_an_additional_previous_loss_value())
+                additional_previous_loss_values_to_keep_until_disk_sync.pop_front();
+
+            // collect all previous loss values we want to consider
```
reunanen: This is pretty much the beef of the trainer code change introduced in the PR summary.

dlib/image_transforms/interpolation.h
```diff
@@ -865,10 +865,18 @@ namespace dlib
...
+                    if (std::is_integral<T>::value)
+                        return static_cast<T>(value + 0.5);
+                    else
+                        return static_cast<T>(value);
+                };
+
+                out_img[r][c]   = convert_to_output_type(fout[0]);
+                out_img[r][c+1] = convert_to_output_type(fout[1]);
+                out_img[r][c+2] = convert_to_output_type(fout[2]);
+                out_img[r][c+3] = convert_to_output_type(fout[3]);
```
reunanen: This chunk is far from being the beef of the PR, but it does make the instance segmentation output look nicer (no holes due to truncated values).

examples/dnn_semantic_segmentation_ex.h
```diff
@@ -34,83 +34,7 @@
...
-    return a.red == b.red && a.green == b.green && a.blue == b.blue;
-}
-
-// ----------------------------------------------------------------------------------------
-
-// The PASCAL VOC2012 dataset contains 20 ground-truth classes + background.  Each class
-// is represented using an RGB color value.  We associate each class also to an index in the
-// range [0, 20], used internally by the network.
-
-struct Voc2012class {
```
reunanen: This stuff simply moved to a new header file (used also by the instance segmentation example).

examples/dnn_semantic_segmentation_train_ex.cpp
```diff
@@ -209,14 +108,14 @@ double calculate_accuracy(anet_type& anet, const std::vector<image_info>& datase
...
-        load_image(rgb_label_image, image_info.label_filename);
+        load_image(rgb_label_image, image_info.class_label_filename);
 
         // Create predictions for each pixel. At this point, the type of each prediction
         // is an index (a value between 0 and 20). Note that the net may return an image
         // that is not exactly the same size as the input.
         const matrix<uint16_t> temp = anet(input_image);
 
-        // Convert the indexes to RGB values.
+        // Convert the RGB values to indexes.
```
reunanen: Fix incorrect comment.

examples/dnn_instance_segmentation_ex.cpp
```diff
@@ -0,0 +1,174 @@
...
+{
+    if (argc != 2)
+    {
+        cout << "You call this program like this: " << endl;
+        cout << "./dnn_instance_segmentation_train_ex /path/to/images" << endl;
+        cout << endl;
+        cout << "You will also need a trained '" << instance_segmentation_net_filename << "' file." << endl;
+        cout << "You can either train it yourself (see example program" << endl;
+        cout << "dnn_instance_segmentation_train_ex), or download a" << endl;
+        cout << "copy from here: http://dlib.net/files/" << instance_segmentation_net_filename << endl;
```
reunanen: I'm uploading a sample file to Dropbox... will update the link here shortly.

dlib/dnn/trainer.h
reunanen: (Certainly not saying these are the exact changes that we need, but at a bare minimum I want to start a discussion here.)

examples/dnn_instance_segmentation_ex.cpp
reunanen: This should work: https://www.dropbox.com/s/tcimn353n3ogax7/instance_segmentation_voc2012net.dnn

dlib/dnn/loss.h
```diff
@@ -1397,12 +1397,34 @@ namespace dlib
...
+                typedef std::deque<const dlib::mmod_options::detector_window_details*> detector_windows;
+                std::map<std::string, detector_windows> detector_windows_by_label;
```
davisking: This would be simpler if declared like this:
```
std::map<std::string,std::vector<mmod_options::detector_window_details>> detector_windows;
```
No pointers needed, and so no reasoning about lifetimes.

dlib/dnn/loss.h
```diff
@@ -1397,12 +1397,34 @@ namespace dlib
 
             out << "detector_windows:(";
             auto& opts = item.options;
-            for (size_t i = 0; i < opts.detector_windows.size(); ++i)
+
             {
```
davisking: Remove { if removing the `for` loop

dlib/dnn/loss.h
```diff
@@ -1397,12 +1397,34 @@ namespace dlib
...
+                for (const auto& i : detector_windows_by_label)
+                {
+                    const auto& label = i.first;
+                    const auto& detector_windows = i.second;
+
+                    if (label_count++ > 0)
+                        out << ";";
+                    out << label << ":";
+
+                    for (size_t i = 0; i < detector_windows.size(); ++i)
```
davisking: Use a variable name other than `i` so it doesn't alias the `i` in the outer scope.

dlib/image_transforms/interpolation.h
```diff
@@ -865,10 +865,18 @@ namespace dlib
...
                 out.store(fout);
 
-                out_img[r][c]   = static_cast<T>(fout[0]);
-                out_img[r][c+1] = static_cast<T>(fout[1]);
-                out_img[r][c+2] = static_cast<T>(fout[2]);
-                out_img[r][c+3] = static_cast<T>(fout[3]);
+                const auto convert_to_output_type = [](float value)
+                {
+                    if (std::is_integral<T>::value)
+                        return static_cast<T>(value + 0.5);
```
davisking: use `std::round()` instead of adding 0.5

dlib/pixel.h
```diff
@@ -125,6 +125,13 @@ namespace dlib
         unsigned char red;
         unsigned char green;
         unsigned char blue;
+
+        inline bool operator == (const rgb_pixel& that) const
+        {
+            return this->red   == that.red
+                && this->green == that.green
+                && this->blue  == that.blue;
+        }
```
davisking: I'm not sure about this one.  It creates a non-uniformity in the pixel types as many others don't allow ==.  Is this used in the code in this PR?

dlib/dnn/trainer.h
davisking: If I understand, the issue this is trying to address is that when we lower the learning rate, the code also kicks a bunch of values out of `previous_loss_values`, which then forces `loss_increased_since_last_disk_sync()` to return `false` for a while no matter what. I agree that that's not great and `loss_increased_since_last_disk_sync()` shouldn't ever be inhibited.

I think we can make this code a lot simpler though.  Recall that the reason lowering the learning rate kicked values out of `previous_loss_values` was just to prevent the learning rate from being immediately lowered again based on the same analysis of `previous_loss_values`. That was a crude way of doing it.  So rather than adding `additional_previous_loss_values_to_keep_until_disk_sync` and all the attendant extra logic, how about when the loss stalls we don't kick any values out of `previous_loss_values` but instead just update a counter that is used to measure the number of steps since the last time we lowered the learning rate.  That counter can then be used to inhibit execution of the code at line 800, by changing it from this:
```
                if (gradient_check_budget > iter_without_progress_thresh && learning_rate_shrink != 1)
 ```
to something like this:
```
                if (gradient_check_budget > iter_without_progress_thresh && learning_rate_shrink != 1 && steps_since_last_learning_rate_shrink > iter_without_progress_thresh)
 ```

This also means the code below there that does this:
```
                            // Empty out some of the previous loss values so that steps_without_progress 
                            // will decrease below iter_without_progress_thresh.  
                            for (unsigned long cnt = 0; cnt < previous_loss_values_dump_amount+iter_without_progress_thresh/10 && previous_loss_values.size() > 0; ++cnt)
                                previous_loss_values.pop_front();
```
would be replaced with something like 
```
steps_since_last_learning_rate_shrink -= previous_loss_values_dump_amount+iter_without_progress_thresh/10;
```

dlib/dnn/loss.h
reunanen: Wanted to avoid making unnecessary copies, and also relying on there being a working copy constructor. But agree that simpler is better – changed now.

(At any rate no need to worry about lifetimes I think, because these pointers clearly have a shorter life.)

dlib/dnn/loss.h
reunanen: Actually wanted to have that new stuff in a scope of its own. But extracted this now to an `operator <<`.

dlib/dnn/loss.h

dlib/image_transforms/interpolation.h
reunanen: Actually, I would rather not. In my experience `std::round()` can be much slower, at least on my machine (compiler). And in this kind of use it may really make a difference, I think.

Why is it slower? Guessing because of error handling, probably similar to `std::floor()` as discussed here: https://stackoverflow.com/a/824233/19254.

dlib/pixel.h
reunanen: Yes, it is used in the handling of PASCAL VOC 2012 classes, which have RGB codes.

If you don't want it here, I can move it to be "local" to the segmentation example code?

dlib/dnn/trainer.h
reunanen: > If I understand, the issue this is trying to address is that when we lower the learning rate, the code also kicks a bunch of values out of previous_loss_values, which then forces loss_increased_since_last_disk_sync() to return false for a while no matter what.

Correct.

> I think we can make this code a lot simpler though.

Thanks for the suggestion – will try it.

dlib/image_transforms/interpolation.h
davisking: Huh, that’s too bad. Yeah. Don’t use round then.

dlib/pixel.h
davisking: Na. I changed my mind. :)

Leave this here. It’s cool.

davisking: Although you don’t need the inline.  And add operator equal for the other pixel types in this file while here so things are basically uniform.

dlib/dnn/trainer.h
reunanen: Hmm. The current `master` code limits the length of `previous_loss_values` to no more than `iter_without_progress_thresh`. Now let's say `iter_without_progress_thresh` is relatively low, when compared to the disk sync interval. The PR, as it stands right now, addresses this issue also (and not only the explicit kick that may happen when the learning rate is shrunk).

dlib/pixel.h
reunanen: Fixed.

dlib/dnn/trainer.h
davisking: I’m not hugely worried about that case. The user is supposed to set the threshold such that it’s long enough to tell what the loss is doing. Either going up or down. 

Although, I would not be adverse to there being two separate loss histories. One used for the check if the loss is going up and another for the check about going down. There isn’t any reason why the one used for the going up check can’t be the entire length since last sync. And that would probably be better, as you point out. 

I just want the code to be simple. Using two totally separate loss histories seems like it would allow a simple implementation.

reunanen: > I’m not hugely worried about that case. The user is supposed to set the threshold such that it’s long enough to tell what the loss is doing. Either going up or down.

Well I already shot myself in the foot, and I'm certainly worried about doing it again. In particular when one thing (disk sync interval) is specified in wall-clock time, and the other (iterations without progress) in steps.

> I just want the code to be simple. Using two totally separate loss histories seems like it would allow a simple implementation.

Yeah, I too want it to be simple. Tried to avoid saving the values two times, but will happily backtrack.

davisking: Not a big deal to save extra stuff in the sync file. It happens rarely and the file never goes anywhere or gets stored long term. So don’t worry about putting redundant stuff in there. Whatever makes the code the simplest is probably the best. :)

reunanen: Simplified the code as discussed, and tested the result. Seems to be ok; pushed.

dlib/dnn/loss.h
```diff
@@ -993,6 +993,36 @@ namespace dlib
         }
     }
 
+    std::ostream& operator<<(std::ostream& out, const std::vector<mmod_options::detector_window_details>& detector_windows)
```
davisking: Global functions like this need to be declared inline if they are in header files, or linker errors will result if they are #included into more than one translation unit.

dlib/dnn/trainer.h
```diff
@@ -1020,6 +1027,23 @@ namespace dlib
...
+                        learning_rate = learning_rate_shrink * learning_rate;
+                        steps_without_progress = 0;
+                        test_steps_without_progress = 0;
+                        // Empty out some of the previous loss values so that steps_without_progress and test_steps_without_progress
+                        // will decrease below iter_without_progress_thresh.  
+                        for (unsigned long cnt = 0; cnt < previous_loss_values_dump_amount+iter_without_progress_thresh/10 && previous_loss_values.size() > 0; ++cnt)
+                            previous_loss_values.pop_front();
+                        for (unsigned long cnt = 0; cnt < test_previous_loss_values_dump_amount+test_iter_without_progress_thresh/10 && test_previous_loss_values.size() > 0; ++cnt)
+                            test_previous_loss_values.pop_front();
+                    }
```
davisking: This same bit of code it duplicated elsewhere.  Can you put this into a function (or actually two functions I think, one for the test and another for the normal loss values seems sensible) and call it instead of copying and pasting the code?

dlib/dnn/trainer.h
```diff
@@ -1057,34 +1081,39 @@ namespace dlib
             if (!std::ifstream(newest_syncfile(), std::ios::binary))
                 return false;
 
-            for (auto x : previous_loss_values)
+            // if we haven't seen much data yet then just say false.
+            if (gradient_updates_since_last_sync < 30)
+                return false;
```
davisking: Move this after the NaN check.  Definitely want to reload when you hit a NaN, even if there are fewer than 30 updates.

dlib/dnn/trainer.h
```diff
@@ -1057,34 +1081,39 @@ namespace dlib
...
-            if (gradient_updates_since_last_sync < 30 || previous_loss_values.size() < 2*gradient_updates_since_last_sync)
-                return false;
-
-            // Now look at the data since a little before the last disk sync.  We will
-            // check if the loss is getting bettor or worse.
             running_gradient g;
-            for (size_t i = previous_loss_values.size() - 2*gradient_updates_since_last_sync; i < previous_loss_values.size(); ++i)
-                g.add(previous_loss_values[i]);
+            for (size_t i = 0; i < previous_loss_values_to_keep_until_disk_sync.size(); ++i)
+                g.add(previous_loss_values_to_keep_until_disk_sync[i]);
```
davisking: Just make this part of the above loop that does the NaN check.  No need to repeat the loop over `previous_loss_values_to_keep_until_disk_sync` again.

examples/dnn_instance_segmentation_ex.cpp
davisking: Cool.  File is live at http://dlib.net/files/instance_segmentation_voc2012net.dnn now :)

examples/dnn_instance_segmentation_ex.cpp
```diff
@@ -0,0 +1,174 @@
...
+            );
+
+            dlib::matrix<uint16_t> resized_mask(
+                static_cast<int>(chip_details.rect.height()),
+                static_cast<int>(chip_details.rect.width())
+            );
+
+            dlib::resize_image(mask, resized_mask);
+
+            for (int r = 0; r < resized_mask.nr(); ++r)
```
davisking: Maybe put some {} on these loops :)

dlib/dnn/loss.h

dlib/dnn/trainer.h
reunanen: Sure; fixed. (Yeah, this is my preference also.)

dlib/dnn/trainer.h
reunanen: Good catch, fixed.

dlib/dnn/trainer.h
reunanen: Fixed.

examples/dnn_instance_segmentation_ex.cpp
reunanen: Added.

davisking:
I'm not sure what's going on with appveyor.  It seems to have gone sideways a few days ago unrelated to this PR.  But travis is erroring on this PR: https://travis-ci.org/davisking/dlib/jobs/611741333 due to non-C++11 stuff.

reunanen:
> But travis is erroring on this PR: https://travis-ci.org/davisking/dlib/jobs/611741333 due to non-C++11 stuff.

Right, [generic lambda appears to be C++14](https://stackoverflow.com/a/48653193/19254). Will fix.

reunanen:
Travis seems to be happier now: https://travis-ci.org/davisking/dlib/builds/611901132

davisking:
Sweet, thanks for another nice PR :)

PR #1600, opened 2018-12-18
Add U-net style skip connections to the semantic-segmentation example

reunanen:
To address the [recent interest](https://github.com/davisking/dlib/issues/1236#issuecomment-447581865) in a U-net implementation, this PR adds a possibly-useful `resize_to_prev` layer, and upgrades the semantic-segmentation example.

Compared to the corresponding network without skip connections (see #943), the updated example should be able to draw more details:

image

However, overall it's not too much more accurate on the PASCAL VOC 2012 data -- this time around, the training should finish like this (note the lower loss but not much better accuracy, when compared to numbers reported in #943):

```
step#: 117901  learning rate: 0.001  average loss: 0.0239372    steps without apparent progress: 4941
step#: 118018  learning rate: 0.001  average loss: 0.0238844    steps without apparent progress: 4878
step#: 118135  learning rate: 0.001  average loss: 0.0235096    steps without apparent progress: 4840
step#: 118252  learning rate: 0.0001  average loss: 0.0232595    steps without apparent progress: 4178
step#: 118369  learning rate: 0.0001  average loss: 0.0243033    steps without apparent progress: 4282
step#: 118486  learning rate: 0.0001  average loss: 0.0214507    steps without apparent progress: 4412
step#: 118603  learning rate: 0.0001  average loss: 0.0246106    steps without apparent progress: 4516
step#: 118720  learning rate: 0.0001  average loss: 0.0239061    steps without apparent progress: 4646
step#: 118837  learning rate: 0.0001  average loss: 0.0219395    steps without apparent progress: 4750
step#: 118954  learning rate: 0.0001  average loss: 0.0220728    steps without apparent progress: 4880
Saved state to pascal_voc2012_trainer_state_file.dat_
step#: 119063  learning rate: 0.0001  average loss: 0.0224262    steps without apparent progress: 4903
Saved state to pascal_voc2012_trainer_state_file.dat
saving network
Testing the network...
train accuracy  :  0.991385
val accuracy    :  0.851742
```

However, on some other datasets the U-net approach can be very useful. (I guess in particular when there are lots of details to draw, but the classification task itself is not too difficult.)

~Note that I still want to find a network architecture that is clean, performs well, and compiles in finite time using MSVC++ (see the comment below). So I'd prefer not to have this merged just yet. But at the same time wanted to make it available already, in case e.g. @goldbattle wants to try it out.~

PR #1938, opened 2019-12-02
Adding Mish activation function

thebhatman:
This PR adds the Mish Activation Function and resolves https://github.com/davisking/dlib/issues/1926


dlib/cuda/cpu_dlib.cpp
```diff
@@ -1467,6 +1467,48 @@ namespace dlib
...
+            const auto g = grad.host();
+            const auto d = dest.host();
+            const auto in = gradient_input.host();
+            if (is_same_object(gradient_input, grad))
+            {
+                for (size_t i = 0; i < dest.size(); ++i)
+                {
+                    auto delta = 2*std::exp(d[i]) + std::exp(2*d[i]) + 2;
+                    auto omega = 4*(d[i] + 1) + 4*std::exp(2*d[i]) + std::exp(3*d[i]) + std::exp(d[i])*(4*d[i] + 6);
+                    g[i] = in[i]*std::exp(d[i])*delta/(omega*omega);
```
davisking: This kind of thing is numerically unstable, in the same way that `log(1+exp(x))` is unstable. If the inputs are not small, exp() will produce inf which will then convert to NaN in the next lines, causing training to fail.  This kind of thing needs to be written in a way that doesn't cause values of `d` with modest values (e.g. a `d` of 5000) to yield NaN.

Take a look at the graph of this function: https://www.wolframalpha.com/input/?i=derivative+of+x*tanh%28log%281%2Bexp%28x%29%29%29. Note how the gradient saturates at 1 and 0.  Again, you can use if statements similar to how `dlib::log1pexp` handles this kind of thing to make the gradient computation numerically stable.

dlib/cuda/cudnn_dlibapi.cpp
```diff
@@ -162,6 +162,12 @@ namespace dlib
             return des.get_handle();
         }
 
+        static cudnnActivationDescriptor_t mish_activation_descriptor()
+        {
+            thread_local cudnn_activation_descriptor des(CUDNN_ACTIVATION_MISH, CUDNN_PROPAGATE_NAN,0);
```
davisking: This isn't available in all versions of cudnn. There should be some `#if` preprocessor check around this to allow old versions of cudnn to be used.  The simplest thing to do is to provide a CUDA method that implements this stuff and call that if cudnn is too old.  You can use a grid stride loop to do this easily by basically just copying and pasting your C++ code into the cuda kernel.  For example, this is the cuda code in dlib that implements prelu:

```
        __global__ void _cuda_prelu(const float* s, float* d, size_t n, const float* pp)
        {
            const float p = *pp; 
            for (auto i : grid_stride_range(0, n))
            {
                if (s[i] > 0) 
                    d[i] = s[i];
                else
                    d[i] = p*s[i];
            }
        }

        void prelu (
            tensor& dest,
            const tensor& src, 
            const tensor& param
        )
        {
            launch_kernel(_cuda_prelu, max_jobs(dest.size()), 
                src.device(), dest.device(), src.size(), param.device());
        }
```
Nothing tricky is required. Honestly it's simpler than calling a lot of the cuDNN functions.

dlib/cuda/cudnn_dlibapi.h
```diff
@@ -434,6 +434,41 @@ namespace dlib
...
+        );
+        /*!
+            requires
+                - have_same_dimensions(dest, src) == true
+            ensures
+                - for all valid i:
+                    - #dest.host()[i] == src.host()[i]*std::tanh(std::log(1+std::exp(src.host()[i])))
+                - This function supports in-place operation, i.e. having
+                  is_same_object(dest, src)==true
+        !*/
```
davisking: Delete this comment.

dlib/cuda/cudnn_dlibapi.h
```diff
@@ -434,6 +434,41 @@ namespace dlib
...
+                - have_same_dimensions(dest,grad) == true
+                - is_same_object(grad,dest) == false
+            ensures
+                - Recalling that dest is the output of mish(dest,SRC) for some SRC tensor,
+                  let f(SRC) == dot(gradient_input,dest)
+                - Then this function computes the gradient of f() with respect to SRC and
+                  assigns it to grad.
+                - This function supports in-place operation, i.e. having
+                  is_same_object(grad, gradient_input)==true
+        !*/
```
davisking: Delete this comment.  

I just noticed that these comments are here for some of these functions (this one is copied and pasted from another one obviously) and they are wrong.  The real ones are in tensor_tools.h, which have the current correct docs.  I just deleted all these from cudnn_dlibapi.h as they are duplicative with tensor_tools.h and often wrong and outdated :(

dlib/cuda/tensor_tools.cpp
```diff
@@ -826,6 +826,33 @@ namespace dlib { namespace tt
 #endif
     }
 
+// ----------------------------------------------------------------------------------------
+
+        void mish (
```
davisking: Fix indenting

dlib/cuda/cudnn_dlibapi.h
thebhatman: I have removed the comments now.

thebhatman:
@davisking Thank you for an detailed review. I have added the test for Mish, and simplified the computation of Mish. I am yet to simplify the gradient computation and add the cuda kernel. I will be doing that soon. The unnecessary comments have been removed and indentation has been fixed.

thebhatman:
I have made the gradients zero when `x > 8` and `x < -8` . Is the current implementation numerically stable enough?

thebhatman:
I had initially written a kernel for mish_gradient by looking at the code for prelu_gradient. pgrad isn't required for Mish right? As it doesn't take in any parameters. So just the cuda_mish kernel should suffice right? I have removed the kernel for mish_gradient.

davisking:
> I had initially written a kernel for mish_gradient by looking at the code for prelu_gradient. pgrad isn't required for Mish right? As it doesn't take in any parameters. So just the cuda_mish kernel should suffice right? I have removed the kernel for mish_gradient.

Right, mish doesn't have any parameters so there is no parameter gradient.

thebhatman:
Is there anything more to be done @davisking ? Thanks!

davisking:
Sorry, I thought you were still working on it. I’ll look at this in a little bit.

davisking:
I just ran the unit tests (via `./dtest  --test_dnn`) with dlib built without cuda and got this error:
```
Error occurred at line 1839.
Error occurred in file /home/davis/source/dlib/dlib/test/dnn.cpp.
Failing expression was res.
Gradient error in data variable #0.  Relative error: 0.970753
expected derivative: 1.54417
output derivative:   0.045163
iteration:           0
```

dlib/cuda/cudnn_dlibapi.cpp
davisking: This needs to have #ifdef around it like I noted earlier.  For instance, I just tried to build this with cudnn 10.1 and it failed to build because this feature is not available.

dlib/cuda/cuda_dlib.cu
```diff
@@ -1351,6 +1351,28 @@ namespace dlib
...
         }
 
+    // ----------------------------------------------------------------------------------------
+
+        __global__ void _cuda_mish(const float* s, float* d, size_t n, const float* pp)
+        {
+            const float p = *pp;
+            for (auto i : grid_stride_range(0, n))
+            {
+                auto delta = 2*std::exp(s[i]) + std::exp(2*s[i]) + 2;
```
davisking: This is more efficiently represented as:
```
    auto e = std::exp(s[i]);
    auto delta = 2*e + e*e + 2;
```

dlib/cuda/cudnn_dlibapi.cpp
davisking: Really though, it's easier to just delete the cudnn calls and use your own cuda code.  You already have the mish function, you just need the gradient and you are good to go.

dlib/test/dnn.cpp
```diff
@@ -1832,6 +1832,12 @@ namespace
             auto res = test_layer(l);
             DLIB_TEST_MSG(res, res);
         }
+        {
+            print_spinner();
+            mish_ l;
+            auto res = test_layer(l);
+            DLIB_TEST_MSG(res, res);
+        }
```
davisking: On second thought, please also add a test that `cuda::mish()` matches the outputs of `cpu::mish()`.  For example, like is done for `cuda::softmax_all()` in this file as well.

dlib/cuda/cudnn_dlibapi.cpp
reunanen: Where is `CUDNN_ACTIVATION_MISH` even available? Downloaded latest cuDNN, but couldn't find anything.

thebhatman: Yeah it's not available. I need to change this and use the cuda code only.

reunanen:
I wanted to test Mish, so tried to fix the remaining issues in [this PR](https://github.com/thebhatman/dlib/pull/1). (It did take me quite a while to debug the test failure.)

@thebhatman, please consider merging my PR to your branch (I guess this PR will then update automatically).

thebhatman:
@reunanen Thanks a lot for your effort and for helping me out.

dlib/test/dnn.cpp
reunanen: Added a simple test.

reunanen:
> I just ran the unit tests (via ./dtest --test_dnn) with dlib built without cuda and got this error:

@davisking , this should now work.

dlib/cuda/cudnn_dlibapi.cpp
reunanen: Fixed by adding a standalone CUDA implementation.

dlib/cuda/cuda_dlib.cu
reunanen: Fixed.

reunanen:
FYI, I [updated the semantic-segmentation example to use Mish](https://github.com/reunanen/dlib/commit/667c2ad764e360239c39eb87a512a7a9b350ce84) (instead of ReLU). The results are pretty much the same as those reported in #1600 – I wouldn't assume that the small differences are statistically significant:

image

```
step#: 119268  learning rate: 0.0001  average loss: 0.0229698    steps without apparent progress: 4026
step#: 119375  learning rate: 0.0001  average loss: 0.0237883    steps without apparent progress: 4130
step#: 119482  learning rate: 0.0001  average loss: 0.0235407    steps without apparent progress: 4234
step#: 119591  learning rate: 0.0001  average loss: 0.0246943    steps without apparent progress: 4364
Saved state to pascal_voc2012_trainer_state_file.dat
step#: 119693  learning rate: 0.0001  average loss: 0.0214533    steps without apparent progress: 1016
step#: 119796  learning rate: 0.0001  average loss: 0.0256509    steps without apparent progress: 4572
step#: 119905  learning rate: 0.0001  average loss: 0.0234571    steps without apparent progress: 4676
step#: 120014  learning rate: 0.0001  average loss: 0.0250673    steps without apparent progress: 4780
step#: 120123  learning rate: 0.0001  average loss: 0.0235702    steps without apparent progress: 4884
step#: 120232  learning rate: 0.0001  average loss: 0.0233183    steps without apparent progress: 4988
Saved state to pascal_voc2012_trainer_state_file.dat_
saving network
Testing the network...
train accuracy  :  0.991318
val accuracy    :  0.847972
```

davisking:
Well, I merged this pull request.  But did it in a way github doesn't like apparently.  It's merged though ¯\_(ツ)_/¯

davisking:
It sort of bothers me that github is saying this isn't merged.  I don't want to close it and make it look like it wasn't merged.  You want to try rebasing and maybe I can get github to mark it as merged?

thebhatman:
I have resolved the merge conflicts. Maybe you can merge the PR now. Or maybe a rebase on master by dropping the previous merge commits and merging this might help.

davisking:
Ha, it’s fine. Merged it :)

Issue #1949, opened 2019-12-19
cv_image issue with OpenCV 4.0 +


mipko:
This is the same issue as #1492 

## Expected Behavior
To be able to pass cv::Mat in cv_image constructor

## Current Behavior
Compilation error with OpenCV version 4.0.0 and above. 

`/dlib/opencv/cv_image.h:37:22: **error**: conversion from ‘const cv::Mat’ to non-scalar type ‘IplImage’ {aka ‘_IplImage’} requested
             IplImage temp = img;
`

## Steps to Reproduce
try to construct cv_image from cv::Mat

## FIX
the same as in #1492 

* **Version**: 19.19 
* **Where did you get dlib**: official site  
* **Platform**:  Ubuntu 19.04 
* **Compiler**:  gcc 8.3.0

davisking:
You need to post more detailed instructions because I am not able to reproduce this.  See also this recent thread: https://github.com/davisking/dlib/pull/1494#issuecomment-565770272 where someone asked the same question, only later to report that this isn't an issue.

mipko:
Thank you very much for the reply.

I will try to explain it a bit detailed.

This is code snippet from my C++ project:

```
 dlib::array2d<unsigned char> dlib_img;
 dlib::assign_image(dlib_img, dlib::cv_image<unsigned char>(_mat));
```
within the **dlib::cv_image** constructor there is a statement

`IplImage temp = img;`

The above statement  causes compilation error with OpenCV 4.x.y.

To fix the problem I have changed the above line from cv_image constructor to this:

`temp = cvIplImage(img);`

I have just cloned the master branch from the OpenCV github.
I have installed dlib 19.19.

davisking:
Thanks for the report.  I just pushed a change to github that should fix this for all versions of OpenCV.

mipko:
Thank you very much for the fix. Everything is OK now.

amir-saniyan:
For resolve compilation error, the following code could be used:

    IplImage ipl_img = cvIplImage(image);
    dlib::cv_image<dlib::bgr_pixel> cimg(&ipl_img);

nwrkbiz:
@amir-saniyan 

or simply:
```
dlib::assign_image(dlib_img, dlib::cv_image<unsigned char>(cvIplImage(ocv_mat)));
``

PR #1958, opened 2019-12-28
Fix cv_image '=' overload for Mat in latest OpenCV

jeffleo:
This addresses same issue as #1949, but for cv_image '=' operator overloading cv::Mat images.

**For example:**
cv_image<bgr_pixel> cimg;
cv::Mat img;
cimg = img;

### DIFF
--- a/dlib/opencv/cv_image.h
+++ b/dlib/opencv/cv_image.h
@@ -142,7 +142,11 @@ namespace dlib
 
         cv_image& operator=( const cv::Mat img)
         {
+#if CV_MAJOR_VERSION > 3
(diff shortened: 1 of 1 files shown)


davisking:
Oh yeah, I forgot about that part.  Thanks for posting this.  There is a simpler solution though, which I just pushed to github, so this PR is not needed now. Thanks for submitting it regardless though :)

jeffleo:
I think it was only a build error because I was actually using the = operator overloading. Did you have to remove that feature?

davisking:
Cool. You can still use operator = if you want to though.

PR #1963, opened 2020-01-06
Fix error for opencv 3.4.9+ over IplImage (#1949)

jeffeDurand:


davisking:
Huh, I've used OpenCV 4 for instance, and a few other versions newer than what this diff sets things to, and it always worked. Near as I could tell, only the very newest OpenCV revision on github needs this.  How did you get it to trip on older versions?

jeffeDurand:
Hi, opencv 3.4.9 is the newest version and introduced the same problem has
with 4.x. (issue 1949). I needed a fix I introduced in it on december and
was surprised by the dlib error in the opencv header.

I used the official revision macro (9+) to only allow newer version of the
3.4.x branch from now on.

I hope this satisfy your curiosity!

JF




Le mar. 7 janv. 2020 21 h 12, Davis E. King <notifications@github.com> a
écrit :

davisking:
Thanks.  I didn't realize they were pushing the same change back into the 3.x versions.

PR #1964, opened 2020-01-07
Add minor sanity check

reunanen:
Add minor sanity check in wake of trainer changes introduced as part of #1918.


davisking:
Nice, thanks for the PR :)

PR #1965, opened 2020-01-09
Fix vector normalizer doc

arrufat:
I've just noticed that the `vector_normalizer` documentation mentions the PCA, which only applies to `vector_normalizer_pca`, so I deleted that part of the documentation.


davisking:
Oh yeah, not sure how that got there.  Thanks for the PR :)

PR #1976, opened 2020-01-15
 Add new loss layer for binary loss per pixel

reunanen:
- Add new loss layer for binary loss per pixel (say, for completeness)
- In the instance-segmentation example, resolve possibly overlapping instances using the per-pixel output intensity (and not result rect size as previously); while at it, also make the instance output look "soft":
image

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1628,11 +1628,11 @@ namespace dlib
 
         __device__ float cuda_log1pexp(float x)
         {
-            if (x <= -37)
+            if (x <= -18)
                 return std::exp(x);
-            else if (-37 < x && x <= 18)
+            else if (-18 < x && x <= 9)
                 return std::log1p(std::exp(x));
-            else if (18 < x && x <= 33.3)
+            else if (9 < x && x <= 16)
                 return x + std::exp(-x);
             else
                 return x;
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1766,7 +1766,7 @@ namespace dlib
 
         void compute_loss_binary_log_per_pixel::
         do_work(
-            cuda_data_void_ptr loss_cuda_work_buffer,
+            cuda_data_ptr<float> loss_cuda_work_buffer,
             const float* truth_buffer,
             const tensor& subnetwork_output,
             tensor& gradient,
@@ -1780,7 +1780,7 @@ namespace dlib
             const double scale = 1.0 / (subnetwork_output.num_samples() * subnetwork_output.nr() * subnetwork_output.nc());
 
             launch_kernel(_cuda_compute_loss_binary_log_per_pixel, max_jobs(gradient.size()),
-                static_cast<float*>(loss_cuda_work_buffer.data()), gradient.device(), truth_buffer, subnetwork_output.device(), gradient.size(), scale);
+                loss_cuda_work_buffer.data(), gradient.device(), truth_buffer, subnetwork_output.device(), gradient.size(), scale);
 
             float floss;
             dlib::cuda::memcpy(&floss, loss_cuda_work_buffer, sizeof(float));
@@ -1789,7 +1789,7 @@ namespace dlib
 
         void compute_loss_multiclass_log_per_pixel::
         do_work(
-            cuda_data_void_ptr loss_cuda_work_buffer,
+            cuda_data_ptr<float> loss_cuda_work_buffer,
             const uint16_t* truth_buffer,
             const tensor& subnetwork_output,
             tensor& gradient,
@@ -1804,7 +1804,7 @@ namespace dlib
             const double scale = 1.0 / (subnetwork_output.num_samples() * subnetwork_output.nr() * subnetwork_output.nc());
 
             launch_kernel(_cuda_compute_loss_multiclass_log_per_pixel, max_jobs(gradient.size()),
-                static_cast<float*>(loss_cuda_work_buffer.data()), gradient.device(), truth_buffer, gradient.size(), gradient.nr()*gradient.nc(), gradient.nr()*gradient.nc()*gradient.k(), gradient.k(), label_to_ignore, scale);
+                loss_cuda_work_buffer.data(), gradient.device(), truth_buffer, gradient.size(), gradient.nr()*gradient.nc(), gradient.nr()*gradient.nc()*gradient.k(), gradient.k(), label_to_ignore, scale);
 
             float floss;
             dlib::cuda::memcpy(&floss, loss_cuda_work_buffer, sizeof(float));
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1709,7 +1709,7 @@ namespace dlib
 
         void compute_loss_binary_log_per_pixel::
         do_work(
-            float* loss_cuda_work_buffer,
+            cuda_data_void_ptr loss_cuda_work_buffer,
             const float* truth_buffer,
             const tensor& subnetwork_output,
             tensor& gradient,
@@ -1723,16 +1723,16 @@ namespace dlib
             const double scale = 1.0 / (subnetwork_output.num_samples() * subnetwork_output.nr() * subnetwork_output.nc());
 
             launch_kernel(_cuda_compute_loss_binary_log_per_pixel, max_jobs(gradient.size()),
-                loss_cuda_work_buffer, gradient.device(), truth_buffer, subnetwork_output.device(), gradient.size(), scale);
+                static_cast<float*>(loss_cuda_work_buffer.data()), gradient.device(), truth_buffer, subnetwork_output.device(), gradient.size(), scale);
 
             float floss;
-            CHECK_CUDA(cudaMemcpy(&floss, loss_cuda_work_buffer, sizeof(float), cudaMemcpyDefault));
+            dlib::cuda::memcpy(&floss, loss_cuda_work_buffer, sizeof(float));
             loss = scale*floss;
         }
 
         void compute_loss_multiclass_log_per_pixel::
         do_work(
-            float* loss_cuda_work_buffer,
+            cuda_data_void_ptr loss_cuda_work_buffer,
             const uint16_t* truth_buffer,
             const tensor& subnetwork_output,
             tensor& gradient,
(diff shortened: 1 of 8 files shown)


dlib/cuda/cuda_dlib.cu
```diff
@@ -1624,6 +1624,48 @@ namespace dlib
             }
         }
 
+    // ----------------------------------------------------------------------------------------
+
+        __device__ float cuda_log1pexp(float x)
```
reunanen: This implementation was copied from [utilities.h](https://github.com/davisking/dlib/blob/d766f5e82e650fb2b24bbbb8c5d9322a9fcef36c/dlib/dnn/utilities.h#L16-L29), but I'm not sure if the magic numbers should be different for `float` input?

examples/dnn_instance_segmentation_ex.cpp
```diff
@@ -16,7 +16,7 @@
        ./dnn_instance_segmentation_ex /path/to/VOC2012-or-other-images
 
     An alternative to steps 2-4 above is to download a pre-trained network
-    from here: http://dlib.net/files/instance_segmentation_voc2012net.dnn
+    from here: http://dlib.net/files/instance_segmentation_voc2012net_v2.dnn
```
reunanen: For the time being at least, this file is available here:
https://www.dropbox.com/s/t7gi4lfw6mrx10v/instance_segmentation_voc2012net_v2.dnn

dlib/test/dnn.cpp
```diff
@@ -3396,6 +3592,9 @@ namespace
             test_multioutput_linear_regression();
             test_simple_autoencoder();
             test_loss_mean_squared_per_channel_and_pixel();
+            test_loss_binary_log_per_pixel_learned_params_on_trivial_two_pixel_task();
+            test_loss_binary_log_per_pixel_outputs_on_trivial_task();
+            test_loss_binary_log_per_pixel_with_noise_and_pixels_to_ignore();
```
reunanen: Not sure if all these tests are really useful...

examples/dnn_instance_segmentation_ex.cpp
```diff
@@ -71,26 +71,21 @@ int main(int argc, char** argv) try
     {
         // Load the input image.
         load_image(input_image, file.full_name());
-
-        // Draw largest objects last
```
reunanen: No need to sort the detections anymore.

dlib/cuda/cuda_dlib.cu
reunanen: Looking at this, I guess the answer is **yes**: https://github.com/JuliaStats/StatsFuns.jl/blob/0beb42294785787ac58fec247cd5d225a99e6d0f/src/basicfuns.jl#L67

Btw, the first Google hit was [this question](https://cs.stackexchange.com/questions/110798/numerically-stable-log1pexp-calculation) :)

reunanen: Fixed.

dlib/cuda/cuda_dlib.cu
```diff
@@ -1663,6 +1705,30 @@ namespace dlib
...
+            sigmoid(gradient, subnetwork_output);
+
+            // The loss we output is the average loss over the mini-batch, and also over each element of the matrix output.
+            const double scale = 1.0 / (subnetwork_output.num_samples() * subnetwork_output.nr() * subnetwork_output.nc());
+
+            launch_kernel(_cuda_compute_loss_binary_log_per_pixel, max_jobs(gradient.size()),
+                loss_cuda_work_buffer, gradient.device(), truth_buffer, subnetwork_output.device(), gradient.size(), scale);
+
+            float floss;
+            CHECK_CUDA(cudaMemcpy(&floss, loss_cuda_work_buffer, sizeof(float), cudaMemcpyDefault));
```
davisking: use `dlib::cuda::memcpy()` instead of `cudaMemcpy()`

dlib/cuda/cuda_dlib.h
```diff
@@ -412,6 +412,62 @@ namespace dlib
 
     // ----------------------------------------------------------------------------------------
 
+        class compute_loss_binary_log_per_pixel
+        {
+            /*!
+                The point of this class is to compute the loss computed by
+                loss_binary_log_per_pixel, but to do so with CUDA.
```
davisking: Should reference `loss_binary_log_per_pixel_` not `loss_binary_log_per_pixel`

dlib/cuda/cuda_dlib.h
```diff
@@ -412,6 +412,62 @@ namespace dlib
...
+                double& loss
+            ) const
+            {
+                const size_t bytes_per_plane = subnetwork_output.nr()*subnetwork_output.nc()*sizeof(float);
+                // Allocate a cuda buffer to store all the truth images and also one float
+                // for the scalar loss output.
+                buf = device_global_buffer(subnetwork_output.num_samples()*bytes_per_plane + sizeof(float));
+
+                cuda_data_void_ptr loss_buf = buf;
+                buf = buf+sizeof(float);
```
davisking: I just realized there isn't any method to cast a `cuda_data_void_ptr` to a `cuda_data_ptr<float>`, which is really what should happen here :|

davisking:
Got to take my kid to school.  I'll look more at this later.

dlib/cuda/cuda_dlib.cu
reunanen: Fixed – also at the source where this was basically copied from.

dlib/cuda/cuda_dlib.h
reunanen: Fixed – also at the source where this was copied from.

dlib/test/dnn.cpp
```diff
@@ -2555,6 +2555,202 @@ namespace
...
+        trainer.set_max_num_epochs(200);
+        trainer.train(x, y);
+
+        // The learning task is easy, so the net should have no problem
+        // getting all the outputs right.
+        const auto response = net(x);
+        for (int ii = 0; ii < num_samples; ++ii)
+            for (int jj = 0; jj < output_height; ++jj)
+                for (int kk = 0; kk < output_width; ++kk)
+                    DLIB_TEST((response[ii](jj,kk) > 0) == (y[ii](jj,kk) > 0));
```
davisking: I'm getting a failure here when I run this test with dlib built to use cuda.  Is it working for you?

dlib/cuda/cuda_dlib.h
davisking: I just pushed `static_pointer_cast()`.  So you can do this now:
```
cuda_data_ptr<float> loss_buf = static_pointer_cast<float>(buf, 1);
```
and `loss_buf.get() == buf.get()` and `loss_buf.size() == 1`.  Then you can use the nicely typed loss_buf rather than void pointers.

dlib/test/dnn.cpp
reunanen: Yes, it works on my machine (CUDA build).

Tried dozens of times now, but couldn't get a failure.

Doubled the number of epochs to train. Does that help?

dlib/cuda/cuda_dlib.h
reunanen: This should look better now.

dlib/test/dnn.cpp
arrufat: I am also getting an error with 400 epochs. Doubling it works for me, but also setting it 500.

reunanen: Thanks @arrufat. On my machine it worked with 200, and the 400 was already a doubled-up number. Now increased further to 800.

dlib/cuda/cuda_dlib.h
```diff
@@ -424,11 +424,67 @@ namespace dlib
...
+                }
+
+                do_work(loss_buf, static_cast<float*>(buf.data()), subnetwork_output, gradient, loss);
+            }
+
+        private:
+
+            static void do_work(
+                cuda_data_ptr<float> loss_cuda_work_buffer,
+                const float* truth_buffer,
```
davisking: Can `truth_buffer` be a cuda_data_ptr too?  The way the code appears suggests `truth_buffer` is a host pointer.

dlib/cuda/cuda_dlib.cu
```diff
@@ -1720,10 +1762,34 @@ namespace dlib
...
+            float floss;
+            dlib::cuda::memcpy(&floss, loss_cuda_work_buffer);
+            loss = scale*floss;
+        }
 
         void compute_loss_multiclass_log_per_pixel::
         do_work(
-            float* loss_cuda_work_buffer,
+            cuda_data_ptr<float> loss_cuda_work_buffer,
             const uint16_t* truth_buffer,
```
davisking: Can `truth_buffer` be a cuda_data_ptr too?  The way the code appears suggests `truth_buffer` is a host pointer.

dlib/cuda/cuda_dlib.cu
```diff
@@ -1720,10 +1762,34 @@ namespace dlib
...
+
+            float floss;
+            dlib::cuda::memcpy(&floss, loss_cuda_work_buffer);
+            loss = scale*floss;
+        }
 
         void compute_loss_multiclass_log_per_pixel::
         do_work(
-            float* loss_cuda_work_buffer,
+            cuda_data_ptr<float> loss_cuda_work_buffer,
```
davisking: Can `truth_buffer` be a cuda_data_ptr too?  The way the code appears suggests `truth_buffer` is a host pointer, since `loss_cuda_work_buffer` is so "hey I'm a cuda buffer!" but `truth_buffer` isn't.  

Also, `loss_cuda_work_buffer` should be named something like `loss_work_buffer`, the type is making it clear it's a cuda buffer.

dlib/cuda/cuda_dlib.h
```diff
@@ -464,13 +520,13 @@ namespace dlib
...
 
-                do_work(static_cast<float*>(loss_buf.data()), static_cast<uint16_t*>(buf.data()), subnetwork_output, gradient, loss);
+                do_work(loss_buf, static_cast<uint16_t*>(buf.data()), subnetwork_output, gradient, loss);
             }
 
         private:
 
             static void do_work(
-                float* loss_cuda_work_buffer,
+                cuda_data_ptr<float> loss_cuda_work_buffer,
```
davisking: Same comments here about cuda naming and `truth_buffer` (i.e. it should be a cuda_data_ptr too so the arguments are consistent)

dlib/test/dnn.cpp
davisking: Cool, working for me now too :)

dlib/cuda/cuda_dlib.cu
reunanen: > Can truth_buffer be a cuda_data_ptr too?

I did think about it, but wasn't sure if I can make it `const` (meaning that the data pointed to won't be changed – obviously the pointer object argument itself can be made `const`).

reunanen: > Also, `loss_cuda_work_buffer` should be named something like `loss_work_buffer`, the type is making it clear it's a cuda buffer.

Fixed.

reunanen: >> Can truth_buffer be a cuda_data_ptr too?

> I did think about it, but wasn't sure if I can make it const

Looking at how [these](https://github.com/davisking/dlib/blob/6bdd289f734839200226de525b78046268905cff/dlib/cuda/cuda_data_ptr.h#L68-L71) are defined, I guess constness is preserved if I pass a `const cuda_data_ptr<T>&`. (?)

davisking:
cuda_data_ptr<const T> should do the trick. Since then the constness will stick through any copying.

reunanen:
> cuda_data_ptr<const T> should do the trick.

Right. Thought we can't do this, because we need to populate the truth buffer as well – but of course that need not happen through this same pointer object. :)

davisking:
Thanks for another nice PR :)

PR #1977, opened 2020-01-16
add visitor to count net parameters

arrufat:
I've just added a simple visitor to count the parameters of a network.

Results for a ResNet 50:
- dlib:
``` c++
count_parameters(net);  // 22803176
```
- pytorch: 
``` python
sum(p.numel() for p in model.parameters() if p.requires_grad)  # 25557032
```
- keras
``` python
model.count_params()  # 25636712
```

There are some discrepancies, I think I'm doing it right, according to the `visit_layer_parameters` documentation and my ResNet 50 implementation.

### DIFF
--- a/dlib/dnn/utilities_abstract.h
+++ b/dlib/dnn/utilities_abstract.h
@@ -120,6 +120,20 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <typename net_type>
+    inline size_t count_parameters(
+        const net_type& net
+    );
+    /*!
+        requires
+            - net_type is an object of type add_layer, add_loss_layer, add_skip_layer, or
+              add_tag_layer.
+        ensures
+            - This function returns the number of parameters of net if it has been properly
+              initialized and 0 otherwise.
+    !*/
+
+// ----------------------------------------------------------------------------------------
 }
 
 #endif // DLIB_DNn_UTILITIES_ABSTRACT_H_ 
--- a/dlib/dnn/utilities.h
+++ b/dlib/dnn/utilities.h
@@ -273,6 +273,35 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    namespace impl
+    {
+        class visitor_count_parameters
+        {
+        public:
+            visitor_count_parameters(size_t& num_parameters_): num_parameters(num_parameters_) {}
+
+            void operator()(size_t, const tensor& t)
+            {
+                num_parameters += t.size();
+            }
+
+        private:
+            size_t& num_parameters;
+        };
+    }
+
+    template <typename net_type>
+    inline size_t count_parameters(
+        const net_type& net
+    )
+    {
+        size_t num_parameters = 0;
+        impl::visitor_count_parameters temp(num_parameters);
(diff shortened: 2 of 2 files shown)


dlib/dnn/utilities_abstract.h
```diff
@@ -120,6 +120,20 @@ namespace dlib
...
+    template <typename net_type>
+    inline size_t count_parameters(
+        const net_type& net
+    );
+    /*!
+        requires
+            - net_type is an object of type add_layer, add_loss_layer, add_skip_layer, or
+              add_tag_layer.
+        ensures
+            - This function returns the number of parameters of net if it has been properly
```
davisking: How about, "this function returns the number of allocated parameters in the net.  E.g. if the network has not been trained then since nothing has been allocated yet it will return 0." instead.

davisking:
Thanks for the PR :)

dlib/dnn/utilities_abstract.h
arrufat: Oh, that's sounds better. I will submit a PR later if you don't do it before (I don't have my computer with me right now)

arrufat: Too late, thank you :)

arrufat:
I don't know if this is the right place to put this, but anyway, I will put it here for reference.

I was a bit concerned about the differences in the number of trainable parameters.
Then, I found out the cause: when performing the downsampling in the residual block, I was using an `avg_pool` layer, so it looked like this:

``` c++
    // a resnet residual that does subsampling on both paths
    template<
        template<long, template<typename> class, int, typename> class BLOCK, // basicblock or bottleneck
        long num_filters,
        template<typename> class BN, // regularization: bn_con or affine
        typename SUBNET
    >
    using residual_down = dlib::add_prev2<dlib::avg_pool<2, 2, 2, 2,
                          dlib::skip1<dlib::tag2<BLOCK<num_filters, BN, 2,
                          dlib::tag1<SUBNET>>>>>>;
```

But Pytorch and Keras they both use a 1x1 convolutional layer with a 2x2 stride, followed by a batch norm layer, so I modified the `residual_down` to look like this:

``` c++
    // a resnet residual that does subsampling on both paths
    template<
        template<long, template<typename> class, int, typename> class BLOCK, // basicblock or bottleneck
        long num_filters,
        template<typename> class BN, // regularization: bn_con or affine
        typename SUBNET
    >
    using residual_down = dlib::add_prev2<BN<dlib::con<4 * num_filters, 1, 1, 2, 2,
                          dlib::skip1<dlib::tag2<BLOCK<num_filters, BN, 2,
                          dlib::tag1<SUBNET>>>>>>>;
```

Note the `4`, which is the expansion factor for bottlenecks used in ResNet(50, 101, 151). The expansion factor is `1` in ResNet{18, 34}.

With that in mind, the number of parameters are:

- dlib: 25,566,440
- pytorch: 25,557,032
- keras: 25,583,592

Also note that in the previous comment I was counting all parameters in the keras implementation, instead of just the trainable ones (which I realized when printing `model.summary()`).

Anyway, I just wanted to be sure that both the PR and my ResNet implementation were correct.

PR #1979, opened 2020-01-16
Add vcpkg installation instructions

PhoebeHui:
Dlib is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build dlib, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for dlib and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,11 @@ mkdir build; cd build; cmake .. -DUSE_AVX_INSTRUCTIONS=1; cmake --build .
 
 Doing so will make some things run faster.
 
+Alternatively, if you are using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager you can download and install dlib with CMake integration in a single command:
+```bash
+vcpkg install dlib
+```
+
 Finally, Visual Studio users should usually do everything in 64bit mode.  By default Visual Studio is 32bit, both in its outputs and its own execution, so you have to explicitly tell it to use 64bits.  Since it's not the 1990s anymore you probably want to use 64bits.  Do that with a cmake invocation like this:
 ```bash
 cmake .. -G "Visual Studio 14 2015 Win64" -T host=x64 


README.md
```diff
@@ -21,6 +21,11 @@ mkdir build; cd build; cmake .. -DUSE_AVX_INSTRUCTIONS=1; cmake --build .
 
 Doing so will make some things run faster.
 
+Alternatively, if you are using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager you can download and install dlib with CMake integration in a single command:
+```bash
+vcpkg install dlib
+```
+
```
davisking: Thanks for the PR.  Isn't the next paragraph somewhat contrary to this?  If someone is using vcpkg they are almost certainly using visual studio correct?  They would then not run cmake as the next paragraph indicates?  This should be clarified and this new text needs to be coherent with the surrounding text. At least by making it clear that this is an either or thing.  Either cmake, or alternatively vcpkg when using visual studio.

PhoebeHui: @davisking, thanks for your reviewing!

VCPKG support to build dlib on Windows, Linux ans OSX. On windows, it use visual studio to build dlib, so if someone using vcpkg, it's not only using visual studio, but also gcc or clang, I considered to add it separately, could you give me some guidelines that where should I add the instructions?

davisking:
I moved the text down a little to a more appropriate place, and it's live on https://github.com/davisking/dlib now :)

PhoebeHui:
@davisking, thanks for your help!

PR #1986, opened 2020-01-28
If nearest-neighbor interpolation is wanted, then don't use an image pyramid

reunanen:
I want to extract image chips using nearest-neighbor interpolation, for segmentation masks and such. (A bicycle still isn't halfway between an aeroplane and a bird.)


davisking:
Nice.

PR #1987, opened 2020-01-28
Do not link to libnsl

jschueller:
Dlib does not use nsl symbols, why was this necessary ?
This make conda-forge build fail


davisking:
This was once upon a time needed on old Unix systems to use part of the BSD sockets api. I would be a little surprised if it’s still needed anywhere. So if it’s causing trouble then it can go. 

Thanks for the PR.

PR #1989, opened 2020-01-30
remove python 2 from travis

arrufat:
python 2 is no longer supported as of Jan 1st 2020, furthermore, it is causing travis checks not to pass.


arrufat:
I clearly don't know how to update my fork of dlib properly... sorry about that...

davisking:
The travis checks seem to be running fine with 2.7, but the appveyor ones are exploding as of a week ago or so. I was hoping appveyor would get fixed, but you are right.  Python 2.7 is done.  I just removed the python 2.7 tests. :)

PR #1991, opened 2020-02-04
Add dnn_introduction3_ex

arrufat:
In this PR I submit a small example of some advance features of the deep learning tools in dlib.

No new features were added, I just exemplified how to create a custom network visitor to modify the learning rates of a network for fine-tuning on other tasks. I've seen some people opening issues on how to do this.

I also provided a file that contains a generic ResNet implementation for all common architectures (18, 34, 50, 101, 152) in under 100 lines. The main motivation was to avoid repeating the network definition with `bn_con` and `affine`.

I hope you like it and people find it useful :)

Thank you!


arrufat:
I was wondering, since the `visitor_learning_rate_multiplier` might be useful, would you like to include in dlib instead of the example?

examples/dnn_introduction3_ex.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    // fully connected layers (with and without bias)
+    template<unsigned long num_outputs, fc_bias_mode bias_mode>
+    void set_learning_rate_multiplier(fc_<num_outputs, bias_mode>& l) const
+    {
+        l.set_learning_rate_multiplier(new_learning_rate_multiplier);
+        if (bias_mode == FC_HAS_BIAS)
+            l.set_bias_learning_rate_multiplier(new_learning_rate_multiplier);
+    }
```
davisking: Instead of explicitly listing the layers that have learning rate multipliers, you could do what `dlib::get_learning_rate_multiplier()` does.  In fact, defining a `dlib::set_learning_rate_multiplier()` and just calling that would be good.  Same for the bias. I just looked and I'm a little surprised I left these set methods out :|

examples/dnn_introduction3_ex.cpp
```diff
@@ -0,0 +1,180 @@
...
+        net.subnet().subnet() = backbone;
+
+        // An alternative way to use the pretrained network on a different network is
+        // to extract the relevant part of the network (we remove loss and fc layers),
+        // stack the new layers on top of it and assign the network.
+        using net_type = loss_metric<fc_no_bias<128, decltype(backbone)>>;
+        net_type net2;
+        net2.subnet().subnet() = backbone;
+    }
+    cin.get();
```
davisking: Remove `cin.get()`?

examples/dnn_introduction3_ex.cpp
```diff
@@ -0,0 +1,180 @@
...
+    // We can use the visit_layers function to modify the learning rate of the whole network:
+    visit_layers(net, visitor_learning_rate_multiplier(0.01));
+
+    // Usually, we want to freeze the network, except for the top layers:
+    visit_layers(net.subnet().subnet(), visitor_learning_rate_multiplier(0));
+
+    // However, sometimes we might want to adjust the learning rate differently thoughout the
+    // network.  Here we show how to use the visit_layers_range to adjust the learning rate
+    // different at the different ResNet50's convolutional blocks:
+    visit_layers_range<  0,   2, decltype(net), visitor_learning_rate_multiplier>(net, visitor_learning_rate_multiplier(1));
```
davisking: Does `visit_layers_range<0,2>(net, visitor_learning_rate_multiplier(1));` not compile?  It should.  Something is funky if it's not.

davisking:
> I was wondering, since the visitor_learning_rate_multiplier might be useful, would you like to include in dlib instead of the example?

Yeah it's certainly generally useful.  It's nice to have it in the example to illustrate how to make a visitor though.  So I'm on the fence about it.  Maybe making the `set_` methods I noted and putting them in dlib proper while leaving the visitor in the example is a good balance.

examples/dnn_introduction3_ex.cpp
arrufat: Oh, yes, it does compile!

examples/dnn_introduction3_ex.cpp
arrufat: Sorry, I left them because I was checking that the VRAM was freed after leaving that scope.

examples/dnn_introduction3_ex.cpp
arrufat: Oh, I see. I've just look at that code and it seems a bit over my C++ skills, so it might take some time to figure it out :)

arrufat:
> > I was wondering, since the visitor_learning_rate_multiplier might be useful, would you like to include in dlib instead of the example?
> 
> Yeah it's certainly generally useful. It's nice to have it in the example to illustrate how to make a visitor though. So I'm on the fence about it. Maybe making the `set_` methods I noted and putting them in dlib proper while leaving the visitor in the example is a good balance.

Yes, I agree that it might be useful to have it as an example, but the visitor that you made for the `set_all_bn_running_stats_window_sizes` helped me a lot to write this one. As you may have noticed, this one is just a simple modification of that one :)

examples/dnn_introduction3_ex.cpp
davisking: No worries.

examples/dnn_introduction3_ex.cpp
davisking: I just mean this stuff:
```
    namespace impl 
    {    
        template <typename T, typename int_<decltype(&T::get_learning_rate_multiplier)>::type = 0> 
        double get_learning_rate_multiplier (
            const T& obj, 
            special_
        ) { return obj.get_learning_rate_multiplier(); }

        template <typename T>
        double get_learning_rate_multiplier ( const T& , general_) { return 1; } 
    }    
    template <typename T>
    double get_learning_rate_multiplier(const T& obj) { return impl::get_learning_rate_multiplier(obj, special_()); }
```
It uses SFINAE (https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error) to select which overload of `get_learning_rate_multiplier` to use. The key bit here is `int_<decltype(&T::get_learning_rate_multiplier)>::type`, which is an error if `T::get_learning_rate_multiplier` doesn't exist.

arrufat: Thank you! I am not very familiar with SFINAE, so I just modified it to something it seemed sensible, but doesn't work yet, I will learn how to do it :) This is what I tried, if you want to shame me :package: 

``` c++
  namespace impl
  {
      template <typename T, typename int_<decltype(&T::set_learning_rate_multiplier)>::type = 0>
      void set_learning_rate_multiplier (
          T& obj,
          special_,
          double new_learning_rate_multiplier
      ) { obj.set_learning_rate_multiplier(new_learning_rate_multiplier); }
  
      template <typename T>
      void set_learning_rate_multiplier (T& , general_) { }
  }
  template <typename T>
  void set_learning_rate_multiplier(
      T& obj,
      double new_learning_rate_multiplier
  )
  {
      impl::set_learning_rate_multiplier(obj, special_(), new_learning_rate_multiplier);
  }
```

arrufat: I found the mistake, I missed the `double` parameter in the general case... I will add now the missing functions :)

arrufat:
I was thinking, how about we add the learning rate visitor to dlib and in the example, we show how to create a custom visitor to change the weight decay? That would tackle both issues :)

arrufat:
I've added the functions to change the learning rate multipliers and modified the visitor in the example to modify the weight decay multiplier instead.

dlib/dnn/utilities.h
```diff
@@ -301,6 +301,65 @@ namespace dlib
...
+        {
+        public:
+            visitor_learning_rate_multiplier(double new_learning_rate_multiplier_) :
+                new_learning_rate_multiplier(new_learning_rate_multiplier_) { }
+
+            template <typename T>
+            void set_new_learning_rate_multiplier(T& l) const
+            {
+                set_learning_rate_multiplier(l, new_learning_rate_multiplier);
+                set_bias_learning_rate_multiplier(l, new_learning_rate_multiplier);
```
davisking: The definition of the bias learning rate multiplier is this:
```
        double get_bias_learning_rate_multiplier(
        ) const;
        /*!
            ensures
                - returns a multiplier number.  The interpretation is that this object is
                  requesting that the learning rate used to optimize its bias parameters be
                  multiplied by get_learning_rate_multiplier()*get_bias_learning_rate_multiplier().
        !*/
```
So only call `set_learning_rate_multiplier`, otherwise the bias learning rate will not result in the desired value.

dlib/dnn/utilities.h
```diff
@@ -301,6 +301,65 @@ namespace dlib
...
+        };
+    }
+
+    template <typename net_type>
+    void set_all_learning_rate_multipliers(
+        net_type& net,
+        double learning_rate_multiplier
+    )
+    {
+        DLIB_CASSERT(learning_rate_multiplier >= 0 && learning_rate_multiplier <= 1);
```
davisking: It's fine for a learning rate multiplier to be > 1

dlib/dnn/utilities.h
```diff
@@ -301,6 +301,65 @@ namespace dlib
...
+
+    template <size_t begin, size_t end, typename net_type>
+    void set_learning_rate_multipliers_range(
+        net_type& net,
+        double learning_rate_multiplier
+    )
+    {
+        static_assert(begin <= end, "Invalid range");
+        static_assert(end <= net_type::num_layers, "Invalid range");
+        DLIB_CASSERT(learning_rate_multiplier >= 0 && learning_rate_multiplier <= 1);
```
davisking: It's fine for a learning rate multiplier to be > 1

examples/dnn_introduction3_ex.cpp
```diff
@@ -0,0 +1,170 @@
...
+public:
+
+    visitor_weight_decay_multiplier(double new_weight_decay_multiplier_) :
+        new_weight_decay_multiplier(new_weight_decay_multiplier_) {}
+
+    template <typename T>
+    void set_new_weight_decay_mulitplier(T& l) const
+    {
+        set_weight_decay_multiplier(l, new_weight_decay_multiplier);
+        set_bias_weight_decay_multiplier(l, new_weight_decay_multiplier);
```
davisking: Same thing here.  The effective bias weight decay is the overall weight decay times the specific bias weight decay.  So only call the first method.

dlib/dnn/core.h
```diff
@@ -47,6 +47,62 @@ namespace dlib
...
+
+        template <typename T>
+        void set_bias_learning_rate_multiplier (T& , general_, double) { }
+    }
+    template <typename T>
+    void set_bias_learning_rate_multiplier(
+        T& obj,
+        double new_bias_learning_rate_multiplier
+    ) { impl::set_bias_learning_rate_multiplier(obj, special_(), new_bias_learning_rate_multiplier); }
+
```
davisking: Looking good.   Be sure to update the docs in core_abstract.h too :)

dlib/dnn/utilities.h
arrufat: Can I assume the same is true for the weight decay multiplier?

arrufat: Sorry I misunderstood you, fixed in new commit

davisking: Yep, see the docs for that too https://github.com/davisking/dlib/blob/master/dlib/dnn/layers_abstract.h#L547.  It's the same deal.

arrufat:
I think it's good to go now :)

davisking:
Sweet, thanks for the PR :)

arrufat:
My pleasure! :D

PR #2007, opened 2020-02-28
imglab: chinese ("automatic") clustering, keyboard shortcuts for zooming

sandsmark:
Using chinese whispers for "automatic" clustering usually works and is nice for lazy people, and scrolling to zoom is a bit annoying at least for me.

In case the diff isn't clear, I just moved the zoom handling code out from the on_wheel_{up,down} into separate functions and call those as appropriate instead.


davisking:
Nice, good idea :)  Thanks for the PR.

PR #2010, opened 2020-03-05
simplify resnet definition by reusing struct template parameter

arrufat:
I realized that I was using the template parameter for `bn_con` and `affine` twice: once for the struct template and one for each template that needed a batch normalization layer inside the struct.
Now everything uses the top level template parameter, leading to more readable code :)

### DIFF
--- a/examples/dnn_introduction3_ex.cpp
+++ b/examples/dnn_introduction3_ex.cpp
@@ -29,7 +29,7 @@ namespace model
     using net_type = loss_metric<
         fc_no_bias<128,
         avg_pool_everything<
-        typename resnet::impl<BN>::template backbone_50<
+        typename resnet::def<BN>::template backbone_50<
         input_rgb_image
         >>>>;
 
--- a/examples/dnn_introduction3_ex.cpp
+++ b/examples/dnn_introduction3_ex.cpp
@@ -29,7 +29,7 @@ namespace model
     using net_type = loss_metric<
         fc_no_bias<128,
         avg_pool_everything<
-        typename resnet<BN>::template backbone_50<
+        typename resnet::impl<BN>::template backbone_50<
         input_rgb_image
         >>>>;
 
@@ -75,7 +75,7 @@ int main() try
     {
         // Now, let's define the classic ResNet50 network and load the pretrained model on
         // ImageNet.
-        resnet<bn_con>::n50 resnet50;
+        resnet::train_50 resnet50;
         std::vector<string> labels;
         deserialize("resnet50_1000_imagenet_classifier.dnn") >> resnet50 >> labels;
 
--- a/examples/resnet.h
+++ b/examples/resnet.h
@@ -8,7 +8,7 @@ namespace resnet
     using namespace dlib;
     // BN is bn_con or affine layer
     template<template<typename> class BN>
-    struct impl
+    struct def
     {
         // the resnet basic block, where BN is bn_con or affine
         template<long num_filters, int stride, typename SUBNET>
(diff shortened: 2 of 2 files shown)


arrufat:
I was thinking also about doing something like this to allow `using namespace dlib;`:
``` cpp
namespace resnet
{
    using namespace dlib;
    template<template<typename> class BN>
    struct impl
    {
        // resnet definition here
    }
    using train_18 = impl<bn_con>::n18;
    using train_34 = impl<bn_con>::n34;
    using train_50 = impl<bn_con>::n50;        
    using train_101 = impl<bn_con>::n101;      
    using train_152 = impl<bn_con>::n152;      
                                           
    using infer_18 = impl<affine>::n18;  
    using infer_34 = impl<affine>::n34;  
    using infer_50 = impl<affine>::n50;  
    using infer_101 = impl<affine>::n101;
    using infer_152 = impl<affine>::n152;
}
```
Then, on the calling side it would be more convenient, too:
``` cpp
resnet::train_50 net;
```
I have already these changes locally, if you want I will push them.
The goal would be to have a generic `resnet.h` file that people can copy and use.

I've also noticed that it would be nice to add tagged layers after each resnet level... I need to check how to do it cleanly :)

davisking:
Yeah that sounds like a good idea too.  Feel free to put that into this PR as well :)

examples/dnn_introduction3_ex.cpp
```diff
@@ -29,7 +29,7 @@ namespace model
     using net_type = loss_metric<
         fc_no_bias<128,
         avg_pool_everything<
-        typename resnet<BN>::template backbone_50<
+        typename resnet::impl<BN>::template backbone_50<
```
davisking: What do you think about `def` or `net` instead of `impl`?  Like maybe:
```
typename resnet::def<BN>::template backbone_50
```

arrufat: Honestly, I don't have any preference. I just used `impl` because I saw that is what you used for internal implementations (although as namespace qualifiers). So, I also think either `def` or `net` would be a better name. Which one do you prefer?

davisking: How about def then. resnet::net feels weird.

arrufat: Done!
With regards doing the tagging after each resnet level (I've found it useful for training multi-scale segmentation models), here's what I am thinking:

- adding the following tags after each level:
``` cpp
template <typename SUBNET> using rtag1 = add_tag_layer<6001, SUBNET>;
template <typename SUBNET> using rtag2 = add_tag_layer<6002, SUBNET>;
template <typename SUBNET> using rtag3 = add_tag_layer<6003, SUBNET>;
template <typename SUBNET> using rtag4 = add_tag_layer<6004, SUBNET>;
```
- adding the following skip layers, too, for convenience:
``` cpp
template <typename SUBNET> using skipr1 = add_skip_layer<rtag1, SUBNET>;
template <typename SUBNET> using skipr2 = add_skip_layer<rtag2, SUBNET>;
template <typename SUBNET> using skipr3 = add_skip_layer<rtag3, SUBNET>;
template <typename SUBNET> using skipr4 = add_skip_layer<rtag4, SUBNET>;
```
(I've chosen 6000 because ResNet has 6 letters)

Then, there comes the tricky part: once I get this model defined, I should iterate over all the layers and do:
``` cpp
layer<i>(new_net).layer_details() = layer<j>(old_net).layer_details();
```
With appropriate `i` and `j`. Once it's done, serialize the new model.

I was wondering if there was an easy way to do this (maybe with a visitor?).
But since I only need to do it once, I think it will take less time to write a macro in my editor that will generate all the assignment lines. How would you do this?
Also, I was wondering if `rtag4` should refer to the level close to the input layer or the loss layer. I've seen that you used `level4` for the one that's close to the input layer (following the same order as the number of the layers) but @reunanen used the inverse order, which makes more sense in the way I think about how the data flows through the network, but it's less consistent with the order of the layers. Sorry for asking too many questions... (:

davisking: I'm not sure what's the best way.  Like you said, it's pretty easy to make a macro in an editor.  Like I use vim and it's super easy to do this kind of thing there.  It depends on how often you do it.  It is sort of awkward to setup something automatic that does this.  You could use a visitor that keeps an int that increments on every layer but tag layers and use that to make the mapping though.

davisking:
Thanks for another nice PR :)

examples/dnn_introduction3_ex.cpp
arrufat: Ok, that's exactly what I thought, I was just wanted the expert's opinion to make sure I was not missing anything.
I guess it would be up to the end user to do it they way they want then :)

PR #2015, opened 2020-03-10
Prevention of compiler warning

hwiesmann:
An exception will be thrown if the actual size is larger than the limit of uint32 but the assignment does not contain an explicit cast. The missing cast may trigger an incorrect compiler warning (e.g. Clang).


davisking:
Cool, thanks for the PR :)

PR #2019, opened 2020-03-11
effect -> affect

scott-vsi:
Should this be affect instead of effect?


davisking:
Thanks for the fix :)

Issue #1999, opened 2020-02-13
C++20 removed std::result_of


215020267:
**Description:**
In /std:c++17 mode, this triggers a deprecation warning in recent versions of MSVC. In /std:c++latest mode, now that microsoft/STL#380 has been merged, this will trigger an error in VS 2019 16.6 Preview 2.

**Reproduce steps:**
1. git clone https://github.com/davisking/dlib  D:\Dlib\src
2. cd D:\Dlib\src
3. set _CL_=/D_HAS_AUTO_PTR_ETC=1 /D_WINSOCKAPI_
5. mkdir build_amd64
6. cd D:\Dlib\src\build_amd64
7. cmake -G "Visual Studio 15 Win64" -DCMAKE_SYSTEM_VERSION=10.0.17134.0 ..
8. cmake -G "Visual Studio 15 Win64" -DCMAKE_SYSTEM_VERSION=10.0.17134.0 ..\dlib\test 
9. msbuild /m /p:Platform=x64 /p:Configuration=Release Project.sln /t:Rebuild /p:PreferredToolArchitecture=x64

**ErrorMessage:**
F:\gitP\davisking\dlib\dlib\threads\async.h(55): error C2039: 'result_of': is not a member of 'std' (compiling source file F:\gitP\davisking\dlib\dlib\filtering\kalman_filter.cpp)

facug91:
For what I've read, since C++17, `std::result_of` should replaced by `std::invoke_result`.  
In dlib, it's only being used in the `async.h` file, and in a way that seems to be wrong (https://en.cppreference.com/w/cpp/types/result_of).  
I think we could use a macro to replace `std::result_of` by `std::invoke_result` in that part, when the compiler is >= C++17. But the potential problem would still be there for C++11 and C++14.  
Maybe someone with better understanding of async.h file, can say if this is safe or not.

davisking:
What's wrong with the current code?  

I'm not super excited about using macros to do this, since compilers pretty reliably don't set `__cplusplus` correctly, especially visual studio. Maybe things are better now, but it wasn't always great.  But if toggling on `__cplusplus` works for MSVC then that's fine, but does it define `__cplusplus` to be `201703L` before it starts warning or even erroring out about `std::result_of`? All my past experience with visual studio makes me expect it will error on `std::result_of` long before `__cplusplus` gets set to `201703L`.

davisking:
Anyway, if macros do work then that would be nice, and a PR that toggled on that would be cool :)

PR #2021, opened 2020-03-12
Replace result_of by invoke_result for C++17 and above #1999

facug91:
As discussed in #1999, std::result_of is deprecated in C++17 and should be replaced by std::invoke_result. In C++20, std::result_of is removed. What I do in this PR is performing this replacement, when using C++17 or C++20. I used almost the same as in this other PR, from another library https://github.com/fmtlib/fmt/pull/1028.  

I also make a minor modification to how std::result_of is use for C++11 and C++14, to comply with how it is recommended (https://en.cppreference.com/w/cpp/types/result_of):  
```cpp
template<class F, class... Args>
std::result_of_t<F&&(Args&&...)> // instead of std::result_of_t<F(Args...)>, which is wrong
  my_invoke(F&& f, Args&&... args) { 
    /* implementation */
}
```

I ran all tests with gcc7, gcc9 and VS2017 (with C++14, and C++latest).

### DIFF
--- a/dlib/threads/async.h
+++ b/dlib/threads/async.h
@@ -40,6 +40,19 @@ namespace dlib
             fun();
             prom.set_value();
         }
+
+        template <typename> struct result_of;
+
+#if (__cplusplus >= 201703L ||                          \
+     (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)) && \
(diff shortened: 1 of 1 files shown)


davisking:
Awesome, thanks for the fix :)

PR #2022, opened 2020-03-12
add loss multiclass log weighted

arrufat:
This PR adds `loss_multiclass_log_weighted` for classification problems where we want to give more/less importance to some classes/samples.

I just took the idea from `loss_multiclass_log_per_pixel_weighted` by @reunanen and created this loss layer based on `loss_multiclass_log` for the sake of consistency.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -369,6 +369,137 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    class loss_multiclass_log_weighted_ 
+    {
+    public:
+
+        struct weighted_label
+        {
+            weighted_label()
+            {}
+
+            weighted_label(unsigned long label, float weight = 1.f)
+                : label(label), weight(weight)
+            {}
+
+            unsigned long label = 0;
+            float weight = 1.f;
+        };
+
+        typedef weighted_label training_label_type;
+        typedef unsigned long output_label_type;
+
+        template <
+            typename SUB_TYPE,
+            typename label_iterator
+            >
+        void to_label (
+            const tensor& input_tensor,
+            const SUB_TYPE& sub,
+            label_iterator iter
+        ) const
+        {
+            const tensor& output_tensor = sub.get_output();
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+            DLIB_CASSERT(output_tensor.nr() == 1 &&
+                         output_tensor.nc() == 1 );
+            DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+
+
+            // Note that output_tensor.k() should match the number of labels.
+
+            for (long i = 0; i < output_tensor.num_samples(); ++i)
+            {
+                // The index of the largest output for this sample is the label.
+                *iter++ = index_of_max(rowm(mat(output_tensor),i));
+            }
+        }
+
+        template <
+            typename const_label_iterator,
+            typename SUBNET
+            >
+        double compute_loss_value_and_gradient (
+            const tensor& input_tensor,
+            const_label_iterator truth,
+            SUBNET& sub
+        ) const
+        {
+            const tensor& output_tensor = sub.get_output();
+            tensor& grad = sub.get_gradient_input();
+
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+            DLIB_CASSERT(input_tensor.num_samples() != 0);
+            DLIB_CASSERT(input_tensor.num_samples()%sub.sample_expansion_factor() == 0);
+            DLIB_CASSERT(input_tensor.num_samples() == grad.num_samples());
+            DLIB_CASSERT(input_tensor.num_samples() == output_tensor.num_samples());
+            DLIB_CASSERT(output_tensor.nr() == 1 &&
+                         output_tensor.nc() == 1);
+            DLIB_CASSERT(grad.nr() == 1 &&
+                         grad.nc() == 1);
+
+            tt::softmax(grad, output_tensor);
+
+            // The loss we output is the average loss over the mini-batch.
+            const double scale = 1.0/output_tensor.num_samples();
+            double loss = 0;
+            float* g = grad.host();
+            for (long i = 0; i < output_tensor.num_samples(); ++i)
+            {
+                const weighted_label wl = (weighted_label)*truth++;
+                const long y = wl.label;
+                const float weight = wl.weight;
+                // The network must produce a number of outputs that is equal to the number
+                // of labels when using this type of loss.
+                DLIB_CASSERT(y < output_tensor.k(), "y: " << y << ", output_tensor.k(): " << output_tensor.k());
+                for (long k = 0; k < output_tensor.k(); ++k)
+                {
+                    const unsigned long idx = i*output_tensor.k()+k;
+                    if (k == y)
+                    {
+                        loss += weight*scale*-safe_log(g[idx]);
+                        g[idx] =weight*scale*(g[idx]-1);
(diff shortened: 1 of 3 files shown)


dlib/dnn/loss.h
```diff
@@ -367,6 +367,137 @@ namespace dlib
     template <typename SUBNET>
     using loss_multiclass_log = add_loss_layer<loss_multiclass_log_, SUBNET>;
 
+// ----------------------------------------------------------------------------------------
+
+    class loss_multiclass_log_weighted_ 
+    {
+    public:
+
+        struct weighted_label
```
reunanen: In that other class `label` is `uint16_t` (here `unsigned long`), but otherwise this is identical, right?

Should there instead be a common class which both can use? (You could make `label` be a template type.)

dlib/dnn/loss_abstract.h
```diff
@@ -371,6 +371,97 @@ namespace dlib
...
+                This object implements the loss layer interface defined above by
+                EXAMPLE_LOSS_LAYER_.  In particular, it implements the multiclass logistic
+                regression loss (e.g. negative log-likelihood loss), which is appropriate
+                for multiclass classification problems.  It is basically just like the
+                loss_multiclass_log except that it lets you define per-sample weights,
+                which might be useful e.g. if you want to emphasize rare classes while
+                training.  (If the classification problem is difficult, a flat weight
+                structure may lead the network to always predict the most common label,
+                in particular if the degree of imbalance is high.  To emphasize a certain
+                class of classes, simply increase the weights of the corresponding samples,
```
reunanen: "class of classes"? :)

In that other class this says "class _or_ classes".

dlib/dnn/loss.h
arrufat: Yes, it is identical. I was also thinking that maybe we could use a common class for both. I can do that :)

dlib/dnn/loss_abstract.h
arrufat: Oops, typo, thank you :)

dlib/dnn/loss.h
arrufat: Ok, I've done what you suggested. I was wondering if `weighted_label` is an appropriate name now that it's in the `dlib` namespace. Also if I should also add serialization support for it...

dlib/dnn/loss.h
```diff
@@ -2832,21 +2966,7 @@ namespace dlib
     {
     public:
 
-        struct weighted_label
```
reunanen: Hmm. Now existing code will break, when it tries to instantiate this nested (now-missing) struct. Not saying it's the end of the world (probably not so many people are using this), but on the other hand I think it's relatively easy to maintain compatibility, so perhaps it's worthwhile to at least give it a try?

I think you can leave here something like `typedef dlib::weighted_label<uint16_t> weighted_label;`? You'll know this works, if you can then revert [the changes you had to make to the test program](https://github.com/davisking/dlib/commit/a9033b58d1858a3271bfa6620f420475f5f2c822#diff-7d2445557c03b31a758c2ef7d310bc53).

If my comment is not clear, let me know and I'll see if I can write it as code instead.

arrufat: I think I got your point :)

dlib/dnn/loss.h
```diff
@@ -367,6 +367,140 @@ namespace dlib
...
+
+            tt::softmax(grad, output_tensor);
+
+            // The loss we output is the average loss over the mini-batch.
+            const double scale = 1.0/output_tensor.num_samples();
+            double loss = 0;
+            float* g = grad.host();
+            for (long i = 0; i < output_tensor.num_samples(); ++i)
+            {
+                const weighted_label<unsigned long> wl = (weighted_label<unsigned long>)*truth++;
```
davisking: `const weighted_label<unsigned long> wl = *truth++;` should be enough. Are you getting a build error without the extra cast?  That would be surprising.

davisking: Even just `const auto wl = *truth++;` should be fine.

dlib/dnn/loss_abstract.h
```diff
@@ -369,6 +369,107 @@ namespace dlib
...
+                EXAMPLE_LOSS_LAYER_.  In particular, it implements the multiclass logistic
+                regression loss (e.g. negative log-likelihood loss), which is appropriate
+                for multiclass classification problems.  It is basically just like the
+                loss_multiclass_log except that it lets you define per-sample weights,
+                which might be useful e.g. if you want to emphasize rare classes while
+                training.  (If the classification problem is difficult, a flat weight
+                structure may lead the network to always predict the most common label,
+                in particular if the degree of imbalance is high.  To emphasize a certain
+                class or classes, simply increase the weights of the corresponding samples,
+                relative to the weights of other pixels.)
```
davisking: Don't need the () around these sentences

dlib/dnn/loss_abstract.h
```diff
@@ -369,6 +369,107 @@ namespace dlib
...
+
+                Note that if you set all the weights equals to 1, then you essentially get
```
davisking: Don't say essentially.  It makes it sound like it's almost, but not quite identical. But it's really identical.

dlib/test/dnn.cpp
```diff
@@ -3224,6 +3224,82 @@ namespace
...
+
+            for (size_t ii = 0; ii < num_samples; ++ii)
+            {
+                if (predictions[ii] == weighted_class)
+                    ++num_weighted_class;
+                else
+                    ++num_not_weighted_class;
+            }
+        }
+    }
```
davisking: This test doesn't end up testing anything.  Don't forget to put some DLIB_TEST statements that check that the network learned the right thing :)

arrufat: Oops, I might have removed by mistake while debugging the test...

dlib/dnn/loss_abstract.h
arrufat: Thank you, done!

dlib/dnn/loss_abstract.h
arrufat: Removed.

dlib/dnn/loss.h
arrufat: Yes, `const auto wl = *truth++;` works fine.

arrufat:
So, I wasn't getting this error before, but today, with gcc 9.2.1 I am getting:
```
../../dnn/loss.h:2969:42: error: declaration of ‘typedef struct dlib::weighted_label<short unsigned int> dlib::loss_multiclass_log_per_pixel_weighted_::weighted_label’ changes meaning of ‘weighted_label’ [-fpermissive]
 2969 |         typedef weighted_label<uint16_t> weighted_label;                                                 
      |                                          ^~~~~~~~~~~~~~                                                  
In file included from ../../dnn.h:26,                                                                            
                 from ../dnn.cpp:12:                                                                             
../../dnn/loss.h:373:12: note: ‘weighted_label’ declared here as ‘struct dlib::weighted_label<short unsigned int>’
  373 |     struct weighted_label                                                                                
      |            ^~~~~~~~~~~~~~
```
Which compiles fine with `-fpermissive`. I don't know why I wasn't getting this error the other day...
So, in order to keep compatibility with the previous `loss_multiclass_log_per_pixel_weighted_`, maybe I should rename `weighted_sample` to `weighted_loss_sample` or something similar... What do you think?

davisking:
Try using `typedef dlib::weighted_label<uint16_t> weighted_label;`

davisking:

arrufat:
Great, thank you :)

Issue #2017, opened 2020-03-11
dlib has a couple of inconsistencies with respect to long, unsigned long and int


hwiesmann:
## Expected Behavior
Proper/documented intentional conversion between different integer types.

## Current Behavior
At a couple of positions in the code longs are converted to int without proper casting. While this may be nowadays relatively uncritical (most of the computers are probably running with 64 bit integers anyway), there might be some issues on older machines where (unsigned) long and int have different sizes.
Though, the program might crash anyway on 32bit systems when trying to use too large datasets.

## Steps to Reproduce
1) Uncritical example: string.h
std::setw(int) has as an int parameter but gets passed in pad_int_with_zeros(int,unsigned long width) an unsigned long parameter "width".

2) Critical example: hash
murmur_hash3(_2) uses for the length also an int but both versions of hash pass a value of type std::string::size_type to murmur_hash3(_2) that may be larger than 32bit (even on 32bit machines). This may result in wrong hash values.

A bigger problem might be simple_hash_map, though, because: simple_hash_map()::insert(0xff..ff00000000,0xff..fe00000000,1) calls internally murmur_hash3_2 with a long but murmur_hash3_2 has as parameter types uint32! So, in my example the hash value is definitely wrong!

simple_hash_map()::insert requires that "a" and "b" are smaller than std::numeric_limits<uint32>::max().
The same holds for the methods accessing the hashed value.

PS: murmur_hash3 and murmur_hash3_2 have different parameter types as well!

3) Critical? matrix_blas_bindings.h parameter inconsistencies
The return type of a lot of functions is int (example get_ld) but often the variable used for returning the result is of type long.

Concrete example:
template <typename T>
        int get_ld (const assignable_ptr_matrix<T>& m) { return m.nc(); }

m.nc() returns a long but get_ld returns an int.

PS: I cannot provide a complete list because the list is 296 issues long! But I suspect that there are more..

dlib 19.19
clang compiler on macOS 64bit

davisking:
Some of these things are just what they are.  Like the CBLAS API deals in ints and there is nothing I can do to change that.  But if you want to submit a PR that makes the simple_hash_map deal completely in 64bit integers that's fine with me.  The same goes for fixing other things.  Please don't submit one massive PR though.  If there are many things you would like to change to do in separate pull requests :)

PR #2024, opened 2020-03-13
Integer conversions generating compiler warnings

hwiesmann:
With respect to issue #2017 I submit a few adjustments that mainly change the type of local variables to prevent potentially wrong integer conversions.


davisking:
Thanks for the PR.

PR #2025, opened 2020-03-13
Prevention of integer overflow

hwiesmann:
Prevention of (unlikely) integer overflow when doing int padding

### DIFF
--- a/dlib/string/string.h
+++ b/dlib/string/string.h
@@ -267,7 +267,7 @@ namespace dlib
     )
     {
         std::ostringstream sout;
-        sout << std::setw(width) << std::setfill('0') << i;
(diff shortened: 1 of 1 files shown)


dlib/string/string.h
```diff
@@ -267,7 +267,7 @@ namespace dlib
     )
     {
         std::ostringstream sout;
-        sout << std::setw(width) << std::setfill('0') << i;
+        sout << std::setw(static_cast<int>(std::min(width,static_cast<unsigned long>(std::numeric_limits<int>::max())))) << std::setfill('0') << i;
```
davisking: This is a bit over the top :)

You can just change the `width` argument to `int`.

Issue #1381, opened 2018-06-26
Calling `.clean()` between forward-passes


cchadowitz-pf:
* **Version**: v19.13
* **Where did you get dlib**: github
* **Platform**: Ubuntu 16.04.4 with CUDA 8.0 and cuDNN7, as well as running via docker (base image: nvidia/cuda:8.0-cudnn7-devel-ubuntu16.04)
* **Compiler**: Built with:
```
cmake .. -DDLIB_NO_GUI_SUPPORT=ON -DDLIB_USE_CUDA=ON && cmake --build . --config Release
```
Using `gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609`

Watching the GPU memory usage via `watch -n0.5 nvidia-smi` while using dlib, it seems like if I deserialize a trained model from disk to memory and use it for inference, over time the GPU memory usage trends up (it fluctuates up and down but overall it slowly rises). However, I think if I first call `net.clean()` before each inference, the memory usage remains more stable and no longer gradually climbs.

I'm not yet sure if there is a leak somewhere or not, but I'd be interested in knowing if calling `.clean()` before each inference is the suggested use when executing inferences on the same model iteratively. 

Some sample code to show what my usage is like (not exact, but should at least show the gist of it):

```
std::vector<dlib::matrix<dlib::rgb_pixel>> inputImages = {.......}; // Assume there are somewhere say 1-50 images stored here
int batchSize = 50; // Just as an example, likely irrelevant for this situation

net_type_faceDetector _faceDetector;
dlib::deserialize(modelfile) >> _faceDetector;

for(int i=0; i < 10000000; i++) {
  _faceDetector.clean(); // This is the line in question
  auto detections = _faceDetector(inputImages, batchSize);
}
```

I guess the tl;dr is should `.clean()` be necessary (or even a good idea) between inference calls? The network architectures of the models I'm using are the same architectures as `mmod_front_and_rear_end_vehicle_detector` or `mmod_human_face_detector`.

Thanks!

davisking:
You shouldn't be calling clean() in a loop, that's just going to slow down
processing since it then deallocates and reallocates GPU resources over and
over.

I doubt there is a memory leak, but you never know.

cchadowitz-pf:
Thanks, that's what I was suspecting about `clean()`.

[...]

PR #2027, opened 2020-03-16
To avoid a GPU memory leak, allow passing thread pools to dnn_trainer from outside

reunanen:
Problem: As discussed starting from [here](https://github.com/davisking/dlib/issues/1381#issuecomment-598777029), the CUDA runtime allocates resources for each thread, and apparently those resources are not freed when the corresponding threads terminate. Therefore, each instantiation of `dnn_trainer` leaks a bit of GPU memory.

Solution: Add possibility to pass thread pools from outside. This way, subsequent `dnn_trainer` instances can use the same threads, and there's no memory leak.

### DIFF
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -75,6 +75,7 @@ namespace dlib
         typedef typename net_type::input_type input_type;
         const static size_t num_computational_layers = net_type::num_computational_layers;
         const static size_t num_layers = net_type::num_layers;
+        using threads = std::vector<std::shared_ptr<thread_pool>>;
     private:
         typedef impl::dnn_job_t<training_label_type> job_t;
     public:
@@ -104,8 +105,9 @@ namespace dlib
         dnn_trainer(
             net_type& net_, 
             const solver_type& solver_,
-            const std::vector<int>& cuda_extra_devices
-        ) : job_pipe(0), net(net_) 
+            const std::vector<int>& cuda_extra_devices,
+            std::shared_ptr<threads> thread_pools_ = std::shared_ptr<threads>()
+        ) : job_pipe(0), net(net_), thread_pools(thread_pools_)
         {
             devices.push_back(std::make_shared<device_data>(dlib::cuda::get_device(), net, solver_));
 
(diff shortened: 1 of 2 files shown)


dlib/dnn/trainer.h
```diff
@@ -667,15 +669,22 @@ namespace dlib
             std::vector<tensor*> reference_params;
             visit_layer_parameters(devices[0]->net, [&](size_t, tensor& t) { reference_params.push_back(&t); });
 
+            // If no external thread pools vector was passed, then create one that will
+            // be automatically destructed as soon as the dnn_trainer object goes out of
+            // scope.
+            if (!thread_pools)
+                thread_pools = std::make_shared<threads>();
+
+            auto& tp = *thread_pools;
```
reunanen: Did like this in order to minimize the diff.

dlib/dnn/trainer_abstract.h
```diff
@@ -96,6 +99,14 @@ namespace dlib
...
                       devices, which you do by putting their device numbers into
                       cuda_extra_devices.
+                - if (thread_pools.get() != nullptr) then
+                    - Any new threads spun within the trainer will execute within the
+                      passed thread pools vector. This means that the same threads can
+                      be re-used across different dnn_trainer instances. Otherwise, the
+                      CUDA runtime may leak memory. This, however, is relevant only if
+                      your program is going to instantiate a large number of trainers,
+                      and generally stay up and running for a very long time. If not,
+                      then you need not worry about this.
```
davisking: Make an extra note that the CUDA runtime memory leak isn't because dlib code is not calling appropriate cuda free calls, but is a limitation of the CUDA runtime we have no control over.

davisking: Also note that any particular thread pool vector should only be passed to one trainer instance at a time.

dlib/dnn/trainer.h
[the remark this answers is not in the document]
reunanen: Hmm, maybe I would still prefer this (as it is now):
`tp[i]->add_task_by_value(...);`

Over something like this:
`(*thread_pools)[i]->add_task_by_value(...);`

What do you think?

dlib/dnn/trainer_abstract.h
reunanen: Added these notes.

dlib/dnn/trainer.h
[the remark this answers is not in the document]
davisking: Yeah, that was bothering me too ¯\_(ツ)_/¯.  I guess `tp` is better.

Issue #2023, opened 2020-03-13
link to parallel openblas on Fedora/Centos


scott-vsi:
In [dlib/find_blas.cmake](https://github.com/davisking/dlib/blob/c8a175f56901e64d741ac43297d3cece858cd31e/dlib/cmake_utils/find_blas.cmake#L195), dlib looks for the libopenblas.so library. On Centos/Fedora, openblas-devel, the equivalent library to libopenblas-dev on Ubuntu, installs libopenblas.so; however, it also installs libopenblasp.so (among other variants). It is my understanding that the p in libopenblasp.so stands for parallel and this is the version of the library that supports multithreading. The numpy rpm for Fedora, for instance, [links](https://src.fedoraproject.org/rpms/numpy/blob/64fa2c8e7504e7f22fa0672b8838a51848954dea/f/numpy.spec#_130) to openblasp.

So, do you think should try to link to libopenblasp.so first on Fedora/Centos?

I know that it is possible to set the CMake variable cblas_lib to the abs path to libopenblasp.so, but the difference between libopenblas.so and libopenblasp.so is not obvious.

dlib v19.19 (or master currently c8a175f) cloned from here
Centos 7 (Linux 5.3.11-200.fc30.x86_64)
g++ 4.8.5 (ancient, I know)

davisking:
That sounds like a better default then.  Can you make a PR with the appropriate change?  And run the dlib unit tests with it to make sure nothing funny happens.

scott-vsi:
Sure. The #2028 is actually very easy. I tried building it on Ubuntu 18.04 and 16.04 and Centos 7 and Fedora 29 (I actually couldn't compile on Fedora 30. I can create an issue if you'd like). 

I simply added another name to the `find_library` NAMES array. CMake's [documentation](https://cmake.org/cmake/help/v3.0/command/find_library.html) states, "When more than one value is given to the NAMES option this command by default will consider one name at a time and search every directory for it." So CMake should prefer libopenblasp.so when available (Centos/Fedora) and then libopenblas.so

I also ran `./dtest --runall`. The test for GPU and GUI were skipped, but those shouldn't involve blas (cublas is not affected). All tests complete successfully.

davisking:
Sweet, thanks.  Don't worry about Fedora 30, unless it matters to you.

PR #2028, opened 2020-03-16
link against openblasp

scott-vsi:
openblasp is a parallel implementation of openblas with pthreads found on Centos/Fedora. See #2023


davisking:
Thanks for the PR :)

Issue #1776, opened 2019-05-21
Example of DCGAN


arrufat:
Hi, I would like to contribute a DCGAN example to dlib.

I have implemented a version of [Pytorch DCGAN for C++](https://github.com/pytorch/examples/blob/master/cpp/dcgan/dcgan.cpp).

However, I would need some guidance with some things I don't know how to do. I am wondering on how I should proceed. Should I attach my current code here (around 150 lines), or make a pull request, even if the code is not able to learn anything? Maybe @edubois can help out, since he stated that he managed to make it work on https://github.com/davisking/dlib/issues/1261

Thanks for your hard work on dlib.

arrufat:
I'm still working on this... I'm going to write about the network architecture
- loss_binary_log
- discriminator
- generator
- input

Here's what they look like:
``` cpp
// the MNIST input image size
constexpr size_t input_size = 28;
// some helper definitions for the noise generation
constexpr size_t noise_size = 100;
using noise_t = std::array<matrix<double, 1, 1>, noise_size>;
// convolution and transposed convolution with custom padding
template<long num_filters, long kernel_size, int stride, int padding, typename SUBNET>
using conp = add_layer<con_<num_filters, kernel_size, kernel_size, stride, stride, padding, padding>, SUBNET>;
template<long num_filters, long kernel_size, int stride, int padding, typename SUBNET>
using contp = add_layer<cont_<num_filters, kernel_size, kernel_size, stride, stride, padding, padding>, SUBNET>;
// the generator
template<typename SUBNET>
using generator_type =
    htan<contp<1, 4, 2, 1,
    relu<bn_con<contp<64, 4, 2, 1,
    relu<bn_con<contp<128, 3, 2, 1,
    relu<bn_con<contp<256, 4, 1, 0,
    SUBNET>>>>>>>>>>>;
// the discriminator
template<typename SUBNET>
using discriminator_type =
    loss_binary_log<
    affine<conp<1, 3, 1, 0,
    prelu<bn_con<conp<256, 4, 2, 1,
    prelu<bn_con<conp<128, 4, 2, 1,
    prelu<conp<64, 4, 2, 1,
    SUBNET>>>>>>>>>>>;
// and the whole network
using net_type =
    discriminator_type<
    // tag5 gets the mask from tag1, multiplies it by the noise and adds it to the image
    tag5<add_prev2<mult_prev4<extract<input_size * input_size * 2, 1, input_size, input_size, skip1<
[...]

PR #2033, opened 2020-03-19
add leaky_relu activation layer

arrufat:
This PR adds the leaky_relu activation layer, which is mostly useful for generative models.
It is my first step towards implementing the DCGAN example in #1776.

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1350,6 +1350,52 @@ namespace dlib
                 grad.device(), src.device(), gradient_input.device(), grad.size(),
                 param.device(), params_grad.device());
         }
+    // ----------------------------------------------------------------------------------------
+
+        __global__ void _cuda_leaky_relu(const float* s, float* d, size_t n, const float alpha)
+        {
+            for (auto i : grid_stride_range(0, n))
+            {
+                if (s[i] > 0)
+                    d[i] = s[i];
+                else
+                    d[i] = alpha * s[i];
+            }
+        }
+
+        void leaky_relu(
+            tensor& dest,
+            const tensor &src,
+            const float alpha
+        )
+        {
+            launch_kernel(_cuda_leaky_relu, max_jobs(dest.size()),
+                src.device(), dest.device(), src.size(), alpha);
+        }
+
+    // ----------------------------------------------------------------------------------------
+
+        __global__ void _cuda_leaky_relu_gradient(float* out, const float* s, const float* gi, size_t n, const float alpha)
+        {
+            for (auto i : grid_stride_range(0, n))
+            {
+                if (s[i] > 0)
+                    out[i] += gi[i];
+                else
+                    out[i] += alpha * gi[i];
+            }
+        }
+
+        void leaky_relu_gradient (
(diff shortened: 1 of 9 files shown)


dlib/dnn/layers_abstract.h
```diff
@@ -2090,6 +2090,53 @@ namespace dlib
...
+
+    class leaky_relu_
+    {
+        /*!
+            WHAT THIS OBJECT REPRESENTS
+                This is an implementation of the EXAMPLE_COMPUTATIONAL_LAYER_ interface
+                defined above.  In particular, it defines a leaky rectified linear
+                layer.  Therefore, it passes its inputs through the function
+                    f(x) = x>0 ? x : alpha*x
+                where f() is applied pointwise across the input tensor and alpha is a scalar.
```
davisking: Say `alpha is a non-learned scalar` and then in the prelu docs above change them to say `alpha is a learned scalar`.  Otherwise the docs for these two things are the same.

dlib/test/dnn.cpp
```diff
@@ -3703,6 +3729,7 @@ namespace
             test_softmax_all();
             test_sigmoid();
             test_mish();
+            test_leaky_relu();
```
davisking: Add a `test_layer()` test for this new layer over in `test_layers()`.

davisking:

dlib/dnn/layers_abstract.h
arrufat: Do you think `alpha` is a good name? I was thinking about something more descriptive, like `slope`?
But again, probably people who are using `leaky_relu` already know what we're referring to.

dlib/test/dnn.cpp
arrufat: Good catch, I forgot to implement the inplace case for the cuda backend...

dlib/dnn/layers_abstract.h
arrufat: I've modified the `leaky_relu` documentation, but `prelu` already says that `p` is not learned.

davisking: Oh right, that it does! :)

dlib/cuda/cuda_dlib.cu
```diff
@@ -1377,12 +1377,25 @@ namespace dlib
 
         __global__ void _cuda_leaky_relu_gradient(float* out, const float* s, const float* gi, size_t n, const float alpha)
         {
-            for (auto i : grid_stride_range(0, n))
+            if (out == gi)
```
davisking: A branch like this is relatively costly in CUDA. If you broke this up into two separate kernels it would likely run measurably faster.  The launching code can then contain the branch which happens once and CUDA never sees it.

arrufat: Thank you for that advice!
One quick question, I've updated the code to get:
```cpp
float* out = grad.device();
const float* gi = gradient_input.device();
```
before making the comparison and launching the kernel. Even if from the documentation of `tensor::to_device()`, it seems that if I call `.to_device()` twice in a row, nothing happens,, it seemed preferable to avoid such calls.

davisking:

PR #2035, opened 2020-03-24
Add DCGAN example

arrufat:
This PR will close issue #1776, as it adds an example on how to train a DCGAN that generates MNIST digits.
The generator will be able to generate good-looking images after around 20'000 training iterations.

 I've also tried it with the Fashion MNIST dataset, and it converges much faster (only 3'000 iterations) and it can be run with no code changes.


arrufat:
I have noticed a "small" issue. The generator sometimes does not converge, so I fixed the pseudo-random generator seed. Then I got convergence after 5'000 steps instead of 20'000+.
However I did:
``` bach
touch dnn_dcgan_train_ex.cpp
```
and then rebuild the program and then it didn't converge.
I repeated the processes about 4 or 5 times until I got convergence again. Is there any way to make the whole program deterministic?.

arrufat:
I've noticed that `randomize_parameters` in the network initialization makes a call to `std::rand`, so I've set a seed to `std::srand`. But it does not make a big difference. I might be missing some random thing elsewhere.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,273 @@
...
+    {
+        deserialize("dcgan_sync") >> generator >> d_trainer;
+    }
+
+    const size_t minibatch_size = 64;
+    const std::vector<float> real_labels(minibatch_size, 1.f);
+    const std::vector<float> fake_labels(minibatch_size, -1.f);
+    dlib::image_window win;
+    size_t iteration = 0;
+    while (iteration < 50'000)
```
davisking: Using ' in integer literals requires C++14, but dlib only requires C++11.  This didn't compile for me when I tried it, I'm not sure why travis is not failing on this, it should be enforcing C++11 :(

Anyway, I want to switch to C++14, but maybe next year.  In my experience it takes 6-7 years before a large enough majority of dlib users upgrade their compilers to allow dlib itself to begin using new compiler features.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,273 @@
...
+};
+
+// Some helper definitions for the noise generation
+const size_t noise_size = 100;
+using noise_t = std::array<matrix<float, 1, 1>, noise_size>;
+
+noise_t make_noise(dlib::rand& rnd)
+{
+    noise_t noise;
+    std::for_each(begin(noise), end(noise),
```
davisking: Use a range based for loop instead of `std::for_each` since it's simpler.  `std::for_each` is a pre-C++11 hack from before range based for loops were in the language.  Although the new C++20 `std::for_each` that supports parallelism might be kinda nice, but that's a different beast.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,273 @@
...
+template<long num_filters, long kernel_size, int stride, int padding, typename SUBNET>
+using conp = add_layer<con_<num_filters, kernel_size, kernel_size, stride, stride, padding, padding>, SUBNET>;
+
+// A transposed convolution to with custom padding
+template<long num_filters, long kernel_size, int stride, int padding, typename SUBNET>
+using contp = add_layer<cont_<num_filters, kernel_size, kernel_size, stride, stride, padding, padding>, SUBNET>;
+
+// The generator is made of a bunch of deconvolutional layers.  Its input is a 1 x 1 x k noise
+// tensor, and the output is the generated image.  The loss layer does not matter for the
+// training, we just stack a compatible one on topi to be able to have a () operator on the
```
davisking: `on top to`

arrufat: Oh, I was already in INSERT mode 😅

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,273 @@
...
+    discriminator(generate_image(generator, make_noise(rnd)));
+    cout << "generator" << endl;
+    cout << generator << endl;
+    cout << "discriminator" << endl;
+    cout << discriminator << endl;
+
+    // The solvers for the generator network.  In this case, we don't need to use a dnn_trainer
+    // for the generator, since we are going to train it manually
+    std::vector<adam> g_solvers(generator.num_computational_layers, adam(0, 0.5, 0.999));
+    auto g_sstack = make_sstack(g_solvers);
```
davisking: Needing to have `make_sstack` in user code is sort of gnarly now that I'm looking at it.  Could you add an overload to `update_parameters()` that took a `std::vector` and just defers to the existing `update_parameters()` using `make_sstack()`?  Then the `generator.subnet().update_parameters(g_sstack, learning_rate);` below could just be passed `g_sovlers`.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+std::vector<matrix<unsigned char>> get_generated_images(generator_type& net)
+{
+    std::vector<matrix<unsigned char>> images;
+    const resizable_tensor &out = layer<1>(net).get_output();
+    for (size_t n = 0; n < out.num_samples(); ++n)
+    {
+        matrix<float> output = image_plane(out, n);
+        matrix<unsigned char> image;
+        assign_image_scaled(image, output);
```
davisking: This might be part of the reason training is unstable.  Like GANs are famously hard to train, but `assign_image_scaled` is multiplying `output` by some number that's jumping around a bunch based on what happens to be in the image.  If instead you used `assign_image(image, 255*output)`, or some other fixed constant, not necessarily 255, I suspect it might be more stable.

davisking:
This is super cool.

There isn't a reliable way to make things deterministic as what happens is dependent on build settings and the particular order that CUDA atomic adds happen (since floating point addition is non-commutative).

In general I don't think this kind of non-determinism is a big deal.  Like you don't want it in unit tests, since it makes automated testing frustrating in some cases.  But the underlying issue is that training is unstable, and finding the magic setting/seed/what-have-you that makes it work in that one case just hides it, which is not great.  The non-determinism exposes this, which adds a silver lining to the non-determinism.

Anyway, you might be able to make it more stable by not using `assign_image_scaled()`.  Even if it's still unstable this is still super cool.  DNN training and GANs in particular are known to be hard to train, so sometimes it just is what it is.

examples/dnn_dcgan_train_ex.cpp
arrufat: Ok, it would be really cool to have some "new" C++ features in dlib. However, if I remember it correctly, you added C++11 as a requirement when you added the deep learning API, which was in 2016 (5 years after C++11). Now it's already 6 years after C++14. But I think, at that point it was more of a necessity, and now it would just be for the fancy stuff :)

examples/dnn_dcgan_train_ex.cpp
arrufat: I don't know what I was thinking when I wrote this...

examples/dnn_dcgan_train_ex.cpp
arrufat: I've changed this, and it seems a bit more stable, thank you for the suggestion!
More specifically, since the output of `htan` is between `-1` and `1`, I've added `1` and then multiplied by `127.5` to make sure the range is between `0` and `255`.

arrufat:
Thank you so much for your kind words and advice! I really want to be like you when I grow up :P

Aside from the `assign_image_scaled`, which seemed to help a little, I've also made sure all floating point values have an explicit type, so that the compiler does not implicitly cast between `float` and `double`.
After this change, I've compiled and run the example many times, and it converged every time after a few thousand iterations.

arrufat:
I don't know why this is failing on Windows, now. Among others, it doesn't recognize the operator `not`, which is [built into the language](https://en.cppreference.com/w/cpp/language/operator_alternative). It was working on the previous commit, and the only thing I changed was the `leaky_relu` documentation (inside of a comment).

davisking:
> I don't know why this is failing on Windows, now. Among others, it doesn't recognize the operator `not`, which is [built into the language](https://en.cppreference.com/w/cpp/language/operator_alternative). It was working on the previous commit, and the only thing I changed was the `leaky_relu` documentation (inside of a comment).

Yeah, visual studio has never been a bastion of standards compliance.  I'm kinda surprised it doesn't know what `not` is though.  But even so, no one uses `not`, the idiomatic thing to do in C++ is to use `!`.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,272 @@
...
+    relu<bn_con<contp<128, 3, 2, 1,
+    relu<bn_con<contp<256, 4, 1, 0,
+    input<noise_t>
+    >>>>>>>>>>>>;
+
+// Now, let's proceed to define the discriminator, whose role will be to decide whether an
+// image is fake or not.
+using discriminator_type =
+    loss_binary_log<
+    htan<conp<1, 3, 1, 0,
```
davisking: Remove the `htan`.  I tried removing it and training converges much faster.  This makes sense because the output of the network, when using `loss_binary_log` is supposed to be a log odds ratio, which is something that can take any real value.  Capping it to [-1, 1] means the highest probability it can assign to anything is 0.73106, which is not very high.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,272 @@
...
+    leaky_relu<conp<64, 4, 2, 1,
+    input<matrix<unsigned char>>
+    >>>>>>>>>>>;
+
+// Some helper functions to generate and get the images from the generator
+matrix<unsigned char> generate_image(generator_type& net, const noise_t& noise)
+{
+    const matrix<float> output = net(noise);
+    matrix<unsigned char> image;
+    assign_image(image, 127.5f * (output + 1.f));
```
davisking: Don't worry about using `f` to make things floats.  The compiler will do something reasonable without it, and generally speaking, if you can get `double` precision you always want it.  Intel chips also operate by doing all (well most, it's complicated) floating point operations in a really high precision and then dropping back to float when writing the result back.

Anyway, let things be in double.  Things on the GPU are in float because it reduces GPU RAM consumption by 2x and the GPU hardware can do float arithmetic a lot faster than double arithmetic.  But that doesn't apply to the constants in this file, as they all run on the CPU.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,272 @@
...
+template<long num_filters, long kernel_size, int stride, int padding, typename SUBNET>
+using contp = add_layer<cont_<num_filters, kernel_size, kernel_size, stride, stride, padding, padding>, SUBNET>;
+
+// The generator is made of a bunch of deconvolutional layers.  Its input is a 1 x 1 x k noise
+// tensor, and the output is the generated image.  The loss layer does not matter for the
+// training, we just stack a compatible one on top to be able to have a () operator on the
+// generator.
+using generator_type =
+    loss_binary_log_per_pixel<
+    htan<contp<1, 4, 2, 1,
```
davisking: Instead of `htan` you could use `sig`, which makes outputs appear between 0 and 1.  Then the code to convert to images is just `255 * out`, which is simpler.  This isn't going to make training converge faster, just make the code more straightforward.

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -0,0 +1,272 @@
...
+            real_samples.push_back(training_images[idx]);
+        }
+        d_trainer.train_one_step(real_samples, real_labels);
+
+        // Train the discriminator with fake images
+        // 1. generate some random noise
+        std::vector<noise_t> noises;
+        while (noises.size() < minibatch_size)
+        {
+            noises.push_back(std::move(make_noise(rnd)));
```
davisking: `std::move` not needed here since the return from `make_noise` is already an rvalue.  So it doesn't do anything.

examples/dnn_dcgan_train_ex.cpp
arrufat: Indeed, way simpler :)

examples/dnn_dcgan_train_ex.cpp
arrufat: Thanks for the clarification!

examples/dnn_dcgan_train_ex.cpp
arrufat: Wow, training converges much faster now, thank you!

arrufat:
Thanks for spending time giving me quality feedback! I really appreciate it.
In case somebody is wondering why I don't use the built-in sync mechanism of the `dnn_trainer`: when the loss increases, it will try to reload from a previous state.

I have a couple of questions:
- since training is much more stable, maybe we can reduce the number of steps to 30000? I don't know of a good way to stop the training: it's hard to tell from the loss values (which look way nicer after the removal of the `htan`), since both networks are improving and the same loss value at different training steps has different meanings.
- I have noticed that the GPU memory usage varies between 496 and 582 MiB. Is there something we can do about it?

davisking:
No problem.

I think it's fine to keep it at 50000, but whatever you want to set it to is fine.  The point of the example is to educate, it doesn't have to be a perfect tool for making a dcgan.  The main thing it does is display the images and losses when it's running.  If the main loop was a `while(true)` loop it would still serve its purpose of being educational and showing this all working.

That memory fluctuation is happening because there are some `resizable_tensor` variables inside the main loop that get created and destroyed over and over.  There are also two cases where you do `const resizable_tensor& thing = ...` which also make a copy, since those expressions are binding to a `tensor`, not a `resizable_tensor`. So a copy is created.  Finally, there is this `device_global_buffer()` idiom inside the conv layers.  Those layers need to allocate scratch buffers, and to not waste memory all instances of conv in a thread share the same global scratch buffer.  Normally this is fine, but what happens is the trainer runs the network in thread A, you then in main() run the network in thread B (the main thread), the conv layers see this happen, and they discard their handles to the device global scratch buffer for thread A.  Importantly, they all discard their handles, causing thread A's global scratch buffer to be deallocated.  But then the trainer runs it all again in thread A and it's reallocated again.

I don't think this reallocation is a huge deal.  But you could just not use the trainer.  Maybe the example would end up being simpler since you wouldn't have to reason about the whole "trainer.get_net() to pause the threads" thing.  Anyway, whatever you think makes the example easiest to read and understand is best here, since this reallocation is really inconsequential.

arrufat:
Again, thank you very much for your explanation. I've decided to leave the number of iterations as it is.

I think also showing how to train both networks manually can be useful, since no other example does it. The training code is much more consistent between the discriminator and the generator: we can see exactly where they differ: `compute_loss(fake_samples_tensor, real_labels.begin())`.

Also, after this change, the memory usage does not fluctuate any more: it stays constant at 486 MiB and the GPU usage has also increased a little bit (from 96% to 97-98%).

Thank you so much for adding this deep learning API to dlib, it is really a joy to use!

As a side note, now that this example is complete, here's a comparison between this implementation and the [PyTorch one](https://github.com/pytorch/examples/blob/master/cpp/dcgan/dcgan.cpp), which was used as a guide.

| Toolkit | GPU |   VRAM  | CPU  | RAM      |
|--------:|:---:|:-------:|------|----------|
|    dlib | 98% | 486 MiB | 100% |  838 MiB |
| PyTorch | 92% | 721 MiB | 100% | 1767 MiB |

davisking:
No problem, and thanks for another cool PR :)

davisking:
All the `.subnet()` calls were bothering me so I just pushed some of these methods into the loss layer as well and updated the example.  This way user interactions with the network are a little more straightforward.

arrufat:
Oh, perfect!

arrufat:
I was thinking about using 2 `running_stats` objects to track the losses of the discriminator and the generator:

``` cpp
running_stats<double> g_loss, d_loss;
```
 And then compute the losses as:
``` cpp
d_loss.add(discriminator.compute_loss(real_samples_tensor, real_labels.begin());
// ...
d_loss.add(discriminator.compute_loss(fake_samples_tensor, fake_labels.begin());
// ...
g_loss.add(discriminator.compute_loss(fake_samples_tensor, real_labels.begin());
```

Then, display `d_loss.mean()` (or `d_loss.mean() * 2`) and `g_loss.mean()`, and clear the stats after each check point. I think it's more representative of what's happening than displaying the raw loss at a particular point. What do you think?

PR #2041, opened 2020-03-30
use running stats to track losses in dcgan example

arrufat:
Feel free to close if you don't like it :)


davisking:
Cool, this is most definitely the right thing to do.  Thanks for another PR.

PR #2045, opened 2020-03-30
remove branch from cuda kernel

arrufat:
As you commented in https://github.com/davisking/dlib/pull/2033#discussion_r395625551, having a branch like that might not be the best thing to do, so I've removed a similar branch from the gradient computation of the mish activation layer. I am not completely happy about having that lambda expression repeated, maybe as a function outside? But I didn't want to pollute the outer scope. What do you think?

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1440,42 +1440,29 @@ namespace dlib
 
     // ----------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


davisking:
Cool, probably worth it.  Put that lambda in a global function and call it from the two kernels though to avoid duplicating it.  Like just put it right next to them in the source file.

davisking:
Thanks for the PR :)

PR #2046, opened 2020-03-31
Fix warnings while running the tests

arrufat:
As I've been running the tests several times these last days, I noticed there were some warnings when building with GCC 9.3.0 and these flags:
``` cmake
if (CMAKE_COMPILER_IS_GNUCXX)
   add_definitions("-W -Wall")
   # I don't care about unused testing functions though.  I like to keep them
   # around.  Don't warn about it.
   add_definitions("-Wno-unused-function -fdiagnostics-color=always -Wno-deprecated-copy -Wextra -pedantic")
endif()
```
I might try the deprecated-copy warnings another time :)

Also, while I was at it, I made the `layers.h` a bit more uniform when there are unused parameters in methods like `serialize`.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -691,14 +691,14 @@ namespace dlib
         const tensor& get_layer_params() const { return params; }
         tensor& get_layer_params() { return params; }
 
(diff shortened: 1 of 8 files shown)


davisking:
Nice, thanks :)

PR #2048, opened 2020-04-02
Remove outdated comment from DCGAN example

arrufat:
That comment was there from when I was using a dnn_trainer to train
the discriminator network.

Sorry for the noise, but that comment might confuse users...


davisking:
No worries, thanks for the PR :)

PR #2049, opened 2020-04-03
Fix warning in dnn_trainer initialization list

arrufat:
The thread pool was initialized after the network, so it lead to a
reorder warning in GCC 9.3.0


davisking:
Thanks :)

PR #2053, opened 2020-04-09
Add cuda implementation for loss_mean_squared_per_channel_and_pixel

arrufat:
Thanks to @reunanen for helping out! :D

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1841,9 +1841,9 @@ namespace dlib
             for (auto i : grid_stride_range(0, n))
             {
                 const float y = truth[i];
-                const float temp1 = y - out_data[i];
-                loss += temp1 * temp1;
-                g[i] = -temp1 * scale;
+                const float temp = y - out_data[i];
+                loss += temp * temp;
+                g[i] = -temp * scale;
             }
             warp_reduce_atomic_add(*loss_out, loss);
         }
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1842,9 +1842,8 @@ namespace dlib
             {
                 const float y = truth[i];
                 const float temp1 = y - out_data[i];
-                const float temp2 = scale * temp1;
-                loss += temp2 * temp1;
-                g[i] = -temp2;
+                loss += temp1 * temp1;
+                g[i] = -temp1 * scale;
             }
             warp_reduce_atomic_add(*loss_out, loss);
         }
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1835,6 +1835,22 @@ namespace dlib
 
     // ----------------------------------------------------------------------------------------
 
+        __global__ void _cuda_compute_loss_mean_squared_per_channel_and_pixel(float* loss_out, float* g, const float* truth, const float* out_data, size_t n, const float scale)
+        {
+            float loss = 0;
+            for (auto i : grid_stride_range(0, n))
+            {
+                const float y = truth[i];
+                const float temp1 = y - out_data[i];
+                const float temp2 = scale * temp1;
+                loss += temp2 * temp1;
+                g[i] = -temp2;
+            }
+            warp_reduce_atomic_add(*loss_out, loss);
+        }
+
+    // ----------------------------------------------------------------------------------------
+
         void compute_loss_binary_log_per_pixel::
         do_work(
             cuda_data_ptr<float> loss_work_buffer,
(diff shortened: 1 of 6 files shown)


dlib/cuda/cuda_dlib.cu
```diff
@@ -1833,6 +1833,21 @@ namespace dlib
...
 
+    // ----------------------------------------------------------------------------------------
+
+        __global__ void _cuda_compute_loss_mean_squared_per_channel_and_pixel(float* loss_out, float* g, const float* truth, const float* out_data, size_t n, const float scale)
+        {
+            float loss = 0;
+            for (auto i : grid_stride_range(0, n))
+            {
+                const float y = truth[i];
+                const float temp1 = y - out_data[i];
```
reunanen, line 1844: Maybe just `temp` now? :)

reunanen:
Did you eventually get some performance boost from using the CUDA version?

arrufat:
Yes, there are some speed improvements. Here's the benchmark code:
```
#include <dlib/dnn.h>
#include <iostream>
#include <numeric>

namespace model
{
    using namespace dlib;
    using net_type =
        loss_mean_squared_per_channel_and_pixel<10, con<10, 1, 1, 1, 1, input_rgb_image>>;
}  // namespace model

auto main() -> int
try
{
    dlib::matrix<dlib::rgb_pixel> dummy(224, 224);
    dlib::assign_all_pixels(dummy, dlib::rgb_pixel(0, 0, 0));
    std::vector<dlib::matrix<dlib::rgb_pixel>> minibatch(1, dummy);
    std::array<dlib::matrix<float>, 10> label;
    for (auto& l : label)
    {
        l = dlib::matrix<float>(224, 224);
        dlib::assign_all_pixels(l, 0);
    }
    std::vector<std::array<dlib::matrix<float>, 10>> labels(1, label);
    dlib::resizable_tensor samples;
    model::net_type net;
    net.to_tensor(minibatch.begin(), minibatch.end(), samples);
    net(minibatch);
    std::cout << net << '\n';
    net.compute_loss(samples, labels.begin());
    std::array<float, 100> times;
    for (auto& time : times)
    {
        const auto t0 = std::chrono::high_resolution_clock::now();
        net.compute_loss(samples, labels.begin());
        const auto t1 = std::chrono::high_resolution_clock::now();
        const auto duration =
            std::chrono::duration_cast<std::chrono::duration<float, std::milli>>(t1 - t0).count();
        time = duration;
    }
    std::cout << "average time: " << std::fixed << std::setprecision(3)
              << std::accumulate(times.begin(), times.end(), 0.f, std::plus<float>()) /
                     times.size()
              << " ms\n";
}
{
    std::cout << e.what() << '\n';
}
```
- cpu version: 0.975 ms
- gpu version: 0.395 ms

Thank you again for your help :D

dlib/cuda/cuda_dlib.cu
arrufat: oops, yes! 😅

dlib/dnn/loss.h
```diff
@@ -389,7 +389,8 @@ namespace dlib
     {
     public:
 
-        typedef weighted_label<unsigned long> training_label_type;
+        typedef dlib::weighted_label<unsigned long> weighted_label;
```
arrufat, line 392: I added this typedef, because maybe users would expect this loss to behave as the `loss_multiclass_log_per_pixel_weighted_`.
Sorry for doing this in this PR...

davisking: Na, seems good.  Update the _abstract.h docs too though.

dlib/cuda/cuda_dlib.cu
```diff
@@ -1882,6 +1897,29 @@ namespace dlib
             loss = scale*floss;
         }
 
+        void compute_loss_mean_squared_per_channel_and_pixel::
+        do_work
+        (
```
davisking, line 1902: Put the ( on the same line as `do_work`

dlib/dnn/loss.h
```diff
@@ -3301,6 +3302,12 @@ namespace dlib
                 }
             }
 
+#ifdef DLIB_USE_CUDA
+            double loss;
+            cuda_compute(truth, output_tensor, grad, loss);
+            return loss;
+#else
+
```
davisking, line 3310: Can you move the stuff in the `else` block into something analogous to the `cuda::compute_loss_mean_squared_per_channel_and_pixel` you added?  Then in the unit tests you can create both the cuda and cpu version, give them the same inputs, and check that both methods give identical results.

arrufat: Sure! Where should I put that? Inside this class or in `dlib/cuda/cpu_dlib.cpp` and then name it `cpu::compute_loss_mean_squared_per_channel_and_pixel`?

arrufat: OK, I think I've done something sensible in the meantime (I think we have 13 hours of time difference).

I don't get identical results for cuda and cpu implementations. However, the difference between them is around 1e-7 or 1r-8. For this reason, I added a small tolerance margin in the tests of 1e-6.

davisking: Sounds good

davisking:

PR #2063, opened 2020-04-20
Update dnn_introduction_ex.cpp

ncoder-1:
Changed C-style cast to static_cast and prefer direct function "using" over using namespace std as per https://isocpp.org/wiki/faq/coding-standards#using-namespace-std


davisking:
All the other examples do this and I think it's fine.  If you want to submit something with just the static_casts then I'll merge that.

Issue #2068, opened 2020-04-22
Doubts about the MMOD examples


arrufat:
Hi, I've been reading the MMOD examples several times, but there are things that I think I am missing.

The first MMOD example talks about a receptive field of a little over 50x50.
https://github.com/davisking/dlib/blob/253745d29fd510620a1bcee0add50a7663a9f5d6/examples/dnn_mmod_ex.cpp#L32-L39

However, the network downsamples the image by a factor of 8, and the sliding window has a size of 6x6 in the feature space, which translates to 48x48 pixels in the image space. So, shouldn't this be `a little under 50x50`?
Regardless of that, the objects that we are trying to detect are 40x40 pixels, which certainly fit inside of the 48x48 sliding windows.

In the DNN Face detection example, the network has been modified to detect 80x80 faces.
https://github.com/davisking/dlib/blob/253745d29fd510620a1bcee0add50a7663a9f5d6/examples/dnn_mmod_face_detection_ex.cpp#L27-L33

However, the sliding window is of size 9x9, which translates to 72x72 in the image space (since the down-sampling factor is 8).

I am guessing there is no problem if the object is a bit bigger or smaller than the sliding window.

Finally, in the car detector example:
https://github.com/davisking/dlib/blob/253745d29fd510620a1bcee0add50a7663a9f5d6/examples/dnn_mmod_train_find_cars_ex.cpp#L221-L229

A similar network is setup, so that the receptive field is 72, but in this case the maximum size is 70, which makes more sense to me than the previous object sizes and network receptive fields.

So, I am certainly missing something here.
I was sure I completely understood the relation between sliding window size, down-sampling of the network and min object size when I read the car examples. But when I read the other two previous examples, I had some doubts.

- Am I understanding correctly what the receptive field means in this case?
- How sensitive is the network to the relation between the min object size and the size of the sliding window? In other words, how much bigger/smaller can objects be with regards to the sliding window in the image space? (my guess is that it depends on how we setup the IoU for scoring the detections)

Thanks in advance :)

Edit:
I know that the literature defines the receptive field of a Iayer (or network) as all the pixels in the input image that contribute to the output of the layer (or network), which is computed in the following way:

![image](https://user-images.githubusercontent.com/1671644/79960981-08d5b800-84c1-11ea-9827-b0ac485324fa.png)

where:
- `rf`: receptive field
- `l`: layer index
- `ks`: kernel size
- `s`: stride
And the receptive field of the first layer is directly the kernel size.
[...]

PR #2070, opened 2020-04-27
fix receptive field comment according to #2068

arrufat:
As you noted on #2068, I changed `a little over` to `approximately`.
Thank you!


davisking:
Thanks :)

PR #2076, opened 2020-05-04
Minor optimization: add shortcut to in-place image resize if size_scale is 1

reunanen:
Just a minor suggestion :)


davisking:
Thanks for another PR :)

Issue #2111, opened 2020-06-22
Broken pkg-config file


remicollet:
Generated dlib-1.pc seems broken

# Current
```
$ pkg-config --libs dlib-1
-ldlib -lpthread /usr/lib64/libX11.so /usr/lib64/libXext.so /usr/lib64/libgif.so /usr/lib64/libpng.so /usr/lib64/libz.so /usr/lib64/libjpeg.so /usr/lib64/libopenblasp.so /usr/lib64/libsqlite3.so 
```

library patch is not expected there, only linker options

# Expected
`$ pkg-config --libs dlib-1 -ldlib -lpthread -lSM -lICE -lX11 -lXext -lgif -lpng -lz -ljpeg -lopenblasp -lsqlite3`

davisking:
How is this broken?

remicollet:
for ex `/usr/lib64/libgif.so ` is not a linker option, so the library is not added to linker command, which result in missing symbols.

Ex: buillging [pdlib](https://github.com/goodspb/pdlib) extension for php:
`modules/pdlib.so: undefined symbol: DGifSlurp`

remicollet:
Notice: libgif is the only needed library, as there is direct call to gif fucntions in `image_loader/load_image.h`, other libraries are private (only used by libdlib)

SO:
```
Libs: -L${libdir} -ldlib -lgif
Libs.private:  -lpthread -lX11 -lXext -lpng -lz -ljpeg -lopenblasp -lsqlite3
```

BTW, an alternative way could be to mode the GIF code in some C file, outside the header, so libgif will also be a private library

remicollet:
See how this output is consumed in 
https://github.com/php/php-src/blob/master/build/php.m4#L368

As you can see, only `-l` and `-L` options are used.

Notice: dlib is the only library among the hundreds one installed on my system to use full paths...

remicollet:
I still think pkgconfig file is not clean.

At least public vs private libraries are not handled as expected
And using libraries full path is not a common/best practice

davisking:
You can make a PR to change it. But it’s just got to avoid breaking existing user’s setups.

PR #2114, opened 2020-06-23
Fix pkg-config file

remicollet:
See #2111

* First commit only fix linker syntax
* Second commit move all libs to private, libgif is the only one I detect as needed (used in headers), 

Foreach loop inspired from libzip.

"libs" are enough for build against the shared library
```
$ pkg-config --libs dlib-1
-ldlib -lgif 

```

"libs.private" are needed for build against the static library
```
$ pkg-config --static --libs dlib-1
-ldlib -lgif -lpthread -lX11 -lXext -lpng -lz -ljpeg -lopenblasp -lsqlite3 

```
P.S. I not a cmake expert, nor a pkg-config expert, perhaps better way exists.

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -878,7 +878,15 @@ if (NOT TARGET dlib)
 
    # Install the library
    if (NOT DLIB_IN_PROJECT_BUILD)
-      string (REPLACE ";" " " pkg_config_dlib_needed_libraries "${dlib_needed_libraries}")
+      foreach(LIB ${dlib_needed_libraries})
+        if(LIB MATCHES "^/")
+          get_filename_component(LIB ${LIB} NAME_WE)
+          string(REGEX REPLACE "^lib" "-l" LIB ${LIB})
+        endif()
+        MESSAGE(LIB=${LIB})
+        set(pkg_config_dlib_needed_libraries "${pkg_config_dlib_needed_libraries} ${LIB}")
+      endforeach()
+
       # Make the -I include options for pkg-config
       foreach (ITR ${dlib_needed_includes})
          set (pkg_config_dlib_needed_includes "${pkg_config_dlib_needed_includes} -I${ITR}")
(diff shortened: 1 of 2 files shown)


remicollet:
Notice: it is probably better to use the "requires" way, when used library provides a .pc file (but of the scope of this PR)

Ex:
`Requires: libpng`

dlib/CMakeLists.txt
```diff
@@ -422,7 +423,7 @@ if (NOT TARGET dlib)
          find_package(GIF QUIET)
          if (GIF_FOUND)
             set (dlib_needed_includes ${dlib_needed_includes} ${GIF_INCLUDE_DIR})
-            set (dlib_needed_libraries ${dlib_needed_libraries} ${GIF_LIBRARY})
```
davisking, line 425: This feeds into more than just pkgconfig.  In particular, into `target_link_libraires`.  So this can't just be elided.

remicollet:
Second commit (about public/private) reverted.

> You can definitely pass a file path to the linker,

Sorry, but this doesn't work with PHP build system which heavily use `pkg-config` for most libraries and parse this output (using `PHP_EVAL_LIBLINE`). dlib is the only one to use such lib path (among the 482 lib installed on my dev system)

remicollet:
hmm Looks like -L also needed is library not in standard path....

Ok closing as I'm really not a CMake expoert, BTW, bug still exists

davisking:
Honestly I think this is properly considered as a bug in the PHP tooling you are using.  That said, it's fine to work around that problem by changing something in dlib.  But it's got to not break other users.

remicollet:
Notice: from "ld" documentation, proper symtax for libraries is -l and using -l:/full/patch/to.so if you want the full patch

And also, from pkg-config documentation (man pc)

```
    Libs    Required linking flags for this package.  
    Libs.private
             Required linking flags for this package that are only required when linking statically.  
```

So diff between public and private library make sense

PR #2118, opened 2020-06-28
Additional documentation for failed dlib::layer<> use.

stoperro:
I got this error few times and each time spent a lot of time guessing what I did wrong. In bigger networks it's easy to by mistake create `add_skip_layer<123, ...>` where there is no `add_tag_layer<123, ...>`. This produces gibberish error in header far from place of mistake.
Hopefully with this comment me and other people will easier get what's wrong.

The error one gets now in MVSC is:
`error C2039: 'subnet': is not a member of 'dlib::input<std::array<dlib::matrix<T,0,0,dlib::default_memory_manager,dlib::assignable_ptr_matrix<T>::layout_type>,3>>'`
... that for just network type definition:
`relu<skip1<tag2<input<::std::array<matrix<float>, 3>>>>> test;`

### DIFF
--- a/dlib/dnn/core.h
+++ b/dlib/dnn/core.h
@@ -2684,6 +2684,9 @@ namespace dlib
         {
             static_assert(i < T::num_layers, "Call to layer() attempted to access non-existing layer in neural network.");
             static T& makeT();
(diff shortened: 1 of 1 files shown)


davisking:
Thanks, good idea :)

PR #2135, opened 2020-07-21
Fix restoration of MSVC warnings in public headers

jbfove:
Previously they were restored to default values, which had the effect of negating the current setting of the calling code (whether set in the compiler options or by pragma previously)


jbfove:
I'm happy to make any changes to this as needed to get it merged. It took me a while to figure out which library was enabling warnings we had previously disabled. People working on other coded bases that include dlib can be spared of that once merged.

davisking:
Yeah, good catch. Thanks for the PR :)

PR #2137, opened 2020-07-24
Added possibility to load PNG images from a data buffer.

yuriio147:


dlib/image_loader/png_loader.cpp
```diff
@@ -108,35 +139,62 @@ namespace dlib
...
-        if (fread( sig, 1, 8, fp ) != 8)
+        memcpy( data, state->buffer_ + state->current_pos_, length );
+        state->current_pos_ += length;
+    }
+
+    void png_loader::read_image( FILE* fp, const unsigned char* image_buffer, size_t buffer_size )
+    {
+        ld_.reset(new LibpngData);
+
+        static const png_size_t png_header_size = 8;
```
davisking, line 157: use `constexpr` instead of `static const`

dlib/image_loader/png_loader.cpp
```diff
@@ -190,25 +260,25 @@ namespace dlib
             color_type_ != PNG_COLOR_TYPE_RGB_ALPHA &&
             color_type_ != PNG_COLOR_TYPE_GRAY_ALPHA)
         {
-            fclose( fp );
+            if ( fp != NULL ) fclose( fp );
             png_destroy_read_struct( &( ld_->png_ptr_ ), &( ld_->info_ptr_ ), &( ld_->end_info_ ) );
-            throw image_load_error(std::string("png_loader: unsupported color type in file ") + filename);
+            throw image_load_error(std::string("png_loader: unsupported color type"));
```
davisking, line 265: Make this something like 
```
throw image_load_error("png_loader: unsupported color type" + file_info);
```
and put `" in file " + filename` in `file_info` when we know the filename.  Otherwise the error is going to be fairly confusing to users if they encounter it.

davisking:
The appveyor failure seems like appveyor flaking out.  I'm rerunning that now.

davisking:

PR #2138, opened 2020-07-24
Add setargv.obj to CMakeLists.txt for wildcard expansion on Windows

lenniea:
Possible improvement on Windows by enabling wildcard expansion in cmdline examples.

Also added some #ifdef COMPILE_OPTIONS to face_detection_ex.cpp and face_landmark_detection_ex.cpp (less useful/generic)

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -105,6 +105,9 @@ target_link_libraries(assignment_learning_ex dlib::dlib)
 # your cmake projects and use the syntax shown above.
 macro(add_example name)
    add_executable(${name} ${name}.cpp)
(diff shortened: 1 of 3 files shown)


davisking:
Thanks for the PR, but this kind of thing isn't appropriate for the example programs.  The purpose of the examples is to be educational, not end user tools.

lenniea:
What about the setargv.obj on Windows?  This makes the Windows
tools/examples behave more like the Linux/Mac version?

-Lennie

davisking:
Still, these things aren't about being useful tools.  And not all compilers on windows know what setargv.obj is and many uses will see it as confusing magic.

PR #2141, opened 2020-07-30
loss multibinary log

arrufat:
Hi! I am planning on adding a multilabel loss layer to dlib.
The main motivation is, obviously, to be able to output multiple labels for a single image.
I know that `loss_multimulticlass` exists, but it is not quite the same, although one could achieve a similar result using as many classifiers as labels and two labels per class (present / not present);

I think the `loss_multilabel_log` I am suggesting might have some advantages:
- easier to use, since we don't need to fiddle with several `std::map<std::string, std::vector<std::string>>`
- it will allow for label smoothing in single label multiclass classifers

However, I have some questions with regards to the implementation.
- should the labels be `std::vector<float>` or `matrix<float>`: I think vectors are more intuitive, but matrices can use `dlib::sigmoid` to convert the output of the layer to a probability without writing your own sigmoid function.

Current implementation uses `std::vector`.

Also, maybe `loss_multilabel_log` is not a good name, maybe `loss_multibinary_log` is a better name? Open to suggestions.

This is still WIP, if you're interested in having this layer, I will add:
- [x] documentation
- [x] tests

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -785,6 +785,119 @@ namespace dlib
     { return rhs == static_cast<const std::string&>(lhs); }
 
 // ----------------------------------------------------------------------------------------
+
+    class loss_multilabel_log_
+    {
+    public:
+        typedef std::vector<float> training_label_type;
+        typedef std::vector<float> output_label_type;
+
+        template <
+            typename SUB_TYPE,
+            typename label_iterator
+            >
+        void to_label (
+            const tensor& input_tensor,
+            const SUB_TYPE& sub,
+            label_iterator iter
+        ) const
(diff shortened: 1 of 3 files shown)


davisking:
Sure, this seems like a good idea.  `loss_multibinary_log` seems like a good name.  Also, using `std::vector` for the labels seems sensible.  Whatever works most naturally in user code is best.

arrufat:
By the way, maybe I should have mentioned: this PR is ready to be reviewed :)

davisking:
Cool, I'll take a look in a bit :)

davisking:

arrufat:
Thank you :)

PR #2149, opened 2020-08-17
Added "get_random_complex_gaussian" to dlib::rand

pfeatherstone:
Added a function for computing a gaussian distributed complex number.
The real version is adapted to use the complex version.


davisking:
Thanks for the PR :)

PR #2151, opened 2020-08-19
Fix warning for zero variadic macro arguments.

samaldana:
When consuming dlib headers and building using gcc/clang with flags
'-Werror -Wpedantic', any inclusion involving DLIB_CASSERT triggers
a compilation error: ISO C++11 requires at least one argument for the
"..." in a variadic macro


davisking:
Cool, thanks for the PR :)

samaldana:
You're welcome :)

PR #2152, opened 2020-08-19
Divide mini-batches more uniformly across devices

reunanen:
**Problem:** I had a rare opportunity to test a 8-GPU machine. I copied my scripts from a one-GPU machine. There I had mini-batch size set to 50.
This caused a crash on the 8-GPU machine, because the mini-batch was allocated as 50 = 7 * 7 + 1, and the batch with size = 1 wasn't good for batch normalization.
Yes, I know that on the 8-GPU machine, I'm better off setting the mini-batch size e.g. to 8 * 50 = 400. However, I didn't immediately realize that this was the problem (the crash error message wasn't particularly helpful), and it took me quite a while to even get debugging started on this new-to-me machine.

**Suggested solution:** Divide the mini-batch more uniformly across the different devices.

### DIFF
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -1198,15 +1198,18 @@ namespace dlib
             job.test_only = test_only;
 
             // chop the data into devs blocks, each of about block_size elements.
-            size_t block_size = (num+devs-1)/devs;
+            const double block_size = num / static_cast<double>(devs);
 
             const auto prev_dev = dlib::cuda::get_device();
+
+            double j = 0;
+
             for (size_t i = 0; i < devs; ++i)
             {
                 dlib::cuda::set_device(devices[i]->device_id);
 
-                size_t start = i*block_size;
-                size_t stop  = std::min(num, start+block_size);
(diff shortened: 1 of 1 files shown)


dlib/dnn/trainer.h
```diff
@@ -1198,15 +1198,18 @@ namespace dlib
             job.test_only = test_only;
 
             // chop the data into devs blocks, each of about block_size elements.
-            size_t block_size = (num+devs-1)/devs;
+            const double block_size = num / static_cast<double>(devs);
```
davisking, line 1201: How about adding something like this as well?
```
CASSERT(num >= devs, "You need to give at least one training sample for each device you want to use.");
```

reunanen: I was thinking about that, too... But I thought it's not even necessary in the generic case (when not using batch normalization, for example)?

davisking: Yeah, on second thought if someone wants to do it we should let them.

davisking:

PR #2154, opened 2020-08-28
Add missing input/output mappings to mult_prev

arrufat:
I just noticed these two methods were missing for `mult_prev`.


davisking:
Good idea, thanks for the PR :)

Issue #2155, opened 2020-08-30
Add visitor to remove bias from layers that are inputs to `bn_` layers


arrufat:
The third introduction to deep neural networks ([dnn_introduction3_ex.cpp](../tree/e9c56fb21aa79d7353a6152aaeb7d8df52adf32f/examples/dnn_introduction3_ex.cpp)) shows how one can define a simple visitor to modify some networks parameters.

The [dnn_dcgan_ex.cpp](../tree/5a715fe24d588469f4a4c0ab2e9d35a6afcef131/examples/dnn_dcgan_train_ex.cpp) shows an example of a visitor that removes bias learning:
https://github.com/davisking/dlib/blob/5a715fe24d588469f4a4c0ab2e9d35a6afcef131/examples/dnn_dcgan_train_ex.cpp#L40-L55

However, as you might now, removing the bias learning from a layer makes a lot of sense if that layer itself is the input of a batch normalization layer, because the batch normalization will remove the mean, anyway.

I have tried to implement a visitor that will do such a thing without success, since, whenever I visit the network layers and I find a `bn_` layer at layer `i`, I need to disable the bias learning for the layer `i+1`. But I can't do that because `i` is not a `constexpr` in the following case:

``` cpp
namespace impl
{
    template <typename net_type>
    class visitor_bn_no_bias_prev
    {
    public:

        visitor_bn_no_bias_prev(net_type& net_) : net(net_) {}

        template <typename T>
        void set_no_bias(size_t , T&) const
        {
            // ignore other layer detail types
        }

        template <layer_mode mode>
        void set_no_bias(size_t i, bn_<mode>& )
        {
            layer<i>(net).layer_details().set_bias_learning_rate_multiplier(0);
            layer<i>(net).layer_details().set_bias_weight_decay_multiplier(0);
        }

        template<typename input_layer_type>
        void operator()(size_t , input_layer_type& ) const
        {
            // ignore other layers
        }

        template <typename T, typename U, typename E>
        void operator()(size_t i, add_layer<T,U,E>& l)
        {
            set_no_bias(i + 1, l.layer_details());
        }

    private:

        net_type& net;
    };
}


template <typename net_type>
void set_all_bn_prev_no_bias (
    net_type& net
)
{
    visit_layers(net, impl::visitor_bn_no_bias_prev<net_type>(net));
}
```

I think I should look at how `visit_layers` and is implemented and specially at `impl::vl_loop`, right? Any advice?
[...]

PR #2156, opened 2020-08-31
Add visitor to remove bias from bn_ layer inputs (#closes 2155)

arrufat:
As I explained in #2155, this PR adds a visitor that will automatically remove bias learning from all layers that are inputs of `bn_` layers.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -1523,12 +1523,11 @@ namespace dlib
             unsigned long new_window_size;
         };
 
-        template <typename net_type>
         class visitor_bn_no_bias_prev
         {
         public:
 
-            visitor_bn_no_bias_prev(net_type& net_) : net(net_) {}
+            visitor_bn_no_bias_prev() = default;
 
             template <typename T>
             void set_no_bias(size_t , T&) const
@@ -1555,9 +1554,6 @@ namespace dlib
                 set_no_bias(i, l);
             }
 
-        private:
-
-            net_type& net;
         };
     }
 
@@ -1575,7 +1571,7 @@ namespace dlib
         net_type& net
     )
     {
-        visit_layers(net, impl::visitor_bn_no_bias_prev<net_type>(net));
+        visit_layers(net, impl::visitor_bn_no_bias_prev());
     }
 
 // ----------------------------------------------------------------------------------------
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -1530,16 +1530,14 @@ namespace dlib
             visitor_bn_no_bias() = default;
 
             template <typename T>
-            void set_no_bias(T&) const
+            void set_input_no_bias(T&) const
             {
                 // ignore other layer types
             }
 
             template <layer_mode mode, typename U, typename E>
-            void set_no_bias(add_layer<bn_<mode>, U, E>& l)
+            void set_input_no_bias(add_layer<bn_<mode>, U, E>& l)
             {
-                l.layer_details().set_bias_learning_rate_multiplier(0);
-                l.layer_details().set_bias_weight_decay_multiplier(0);
                 l.subnet().layer_details().set_bias_learning_rate_multiplier(0);
                 l.subnet().layer_details().set_bias_weight_decay_multiplier(0);
             }
@@ -1553,9 +1551,8 @@ namespace dlib
             template <typename T, typename U, typename E>
             void operator()(size_t , add_layer<T,U,E>& l)
             {
-                set_no_bias(l);
+                set_input_no_bias(l);
             }
-
         };
     }
 
@@ -1569,7 +1566,7 @@ namespace dlib
     }
 
     template <typename net_type>
-    void set_all_bn_no_bias (
+    void set_all_bn_inputs_no_bias (
         net_type& net
     )
     {
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -1536,8 +1536,8 @@ namespace dlib
             template <layer_mode mode, typename U, typename E>
             void set_input_no_bias(add_layer<bn_<mode>, U, E>& l)
             {
-                l.subnet().layer_details().set_bias_learning_rate_multiplier(0);
-                l.subnet().layer_details().set_bias_weight_decay_multiplier(0);
+                set_bias_learning_rate_multiplier(l.subnet().layer_details(), 0);
+                set_bias_weight_decay_multiplier(l.subnet().layer_details(), 0);
             }
 
             template<typename input_layer_type>
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -1522,6 +1522,43 @@ namespace dlib
 
             unsigned long new_window_size;
         };
+
+        template <typename net_type>
+        class visitor_bn_no_bias_prev
+        {
+        public:
+
+            visitor_bn_no_bias_prev(net_type& net_) : net(net_) {}
+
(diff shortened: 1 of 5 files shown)


arrufat:
I think it's ready for review :)

dlib/dnn/layers.h
```diff
@@ -1522,6 +1522,36 @@ namespace dlib
...
+            void set_input_no_bias(T&) const
+            {
+                // ignore other layer types
+            }
+
+            template <layer_mode mode, typename U, typename E>
+            void set_input_no_bias(add_layer<bn_<mode>, U, E>& l)
+            {
+                l.subnet().layer_details().set_bias_learning_rate_multiplier(0);
+                l.subnet().layer_details().set_bias_weight_decay_multiplier(0);
```
davisking, line 1540: This should be this instead right?
```
        set_bias_learning_rate_multiplier(l.subnet().layer_details(), 0);
        set_bias_weight_decay_multiplier(l.subnet().layer_details(), 0);
```
just in case someone sticks a batch norm layer on top of something that doesn't have a `.set_bias_learning_rate_multiplier`

dlib/dnn/layers.h
```diff
@@ -1533,6 +1563,14 @@ namespace dlib
...
     }
 
+    template <typename net_type>
+    void set_all_bn_inputs_no_bias (
+        net_type& net
+    )
+    {
+        visit_layers(net, impl::visitor_bn_input_no_bias());
+    }
+
```
davisking, line 1573: This should probably go in core.h next to the other visitors that do this kind of thing.  Like right below `set_bias_learning_rate_multiplier` is reasonable.

arrufat: I wanted to put it there in the first place, but I just put it here because the template parameter is `layer_mode` is defined in this file. I guess that's also why you put here the `visitor_bn_running_stats_window_size` and the `set_all_bn_running_stats_window_sizes`.

dlib/dnn/layers.h
arrufat: Oh, good catch! thank you :)

arrufat:
> Thanks for the PR :)
> 
> As an side, you might want to add a `.disable_bias()` to the layers that makes them not execute the bias addition at all, since that will run faster. Then instead of calling `set_bias_learning_rate_multiplier()` you could call an equivalent `disable_bias()`.

Yes, that was bothering me, the fact that even all the values are 0, we're still performing the computation. I stared looking at the `fc_no_bias` to see how you disabled the bias there. I will try to do it :)

dlib/dnn/layers.h
davisking: Oh right, I forgot about `set_all_bn_running_stats_window_sizes`. Yeah, putting it here seems good :+1:

davisking:
> > Thanks for the PR :)
> > As an side, you might want to add a `.disable_bias()` to the layers that makes them not execute the bias addition at all, since that will run faster. Then instead of calling `set_bias_learning_rate_multiplier()` you could call an equivalent `disable_bias()`.
> 
> Yes, that was bothering me, the fact that even all the values are 0, we're still performing the computation. I stared looking at the `fc_no_bias` to see how you disabled the bias there. I will try to do it :)

Na, don't do it like `fc_no_bias`.  It's probably nicer and simpler if this is just a normal runtime `bool` in each applicable layer that controls if the bias is added or not. Then a member function can just set it to true or false.

arrufat:
Done, I hope this is what you expected. I had to modify the serialize functions as well, but I made sure to keep backwards compatibility: I've successfully loaded and run the following models:
-  resnet34 model, which contains `fc` layers
- dlib_face_recognition_resnet_model, which contains `fc_no_bias` and `con` layers
- a saved sync file from the dcgan example, which contains `con` and `cont` layers.

I've also checked that the DCGAN example still runs. I've modified a little bit to print out the number of trainable parameters, to ensure the bias are not allocated when disabling the bias.

dlib/dnn/layers.h
```diff
@@ -1522,6 +1564,37 @@ namespace dlib
...
+            {
+                // ignore other layer types
+            }
+
+            template <layer_mode mode, typename U, typename E>
+            void set_input_no_bias(add_layer<bn_<mode>, U, E>& l)
+            {
+                disable_bias(l.subnet().layer_details());
+                set_bias_learning_rate_multiplier(l.subnet().layer_details(), 0);
+                set_bias_weight_decay_multiplier(l.subnet().layer_details(), 0);
```
arrufat, line 1583: I know these two functions are redundant, but I thought it was good to keep them any way.

dlib/dnn/layers.h
```diff
@@ -289,8 +305,11 @@ namespace dlib
                 << ")";
             out << " learning_rate_mult="<<item.learning_rate_multiplier;
             out << " weight_decay_mult="<<item.weight_decay_multiplier;
-            out << " bias_learning_rate_mult="<<item.bias_learning_rate_multiplier;
-            out << " bias_weight_decay_mult="<<item.bias_weight_decay_multiplier;
+            if (item.use_bias)
+            {
+                out << " bias_learning_rate_mult="<<item.bias_learning_rate_multiplier;
+                out << " bias_weight_decay_mult="<<item.bias_weight_decay_multiplier;
+            }
```
davisking, line 312: `to_xml()` needs to be updated as well.  There adding a `use_bias=true` or `use_bias=false` attribute is reasonable.

dlib/dnn/layers.h
davisking: Yeah, seems reasonable.

dlib/dnn/layers_abstract.h
```diff
@@ -573,6 +573,13 @@ namespace dlib
                 - #get_bias_weight_decay_multiplier() == val
         !*/
 
+        void disable_bias(
+        );
+        /*!
+            ensures
+                - disables bias for this layer
```
davisking, line 580: Add a `bool bias_is_disabled()` method too so that someone can tell if a layer has had its bias disabled or not.  Then make the docs for the `disable_bias()` methods note that they set `bias_is_disabled()` to false.

arrufat:
Do you think I need update the `operator<<` so that I show the bias status? Right now, the only way to know is by looking at whether the bias learning rates are printed.

davisking:
> Do you think I need update the `operator<<` so that I show the bias status? Right now, the only way to know is by looking at whether the bias learning rates are printed.

I thought about that earlier and I'm on the fence.  `operator<<` output is just for humans to eyeball and at some level adding more and more stuff to it makes it busy and hard to read.  Take a look at it for a printout of a network and set it however you think looks the most clear in that context.  But likely what you have in there now is fine.

arrufat:
I think it's more clear this way:
- if we are using bias, keep it as it was, that is, print the bias multipliers
- otherwise, do not print those, and print `use_bias=false`, so that users don't think there has been an error, or the output has been truncated.

I think it's a good trade-off, since we are not making the lines with bias even longer, and the ones without are still short in comparison.

davisking:
> I think it's more clear this way:
> 
> * if we are using bias, keep it as it was, that is, print the bias multipliers
> * otherwise, do not print those, and print `use_bias=false`, so that users don't think there has been an error, or the output has been truncated.
> 
> I think it's a good trade-off, since we are not making the lines with bias even longer, and the ones without are still short in comparison.

Sounds good :)

davisking:
Sweet, thanks for another cool PR :)

arrufat:
Thank you! I am really learning a lot by doing these things :)

Issue #2153, opened 2020-08-26
Training mmod detector crashes when using bounding box regression


arrufat:
I am training a normal mmod detector like the ones in the examples, but I enable bounding box regression with:
``` cpp
net.subnet().layer_details().set_num_filters(5 * options.detector_windows.size());
```

## Current Behavior
I get the following error really often:
```
Error detected at line 1591.                                       
Error detected in file ../../external/dlib/dlib/../dlib/dnn/loss.h.
```
which refers to: https://github.com/davisking/dlib/blob/b401185aa5a59bfff8eb5f4675a7e4802c37b070/dlib/dnn/loss.h#L1591-L1592

I was not getting this error with the same code a while ago, but many things have changed since (especially CUDA versions).
Most of the time, this error happens at the beginning of the training, when everything is very chaotic. However, sometimes I get the error after the loss has stabilized, as well.

I tried updating the gradient for `h` and `w` only when they are positive and let the gradient be 0 otherwise.
This avoids the crashing, but messes up the training (loss goes to `inf`).

I've also noticed this didn't happen when I changed the lambda value to much lower values, such as 1, instead of the default 100. Maybe it's just that the lambda is too big?
EDIT: it also happens.

How would you proceed?

* **Version**: dlib master (19.21.99)
* **Where did you get dlib**: github
* **Platform**: Linux 64 bit, CUDA 11.0.2, CUDNN 8.0.2.39
* **Compiler**: GCC-9.3.0 for C (in order to enable CUDA) and GCC-10.2 for C++

arrufat:
I think I have figured a way to make it work:

1.  train detector without bounding box regression (bbr)
1. retrain the previous detector with bbr (I set `lambda_bbr = 10`) and initialize all but final `con` layer with previous detector, as number of filters don't match (basically: `net_bbr10.subnet().subnet() = net_nobbr.subnet().subnet();`
1. finetune the previous detector with `lambda_bbr = 100`. This time we start from the whole network and a smaller learning rate (basically: `net_bbr100.subnet() = net.bbr10.subnet();`)

I've tried this so far and it does not crash because of that assertion. It would be nice though to have it working without these intermediate steps.

I am open to suggestions on how we could tackle this :)

davisking:
Na, you shouldn't have to do any of that hacky stuff. 

I just looked at the code and oops, it's totally wrong.  It should be like this:
```
diff --git a/dlib/dnn/loss.h b/dlib/dnn/loss.h
[...]

PR #2157, opened 2020-09-02
Add missing input/output mappings to scale layer

arrufat:
These are needed if we want to use a detector that has some kind of channel attention (I found it the hard way :P)

### DIFF
--- a/dlib/dnn/layers_abstract.h
+++ b/dlib/dnn/layers_abstract.h
@@ -2554,6 +2554,8 @@ namespace dlib
         template <typename SUBNET> void setup (const SUBNET& sub);
         template <typename SUBNET> void forward(const SUBNET& sub, resizable_tensor& output);
         template <typename SUBNET> void backward(const tensor& gradient_input, SUBNET& sub, tensor& params_grad);
+        dpoint map_input_to_output(dpoint p) const;
+        dpoint map_output_to_input(dpoint p) const;
         const tensor& get_layer_params() const; 
         tensor& get_layer_params(); 
         /*!
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2547,6 +2547,9 @@ namespace dlib
         const tensor& get_layer_params() const { return params; }
         tensor& get_layer_params() { return params; }
 
+        inline dpoint map_input_to_output (const dpoint& p) const { return p; }
+        inline dpoint map_output_to_input (const dpoint& p) const { return p; }
+
         friend void serialize(const scale_& item, std::ostream& out)
         {
             serialize("scale_", out);


dlib/dnn/layers_abstract.h
```diff
@@ -2554,6 +2554,8 @@ namespace dlib
         template <typename SUBNET> void setup (const SUBNET& sub);
         template <typename SUBNET> void forward(const SUBNET& sub, resizable_tensor& output);
         template <typename SUBNET> void backward(const tensor& gradient_input, SUBNET& sub, tensor& params_grad);
+        dpoint map_input_to_output(dpoint p) const;
+        dpoint map_output_to_input(dpoint p) const;
```
davisking, line 2558: Doesn't this do something you don't want since the input here is the `SCALES` tensor, which has only 1 row and column?

arrufat: Oh, I didn't think about that, I was trying to define this block, while implementing the effective Squeeze and Excitation module from the [VoVNetV2](https://arxiv.org/abs/1911.06667) backbone:

```
template <long num_filters, typename SUBNET>
        using ese_module =
        scale1<
        sig<con<num_filters, 1, 1, 1, 1,
        avg_pool_everything<
        tag1<SUBNET>>>>>;
```

Then dlib complained that `scale_` doesn't have those methods, so I just added them, since they just modify the channels.
The plots in #2153 are generated using a detector that has this backbone. I thought, since it just modifies the channels of `SUBNET`, it should not cause trouble. Am I missing something?

arrufat: Yes, you're right, I was not using the `scale_` layer for those training experiments.
I realized I have a residual block in the network that goes around the `ese_module`, maybe that is why I was still able to learn a detector. These networks have so many things and parameters, than even when doing something wrong, they might still work...

So how should one go about this? or is it not possible to use the scale layer with the mmod loss?

davisking: Yeah, `scale_` is kinda borked for this.  Really the two inputs to `scale_` should probably have been switched, so that the direct input is the shape of its output.  Then this would work the way you want.   I don't think we can change it now though since `scale_` has been around a while and users likely depend on it being the way it is.

You could add a new scale layer that works differently ¯\_(ツ)_/¯

arrufat: This new scale layer would just need to swap its inputs right? I think I can handle that... Maybe the hardest part is coming up with a name...

Maybe it's a bit dirty, but could we handle it with the layer version, let's say if we deserialize `scale_` version, then it behaves as now, otherwise it swaps its inputs, but this might be very confusing...

arrufat: That is weird, I just retrained a detector with and without `ese_module` with the `scale_` layer, and it's working in both cases on the same dataset as I used in #2153.
- Without eSE module (VoVNet_V1):
```
  set |     prec.   recall      mAP
------+----------------------------
train |         1 0.957447 0.957447 
```
- With eSE module (VoVNet_V2)
```
  set |     prec.   recall      mAP
------+----------------------------
train |         1 0.978723 0.978723 
```
Here's a result of the VoVNet_V2 detector (using the eSE module, and therefore the `scale_` layer) on an image in the training set. (I am not interested in horse racing or anything, I just built a small dataset to test detectors where some objects could appear inside others, as in this example).

detections

However, if I understand it correctly, it should not be possible to get a working detector using this layer, right? Also, during the training I should see how all objects are mapped to the same coordinates in the feature space, which I don't, so maybe it's working?

davisking: You sure you didn't have a skip layer that skipped over the scale layers?  Like `input_tensor_to_output_tensor()` traces the network from input to output but follows skip layers, so if this stuff was skipped then it would have worked out.

arrufat: I am quite sure, after checking the network definition (I will publish it once I know it's right).

I implemented the One-Shot Aggregation (OSA) module from [VoVNet](https://arxiv.org/abs/1904.09730) architecture:
image

Then I added the the effective Sequeeze and Excitation module (eSE) from [VoVNet V2](https://arxiv.org/abs/1911.06667) (right-most diagram without the identity mapping, aka residual):

image

You can see the printed architecture here, as well as the tensor sizes at each layer for an input of 224x224 after a `pyramid_down<6>`. The gist of it is:
- downsample by 4 with 2 `relu<bn_con<con<M, 3, 3, 2, 2, SUBNET>>>`
- OSA Module:
  - concatenate the input + the output of 3 sequential `relu<bn_con<con<M, 3, 3, 1, 1, SUBNET>>>`
  - eSE module: scale the previous output with the result of `sig<con<N, 1, 1, 1, 1,avg_pool_everything, SUBNET>>`
- max_pool, OSA Module, max_pool, OSA module
- detection head
```
layer<0>	loss_mmod	 (detector_windows:(horse:80x74,55x80;person:74x80,50x80), loss per FA:1, loss per miss:1, truth match IOU thresh:0.5, use_bounding_box_regression:1, bbr_lambda:100, overlaps_nms:(0.5,1), overlaps_ignore:(0.5,0.95))
layer<1>	output size=(num:1, k:20,   nr:54,  nc:16)	con	 (num_filters=20, nr=5, nc=5, stride_y=1, stride_x=1, padding_y=2, padding_x=2) learning_rate_mult=1 weight_decay_mult=1 bias_learning_rate_mult=1 bias_weight_decay_mult=0
layer<2>	output size=(num:1, k:384,  nr:54,  nc:16)	scale1
layer<3>	output size=(num:1, k:384,  nr:1,   nc:1) 	sig
layer<4>	output size=(num:1, k:384,  nr:1,   nc:1) 	con	 (num_filters=384, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 bias_learning_rate_mult=1 bias_weight_decay_mult=0
layer<5>	output size=(num:1, k:384,  nr:1,   nc:1) 	avg_pool (nr=0, nc=0, stride_y=1, stride_x=1, padding_y=0, padding_x=0)
layer<6>	output size=(num:1, k:384,  nr:54,  nc:16)	tag1
layer<7>	output size=(num:1, k:384,  nr:54,  nc:16)	relu
layer<8>	output size=(num:1, k:384,  nr:54,  nc:16)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<9>	output size=(num:1, k:384,  nr:54,  nc:16)	con	 (num_filters=384, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<10>	output size=(num:1, k:544,  nr:54,  nc:16)	concat	 (6000,6001,6002,6003)
layer<11>	output size=(num:1, k:96,   nr:54,  nc:16)	tag6003
layer<12>	output size=(num:1, k:96,   nr:54,  nc:16)	relu
layer<13>	output size=(num:1, k:96,   nr:54,  nc:16)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<14>	output size=(num:1, k:96,   nr:54,  nc:16)	con	 (num_filters=96, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<15>	output size=(num:1, k:96,   nr:54,  nc:16)	tag6002
layer<16>	output size=(num:1, k:96,   nr:54,  nc:16)	relu
layer<17>	output size=(num:1, k:96,   nr:54,  nc:16)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<18>	output size=(num:1, k:96,   nr:54,  nc:16)	con	 (num_filters=96, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<19>	output size=(num:1, k:96,   nr:54,  nc:16)	tag6001
layer<20>	output size=(num:1, k:96,   nr:54,  nc:16)	relu
layer<21>	output size=(num:1, k:96,   nr:54,  nc:16)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<22>	output size=(num:1, k:96,   nr:54,  nc:16)	con	 (num_filters=96, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<23>	output size=(num:1, k:256,  nr:54,  nc:16)	tag6000
layer<24>	output size=(num:1, k:256,  nr:54,  nc:16)	max_pool (nr=3, nc=3, stride_y=2, stride_x=2, padding_y=1, padding_x=1)
layer<25>	output size=(num:1, k:256,  nr:108, nc:31)	scale1
layer<26>	output size=(num:1, k:256,  nr:1,   nc:1) 	sig
layer<27>	output size=(num:1, k:256,  nr:1,   nc:1) 	con	 (num_filters=256, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 bias_learning_rate_mult=1 bias_weight_decay_mult=0
layer<28>	output size=(num:1, k:256,  nr:1,   nc:1) 	avg_pool (nr=0, nc=0, stride_y=1, stride_x=1, padding_y=0, padding_x=0)
layer<29>	output size=(num:1, k:256,  nr:108, nc:31)	tag1
layer<30>	output size=(num:1, k:256,  nr:108, nc:31)	relu
layer<31>	output size=(num:1, k:256,  nr:108, nc:31)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<32>	output size=(num:1, k:256,  nr:108, nc:31)	con	 (num_filters=256, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<33>	output size=(num:1, k:352,  nr:108, nc:31)	concat	 (6000,6001,6002,6003)
layer<34>	output size=(num:1, k:80,   nr:108, nc:31)	tag6003
layer<35>	output size=(num:1, k:80,   nr:108, nc:31)	relu
layer<36>	output size=(num:1, k:80,   nr:108, nc:31)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<37>	output size=(num:1, k:80,   nr:108, nc:31)	con	 (num_filters=80, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<38>	output size=(num:1, k:80,   nr:108, nc:31)	tag6002
layer<39>	output size=(num:1, k:80,   nr:108, nc:31)	relu
layer<40>	output size=(num:1, k:80,   nr:108, nc:31)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<41>	output size=(num:1, k:80,   nr:108, nc:31)	con	 (num_filters=80, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<42>	output size=(num:1, k:80,   nr:108, nc:31)	tag6001
layer<43>	output size=(num:1, k:80,   nr:108, nc:31)	relu
layer<44>	output size=(num:1, k:80,   nr:108, nc:31)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<45>	output size=(num:1, k:80,   nr:108, nc:31)	con	 (num_filters=80, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<46>	output size=(num:1, k:112,  nr:108, nc:31)	tag6000
layer<47>	output size=(num:1, k:112,  nr:108, nc:31)	max_pool (nr=3, nc=3, stride_y=2, stride_x=2, padding_y=1, padding_x=1)
layer<48>	output size=(num:1, k:112,  nr:216, nc:62)	scale1
layer<49>	output size=(num:1, k:112,  nr:1,   nc:1) 	sig
layer<50>	output size=(num:1, k:112,  nr:1,   nc:1) 	con	 (num_filters=112, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 bias_learning_rate_mult=1 bias_weight_decay_mult=0
layer<51>	output size=(num:1, k:112,  nr:1,   nc:1) 	avg_pool (nr=0, nc=0, stride_y=1, stride_x=1, padding_y=0, padding_x=0)
layer<52>	output size=(num:1, k:112,  nr:216, nc:62)	tag1
layer<53>	output size=(num:1, k:112,  nr:216, nc:62)	relu
layer<54>	output size=(num:1, k:112,  nr:216, nc:62)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<55>	output size=(num:1, k:112,  nr:216, nc:62)	con	 (num_filters=112, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<56>	output size=(num:1, k:320,  nr:216, nc:62)	concat	 (6000,6001,6002,6003)
layer<57>	output size=(num:1, k:64,   nr:216, nc:62)	tag6003
layer<58>	output size=(num:1, k:64,   nr:216, nc:62)	relu
layer<59>	output size=(num:1, k:64,   nr:216, nc:62)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<60>	output size=(num:1, k:64,   nr:216, nc:62)	con	 (num_filters=64, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<61>	output size=(num:1, k:64,   nr:216, nc:62)	tag6002
layer<62>	output size=(num:1, k:64,   nr:216, nc:62)	relu
layer<63>	output size=(num:1, k:64,   nr:216, nc:62)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<64>	output size=(num:1, k:64,   nr:216, nc:62)	con	 (num_filters=64, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<65>	output size=(num:1, k:64,   nr:216, nc:62)	tag6001
layer<66>	output size=(num:1, k:64,   nr:216, nc:62)	relu
layer<67>	output size=(num:1, k:64,   nr:216, nc:62)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<68>	output size=(num:1, k:64,   nr:216, nc:62)	con	 (num_filters=64, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<69>	output size=(num:1, k:128,  nr:216, nc:62)	tag6000
layer<70>	output size=(num:1, k:128,  nr:216, nc:62)	relu
layer<71>	output size=(num:1, k:128,  nr:216, nc:62)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<72>	output size=(num:1, k:128,  nr:216, nc:62)	con	 (num_filters=128, nr=3, nc=3, stride_y=2, stride_x=2, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<73>	output size=(num:1, k:64,   nr:432, nc:123)	relu
layer<74>	output size=(num:1, k:64,   nr:432, nc:123)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<75>	output size=(num:1, k:64,   nr:432, nc:123)	con	 (num_filters=64, nr=3, nc=3, stride_y=1, stride_x=1, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<76>	output size=(num:1, k:64,   nr:432, nc:123)	relu
layer<77>	output size=(num:1, k:64,   nr:432, nc:123)	bn_con   eps=0.0001 running_stats_window_size=1000 learning_rate_mult=1 weight_decay_mult=0 bias_learning_rate_mult=1 bias_weight_decay_mult=1
layer<78>	output size=(num:1, k:64,   nr:432, nc:123)	con	 (num_filters=64, nr=3, nc=3, stride_y=2, stride_x=2, padding_y=1, padding_x=1) learning_rate_mult=1 weight_decay_mult=1 use_bias=false
layer<79>	input_rgb_image_pyramid(122.782,117.001,104.298) pyramid_padding=10 pyramid_outer_padding=11
number of parameters: 1599300
```

From what I see, the `scale_` layer already does some kind of skip connection in dlib, as it's direct input has to be `n x k x 1 x 1`, but the tagged layer is of an arbitrary size, and its output is the same size as the tagged layer. The curious thing is that not only does it work, but it also gives better results than without it (which makes me think that channel attention is doing something useful).

arrufat:
as an additional experiment, I am training the `dnn_mmod_train_find_cars_ex` program with only one change: I added the following just after the last convolutional layer
```
scale1<sig<con<55,1,1,1,1,avg_pool_everything<tag1<
```

So, the network architecture now looks like this:
```
using net_type = loss_mmod<con<1,9,9,1,1,
scale1<sig<con<55,1,1,1,1,avg_pool_everything<tag1<
rcon5<rcon5<rcon5<downsampler<input_rgb_image_pyramid<pyramid_down<6>>>>>>>>>>>>>;
```
I will let you know how it goes.

davisking:
Looking at the network you pasted, you have this in it
```
layer<2>	output size=(num:1, k:384,  nr:54,  nc:16)	scale1
layer<3>	output size=(num:1, k:384,  nr:1,   nc:1) 	sig
layer<4>	output size=(num:1, k:384,  nr:1,   nc:1) 	con	 (num_filters=384, nr=1, nc=1, stride_y=1, stride_x=1, padding_y=0, padding_x=0) learning_rate_mult=1 weight_decay_mult=1 bias_learning_rate_mult=1 bias_weight_decay_mult=0
layer<5>	output size=(num:1, k:384,  nr:1,   nc:1) 	avg_pool (nr=0, nc=0, stride_y=1, stride_x=1, padding_y=0, padding_x=0)
layer<6>	output size=(num:1, k:384,  nr:54,  nc:16)	tag1
```
The stuff between the `scale1` and `tag1` applies a bunch of identity transforms, so it just happens to work out to the mapping you want.  But that's just sort of lucky.  If for instance you had some conv with a stride other than 1 in there it would get all messed up.

davisking:
There should probably be a `scale_prev` layer that's just like `scale` except the two input tensors are reversed.

arrufat:
Oh, I see! Thank you for the insight!

The `dnn_mmod_train_find_cars_ex` didn't converge with my modification, so finally I got something that doesn't work.

With regards, of the `scale_prev` that is actually what I am doing right now. I am retraining with a scale_ layer with swapped inputs, so now it looks like this:

```
...
scale_prev2<skip1<tag2<sig<con<55,1,1,1,1,avg_pool_everything<tag1<
...
```
If that works, I will modify this PR to include the `scale_prev` layer. Unless you want me to close this one and make a new one.

davisking:
Cool. However you want to organize it is fine.

PR #2161, opened 2020-09-05
Adding move semantics to type_safe_union

pfeatherstone:
Not quite ready for merge. Really need @davisking to have a look. I'm of the opinion that swaps should be written in terms of moves, not the other way round. And this would need some heavy unit testing.

### DIFF
--- a/dlib/algs.h
+++ b/dlib/algs.h
@@ -115,7 +115,6 @@ namespace std
 #include "enable_if.h"
 #include "uintn.h"
 #include "numeric_constants.h"
-#include "platform.h" // for WIN32
 #include "memory_manager_stateless/memory_manager_stateless_kernel_1.h" // for the default memory manager
 
 
@@ -426,7 +425,6 @@ namespace dlib
     
 // ----------------------------------------------------------------------------------------
     
-#ifndef WIN32 //For some reason, the visual studio compiler throws error : error C2210: 'to': pack expansions cannot be used as arguments to non-packed parameters in alias templates
     /*!A is_any 
 
         This is a template where is_any<T,U,V>::value == true when T is the same type as U or V and false otherwise.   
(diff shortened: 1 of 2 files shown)


pfeatherstone:
Observation : each unit test file takes roughly 15s to compile. Wow. I wish C++ modules came sooner.

pfeatherstone:
Looks live Visual Studio compiler complains about this:
```error C2210: 'to': pack expansions cannot be used as arguments to non-packed parameters in alias templates```

davisking:
I'm not totally sure I understand what you are working towards in this PR, but I just now pushed some minor changes to `type_safe_union` that add what I would think of as move semantics support.  If there is something else you want it to do then feel free to update the PR to add that as well.

pfeatherstone:
Yep. I got it all working but i was also doing some stuff with enable_if instead of using validate_type, creating variadic versions of std::is_same and std::is_convertible, and using the latter two in type_safe_union to reduce code.Sod it, what you've done works.

pfeatherstone:
And the variadic stuff wasn't playing ball with the visual studio compiler. Otherwise all the tests run fine on my PC.

pfeatherstone:
I can't be bothered to battle with compilers that don't support full C++11

pfeatherstone:
Does it make sense for type_safe_union to be noncopyable if the copy constructors for the underlying template parameters are well defined? May as well define the copy constructor and copy assign operator for type_safe_union.

pfeatherstone:
If you're worried about it supporting non-copyable types, you can use enable_if with std::is_copy_constructible on the copy constructors.

davisking:
`enable_if` isn't needed to deal with non-copyable types.  It's just a matter of adding the appropriate copy constructor and copy assignment operator.  I'm somewhat indifferent to this.  I like non-copyable types because it forces people to think more carefully about what they are doing and they end up doing many less unnecessary copies.  But if you want to update this PR to add the two missing functions to make this all copyable that's fine with me as well.

PR #2162, opened 2020-09-06
Show how to use the new visitors with lambdas

arrufat:
Following your example in this [reddit thread](https://www.reddit.com/r/dlib/comments/in0upc/setup_lealy_reluprelu_activation_layers_in_a/), I've updated the dcgan example to show how to setup `leaky_relu_` layers using the new style visitors with a lambda.

I think it's worth having because the 2nd introduction already shows how to do with the network constructor.


davisking:
Cool :)

Issue #2163, opened 2020-09-06
I think set_all_bn_inputs_no_bias broke something


arrufat:
So, I was trying to stress test the `set_all_bn_inputs_no_bias` to make sure it works everywhere.
I've tried many architectures and setups and it works. However, when I tackled the imagenet training example, I got this error:
```
SCANNING IMAGENET DATASET

images in dataset: 1281167
step#: 0     learning rate: 0.1   average loss: 0            steps without apparent progress: 0

**************************** FATAL ERROR DETECTED ****************************

Error detected at line 938.
Error detected in file /home/adria/imagenet/dlib/dlib/cuda/cuda_dlib.cu.
Error detected in function void dlib::cuda::affine_transform(dlib::tensor&, const dlib::tensor&, const dlib::tensor&, float, float).

Failing expression was dest.size()==src2.size().


******************************************************************************

Aborted (core dumped)
```

The only changes I made to the example program are:
```
diff --git a/examples/dnn_imagenet_train_ex.cpp b/examples/dnn_imagenet_train_ex.cpp
index e672018d..2f1cee88 100644
--- a/examples/dnn_imagenet_train_ex.cpp
+++ b/examples/dnn_imagenet_train_ex.cpp
@@ -214,7 +214,8 @@ int main(int argc, char** argv) try
 
     cout << "\nSCANNING IMAGENET DATASET\n" << endl;
 
-    auto listing = get_imagenet_train_listing(string(argv[1])+"/Data/CLS-LOC/train/");
+    // auto listing = get_imagenet_train_listing(string(argv[1])+"/Data/CLS-LOC/train/");
+    auto listing = get_imagenet_train_listing(string(argv[1]));
     cout << "images in dataset: " << listing.size() << endl;
     const auto number_of_classes = listing.back().numeric_label+1;
     if (listing.size() == 0 || number_of_classes != 1000)
@@ -231,7 +232,7 @@ int main(int argc, char** argv) try
     const double momentum = 0.9;
 
     net_type net;
-    dnn_trainer<net_type> trainer(net,sgd(weight_decay, momentum));
+    dnn_trainer<net_type> trainer(net,sgd(weight_decay, momentum), {0, 1});
     trainer.be_verbose();
     trainer.set_learning_rate(initial_learning_rate);
     trainer.set_synchronization_file("imagenet_trainer_state_file.dat", std::chrono::minutes(10));
@@ -241,6 +242,7 @@ int main(int argc, char** argv) try
[...]

PR #2165, opened 2020-09-06
Add error message for mismatched tensor sizes in dnn_trainer

arrufat:
As discussed in #2163, this PR adds a check and a helpful message to inform the user that the tensor sizes in the devices have different size, probably due to an external modification of some network parameters after trainer construction


davisking:

PR #2171, opened 2020-09-09
Add scale_prev layer

arrufat:
As discussed in #2157, this PR adds a new layer (`scale_prev_`) which is the same as `scale_` except that the inputs are swapped, which is useful since it allows mapping between inputs and outputs of this layer.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2764,9 +2764,6 @@ namespace dlib
         {
             auto&& src = sub.get_output();
             auto&& scales = layer<tag>(sub).get_output();
-            // The gradient just flows backwards to the two layers that forward()
-            // read from.
-            // tt::scale_channels(true, layer<tag>(sub).get_gradient_input(), gradient_input, scales);
             tt::scale_channels(true, sub.get_gradient_input(), gradient_input, scales);
 
             if (reshape_src.num_samples() != src.num_samples())
@@ -2775,7 +2772,7 @@ namespace dlib
                 reshape_src = alias_tensor(src.num_samples()*src.k(),src.nr()*src.nc());
             }
 
-            auto&& scales_grad = sub.get_gradient_input();
+            auto&& scales_grad = layer<tag>(sub).get_gradient_input();
             auto sgrad = reshape_scales(scales_grad);
             tt::dot_prods(true, sgrad, reshape_src(src), reshape_src(gradient_input));
         }
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2725,6 +2725,131 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <
+        template<typename> class tag
+        >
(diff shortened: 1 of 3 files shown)


dlib/dnn/layers.h
```diff
@@ -2723,6 +2723,131 @@ namespace dlib
...
+        }
+
+        template <typename SUBNET>
+        void backward(const tensor& gradient_input, SUBNET& sub, tensor& /*params_grad*/)
+        {
+            auto&& src = sub.get_output();
+            auto&& scales = layer<tag>(sub).get_output();
+            // The gradient just flows backwards to the two layers that forward()
+            // read from.
+            // tt::scale_channels(true, layer<tag>(sub).get_gradient_input(), gradient_input, scales);
```
davisking, line 2769: Remove this comment.

dlib/dnn/layers.h
```diff
@@ -2723,6 +2723,131 @@ namespace dlib
...
+            tt::scale_channels(true, sub.get_gradient_input(), gradient_input, scales);
+
+            if (reshape_src.num_samples() != src.num_samples())
+            {
+                reshape_scales = alias_tensor(src.num_samples()*src.k());
+                reshape_src = alias_tensor(src.num_samples()*src.k(),src.nr()*src.nc());
+            }
+
+            auto&& scales_grad = sub.get_gradient_input();
```
davisking, line 2778: Shouldn't this be this?
```
auto&& scales_grad = layer<tag>(sub).get_gradient_input();
```

davisking (changes requested):
Thanks for another PR :)

It looks like the gradient isn't quite right.  Can you also add a small test that exercises this?

dlib/dnn/layers.h
arrufat: Oops, I did it too late at night yesterday...

dlib/dnn/layers.h
arrufat: > Thanks for another PR :)
> 
> It looks like the gradient isn't quite right. Can you also add a small test that exercises this?

Sure, test coming up! :)

arrufat:
> Thanks for another PR :)
> 
> It looks like the gradient isn't quite right. Can you also add a small test that exercises this?

Sure, test coming up! :)

arrufat:
I've added a simple test that makes sure one can reproduce the behavior of `scale_` with `scale_prev_`.

davisking:
Looks good, thanks for another nice PR :)

arrufat:
Thanks.

Quick question, if I use this new `scale_prev` layer to replicate the behavior of the `scale` layer (using one skip layer and extra tag layer, like in the test) is it going to impact performance? That is, memory usage and speed, since with tag and skip layers make copies of tensors, right?

davisking:
They don't make copies or do anything at runtime, so there is no
performance implication.

arrufat:
Awesome, thank you :)

PR #2173, opened 2020-09-09
Serialization macros inspired by nlohmann/json + serialise to/from memory

pfeatherstone:
Added serialisation macros. Unit tests to come.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2725,6 +2725,128 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <
+        template<typename> class tag
+        >
+    class scale_prev_
+    {
+    public:
+        const static unsigned long id = tag_id<tag>::id;
+
+        scale_prev_()
+        {
+        }
+
+        template <typename SUBNET>
+        void setup (const SUBNET& /*sub*/)
+        {
+        }
+
+        template <typename SUBNET>
+        void forward(const SUBNET& sub, resizable_tensor& output)
+        {
+            auto&& src = sub.get_output();
+            auto&& scales = layer<tag>(sub).get_output();
+            DLIB_CASSERT(scales.num_samples() == src.num_samples() &&
+                         scales.k()           == src.k() &&
+                         scales.nr()          == 1 &&
+                         scales.nc()          == 1,
+                         "scales.k(): " << scales.k() <<
+                         "\nsrc.k(): " << src.k()
+                         );
+
+            output.copy_size(src);
+            tt::scale_channels(false, output, src, scales);
+        }
+
+        template <typename SUBNET>
+        void backward(const tensor& gradient_input, SUBNET& sub, tensor& /*params_grad*/)
+        {
+            auto&& src = sub.get_output();
+            auto&& scales = layer<tag>(sub).get_output();
+            tt::scale_channels(true, sub.get_gradient_input(), gradient_input, scales);
+
+            if (reshape_src.num_samples() != src.num_samples())
+            {
+                reshape_scales = alias_tensor(src.num_samples()*src.k());
+                reshape_src = alias_tensor(src.num_samples()*src.k(),src.nr()*src.nc());
+            }
+
+            auto&& scales_grad = layer<tag>(sub).get_gradient_input();
+            auto sgrad = reshape_scales(scales_grad);
+            tt::dot_prods(true, sgrad, reshape_src(src), reshape_src(gradient_input));
+        }
+
+        const tensor& get_layer_params() const { return params; }
+        tensor& get_layer_params() { return params; }
+
+        inline dpoint map_input_to_output (const dpoint& p) const { return p; }
+        inline dpoint map_output_to_input (const dpoint& p) const { return p; }
+
+        friend void serialize(const scale_prev_& item, std::ostream& out)
+        {
+            serialize("scale_prev_", out);
+            serialize(item.reshape_scales, out);
+            serialize(item.reshape_src, out);
+        }
+
+        friend void deserialize(scale_prev_& item, std::istream& in)
+        {
+            std::string version;
+            deserialize(version, in);
+            if (version != "scale_prev_")
+                throw serialization_error("Unexpected version '"+version+"' found while deserializing dlib::scale_prev_.");
+            deserialize(item.reshape_scales, in);
+            deserialize(item.reshape_src, in);
+        }
+
+        friend std::ostream& operator<<(std::ostream& out, const scale_prev_& /*item*/)
+        {
+            out << "scale_prev"<<id;
+            return out;
+        }
+
+        friend void to_xml(const scale_prev_& /*item*/, std::ostream& out)
+        {
+            out << "<scale_prev tag='"<<id<<"'/>\n";
+        }
+
+    private:
+        alias_tensor reshape_scales;
+        alias_tensor reshape_src;
+        resizable_tensor params;
+    };
+
+    template <
+        template<typename> class tag,
+        typename SUBNET
+        >
+    using scale_prev = add_layer<scale_prev_<tag>, SUBNET>;
+
+    template <typename SUBNET> using scale_prev1  = scale_prev<tag1, SUBNET>;
+    template <typename SUBNET> using scale_prev2  = scale_prev<tag2, SUBNET>;
+    template <typename SUBNET> using scale_prev3  = scale_prev<tag3, SUBNET>;
+    template <typename SUBNET> using scale_prev4  = scale_prev<tag4, SUBNET>;
+    template <typename SUBNET> using scale_prev5  = scale_prev<tag5, SUBNET>;
+    template <typename SUBNET> using scale_prev6  = scale_prev<tag6, SUBNET>;
+    template <typename SUBNET> using scale_prev7  = scale_prev<tag7, SUBNET>;
+    template <typename SUBNET> using scale_prev8  = scale_prev<tag8, SUBNET>;
+    template <typename SUBNET> using scale_prev9  = scale_prev<tag9, SUBNET>;
+    template <typename SUBNET> using scale_prev10 = scale_prev<tag10, SUBNET>;
+
+    using scale_prev1_  = scale_prev_<tag1>;
+    using scale_prev2_  = scale_prev_<tag2>;
+    using scale_prev3_  = scale_prev_<tag3>;
+    using scale_prev4_  = scale_prev_<tag4>;
+    using scale_prev5_  = scale_prev_<tag5>;
+    using scale_prev6_  = scale_prev_<tag6>;
+    using scale_prev7_  = scale_prev_<tag7>;
+    using scale_prev8_  = scale_prev_<tag8>;
+    using scale_prev9_  = scale_prev_<tag9>;
+    using scale_prev10_ = scale_prev_<tag10>;
+
+// ----------------------------------------------------------------------------------------
+
     class relu_
     {
     public:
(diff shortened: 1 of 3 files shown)


pfeatherstone:
Can now serialise to and from memory using:
```
vector<char> buf;
dlib::serialize(buf) << ...;
dlib::deserialize(buf) >> ...;
```

pfeatherstone:
@davisking I've added the versioning as you requested and it models what you've done [here](https://github.com/davisking/dlib/blob/master/dlib/cuda/tensor.h#L657) exactly. (I'm still not 100% convinced it's useful but it's there). As it stands, there is a working implementation using macros only. I'm inclined to think that, since the user should never have to worry about how something is implemented and so long as the API is good, then it doesn't matter whether it is 100% macro or not. Or do you strongly object to this macro impl? Or do you strongly object to this whole idea?

pfeatherstone:
@davisking regarding what you describe here:
`Regarding using a variadic template, no magic is needed. Just make a method in the class that refers to the member variables. Then you don't have to prefix them by item.`
I'm still not sure what you mean if the impl were to look something like this:
```
namespace dlib
{
    namespace detail
    {
        template<typename T>
        void serialize_many(std::ostream& out, const T& x)
        {
            serialize(x, out);
        }
        
        template<typename T, typename... Rest>
        void serialize_many(std::ostream& out, const T& x, const Rest& ... rest)
        {
            serialize_many(out, x);
            serialize_many(out, rest...);
        }
        
        template<typename T>
        void deserialize_many(std::istream& in, T& x)
        {
            deserialize(x, in);
        }
        
        template<typename T, typename... Rest>
        void deserialize_many(std::istream& in, T& x, Rest& ... rest)
        {
            deserialize_many(in, x);
            deserialize_many(in, rest...);
        }
    }
}

#define DLIB_DEFINE_DEFAULT_SERIALIZATION(Type, ...)                    \
    inline friend void serialize(const Type& item, std::ostream& out)   \
    {                                                                   \
        using dlib::serialize;                                          \
        try                                                             \
        {                                                               \
            dlib::detail::serialize_many(out, item.__VA_ARGS__)         \
        }                                                               \
        catch (dlib::serialization_error& e)                            \
        {                                                               \
            throw dlib::serialization_error(e.info + "\n   while serializing object of type " #Type); \
        }                                                               \
    }                                                                   \
    inline friend void deserialize(Type& item, std::istream& in)        \
    {                                                                   \
        using dlib::deserialize;                                        \
        try                                                             \
        {                                                               \
            dlib::detail::deserialize_many(in, item.__VA_ARGS__)        \
        }                                                               \
        catch (dlib::serialization_error& e)                            \
        {                                                               \
            throw dlib::serialization_error(e.info + "\n   while deserializing object of type " #Type); \
        }                                                               \
    }
```

Do you mind spelling it out a bit more please (in code) ?

pfeatherstone:
@davisking As an alternative i have this, but it could potentially create a lot of member functions. Not sure i like it:

```
#define DLIB_DEFINE_DEFAULT_SERIALIZATION(Type, ...)                    \
    template<typename T>                                                \
    void serialize_me(std::ostream& out, const T& x) const              \
    {                                                                   \
        using dlib::serialize;                                          \
        serialize(x, out);                                              \
    }                                                                   \
                                                                        \
    template<typename T, typename... Rest>                              \
    void serialize_me(std::ostream& out, const T& x, const Rest& ... rest) const  \
    {                                                                   \
        serialize_me(out, x);                                           \
        serialize_me(out, rest...);                                     \
    }                                                                   \
                                                                        \
    template<typename T>                                                \
    void deserialize_me(std::istream& in, T& x)                         \
    {                                                                   \
        using dlib::deserialize;                                        \
        deserialize(x, in);                                             \
    }                                                                   \
                                                                        \
    template<typename T, typename... Rest>                              \
    void deserialize_me(std::istream& in, T& x, Rest& ... rest)         \
    {                                                                   \
        deserialize_me(in, x);                                          \
        deserialize_me(in, rest...);                                    \
    }                                                                   \
                                                                        \
    void serialize_all(std::ostream& out) const                         \
    {                                                                   \
        try                                                             \
        {                                                               \
            serialize_me(out, __VA_ARGS__);                             \
        }                                                               \
        catch (dlib::serialization_error& e)                            \
        {                                                               \
            throw dlib::serialization_error(e.info + "\n   while serializing object of type " #Type); \
        }                                                               \
    }                                                                   \
                                                                        \
    void deserialize_all(std::istream& in)                              \
    {                                                                   \
        try                                                             \
        {                                                               \
            deserialize_me(in, __VA_ARGS__);                            \
        }                                                               \
        catch (dlib::serialization_error& e)                            \
        {                                                               \
            throw dlib::serialization_error(e.info + "\n   while deserializing object of type " #Type); \
        }                                                               \
    }                                                                   \
    inline friend void serialize(const Type& item, std::ostream& out)   \
    {                                                                   \
        item.serialize_all(out);                                        \
    }                                                                   \
    inline friend void deserialize(Type& item, std::istream& in)        \
    {                                                                   \
        item.deserialize_all(in);                                       \
    }

```

pfeatherstone:
@davisking Please let me know what you think.

pfeatherstone:
I would rather use something like:
```
auto tied = std::tie(__VA_ARGS__);
```
and use a version of ```for_each``` that works on ```std::tuple```. I tried this and i have to create C++11 versions of ```index_sequence``` etc like so (again, borrowed from nlohmann::json):

```
namespace dlib
{
    template<std::size_t... Ints>
    struct index_sequence
    {
        using type = index_sequence;
        using value_type = std::size_t;
        static constexpr std::size_t size() noexcept
        {
            return sizeof...(Ints);
        }
    };

    template<class Sequence1, class Sequence2>
    struct merge_and_renumber;

    template<std::size_t... I1, std::size_t... I2>
    struct merge_and_renumber<index_sequence<I1...>, index_sequence<I2...>>
            : index_sequence < I1..., (sizeof...(I1) + I2)... > {};

    template<std::size_t N>
    struct make_index_sequence
        : merge_and_renumber < typename make_index_sequence < N / 2 >::type,
          typename make_index_sequence < N - N / 2 >::type > {};

    template<> struct make_index_sequence<0> : index_sequence<> {};
    template<> struct make_index_sequence<1> : index_sequence<0> {};

    template<typename... Ts>
    using index_sequence_for = make_index_sequence<sizeof...(Ts)>;
    
    template< class T >
    using remove_reference_t = typename std::remove_reference<T>::type;
}
```

Then i borrowed the aforementioned version of ```for_each``` from fluentc++ using:
```
template <class Tuple, class F, std::size_t... I>
    constexpr F for_each_impl(Tuple&& t, F&& f, dlib::index_sequence<I...>)
    {
        return (void)std::initializer_list<int>{(std::forward<F>(f)(std::get<I>(std::forward<Tuple>(t))),0)...}, f;
    }

    template <class Tuple, class F>
    constexpr F for_each(Tuple&& t, F&& f)
    {
        return for_each_impl(std::forward<Tuple>(t), std::forward<F>(f),
                             dlib::make_index_sequence<std::tuple_size<dlib::remove_reference_t<Tuple>>::value>{});
    }
```

But couldn't get it to work.

Alternatively, instead of using ```std::tuple```, could use ```dlib::tuple``` that already has serialisation support, but there is no dlib version of ```std::tie```, or a variadic constructor for ```dlib::tuple```.

pfeatherstone:
Being able to do:
```
auto tied = dlib::tied(__VA_ARGS__);
```
which returned a ```dlib::tuple``` object would be the best option for me because of built-in serialisation support in ```dlib::tuple```

pfeatherstone:
There is another version that uses std::tie and a version of for_each that works with std::tuple.

pfeatherstone:
Tests run fine on my machine so not quite sure what's failing on appveyor.

pfeatherstone:
Also, still need to fix ```vectorstream``` properly. I'm not an iostream person, so some help would be welcome.

dlib/serialize.h
```diff
@@ -1772,8 +1889,321 @@ namespace dlib
...
+#define DLIB_PASTE64(func, v1, v2, v3, v4, v5, v6, v7, v8, v9, v10, v11, v12, v13, v14, v15, v16, v17, v18, v19, v20, v21, v22, v23, v24, v25, v26, v27, v28, v29, v30, v31, v32, v33, v34, v35, v36, v37, v38, v39, v40, v41, v42, v43, v44, v45, v46, v47, v48, v49, v50, v51, v52, v53, v54, v55, v56, v57, v58, v59, v60, v61, v62, v63) DLIB_PASTE2(func, v1) DLIB_PASTE63(func, v2, v3, v4, v5, v6, v7, v8, v9, v10, v11, v12, v13, v14, v15, v16, v17, v18, v19, v20, v21, v22, v23, v24, v25, v26, v27, v28, v29, v30, v31, v32, v33, v34, v35, v36, v37, v38, v39, v40, v41, v42, v43, v44, v45, v46, v47, v48, v49, v50, v51, v52, v53, v54, v55, v56, v57, v58, v59, v60, v61, v62, v63)
+
+#define DLIB_TO(v1)  serialize(item.v1, out);
+#define DLIB_FROM(v1) deserialize(item.v1, in);
+
+#define DLIB_DEFINE_DEFAULT_SERIALIZATION(Type, version, ...)           \
```
davisking, line 2056: The user should not set the version.  It's pointless for them to do that since they can't make any use of it when using this macro.  It should be hard coded in the macro to the value 1.  Then if someone needs to deviate from what the macro does they will write their own serialize and deserialize and have access to the version field in existing serialized files, allowing them to distinguish between old and new formats.

dlib/serialize.h
```diff
@@ -1772,8 +1889,321 @@ namespace dlib
...
+        using dlib::deserialize;                                        \
+        deserialize(x, in);                                             \
+    }                                                                   \
+                                                                        \
+    template<typename T, typename... Rest>                              \
+    void deserialize_me(std::istream& in, T& x, Rest& ... rest)         \
+    {                                                                   \
+        deserialize_me(in, x);                                          \
+        deserialize_me(in, rest...);                                    \
+    }                                                                   \
```
davisking, line 2117: These methods don't need to be class members.  Just make them global (and rename them to `serialize_these` or some other more appropriate name.  Then this version of DLIB_DEFAULT_SERIALIZATION_METHOD will be even simpler than DLIB_DEFAULT_SERIALIZATION_METHOD 2.

davisking:
> There are now three versions of the macro: using 100% macro expansion, using variadic template member functions, and using std::tie + a templated for_each that works with std::tuple. The default is set to the latter. Feedback would be great. I don't like the second one as it introduces new member functions which users could use in their custom types accidentally. I like the third one as it's super clean. I'm indifferent to the first one. It's only drawback is that there is an upper bound on number of arguments.

The right thing is definitely to do method 1.  Method 0 is a pre-c++11 hack.  I've done plenty of stuff like that myself too, but only when C++11 wasn't available.  Method 2 is overly complex.   Method 1 is good, you just need to move the serialize_me methods out of the macro.  Those methods don't need to be in the class.  Such a change makes method 1 simpler than all the others and doesn't add extraneous stuff to the class.

Anyway, there is way too much stuff happening in this pull request.  It's full of lots of unrelated things that could be committed separately.  Please split it up into separate pull requests.

pfeatherstone:
@davisking Ok will split this into different PRs. I quite liked the ```std::tie``` version but yes you're right, keep it simple. If you don't mind i will still submit a different PR with all the stuff relating to ```for_each``` so we can add serialisation support for ```std::tuple```. I will submit a PR realting to type_trait cleaning up. So ```dlib::is_same_type``` becomes an alias for ```std::is_same``` and all that stuff. Or do you think that's pointless? It reduces dlib code while maintaining compatibility with existing code out there. I will also submit a PR with all the ```proxy_serialize``` stuff. Though i might need some assistance with ```dlib::vectorstream```.

davisking:
That all sounds like good stuff.  The only comment I have is there is a simpler way to define serialization for `std::tuple`.  See for instance https://stackoverflow.com/a/6894436/310086

PR #2178, opened 2020-09-13
Added (de)serialization via stringstream, istringstream, ostringstream and vector<char>

pfeatherstone:
Extended the proxy_serialize family of objects

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -136,6 +136,46 @@
         then serialize the exponent and mantissa values using dlib's integral serialization
         format.  Therefore, the output is first the exponent and then the mantissa.  Note that
         the mantissa is a signed integer (i.e. there is not a separate sign bit).
+
+
+    MAKING YOUR OWN CUSTOM OBJECTS SERIALIZABLE
+        Suppose you create your own type, my_custom_type, and you want it to be serializable.  I.e.
+        you want to be able to use the tools above to save and load it.  E.g. maybe you have a
+        std::vector<my_custom_type> you wish to save.
+
+        To make my_custom_type properly serializable all you have to do is define global serialize
+        and deserialize functions **IN THE SAME NAMESPACE AS MY_CUSTOM_TYPE**.  You must define them
+        in your namespace so that argument dependent lookup will be able to find them.  So your code
+        might look like this:
+
+            namespace your_namespace 
+            {
+                struct my_custom_type
+                {
+                    int a;
+                    float b;
+                    std::vector<float> c;
+                };
+                void serialize (const my_custom_type& item, std::ostream& out);
+                void deserialize (my_custom_type& item, std::istream& in);
+            }
+       
+        That's all you need to do.  You may optionally avail yourself of the
+        DLIB_DEFINE_DEFAULT_SERIALIZATION macro, which generates global friend serialize and
+        deserialize functions for you.  In that case you would do this instead:
+
+            namespace your_namespace 
+            {
+                struct my_custom_type
+                {
+                    int a;
+                    float b;
+                    std::vector<float> c;
+
+                    DLIB_DEFINE_DEFAULT_SERIALIZATION(my_custom_type, a, b, c);
+                };
+            }
+
 !*/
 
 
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  Please put the vectorstream updates into their own PR though and update the vectorstream unit tests to exercise these new code pathways.

As for the serialization stuff, this could all be done more simply by holding a pointer to an `ostream` or `istream`. You don't need to know the concrete type (e.g. sstringstream, vectorstream, or whatever).  So no additional classes or visual functions or any of that complication is needed.

pfeatherstone:
Ok will do. I've fixed the ```proxy_serialize``` and ```proxy_deserialize``` objects. I'm having to use ```new``` and ```delete``` with a boolean to control whether the pointer should be freed. A bit ugly but i think it's simpler than using a smart pointer hack. Now I need to go through the pain of undoing all of that, submit a PR for vectorstream + unit test, then recommit+push all the proxy stuff, then new PR. Sigh... But thanks for the great advice! Incremental steps is always good and keeping it simple is always better. As a warning, I was going to submit another PR with enhanced support for (de)serialization of stl containers (std::unodered_map, std::multimap, std::unordered_multimap, std::unordered_set, std::multiset, std::unordered_multiset, std::list, std::forward_list, std::tuple, std::variant (if c++17 is enabled), ... , and std::unique_ptr, std::shared_ptr. Will you want a separate PR for each container or can we do all of them in one go? Also, I did all of this yesterday (+ unit testing), while waiting for this PR to be merged, and ```serialize.h``` got very crowded and slowed down compilation of dlib because the entire library includes ```serialize.h``` and therefore almost every cpp file in dlib is including every single std container header file. I will ask for some advice in a bit when we get to that. But, so that i don't waste my time doing something you will object to, do you think it's a good idea to separate ```serialize.h``` into several header files: header for the basic serialize functions, separate header for each std container (eg. ```dlib/serialize/std/vector.h```, ```dlib/serialize/std/map.h```, ..., another header that includes all of the preceding headers (eg. ```dlib/serialize/std_types.h``` which includes ```dlib/serialize/std/vector.h```, ```dlib/serialize/std/map.h```, ...,). This means that if something in dlib needs serialization support, which it almost certainly does, it might only need to include the basic header, and not all the std containers. This should improve compilation speed. However if someone has lots of classes they want to serialize and they all need different flavours of std containers, they can just include ```dlib/serialize/std_types.h``` for convenience. (This idea is pinched from yas and cereal)

pfeatherstone:
Yep, dlib::vectorstream ain't ready. Running test1() in ```dlib/test/vectorstream.cpp``` by changing
```
std::vector<char> buf;
vectorstream s(buf);
```
to
```
std::vector<char> buf;
vectorstream s_ori(buf);
std::iostream& s(s_ori);
```
Comes up with a few errors.

davisking:
> Ok will do. I've fixed the `proxy_serialize` and `proxy_deserialize` objects. I'm having to use `new` and `delete` with a boolean to control whether the pointer should be freed. A bit ugly but i think it's simpler than using a smart pointer hack. 
Don't do that.  The pointer should be non-owning, as you are holding a pointer to some vectorstream or stringstream reference the user provided.  And then use a separate smart pointer to hold the fstream, if one is actually used.

As a warning, I was going to submit another PR with enhanced support for (de)serialization of stl containers (std::unodered_map, std::multimap, std::unordered_multimap, std::unordered_set, std::multiset, std::unordered_multiset, std::list, std::forward_list, std::tuple, std::variant (if c++17 is enabled), ... , and std::unique_ptr, std::shared_ptr. Will you want a separate PR for each container or can we do all of them in one go? 

They should be trivial so all in one is fine.  Put them in a separate header though.  I realize serialize.h is big, but I don't want to make it worse by pulling in the whole STL.

Also, I did all of this yesterday (+ unit testing), while waiting for this PR to be merged, and `serialize.h` got very crowded and slowed down compilation of dlib because the entire library includes `serialize.h` and therefore almost every cpp file in dlib is including every single std container header file. I will ask for some advice in a bit when we get to that. But, so that i don't waste my time doing something you will object to, do you think it's a good idea to separate `serialize.h` into several header files: 

Whatever you do you can't break existing dlib users.

pfeatherstone:
So i think there are some missing functions in ```vectorstream``` that need to override virtual functions in ```std::iostream```. Or there are some missing functions in ```vectorstream::vector_streambuf``` that need to override virtual functions in ```std::streambuf```.

pfeatherstone:
Candidates: ```sync()```, ```imbue()```, ```setbuf()```, ```showmanyc```.

pfeatherstone:
Or, ```std::iostream::seekg``` does nothing when the bad bit is set which causes the error. We don't observe this when calling ```dlib::vectorstream::seekg```.

pfeatherstone:
yep that's it: http://www.cplusplus.com/reference/istream/istream/seekg/

pfeatherstone:
So the fixes are : removing ```DLIB_ASSERT``` statements from ```vectorstream.h```, and fixing a bug in the unit test. You need to call ```clear()``` before ```seekg()```. Otherwise, if the bad bit is set ```seekg()``` does nothing.

PR #2179, opened 2020-09-15
vectorstream updates: added seekoff and seekpos

pfeatherstone:
These updates ensure that casting ```dlib::vectorstream&``` to ```std::iostream&``` still works fine.

### DIFF
--- a/dlib/test/vectorstream.cpp
+++ b/dlib/test/vectorstream.cpp
@@ -21,6 +21,116 @@ namespace
 
 
     logger dlog("test.vectorstream");
+    
+    namespace helper 
+    {        
+        /*
+        * This class helps test the functionality of dlib::vectorstream when an object reference is treated as:
+        * - dlib::vectorstream&
+        * - std::iostream&
+        * Depending on which is used, the underlying function calls might have different call stacks.
+        */
+        struct test1_helper
+        {
+            test1_helper(
+                std::vector<char>& buf_
+            ) : buf(buf_)
+            {}
+
+            template <typename T>
+            void operator()(T& s)
+            {
+                for (int i = -1000; i <= 1000; ++i)
+                {
+                    char ch = i;
+                    s.put(ch);
+                }
+
+                DLIB_TEST(buf.size() == 2001);
+
+                int cnt = -1000;
+                for (unsigned long i = 0; i < buf.size(); ++i)
+                {
+                    char ch = cnt;
+                    DLIB_TEST(buf[i] == ch);
+                    ++cnt;
+                }
+
+                for (int i = -1000; i <= 1000; ++i)
+                {
+                    DLIB_TEST(s.peek() != EOF);
+                    char ch1 = i;
+                    char ch2 = s.get();
+                    DLIB_TEST(ch1 == ch2);
+                }
+
+                DLIB_TEST(s.peek() == EOF);
+                DLIB_TEST(s.get() == EOF);
+
+                s.clear();
+                s.seekg(6);
+
+                for (int i = -1000+6; i <= 1000; ++i)
+                {
+                    DLIB_TEST(s.peek() != EOF);
+                    char ch1 = i;
+                    char ch2 = s.get();
+                    DLIB_TEST(ch1 == ch2);
+                }
+
+                DLIB_TEST(s.peek() == EOF);
+                DLIB_TEST(s.get() == EOF);
+
+                std::string temp;
+                temp = "one two three!";
+
+                s.clear();
+                s.seekg(0);
+                buf.clear();
+
+                serialize(temp, s);
+                std::string temp2;
+                deserialize(temp2, s);
+                DLIB_TEST(temp2 == temp);
+
+                s.put('1');
+                s.put('2');
+                s.put('3');
+                s.put('4');
+                DLIB_TEST(s.get() == '1');
+                DLIB_TEST(s.get() == '2');
+                DLIB_TEST(s.get() == '3');
+                DLIB_TEST(s.get() == '4');
+
+                s.putback('4');
+                DLIB_TEST(s.get() == '4');
+                s.putback('4');
+                s.putback('3');
+                s.putback('2');
+                s.putback('1');
+                DLIB_TEST(s.get() == '1');
+                DLIB_TEST(s.get() == '2');
+                DLIB_TEST(s.get() == '3');
+                DLIB_TEST(s.get() == '4');
+                DLIB_TEST(s.good() == true);
+                DLIB_TEST(s.get() == EOF);
+                DLIB_TEST(s.good() == false);
+
+                // make sure seeking to a crazy offset doesn't mess things up
+                s.clear();
+                s.seekg(1000000);
+                DLIB_TEST(s.get() == EOF);
+                DLIB_TEST(s.good() == false);
+                s.clear();
+                s.seekg(1000000);
+                char sbuf[100];
+                s.read(sbuf, sizeof(sbuf));
+                DLIB_TEST(s.good() == false);
+            }
+
+            std::vector<char>& buf;
+        };
+    }
 
 // ----------------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


dlib/test/vectorstream.cpp
```diff
@@ -21,102 +21,137 @@ namespace
...
+                DLIB_TEST(s.good() == false);
+            }
+
+            std::vector<char>& buf;
+        };
+    }
 
 // ----------------------------------------------------------------------------------------
 
     void test1()
```
davisking, line 137: Please don't move the code in `test1()` into a class.   It makes the code harder to follow than needed.  Instead, just do
```
template <typename stream>
void test1(std::vector<char>& buf, stream& s) 
{
    <then all the code that's there is still there just like it is, with whatever extra you need to add to test the seeking>
}
```

Also add tests that call `s.seekg()` with explicit `beg`, `end`, and `cur` modes to test each path.

dlib/vectorstream/vectorstream.h
```diff
@@ -43,6 +42,32 @@ namespace dlib
...
+            
+            pos_type seekpos(pos_type pos, std::ios_base::openmode mode = std::ios_base::in | std::ios_base::out)
+            {
+                return seekoff(pos - pos_type(off_type(0)), std::ios_base::beg, mode);
+            }
+
+            pos_type seekoff(off_type off, std::ios_base::seekdir dir,
+                             std::ios_base::openmode mode = std::ios_base::in | std::ios_base::out )
+            {
+                (void)mode; //don't care what mode is. This class handles both
```
davisking, line 54: `mode` can't be ignored.  `mode` might be indicating that only the write pointer is moved, which is just not something the vectorstream supports.  You at least need a `DLIB_ASSERT` on this to claim that `mode` is `in`.

dlib/vectorstream/vectorstream.h
```diff
@@ -120,14 +145,15 @@ namespace dlib
...
-        }
+            
+        vectorstream(const vectorstream& ori) = delete;
+            
+        vectorstream(
+            vectorstream&& item
+        ) : 
+            std::iostream(std::move(item)), 
+            buf(std::move(item.buf))
+        {}
```
davisking, line 156: Why add this?  This seems to be asking to break user code since this class doesn't own the vector buffer.  This change would allow someone to make a vectorstream in a function and then return it.  In such a case it is very likely they created the vectorstream referencing a local buffer that is no defunct.

davisking (changes requested):
Thanks for another PR, just a few things to fix up and then it looks like it will be good to go.

dlib/vectorstream/vectorstream.h
pfeatherstone: I initially needed the move constructor of vectorstream for proxy_serialize to work (the way i initially implemented it). Now, using a unique_ptr, it's not needed. Yep you're right. So both the copy constructor and the move constructor should be marked deleted. Or just leave it as it was, i.e. default implementation

pfeatherstone: Given your argument, it sounds like explicitly deleting both copy and move constructors is the safest thing to do.

dlib/vectorstream/vectorstream.h
pfeatherstone: Yep agreed

dlib/test/vectorstream.cpp
pfeatherstone: Yep agreed. Don't know what i was thinking.

pfeatherstone:
@davisking Hopefully the changes are what you are looking for. With regards to vectorstream not owning the buffer, a future PR could possibly include vectorstream owning the buffer, and add a function which allows you to swap vectorstream's own buffer with yours. So add something like ```void swap_vector(std::vector<char>& buf)``` to vectorstream instead of passing a buffer by reference. Possibly also add a function ```const std::vector<char>& get_vector() const``` if a swap isn't required. This would have to be done in a different class to avoid breaking user code.

dlib/test/vectorstream.cpp
```diff
@@ -59,7 +60,7 @@ namespace
         DLIB_TEST(s.get() == EOF);
 
         s.clear();
-        s.seekg(6);
+        s.seekg(6, std::ios_base::beg);
```
davisking, line 63: I think `seekg(6)` does (or at least can) execute a different codepath through the iostreams than `seekg(6, beg)`.  So this line should remain `seekg(6)`.  Then later on you should call the 3 other variants of `seekg()` and verify that the right thing happened.

dlib/test/vectorstream.cpp
```diff
@@ -119,6 +149,26 @@ namespace
...
+        {
+            std::vector<char> buf;
+            vectorstream s1(buf);
+            test1_variant(buf, s1);
+        }
+        
+        {
+            vector<char> buf;
+            dlib::vectorstream s1(buf);
+            std::iostream& s2(s1);
```
davisking, line 167: It's more idiomatic to say `std::iostream& s2 = s1;`

dlib/test/vectorstream.cpp
```diff
@@ -71,13 +72,42 @@ namespace
 
         DLIB_TEST(s.peek() == EOF);
         DLIB_TEST(s.get() == EOF);
+        
+        s.clear();
+        s.seekg(1000, std::ios_base::beg); //read_pos should be 1000
```
davisking, line 77: Add a test right after this line to check that the stream isn't in a funny state.  E.g. you should be able to read some values and see they are what is expected.

dlib/vectorstream/vectorstream.h
```diff
@@ -43,6 +43,32 @@ namespace dlib
             {
                 read_pos = pos;
             }
+            
+            pos_type seekpos(pos_type pos, std::ios_base::openmode mode = std::ios_base::in | std::ios_base::out)
+            {
+                return seekoff(pos - pos_type(off_type(0)), std::ios_base::beg, mode);
```
davisking, line 49: Are you getting a warning when you don't do `- pos_type(off_type(0))`?  I get what it's doing with regard to type conversion, but it really shouldn't be doing anything at runtime and seems unneeded.  But if it suppresses compiler warnings then this is fine.

dlib/vectorstream/vectorstream.h
```diff
@@ -120,15 +146,12 @@ namespace dlib
...
-        {
-            buf.seekg(pos);
-            return *this;
-        }
-
+            
+        vectorstream(const vectorstream& ori) = delete;
+        vectorstream(vectorstream&& item) = delete;
+        vectorstream& operator=(const vectorstream& ori) = delete;
+        vectorstream& operator=(vectorstream&& item) = delete;
```
davisking, line 153: You don't need to delete `&&` versions.  They are deleted by default when the normal ones are deleted.

davisking (changes requested):

dlib/test/vectorstream.cpp
```diff
@@ -21,16 +21,17 @@ namespace
...
-// ----------------------------------------------------------------------------------------
-
-    void test1()
+          
+    /*
+    * This class helps test the functionality of dlib::vectorstream when an object reference is treated as:
+    * - dlib::vectorstream&
+    * - std::iostream&
+    * Depending on which is used, the underlying function calls might have different call stacks.
+    */
```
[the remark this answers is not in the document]
pfeatherstone: Sorry that's left over from before. Should i just remove the comments? The user doesn't really need any explanation here.

dlib/vectorstream/vectorstream.h
pfeatherstone: To be perfectly honest, a lot of stream objects implement ```seekpos``` like this, verbatim. So i copied it.

dlib/vectorstream/vectorstream.h
pfeatherstone: So shall we just make ```vectorstream``` inherit from```dlib::noncopyable``` ?

dlib/test/vectorstream.cpp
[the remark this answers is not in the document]
davisking: Yeah removing it is fine.

dlib/vectorstream/vectorstream.h
davisking: Leave the two deleted ones. They will give slightly better compile time errors than what users receive if noncopyable is used.

davisking:

PR #2181, opened 2020-09-17
Extended proxy_(de)serialize objects to work with stringstream, ostringstream, istringstream and vector<char>

pfeatherstone:


dlib/serialize.h
```diff
@@ -1655,34 +1707,31 @@ namespace dlib
                 if (objects_read == 0)
                 {
                     throw serialization_error("An error occurred while trying to read the first" 
-                        " object from the file " + filename + ".\nERROR: " + e.info + "\n" + suffix);
+                        " object from the stream.\nERROR: " + e.info + "\n" + suffix);
```
davisking, line 1710: Definitely don't remove the filename.  That's a hugely important part of the error message.

dlib/serialize.h
```diff
@@ -1583,47 +1584,86 @@ namespace dlib
...
-            // file.
-            fin->read(file_header,4);
-            fin->clear();
-            fin->seekg(0);
+            init();
         }
-
+        
+        explicit proxy_deserialize (
+            std::istringstream& ss
```
davisking, line 1644: Why not just make this `std::istream`?

dlib/serialize.h
```diff
@@ -1583,47 +1584,86 @@ namespace dlib
...
+        ) : fin_optional_owning_ptr(nullptr),
+            fin(ss)
+        {
+            init();
+        }
+        
+        explicit proxy_deserialize (
+            std::stringstream& ss
```
davisking, line 1652: Same here.  This isn't needed if the constructor above just takes `std::istream`

davisking:
Looks good, aside from the few things I noted.

dlib/serialize.h
pfeatherstone: Yes good point.

dlib/serialize.h
pfeatherstone: Yep

davisking (changes requested):

davisking:

PR #2182, opened 2020-09-18
Use DLIB_CASSERT to avoid unused variable warning in release compilation

arrufat:
After updating to latest dlib I got a warning about unused variable, so I changed the `DLIB_ASSERT` to `DLIB_CASSERT`. I am guessing we always want to check the `mode` we're in.


davisking:
Yeah, good call.

PR #2185, opened 2020-09-20
DLIB (de)serialization : enhanced STL container support

pfeatherstone:


pfeatherstone:
@davisking Is this PR ok?

dlib/serialize.h
```diff
@@ -933,6 +1092,88 @@ namespace dlib
...
+        deserialize_tuple_helper(std::istream& in_) : in(in_) {}
+        
+        template<typename T>
+        void operator()(T& item)
+        {
+            deserialize(item, in);
+        }
+                
+        std::istream& in;
+    };
```
davisking, line 1145: Use lambdas instead of writing out function classes.

dlib/serialize.h
```diff
@@ -983,6 +1224,144 @@ namespace dlib
...
 
+    template <typename domain, typename range, typename hash, typename keyEqual, typename alloc>
+    void serialize (
+        const std::unordered_map<domain, range, hash, keyEqual, alloc>& item,
+        std::ostream& out
+    )
+    {
+        try
+        { 
+            const unsigned long size = static_cast<unsigned long>(item.size());
```
davisking, line 1235: Don't need this cast, just `serialize(item.size(), out)` is fine.

dlib/serialize.h
```diff
@@ -983,6 +1224,144 @@ namespace dlib
...
+    template <typename domain, typename range, typename hash, typename keyEqual, typename alloc>
+    void deserialize (
+        std::unordered_map<domain, range, hash, keyEqual, alloc>& item,
+        std::istream& in
+    )
+    {
+        try 
+        { 
+            item.clear();
+            unsigned long size;
```
davisking, line 1256: And then here `size_t size;`

dlib/serialize.h
```diff
@@ -1592,6 +2168,93 @@ namespace dlib
...
+
+    template <typename T>
+    void serialize (
+        const std::shared_ptr<T>& item,
+        std::ostream& out
+    )
+    {
+        try
+        {
+            bool is_non_empty = item.operator bool();
```
davisking, line 2225: It's really weird to call `.operator bool()`.  Compare against `nullptr` or just assign to `bool` or explicitly cast to `bool` if you want.

dlib/serialize.h
```diff
@@ -1592,6 +2168,93 @@ namespace dlib
...
+        std::shared_ptr<T>& item,
+        std::istream& in
+    )
+    {
+        try
+        {
+            //when deserializing shared_ptr, this is fresh state, so reset the pointers, even if item is non-empty
+            bool is_non_empty;
+            deserialize(is_non_empty, in);
+            item.reset(is_non_empty ? new T() : nullptr);
```
davisking, line 2247: Use `make_shared`, it is more efficient than calling `new` and `reset()`.

dlib/test/serialize.cpp
```diff
@@ -401,18 +401,52 @@ namespace
         dlib::deserialize(item.b_false,in);
     }
     
+    template<typename T>
+    bool pointers_equal(const unique_ptr<T>& l, const unique_ptr<T>& r)
+    {
+        bool ptrs_equal = l.operator bool() == r.operator bool();
+        if (ptrs_equal && l)
+            ptrs_equal &= (*l == *r);
```
davisking, line 409: This is doing a bitwise operation on a bool, which is odd and I don't know if the standard actually guarantees this is portable.

And don't call `.operator bool()` directly.  Write something like
```
if (l && r) 
   return *l == *r;
else
   return l == r;
```

dlib/test/serialize.cpp
```diff
@@ -401,18 +401,52 @@ namespace
...
+        return ptrs_equal;
+    }
+
+    template<typename T>
+    bool pointers_equal(const shared_ptr<T>& l, const shared_ptr<T>& r)
+    {
+        bool ptrs_equal = l.operator bool() == r.operator bool();
+        if (ptrs_equal && l)
+            ptrs_equal &= (*l == *r);
+        return ptrs_equal;
```
davisking, line 419: same thing here

davisking (changes requested):
This looks generally ok except for the things I noted.  I do worry about compile times though.  How does the total compile time of all the example programs change, for instance, with this change?  It seems like it would be better to put these additional routines in a dlib/serialize_stl.h header instead.

dlib/serialize.h
pfeatherstone: Can I use lambdas here in C++11? C++11 doesn't support generic lambdas. I must be missing something here. Can you be more specific?

dlib/serialize.h
pfeatherstone: Okidok. I was simply following what other areas of the code were doing in ```serialize.h```. I assumed there was maybe some subtle reason for doing this that you'd found.

dlib/serialize.h
pfeatherstone: Okidok. Again, I was following what you'd done in other areas like ```std::map```

dlib/serialize.h
pfeatherstone: Yep ok. I did this just so i was 100% certain it was calling the ```bool()``` operator overload. I didn't want any surprises.

dlib/test/serialize.cpp
pfeatherstone: ```l  == r``` will check if the underlying pointers are equal according to cppreference. We don't want that surely.

pfeatherstone: I shouldn't have used the variable name ```ptrs_equal```. I should have used ```ptr_values_equal```

dlib/test/serialize.cpp
pfeatherstone: Again, ```l == r```, is the same as ```l.get() == r.get()``` i think. We don't care if the raw underlying pointers are equal.

pfeatherstone:
Regarding putting everything in new headers, I'd thought of that but dumping everything in ```serialize.h``` was the simplest solution that didn't break internal dlib code or user code. Also, it is highly likely that someone wanting to use dlib for serializing their own types will want a one-stop header for all things serialization. So it seemed that seperating ```serialize.h``` into several headers was a bit of a waste of time. But i agree that compile times might bloat. But when you look at all the headers already included in a .cpp file when including ```serialize.h```, not just the ones explicitly listed in ```serialize.h```, but all the ones you get if you follow the include paths, it's already quite a lot. Correct me if I'm wrong, but i don't think dlib took an aggressive stance on minimising ```#include```s and using forward declaration to volume 11. For example, using dlib's dnn module, compiling a network like resnet150 takes a huge amount of time because of the template API. So i thought adding ```<unordered_map>```, ```<unordered_set>```, ```<forward_list>``` and ```<tuple>``` wasn't a big deal for all these reasons.  BUT, i know compilation time can be an issue in large projects and in general it is a ballache. But if it works...

pfeatherstone:
Also i didn't notice a noticeable increase in compile time for the example programs.

pfeatherstone:
If we are going down the route of splitting headers, i would suggest splitting properly. So a base header ```serialize_common.h``` that includes serialization of primitive types, individual headers for all containers, including std::vector, std::deque, std::map, ... (all those currently in ```serialize.h```). Then a convenience header like ```serialize_stl.h``` that includes all the individual container headers, then a one-stop header ```serialize.h``` that includes everything. So this wouldn't break user code. Then i would change internal areas of dlib to only include the serialization headers they need. I'm willing to bet, most of them will only need ```serialize_common.h```. Then the overall effect would reduce the compilation time of dlib and everyone is happy. But that is substantially more work and a PR of this nature would take weeks to go through given that we are in different time zones and we are only managing 1 message per day.

dlib/serialize.h
pfeatherstone: Also i found this in ```dlib/dlib/tuple/tuple.h```:

```
funct(a.template get<i>());
```

So i thought maybe Davis doesn't mind weird notation too much.

dlib/serialize.h
pfeatherstone: yep ok fair. I did think of doing that, but i didn't for the sake of symmetry with ```unique_ptr```'s deserialize function.

dlib/serialize.h
davisking: Oh right, yeah. Forgot that’s c++14 :(

dlib/serialize.h
davisking: Long long ago I used to run into a compiler that would fail to compile when it saw size_t. That’s why it’s not as used as it should be.

dlib/test/serialize.cpp
davisking: If the pointer values are equal then the values are equal too.

davisking:
I’m just saying to time the build time of all the examples with and without this PR. If it’s not really changed then this is fine and preferable to have it in one header. 

But whatever we do we must not break existing user builds.

dlib/serialize.h
pfeatherstone: Let me guess, windows

dlib/test/serialize.cpp
pfeatherstone: Oh i see. Thank you

pfeatherstone:
Okidok i will benchmark the build times of the examples on my machine and post here. Then we can decide what to do. If build times are slower, i still think we should do it properly and do a full decomposition of ```serialize.h``` and include only the relevant parts across dlib. But that might involve scores of small changes across dlib. I don't like the prospect of that. So i'm hoping build times aren't affected that much.

dlib/serialize.h
pfeatherstone: Do you want this change applied to all containers? It's not just unordered_map that casts to unsigned long...

pfeatherstone: I'm going to assume yes

pfeatherstone:
Right on my machine, where the output of ```lsb_release -a``` is:
```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.4 LTS
Release:	18.04
Codename:	bionic
```
output of ```lscpu``` is:
```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              12
On-line CPU(s) list: 0-11
Thread(s) per core:  2
Core(s) per socket:  6
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               158
Model name:          Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Stepping:            10
CPU MHz:             800.009
CPU max MHz:         4500.0000
CPU min MHz:         800.0000
BogoMIPS:            5199.98
Virtualisation:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            12288K
NUMA node0 CPU(s):   0-11
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid mpx rdseed adx smap clflushopt intel_pt xsaveopt xsavec xgetbv1 xsaves dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp md_clear flush_l1d
```

building the examples on master takes 9m48s and building the examples on my fork takes 9m52s. The 4s difference is in the noise I think. So it would seem that adding the features in ```serialize.h``` is ok at the moment.

pfeatherstone:
Sorry for the delay, i've been busy with documentation this week.

dlib/serialize.h
davisking: Yeah, that's a good change.  I forgot those casts were even there.  I honestly don't remember what compiler it was that had problems with size_t.  It was probably visual studio though.  You can see lots of comments in dlib in front of bizarre code to the effect of "we do this silly thing to avoid a bug in visual studio".

dlib/serialize.h
davisking: Well, `template` has to be there to compile.

davisking:
Yeah, I timed this myself and I think the build time is fine.  So looks good :+1:

PR #2187, opened 2020-09-21
Add label_to_ignore to loss_multiclass_log_per_pixel_weighted

arrufat:


arrufat:
On a second thought, this might not be necessary as one can set the weight to 0 for the labels we want to ignore...

PR #2194, opened 2020-09-27
Add loss multiclass log per pixel weighted cuda

arrufat:
I am not entirely happy about the location of the `weighted_label` definition... But it was a header file that was already included everywhere I needed the definition...

I've moved the cpu implementation of the loss from the `dnn/loss.h` to `cuda/cpu_dlib.h` so that I can run a test between both implementations.

Do you want to make another PR to do the same thing for all the losses that have CUDA implementations (move them from `dnn/loss.h` to `cuda/cpu_dlib.h`, so that we can actually test them?

### DIFF
--- a/dlib/cuda/tensor_abstract.h
+++ b/dlib/cuda/tensor_abstract.h
@@ -720,6 +720,36 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <typename label_type>
+    struct weighted_label
+    {
+        /*!
+            WHAT THIS OBJECT REPRESENTS
+                This object represents the truth label of a single sample, together with
+                an associated weight (the higher the weight, the more emphasis the
+                corresponding sample is given during the training).
+                This object is used in the following loss layers:
+                    - loss_multiclass_log_weighted_ with unsigned long as label_type
+                    - loss_multiclass_log_per_pixel_weighted_ with uint16_t as label_type,
+                      since, in semantic segmentation, 65536 classes ought to be enough for
+                      anybody.
+        !*/
+        weighted_label()
+        {}
+
+        weighted_label(label_type label, float weight = 1.f)
+            : label(label), weight(weight)
+        {}
+
+        // The ground truth label
+        label_type label{};
+
+        // The weight of the corresponding sample
+        float weight = 1.f;
+    };
+
+// ----------------------------------------------------------------------------------------
+
 }
 
 #endif // DLIB_DNn_TENSOR_ABSTRACT_H_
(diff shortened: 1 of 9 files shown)


arrufat:
By way, I timed both losses in the test file (I changed the input size from 5x7 to 50x70):
- cpu: 19.81 ms
- gpu: 1.23 ms (x16 speedup)

Also, during a training, I noticed around 50-60% speedup for the `train_one_step()` calls.

dlib/cuda/tensor_abstract.h
```diff
@@ -718,6 +718,36 @@ namespace dlib
         provides serialization support for alias_tensor.  
     !*/
 
+// ----------------------------------------------------------------------------------------
+
+    template <typename label_type>
+    struct weighted_label
```
davisking, line 724: Yeah, this is a weird place for it.  Put it in dnn/utilities.h instead.  It probably makes the most sense to leave the documentation for it in loss.h though since that's right next to the docs that depend on it, so it will be easily located by readers.  We are only talking about moving this because of include order stuff after all, not because it logically makes sense to take it out of loss.h

davisking:
> By way, I timed both losses in the test file (I changed the input size from 5x7 to 50x70):
> 
> * cpu: 19.81 ms
> * gpu: 1.23 ms (x16 speedup)
> 
> Also, during a training, I noticed around 50-60% speedup for the `train_one_step()` calls.

Awesome

davisking:
> Do you want to make another PR to do the same thing for all the losses that have CUDA implementations (move them from `dnn/loss.h` to `cuda/cpu_dlib.h`, so that we can actually test them?

I'm sort of indifferent about it.  You certainly can if you like though.

dlib/cuda/tensor_abstract.h
arrufat: Agreed. Maybe I'm missing something, but if I just put it in `dnn/utilities.h` and `#include "../dnn/utilites.h" in the `cuda/cuda_dlib.h` I get a compilation error:
```
╭───────────────────────────────────────────────────────────────────────────────────────╮
│ • In included file: main file cannot be included recursively when building a preamble │
│                                                                                       │
│ /home/adria/Projects/dlib/dlib/cuda/../dnn/../cuda/tensor_tools.h:12:10:              │
│ note: error occurred here                                                             │
╰───────────────────────────────────────────────────────────────────────────────────────╯
```
That is why I put it outside of the `dnn` directory... maybe add a new file in the `cuda` directory? It's really weird...

arrufat:
> > Do you want to make another PR to do the same thing for all the losses that have CUDA implementations (move them from `dnn/loss.h` to `cuda/cpu_dlib.h`, so that we can actually test them?
> 
> I'm sort of indifferent about it. You certainly can if you like though.

Ok, I will do that in a bit. I think it's good to make sure CPU an CUDA versions don't diverge...

davisking:

PR #2199, opened 2020-09-30
Test cuda losses

arrufat:
This PR adds tests to ensure the untested cuda implementations for:
- `loss_binary_log_per_pixel`
- `loss_multiclass_log_per_pixel`

I've also added a couple of things to `misc.h` that I needed to be accessible from the cuda directory.

Also, I've added a new `.index` method to tensor. I know the memory layout is not specified, but I saw many losses relying on the convention and defining their own `tensor_index` method, so I think it will be more convenient.

### DIFF
--- a/dlib/cuda/tensor.h
+++ b/dlib/cuda/tensor.h
@@ -50,15 +50,6 @@ namespace dlib
         long long nr() const { return m_nr; }
         long long nc() const { return m_nc; }
         size_t size() const { return m_size; }
-        size_t index(
-            const long sample,
-            const long k,
-            const long r,
-            const long c
-        ) const
-        {
-            return ((sample * m_k + k) * m_nr + r) * m_nc + c;
-        }
 
         typedef float* iterator;
         typedef const float* const_iterator;
--- a/dlib/cuda/tensor.h
+++ b/dlib/cuda/tensor.h
@@ -50,6 +50,15 @@ namespace dlib
         long long nr() const { return m_nr; }
         long long nc() const { return m_nc; }
         size_t size() const { return m_size; }
+        size_t index(
+            const long sample,
+            const long k,
+            const long r,
+            const long c
+        ) const
+        {
+            return ((sample * m_k + k) * m_nr + r) * m_nc + c;
+        }
 
         typedef float* iterator;
         typedef const float* const_iterator;
--- a/dlib/test/dnn.cpp
+++ b/dlib/test/dnn.cpp
@@ -2683,9 +2683,14 @@ namespace
         cpu::compute_loss_mean_squared_per_channel_and_pixel cpu_compute;
         double cuda_loss, cpu_loss;
         const tensor& output_tensor = net.subnet().get_output();
-        tensor& grad = net.subnet().get_gradient_input();
-        cuda_compute(labels.begin(), output_tensor, grad, cuda_loss);
-        cpu_compute(labels.begin(), output_tensor, grad, cpu_loss);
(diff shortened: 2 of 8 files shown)


dlib/cuda/tensor.h
```diff
@@ -50,6 +50,15 @@ namespace dlib
...
         size_t size() const { return m_size; }
+        size_t index(
+            const long sample,
+            const long k,
+            const long r,
+            const long c
+        ) const
+        {
+            return ((sample * m_k + k) * m_nr + r) * m_nc + c;
+        }
```
davisking, line 61: This really doesn't belong here.  If we wanted to do this it would be better to have tensor work like ```
cpu_tensor_view view = my_tensor.cpu_view();
view(sample,k,r,c) = 123;
```
The point of `cpu_tensor_view` is to incur the cost of calling `host()` once and to provide nicer indexed access.  Could also have a `gpu_tensor_view` that's the same thing basically but calls `device()` when created.  Then that could be passed to cuda kernels instead of pointers and ints.

File: dlib/test/dnn.cpp [hunk 2]
```diff
@@ -2883,6 +2883,18 @@ namespace
...
+#if DLIB_USE_CUDA
+        cuda::compute_loss_binary_log_per_pixel cuda_compute;
+        cpu::compute_loss_binary_log_per_pixel cpu_compute;
+        double cuda_loss, cpu_loss;
+        const tensor& output_tensor = net.subnet().get_output();
+        tensor& grad = net.subnet().get_gradient_input();
+        cuda_compute(y.begin(), output_tensor, grad, cuda_loss);
+        cpu_compute(y.begin(), output_tensor, grad, cpu_loss);
+        const auto err = abs(cuda_loss - cpu_loss) / cpu_loss;
+        DLIB_TEST_MSG(err < 1e-6, "binary log per pixel cuda and cpu losses differ");
```
davisking, line 2896: Don't forget to check that the `grad` outputs are the same as well.

davisking:
Thanks, looks good aside from the two things I mentioned.

dlib/cuda/tensor.h
arrufat: I see, maybe for another PR, then... for now I will just make a static function `tensor_index` in the `misc.h` file and use it across loss functions, instead of each one having their own. Would that be OK?

davisking: Yeah, that's fine.

dlib/test/dnn.cpp
```diff
@@ -2683,9 +2683,14 @@ namespace
         cpu::compute_loss_mean_squared_per_channel_and_pixel cpu_compute;
         double cuda_loss, cpu_loss;
         const tensor& output_tensor = net.subnet().get_output();
-        tensor& grad = net.subnet().get_gradient_input();
-        cuda_compute(labels.begin(), output_tensor, grad, cuda_loss);
-        cpu_compute(labels.begin(), output_tensor, grad, cpu_loss);
+        resizable_tensor cuda_grad(output_tensor), cpu_grad(output_tensor);
+        cuda_compute(labels.begin(), output_tensor, cuda_grad, cuda_loss);
+        cpu_compute(labels.begin(), output_tensor, cpu_grad, cpu_loss);
+        DLIB_CASSERT(cuda_grad.size() == cpu_grad.size());
```
davisking, line 2689: Tests should use `DLIB_TEST` instead of `DLIB_CASSERT`.

arrufat: Oops, done!

davisking:

arrufat:
Thank you! :)

PR #2204, opened 2020-10-07
Add GELU activation layer

arrufat:
Hi, in this PR I implemented the [Gaussian Error Linear Unit (GELU)](https://arxiv.org/abs/1606.08415) activation layer.
This layer is often used in [Transformers](https://arxiv.org/abs/1706.03762).

The gelu function is defined as:
image

I choose the sigmoid approximation for this implementation. For the [derivative](https://www.wolframalpha.com/input/?i=derivative+of+x+*+sigmoid%28x+*+1.702%29), I used WolframAlpha... 😅

image

I would like dlib to support Transformer architectures some day, since they are proving to be really good at NLP tasks (they've made RNNs obsolete).

I will certainly give it a go in my spare time, but I can't promise anything or give a deadline. This is my first baby step (I know, I started building the house from the roof).

I think most of the building blocks are already there. I just need to figure out how to do the positional encoding in a nice way (maybe a custom `input` layer that does that automatically? Anyway, I will try unless somebody has a clear idea on how it can be implemented nicely.

### DIFF
--- a/dlib/cuda/cpu_dlib.cpp
+++ b/dlib/cuda/cpu_dlib.cpp
@@ -1711,9 +1711,9 @@ namespace dlib
             const tensor& gradient_input
         )
         {
-            const auto compute_gradient = [](float x)
+            const float beta = 1.0f / std::sqrt(pi) / sqrt_2;
+            const auto compute_gradient = [beta](float x)
             {
-                const float beta = M_2_SQRTPI * M_SQRT1_2 * 0.5f;
                 const float cdf = 0.5f*(1.0f + std::erf(x/sqrt_2));
                 const float pdf = beta*std::exp(-0.5f*x*x);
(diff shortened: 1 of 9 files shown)


arrufat:
I've noticed this implementation using the simoid is slower and less precise than the one using the tanh approximation.
Also, I've cheated a little bit by looking at how PyTorch implemented this. Here's my testing for a 1x3x224x224 tensor on the gpu:

- dlib: gelu with sigmoid approx: 30 µs
- dlib: gelu with tanh approx: 26 µs
- dlib: gelu with normcdf: 8 µs (based on PyTorch implementation)
- pytorch gelu with normcdf: 12 µs (maybe slower because of compilation flags and python overhead)

davisking:
Sweet, thanks for another cool PR :)

PR #2213, opened 2020-10-12
Add Layer Normalization

arrufat:
This is a second step towards implementing transformers in dlib.

This PR adds [Layer Normalization](https://arxiv.org/abs/1607.06450) support to dlib.

I have added the CPU implementation and the forward GPU, but I have some doubts, so I would really appreciate if somebody could have a look to the CUDA implementation.

It runs slower than the CPU version and for big inputs (2, 3, 224, 224) results are not good, but for small inputs, it works.

Here's how I tried it, in case anyone wants to try

``` c++
#include <dlib/dnn.h>

using namespace dlib;

int main() try
{

    auto print_tensor_and_avg = [](const tensor& t)
    {
        float sum = 0;
        const float* p_t = t.host();
        for (long n = 0; n < t.num_samples(); ++n)
        {
            // std::cout << "sample: " << n << '\n';
            for (long k = 0; k < t.k(); ++k)
            {
                // std::cout << "\tk: " << k << '\n';
                for (long r = 0; r < t.nr(); ++r)
                {
                    for (long c = 0; c < t.nc(); ++c)
                    {
                        // std::cout << '\t' << p_t[tensor_index(t, n, k, r, c)] << ' ';
                        sum += p_t[tensor_index(t, n, k, r, c)];
                    }
                    // std::cout << '\n';
                }
            }
        }
        std::cout << "avg: " << sum / t.size() << '\n';
    };

    resizable_tensor x(2, 3, 224, 224);
    resizable_tensor y_cpu(x), y_cuda(x);
    tt::tensor_rand rnd(0);
    rnd.fill_uniform(x);
    resizable_tensor means(x.num_samples()), invstds(x.num_samples());
    resizable_tensor gamma(x.num_samples()), beta(x.num_samples());
    gamma = 1;
    beta = 0;

    print_tensor_and_avg(x);

    float eps = 1e-5;
    running_stats<float> rs;
    cpu::layer_normalize(eps, y_cpu, means, invstds, x, gamma, beta);
    print_tensor_and_avg(y_cpu);
    for (int i = 0; i < 1000; ++i)
    {
        const auto t0 = std::chrono::steady_clock::now();
        cpu::layer_normalize(eps, y_cpu, means, invstds, x, gamma, beta);
        const auto t1 = std::chrono::steady_clock::now();
        rs.add(std::chrono::duration_cast<std::chrono::duration<float, std::micro>>(t1 - t0).count());
    }
    std::cout << "inference time: " << rs.mean() << " ± " << rs.stddev() << " µs\n";

    rs.clear();
    cuda::layer_normalize(eps, y_cuda, means, invstds, x, gamma, beta);
    print_tensor_and_avg(y_cuda);
    for (int i = 0; i < 1000; ++i)
    {
        const auto t0 = std::chrono::steady_clock::now();
        cuda::layer_normalize(eps, y_cuda, means, invstds, x, gamma, beta);
        const auto t1 = std::chrono::steady_clock::now();
        rs.add(std::chrono::duration_cast<std::chrono::duration<float, std::micro>>(t1 - t0).count());
    }
    std::cout << "inference time: " << rs.mean() << " ± " << rs.stddev() << " µs\n";

    std::cout << "difference: " << sum(squared(mat(y_cpu) - mat(y_cuda))) << '\n';

    return EXIT_SUCCESS;
}
{
    std::cout << e.what() << '\n';
    return EXIT_FAILURE;
}
```
With a size of (2, 3, 4, 5), I get this:
``` c++
// cpu
avg: 7.05322e-08
inference time: 29.3967 ± 0.475304 µs
// cuda
avg: 6.25849e-08
inference time: 49.7305 ± 1.50192 µs
difference: 2.39537e-12
```
But with a size of (2, 3, 224, 224):
``` c++
// cpu
avg: -4.32392e-06
inference time: 777.195 ± 16.5298 µs
// cuda
avg: -379502
inference time: 4188.05 ± 267.693 µs
difference: nan
```

Thanks in advance for any help

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1754,9 +1754,9 @@ namespace dlib
         __global__ void _cuda_layer_normalize(float* out, const float* s, float* m, float* v, const float* g, const float* b, float eps, size_t ns, size_t num)
         {
            // compute means and sum of squares
-            for (auto n : grid_stride_range(0, ns))
+            for (auto n : grid_stride_range_y(0, ns))
             {
-                for (auto i : grid_stride_range_y(0, num))
+                for (auto i : grid_stride_range(0, num))
                 {
                     const float val = s[n*num+i];
                     m[n] += val;
@@ -1764,9 +1764,9 @@ namespace dlib
                 }
             }
             __syncthreads();
-            for (auto n : grid_stride_range(0, ns))
+            for (auto n : grid_stride_range_y(0, ns))
             {
-                for (auto i : grid_stride_range_y(0, 1))
+                for (auto i : grid_stride_range(0, 1))
                 {
                     m[n] /= num;
                     v[n] /= num;
@@ -1774,20 +1774,19 @@ namespace dlib
             }
             __syncthreads();
             // compute variances
-            for (auto n : grid_stride_range(0, ns))
+            for (auto n : grid_stride_range_y(0, ns))
             {
-                for (auto i : grid_stride_range_y(0, 1))
+                for (auto i : grid_stride_range(0, 1))
                 {
                     auto var = v[n] - m[n] * m[n];
                     v[n] = 1.0f / std::sqrt(var + eps);
                 }
             }
             __syncthreads();
-            for (auto n : grid_stride_range(0, ns))
+            for (auto n : grid_stride_range_y(0, ns))
             {
-                for (auto i : grid_stride_range_y(0, num))
+                for (auto i : grid_stride_range(0, num))
                 {
-                    // printf("%d, %d: %f, %f, %f\n", (int)n, (int)i, s[n*num+i], m[n], v[n]);
                     const float val = (s[n*num+i]-m[n])*v[n];
                     out[n*num+i] = val*g[n]+b[n];
                 }
@@ -1832,7 +1831,7 @@ namespace dlib
             invstds.set_size(src.num_samples());
             means = 0;
             invstds = 0;
-            launch_kernel(_cuda_layer_normalize, max_jobs(num, src.num_samples()), dest.device(), src.device(),
+            launch_kernel(_cuda_layer_normalize, max_jobs(src.num_samples(), num), dest.device(), src.device(),
                           means.device(), invstds.device(), gamma.device(), beta.device(),
                           eps, src.num_samples(), num);
         }
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1756,23 +1756,19 @@ namespace dlib
            // compute means and sum of squares
             for (auto n : grid_stride_range_y(0, ns))
             {
+                auto p = s + n * num;
+                float means = 0;
+                float invstds = 0;
                 for (auto i : grid_stride_range(0, num))
                 {
-                    const float val = s[n*num+i];
-                    m[n] += val;
-                    v[n] += val*val;
+                    means += p[i];
+                    invstds += p[i] * p[i];
                 }
+                warp_reduce_atomic_add(m[n], means/num);
+                warp_reduce_atomic_add(v[n], invstds/num);
             }
             __syncthreads();
-            for (auto n : grid_stride_range_y(0, ns))
-            {
-                for (auto i : grid_stride_range(0, 1))
-                {
-                    m[n] /= num;
-                    v[n] /= num;
-                }
-            }
-            __syncthreads();
+
             // compute variances
             for (auto n : grid_stride_range_y(0, ns))
             {
@@ -1783,6 +1779,7 @@ namespace dlib
                 }
             }
             __syncthreads();
+
             for (auto n : grid_stride_range_y(0, ns))
             {
                 for (auto i : grid_stride_range(0, num))
@@ -1831,9 +1828,8 @@ namespace dlib
             invstds.set_size(src.num_samples());
             means = 0;
             invstds = 0;
-            launch_kernel(_cuda_layer_normalize, max_jobs(src.num_samples(), num), dest.device(), src.device(),
-                          means.device(), invstds.device(), gamma.device(), beta.device(),
-                          eps, src.num_samples(), num);
+            launch_kernel(_cuda_layer_normalize, max_jobs(num, src.num_samples()), dest.device(), src.device(),
+                          means.device(), invstds.device(), gamma.device(), beta.device(), eps, src.num_samples(), num);
         }
 
         void layer_normalize_gradient (
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1792,7 +1792,52 @@ namespace dlib
 
         __global__ void _cuda_layer_normalize_gradient(float* out, float* gg, float* bg, const float* s, const float* gi, const float* m, const float* v, const float* g, float* dm, float* dv, float eps, size_t ns, size_t num)
         {
-            // TODO
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                float temp_bg = 0;
+                float temp_gg = 0;
+                float temp_dv = 0;
+                for (auto i : grid_stride_range(0, num))
+                {
+                    auto idx = n*num+i;
+                    const float x_hat = (s[idx] - m[n])*v[n];
+                    temp_bg += gi[idx];
+                    temp_gg += gi[idx]*x_hat;
+
+                    const float dx = gi[idx] * g[n];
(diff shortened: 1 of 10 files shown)


dlib/cuda/cuda_dlib.cu
```diff
@@ -1749,6 +1749,129 @@ namespace dlib
...
         }
 
+    // ----------------------------------------------------------------------------------------
+
+        __global__ void _cuda_layer_normalize(float* out, const float* s, float* m, float* v, const float* g, const float* b, float eps, size_t ns, size_t num)
+        {
+           // compute means and sum of squares
+            for (auto n : grid_stride_range(0, ns))
+            {
+                for (auto i : grid_stride_range_y(0, num))
```
davisking, line 1759: The outer loop needs to use `grid_stride_range_y` and the inner `grid_stride_range`.  The GPU really doesn't like it the other way around.

arrufat: Yes, I've tried that, but then I get even weirder results, like the average of the output tensor is not even 0, so there must be a problem somewhere else. I've tried swapping the numbers inside of the `max_jobs` when doing that, no luck... I will keep trying

dlib/dnn/layers.h
```diff
@@ -1587,6 +1722,14 @@ namespace dlib
...
             }
 
+            template <typename U, typename E>
+            void set_input_no_bias(add_layer<layer_norm_, U, E>& l)
+            {
+                disable_bias(l.subnet().layer_details());
+                set_bias_learning_rate_multiplier(l.subnet().layer_details(), 0);
+                set_bias_weight_decay_multiplier(l.subnet().layer_details(), 0);
+            }
+
```
arrufat, line 1732: I think it makes sense to update this visitor, to remove bias from inputs of `layer_norm`. However, I think it should be renamed. Something like: `remove_superfluous bias`, so that it can still make sense if we add more layers that can be added to the visitor. Any better idea?

arrufat:
Ready for review!
In this implementation, these are the results for a (2, 3, 224, 224)-sized tensor

```
 cpu forward time: 780.407 ± 12.5129 µs
cuda forward time: 87.3476 ± 2.33944 µs

 cpu backward time: 3706.97 ± 45.1868 µs
cuda backward time: 448.009 ± 31.0569 µs
```

dlib/dnn/layers.h
davisking: Yeah, how about `remove_duplicative_biases`?

dlib/test/dnn.cpp
```diff
@@ -470,6 +470,59 @@ namespace
...
+        resizable_tensor means(x.num_samples()), invstds(x.num_samples());
+        resizable_tensor gamma(x.num_samples()), beta(x.num_samples());
+        gamma = 1;
+        beta = 0;
+        const float eps = 1e-5;
+        resizable_tensor gradient_input(x), src_grad_cpu(x);
+        resizable_tensor gamma_grad_cpu(x.num_samples()), beta_grad_cpu(x.num_samples());
+        src_grad_cpu = 0;
+        rnd.fill_gaussian(gradient_input);
+        cpu::layer_normalize(eps, y_cpu, means, invstds, x, gamma, beta);
```
davisking, line 490: It looks like `y_cpu` is uninitialized.  Fill it with something like uniform random numbers, so it's got non-zero mean and non-unit variance.

davisking: Oh wait, nevermind.  `y_cpu` is a pure output.

dlib/test/dnn.cpp
```diff
@@ -470,6 +470,59 @@ namespace
...
+                }
+            }
+            DLIB_TEST(::std::abs(rs.mean()) < 1e-6);
+            DLIB_TEST(::std::abs(rs.stddev() - 1.0f) < 0.01);
+        }
+        // check that the CPU and the CUDA implementation are equivalent
+#if DLIB_USE_CUDA
+        resizable_tensor y_cuda(x);
+        cuda::layer_normalize(eps, y_cuda, means, invstds, x, gamma, beta);
+        DLIB_TEST(max(abs(mat(y_cpu) - mat(y_cuda))) < 1e-5);
```
davisking, line 513: Check that the cuda and cpu versions of `means` and `invstds` match too.

davisking:
Cool, looks good.  Thanks for another PR :)

Rename that function you noted and updated the docs and add that one test and this is good to go :)

dlib/dnn/layers.h
arrufat: `disable_duplicative_biases`? Since we call `disable_bias` on the layers...

davisking: Oh yeah, that’s a better name :)

dlib/dnn/layers.h
```diff
@@ -1712,32 +1712,40 @@ namespace dlib
...
             }
 
+            template <layer_mode mode, size_t N, template <typename> class R, typename U, typename E>
+            void disable_input_bias(add_layer<bn_<mode>, repeat<N, R, U>, E>& l)
+            {
+                disable_bias(l.subnet().get_repeated_layer(0).layer_details());
+                set_bias_learning_rate_multiplier(l.subnet().get_repeated_layer(0).layer_details(), 0);
+                set_bias_weight_decay_multiplier(l.subnet().get_repeated_layer(0).layer_details(), 0);
+            }
+
```
arrufat, line 1748: Not sure about this... as I said in #2216, it works for some cases, but not all...

dlib/dnn/layers.h
```diff
@@ -1577,24 +1712,61 @@ namespace dlib
...
+            template <layer_mode mode, unsigned long ID, typename E, typename F>
+            void disable_input_bias(add_layer<bn_<mode>, add_tag_layer<ID, impl::repeat_input_layer, E>, F>& )
+            {
+            }
+
+            template <unsigned long ID, typename E, typename F>
+            void disable_input_bias(add_layer<layer_norm_, add_tag_layer<ID, impl::repeat_input_layer, E>, F>& )
+            {
+            }
+
```
arrufat, line 1769: I've added these overloads to handle repeat layers with tag layers. Since the `impl::repeat_input_layer` does not have any method, we can't do anything here.

dlib/test/dnn.cpp
```diff
@@ -3846,6 +3906,46 @@ namespace
...
+        DLIB_TEST(layer<3>(net).layer_details().bias_is_disabled() == true);
+        DLIB_TEST(layer<6>(net).layer_details().bias_is_disabled() == true);
+        DLIB_TEST(layer<9>(net).layer_details().bias_is_disabled() == true);
+        DLIB_TEST(layer<12>(net).layer_details().bias_is_disabled() == false);
+        DLIB_TEST(layer<15>(net).layer_details().bias_is_disabled() == true);
+        DLIB_TEST(layer<21>(net).layer_details().bias_is_disabled() == false);
+        DLIB_TEST(layer<24>(net).layer_details().bias_is_disabled() == true);
+        DLIB_TEST(layer<31>(net).layer_details().bias_is_disabled() == true);
+    }
+
```
arrufat, line 3948: Just a weird network that has some interesting cases to test.

davisking:
Nice, looks good :)

examples/dnn_dcgan_train_ex.cpp
```diff
@@ -134,8 +134,8 @@ int main(int argc, char** argv) try
     // setup all leaky_relu_ layers in the discriminator to have alpha = 0.2
     visit_computational_layers(discriminator, [](leaky_relu_& l){ l = leaky_relu_(0.2); });
     // Remove the bias learning from all bn_ inputs in both networks
-    set_all_bn_inputs_no_bias(generator);
-    set_all_bn_inputs_no_bias(discriminator);
+    disable_duplicative_bias(generator);
+    disable_duplicative_bias(discriminator);
```
arrufat, line 138: I noticed that we agreed to call it `disable_duplicative_biases`, but somehow I just called like this... Should we rename it before the stable release? I think it should be plural, too.

arrufat, line 138: @davisking, I noticed that we agreed to call it `disable_duplicative_biases`, but somehow I just called like this... Should we rename it before the stable release? I think it should be plural, too.

davisking: Huh, yeah.  `disable_duplicative_biases` is probably a better name.  If you want to do that rename in a separate PR that would be cool.

arrufat: Sorry about that...

davisking: No worries :)

PR #2217, opened 2020-10-21
Remove an unused variable and old commented code in layer_norm

arrufat:
I didn't notice this when I submitted the PR. `nvcc` complained about an unreferenced variable, so I removed it, as well as some old code from when I was implementing/debugging this layer.

Sorry about that :(


davisking:
No worries.  I noticed this too in my first review and then forgot about it later :)

PR #2218, opened 2020-10-24
[WIP] Use dpoint instead of point in full_object_detection

ejoebstl:


ejoebstl:
This currently failes compilation with 

``` c++
/Projects/dlib/tools/python/src/image_dataset_metadata.cpp:192:85:   required from here
/Projects/dlib/dlib/external/pybind11/include/pybind11/cast.h:1382:61: error: static assertion failed: Holder classes are only supported for custom types
 1382 |     static_assert(std::is_base_of<base, type_caster<type>>::value,
      |                                                             ^~~~~
```

Any help would be appreciated - I have not worked with pybind11 before.

PR #2227, opened 2020-11-03
Update setup.py

GitHubEmploy:
system code 0 is more suitable here


davisking:
Why?  Giving --yes is actually an error.

GitHubEmploy:
Exit code 0 will also kill background processes, code 1 will only kill the code.

PR #2233, opened 2020-11-12
[Fix] Same branches for inconsistent build check

topin89:
Inconsistent build configuration was checked regardless of ENABLE_ASSERTS.
It looks like an error, and dropping it helps with header only use.

### DIFF
--- a/dlib/test_for_odr_violations.h
+++ b/dlib/test_for_odr_violations.h
@@ -19,9 +19,6 @@ extern "C"
 #ifdef ENABLE_ASSERTS
     const extern int USER_ERROR__inconsistent_build_configuration__see_dlib_faq_1;
     const int DLIB_NO_WARN_UNUSED dlib_check_assert_helper_variable = USER_ERROR__inconsistent_build_configuration__see_dlib_faq_1;
-#else
-    const extern int USER_ERROR__inconsistent_build_configuration__see_dlib_faq_1_;
-    const int DLIB_NO_WARN_UNUSED dlib_check_assert_helper_variable = USER_ERROR__inconsistent_build_configuration__see_dlib_faq_1_;
 #endif
 
 


davisking:
This is here on purpose.  The logic of the check doesn't work if you remove that code.  Right, this would allow someone to build object files with asserts enabled and then later build other object files against headers with asserts disabled.  That would be an ODR violation.

davisking:
I do realize this is irritating for things that would otherwise work header only.  But before I added these checks I would constantly get posts from users who messed up their build systems and created ODR violations.  There isn't anything I can do to stop users from doing that.  Like they would setup some broken build system, and they happen to also be using dlib and so they complain to me about it.  But it's got nothing to do with dlib.

I'm open to suggestions for how to make this less annoying.  But I don't really want to remove the error checking.  Maybe there could be some option that disables this, like #define DLIB_HEADER_ONLY or something.  Needing that is also sort of irritating though and it wouldn't surprise me if I then start getting messages from users like "I'm using #define DLIB_HEADER_ONLY and I then went and built the dlib .cpp files into my program anyway and created ODR violations!".   Or people would get one of these "you are making an ODR violation, fix your build system!" errors.  And their solution is go to stick `#define DLIB_HEADER_ONLY` at the top of their .cpp files to make it go away.

topin89:
No problem there. I thought it was a mistake because it's the same two lines in two branches. Maybe just remove `#ifdef ENABLE_ASSERTS` altogether then?

davisking:
Na, those are different lines.

topin89:
Oh. One tiny underscore. Didn't notice. Still, maybe at least a change in the FAQ?

davisking:
What change would you suggest?

topin89:
So, for [this question](http://dlib.net/faq.html#WhydoIgetUSERERRORinconsistentbuildconfigurationseedlibfaq1), I suggest something like that:


> If you absolutely sure you have to use dlib as headless library (strongly discouraged), e.g. to use [pipe](http://dlib.net/dlib/pipe/pipe_kernel_abstract.h.html), you should comment out macroses `ENABLE_ASSERTS` and `DLIB_CHECK_FOR_VERSION_MISMATCH ` in config.h and this line:
> ```C++
>     const int USER_ERROR__inconsistent_build_configuration__see_dlib_faq_1_ = 0;
> ```
> in one (and only one) of your cpp file.
> 
> Before you do that, make sure you actually need this. If you don't wand to use dlib as a shared library, you can simply add `source.cpp` to a list of `*.cpp` files.

This definitely worked for `pipe` header-only use. Of course, then I thought these "identical" lines is a bug, and "fixed" that. As for why I didn't just include `source.cpp` myself, it was a fix of [another project](https://github.com/cameron314/concurrentqueue) with five Studio projects, one xcode project and one Makefile. At that point, just adding one line to an existing file, and then removing two from another one feels less of a hassle.

davisking:
It honestly sounds like you are talking about creating an actual ODR violation though.  Like if you are using `dlib::pipe` then you need to be building the threading `.cpp` files or it won't work.  It's not header only.  And if you are doing that and getting errors from things like `USER_ERROR__inconsistent_build_configuration__see_dlib_faq_1_` then it's because you have done something in your build setup to violate ODR.

topin89:
An yet it works: https://github.com/topin89/dlib_pipe_is_headless/

Looks like none of the function in threading `.cpp`s are actually used, and unused function declarations are allowed. Luck, I guess, but it works anyway.

Still, as I understand it, headless use is unsupported, or at least limited to a few libs. Maybe this should work for the FAQ:

> Maybe you're trying to use dlib as a headless library. This is not supported <or supported only for ...> and you're on your own, so don't ask for help if something is not working as expected.
> 
> <previous quote snip>

davisking:
Ah, yeah, turns out the `dlib::pipe` does work with just the header.  Header only use is fine and something I expect people to be able to do for some things.  If you want to do it you can just delete the contents of `dlib/test_for_odr_violations.h`.  Nothing bad should happen for actually header only use.  The problem happens when people then go build in dlib/all/source.cpp with different settings or do some other ODR violating thing.

But if you really aren't building any dlib .cpp files into your application then you can't get into any ODR violation situations.

topin89:
> Header only use is fine and something I expect people to be able to do for some things. If you want to do it you can just delete the contents of `dlib/test_for_odr_violations.h`

Cool! It _is_ worth mentioning in FAQ, I think. Or maybe another approach:

Another macro, `ENABLE_HEADLESS`. So, if enabled, `test_for_odr_violations.h` is skipped entirely, but another header, e.g. `test_header_only.h`, with something like

```C++
#ifdef ENABLE_HEADLESS
extern const int USER_ERROR__source_cpp_is_compiled_despite_using_headless;
#endif
```

included in `source.cpp`.

davisking:
Yeah, an `ENABLE_HEADER_ONLY` macro could be ok.  That seems kinda funny too though.  I'm more happy to just let someone who knows what they are doing delete the contents of `test_for_odr_violations.h`.  I'm hesitant to put that in the FAQ though, because lots of users will cargo cult that into their programs without understanding and I'll be right back to the original problem.

PR #2234, opened 2020-11-13
Add custom ostream to console progress indicator (closes #2232)

### LINKED ISSUE #2232: Question about modifying console_progress_indicator
Hi, I would like to add an option so that it can print to std::cerr (or a custom std::ostream). This can be useful when piping the output, to a file, since we don't want the output of the progress.

So I was wondering if you'll be okay with that. I was thinking about doing it like you do it in the `cmd_line_parser::print_options`, which has an overload with a custom `std::ostream`. Would that be ok?

arrufat:
Would this be enough to address #2232?

### DIFF
--- a/dlib/console_progress_indicator.h
+++ b/dlib/console_progress_indicator.h
@@ -66,15 +66,16 @@ namespace dlib
 
         inline bool print_status (
             double cur,
-            bool always_print = false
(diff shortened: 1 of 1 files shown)


davisking:
Looks good :)

PR #2236, opened 2020-11-16
Make dnn_trainer print the minibatch size to ostream

arrufat:
So, maybe this a bit controversial...

When I train networks, I like to fill all the fields of the `dnn_trainer` so that they all appear in my training log, and then access them from the trainer (`get_min_learning_rate()`, `get_mini_batch_size()`, etc...)
I found myself logging the minibatch-size manually many times, so finally I decided to make the `dnn_trainer` print it.

I was hesitating to submit this PR because, when you build your minibatches manually, and print the trainer to the screen, users might be surprised to see a `128` for the minibatch size (default value). But then I thought that it is already the case with the minimum learning rate, if you train with your own custom loop.

Anyway, feel free to reject it if you think it's not a good idea :)


davisking:
Na, it's cool :)

PR #2238, opened 2020-11-17
Expose get_face_chip_details to Python

frankier:
I found I needed this for:

1. Visualizing the positions of the chips on video
2. Calibrating chip extraction based on different facial keypoint models


davisking:
Thanks for the PR :)

PR #2240, opened 2020-11-18
Fix unused parameter warning in visitor_net_to_xml

arrufat:
Minor fix, but when compiling with `-Wall -Wextra -pedantic`, I got a really huge nested warning when converting big networks to xml.

EDIT: I hope I am not bothering you with all these tiny PR...


davisking:
Na, keep them coming.  These are good PRs :)

PR #2241, opened 2020-11-19
Add support for matrix serialization to python API

arrufat:
I've added serialization support for matrices (I know they are already pickable).

I was facing the following problem: I had to load some numpy arrays, but I wanted to work with them in C++. After looking online on how to load them, I realized that the easiest way was to make dlib do it for me:

``` python
import numpy as np
from dlib import matrix
# get a numpy array
arr = np.random.rand(10, 2)
# convert it to a dlib.matrix and serialize it to disk
matrix(arr).serialize("rand.dat")
# create a new matrix
m = matrix()
# deserialize the previous matrix into this one
m.deserialize("rand.dat")
```
The good thing is that now I can deserialize the `rand.dat` from C++.

Maybe there's a better way of doing it, but I think having this is pretty convenient.

### DIFF
--- a/tools/python/src/matrix.cpp
+++ b/tools/python/src/matrix.cpp
@@ -108,6 +108,16 @@ long matrix_double__len__(matrix<double>& c)
     return c.nr();
 }
 
+void matrix_serialize(const matrix<double>& m, const std::string& file)
+{
+    serialize(file) << m;
+}
+
+void matrix_deserialize(matrix<double>& m, const std::string& file)
+{
+    deserialize(file) >> m;
+}
+
 struct mat_row
 {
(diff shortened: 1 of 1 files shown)


davisking:
Looks good :+1:

pfeatherstone:
This is fantastic. Any suggestions on how to move forward to support complex types? If I try with a complex number array, i get warning:
```
ComplexWarning: Casting complex values to real discards the imaginary part
```

pfeatherstone:
Maybe copy paste the contents of ```tools/python/src/matrix.cpp``` to support ```dlib::matrix<std::complex<double>>``` and bind to python object ```matrixcplx``` maybe.

Issue #2243, opened 2020-11-20
vector<complex<float>> not (de)serializable


pfeatherstone:
## Expected Behavior

```vector<complex<float>>``` is not (de)serializable. Here is a minimal example that demonstrates this.

```
vector<complex<float>> d;
dlib::serialize("bug.txt") << d;
```

I don't think this is expected behaviour

## Current Behavior

Using clang 6 i get the following errors:
```
In file included from main.cpp:5:
/usr/local/include/dlib/array/../serialize.h:1578:17: error: call to function 'serialize' that is neither visible in the template definition nor found by argument-dependent lookup
                serialize(item[i],out);
                ^
/usr/local/include/dlib/array/../serialize.h:2281:13: note: in instantiation of function template specialization 'dlib::serialize<std::complex<float>, std::allocator<std::complex<float> > >' requested here
            serialize(item, fout);
            ^
main.cpp:30:32: note: in instantiation of function template specialization 'dlib::proxy_serialize::operator<<<std::vector<std::complex<float>, std::allocator<std::complex<float> > > >' requested here
    dlib::serialize("bug.txt") << d;
                               ^
/usr/local/include/dlib/array/../serialize.h:2126:17: note: 'serialize' should be declared prior to the call site
    inline void serialize (
                ^
1 error generated.
```

## Steps to Reproduce

* **Version**: latest
* **Where did you get dlib**: github (built from source)
* **Platform**: 
```
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.5 LTS
Release:	18.04
Codename:	bionic
```

* **Compiler**: 
```
 clang++ --version
clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

**Comment by pfeatherstone** [comment]
Note that moving 
```
    template <
        typename T
        >
    inline void serialize (
        const std::complex<T>& item,
        std::ostream& out
    )
    {
        try
        {
            serialize(item.real(),out);
            serialize(item.imag(),out);
        }
        catch (serialization_error& e)
        {
            throw serialization_error(e.info + "\n   while serializing an object of type std::complex");
        }
    }

// ----------------------------------------------------------------------------------------

    template <
        typename T
        >
    inline void deserialize (
        std::complex<T>& item,
        std::istream& in
    )
    {
        try
        {
            T real, imag;
            deserialize(real,in); 
[...]

PR #2244, opened 2020-11-21
Fixes bug when (de)serializing vector<complex<float>>

pfeatherstone:
This resolves issue https://github.com/davisking/dlib/issues/2243

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -2583,7 +2583,7 @@ namespace dlib
         {                                                               \
             int dlibDefaultSer$_version = 0;                            \
             deserialize(dlibDefaultSer$_version, dlibDefaultSer$_in);   \
-            if (version != 1)                                           \
+            if (dlibDefaultSer$_version != 1)                           \
                 throw dlib::serialization_error("Unexpected version found while deserializing " #Type); \
             deserialize_these(dlibDefaultSer$_in, __VA_ARGS__);         \
         }                                                               \
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -2555,7 +2555,7 @@ namespace dlib
     }  
     
     #define DLIB_DEFINE_DEFAULT_SERIALIZATION(Type, ...)                \
-    void serialize_to(std::ostream& __out) const                        \
+    void serialize_to(std::ostream& dlibDefaultSer$_out) const          \
     {                                                                   \
         using dlib::serialize;                                          \
         using dlib::serialize_these;                                    \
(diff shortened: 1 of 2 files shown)


dlib/serialize.h
```diff
@@ -2555,7 +2555,7 @@ namespace dlib
     }  
     
     #define DLIB_DEFINE_DEFAULT_SERIALIZATION(Type, ...)                \
-    void serialize_to(std::ostream& out) const                          \
+    void serialize_to(std::ostream& __out) const                        \
```
arrufat, line 2558: if I remember correctly, identifiers starting with double underscore are reserved, and may cause ub.

pfeatherstone: What convention do you prefer? ```out``` and ```in``` need to be refactored to something sensible that would likely not conflict with member variables of custom types.

pfeatherstone: @davisking Might have an opinion on appropriate names.

arrufat: I would probably just use `out_` and `in_`, or `out_param_` and `in_param_` if you want to reduce the collision odds.

pfeatherstone: I think i prefer ```out_ref_``` and ```in_ref_``` as private member variables are often either prefixed or suffixed with ```_```.

pfeatherstone: I don't know if a silly name like ```out_avoid_conflict_with_member_variable_name``` is ok?

pfeatherstone: I've seen @davisking use names like that when writing visual studio workarounds.

davisking: Yeah, it is illegal to use names with double underscores.  But yeah, these names are not good.  It's also a problem that `version` is simple name too.  I would prefix in, out, and version with some crazy string.  Like turn `out` into `dlibDefaultSer$_out`.   No one will ever type `dlibDefaultSer$_`.

dlib/test/serialize.cpp
```diff
@@ -431,16 +432,17 @@ namespace
         std::multiset<string> n;
         std::unordered_multiset<string> o;
         std::shared_ptr<string> ptr_shared1;
-        std::shared_ptr<string> ptr_shared2;    
+        std::shared_ptr<string> ptr_shared2;
+	std::vector<std::complex<double>> p;
```
davisking, line 436: Can you remove the tabs from this file?

pfeatherstone: Yep

dlib/test/serialize.cpp
```diff
@@ -1119,6 +1121,9 @@ namespace
         t1.o.insert("hello from unordered_multiset");
         t1.o.insert("hello from unordered_multiset");
         t1.ptr_shared1 = make_shared<string>("hello from shared_ptr");
+	dlib::rand rng(std::time(NULL));
```
davisking, line 1124: Please remove all the other tabs too though :)

pfeatherstone: Sorry. Hopefully this is it. 24 hour turnaround is long between reviews. Sodding time zones... It's almost as long as waiting for some VHDL to build :)

davisking:

PR #2245, opened 2020-11-21
Add CIFAR-10 dataset loader

arrufat:
I thought this would be nice to have to test easily on CIFAR-10.

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -315,6 +315,7 @@ if (NOT TARGET dlib)
          cuda/tensor_tools.cpp
          data_io/image_dataset_metadata.cpp
          data_io/mnist.cpp
+         data_io/cifar.cpp
          global_optimization/global_function_search.cpp
          filtering/kalman_filter.cpp
          svm/auto.cpp
--- a/dlib/data_io/cifar.cpp
+++ b/dlib/data_io/cifar.cpp
@@ -1,0 +1,90 @@
+// Copyright (C) 2020  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_CIFAR_CPp_
+#define DLIB_CIFAR_CPp_
+
+#include "cifar.h"
+#include <fstream>
+
+// ----------------------------------------------------------------------------------------
+
+namespace dlib
+{
+    void load_cifar_10_dataset (
+        const std::string& folder_name,
+        std::vector<matrix<rgb_pixel>>& training_images,
+        std::vector<unsigned long>& training_labels,
+        std::vector<matrix<rgb_pixel>>& testing_images,
+        std::vector<unsigned long>& testing_labels
+    )
+    {
+        using namespace std;
+        std::vector<string> file_names{
+            folder_name + "/data_batch_1.bin",
+            folder_name + "/data_batch_2.bin",
+            folder_name + "/data_batch_3.bin",
+            folder_name + "/data_batch_4.bin",
+            folder_name + "/data_batch_5.bin",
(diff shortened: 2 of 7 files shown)


dlib/data_io/cifar.cpp
```diff
@@ -73,7 +73,7 @@ namespace dlib
                 }
 
                 std::array<unsigned char, image_size> buffer;
-                fins[i].read((char*)buffer.begin(), buffer.size());
+                fins[i].read((char*)(&buffer[0]), buffer.size());
```
davisking, line 76: `buffer.data()` is preferable to `buffer.data()`

dlib/data_io/cifar.cpp
```diff
@@ -0,0 +1,108 @@
...
+        using namespace std;
+        std::vector<string> file_names{
+            "data_batch_1.bin",
+            "data_batch_2.bin",
+            "data_batch_3.bin",
+            "data_batch_4.bin",
+            "data_batch_5.bin",
+            "test_batch.bin"
+        };
+        std::vector<ifstream> fins;
```
davisking, line 30: It's kinda funny to make an array of ifstreams.  You should instead make them as you need them.  E.g.
```
for (const string filename : {"file1.bin", "file2.bin"})
{
    ifstream fin(folder_name + "/" + filename), ios::binary);
    ...
} 
```

dlib/data_io/cifar.cpp
```diff
@@ -0,0 +1,108 @@
...
+        const long image_size = 3 * plane_size;
+
+        training_images.resize(images_per_batch * training_batches);
+        training_labels.resize(images_per_batch * training_batches);
+        testing_images.resize(images_per_batch * testing_batches);
+        testing_labels.resize(images_per_batch * testing_batches);
+
+        for (size_t i = 0; i < fins.size(); ++i)
+        {
+            bool test_batch = i == fins.size() - 1;
```
davisking, line 56: Rather than having this logic, make a function that does the reading, then make a loop that loops over the training files and calls that function. Then call that function again for the testing file.

dlib/CMakeLists.txt
```diff
@@ -315,6 +315,7 @@ if (NOT TARGET dlib)
          cuda/tensor_tools.cpp
          data_io/image_dataset_metadata.cpp
          data_io/mnist.cpp
+         data_io/cifar.cpp
```
davisking, line 318: Add this to dlib/all/source.cpp too, just like is done for mnist.cpp.

dlib/data_io/cifar.cpp
arrufat: Yes, I was not totally happy about it either, but it made the code so much shorter... 😅

dlib/data_io/cifar.cpp
```diff
@@ -0,0 +1,106 @@
...
+        void load_cifar_10_batch (
+            const std::string& folder_name,
+            const std::string& batch_name,
+            const size_t first_idx,
+            const size_t images_per_batch,
+            std::vector<matrix<rgb_pixel>>& images,
+            std::vector<unsigned long>& labels
+        )
+        {
+            std::ifstream fin((folder_name + "/" + batch_name).c_str(), std::ios::binary);
```
davisking, line 24: Since C++11 the `.c_str()` is not needed anymore.

davisking:
Thanks for another PR :)

dlib/data_io/cifar.cpp
arrufat: I actually never used in my code, but saw that the MNIST code used it (I started late with C++, so not really aware of the pre C++11 era). Now I know, thank you :)

PR #2247, opened 2020-11-27
Adding exponential and weibull distributions

pfeatherstone:


pfeatherstone:
for unit tests, i was thinking of calculating the mean and variance of these distributions for a range of parameters and checking they are correct. Formulae for these are on wikipedia. the only thing is that these require the gamma function. which we can add as well. But then you need unit tests for the gamma function. @davisking what do you think? Is this sound? Or can you think of better unit tests for these distributions that don't require gamma functions.

dlib/rand/rand_kernel_1.h
```diff
@@ -265,6 +265,31 @@ namespace dlib
...
+            )
+            {
+                DLIB_ASSERT(alpha > 0, "alpha must be greater than zero");
+                DLIB_ASSERT(beta > 0, "beta must be greater than zero");
+                double u = 0.0;
+                while (u == 0.0)
+                    u = get_random_double();
+                return gamma + beta*std::pow(-std::log(u), 1.0 / alpha);
+            }
+
```
davisking, line 292: Please update the docs in the _abstract.h file too.

davisking: It's also a good idea to make the parameter names match the names used on wikipedia, since that's what most people will expect.  E.g. https://en.wikipedia.org/wiki/Weibull_distribution

pfeatherstone: Shall i refactor alpha and beta to k and lambda respectively?

davisking: Yeah, seems like a good idea :)

davisking:

PR #2252, opened 2020-12-06
Release GIL before detecting faces

akshaym:
A small modification to release the GIL in the face_detector. Since the face_detector seemed to be single-threaded, I made this change. 

I wrote some simple code to benchmark:
```python
files = sys.argv[1:] * 16

def basic():
    detector = dlib.get_frontal_face_detector()
    for f in files:
        img = dlib.load_rgb_image(f)
        # The 1 in the second argument indicates that we should upsample the image
        # 1 time.  This will make everything bigger and allow us to detect more
        # faces.
        dets = detector(img, 1)

def threads(num_threads=8):
  file_chunks = np.array_split(np.asarray(files), num_threads)

  def work(files):
      detector = dlib.get_frontal_face_detector()

      for f in files:
          img = dlib.load_rgb_image(f)
          # The 1 in the second argument indicates that we should upsample the image
          # 1 time.  This will make everything bigger and allow us to detect more
          # faces.
          dets = detector(img, 1)

  threads = []
  for i in range(num_threads):
    threads.append(threading.Thread(target=work, args=(file_chunks[i],)))
    threads[-1].start()
  for thread in threads:
    thread.join()

import os; print(os.cpu_count())
num = 1
print("basic", timeit.timeit("basic()", number=num, globals=globals()))
print("threads", timeit.timeit("threads()", number=num, globals=globals()))
```

For the [image](https://github.com/davisking/dlib/blob/master/examples/faces/bald_guys.jpg) which is processed 16 times by the face detector, I get the following results:
```
basic 19.069416782
threads 5.340275335000001
```

(Note that, different numbers of threads provide different amount of speedups, this is run with the code above).

### DIFF
--- a/tools/python/src/simple_object_detector_py.h
+++ b/tools/python/src/simple_object_detector_py.h
@@ -52,16 +52,21 @@ namespace dlib
         if (is_image<unsigned char>(img))
         {
             array2d<unsigned char> temp;
+            auto img_view = make_image_view(numpy_image<unsigned char>(img));
+
+            // Release the GIL so that this code can be run in parallel.
+            py::gil_scoped_release release;
+
             if (upsampling_amount == 0)
             {
-                detector(numpy_image<unsigned char>(img), rect_detections, adjust_threshold);
+                detector(img_view, rect_detections, adjust_threshold);
                 split_rect_detections(rect_detections, rectangles,
                                       detection_confidences, weight_indices);
                 return rectangles;
             }
             else
             {
-                pyramid_up(numpy_image<unsigned char>(img), temp, pyr);
+                pyramid_up(img_view, temp, pyr);
                 unsigned int levels = upsampling_amount-1;
                 while (levels > 0)
                 {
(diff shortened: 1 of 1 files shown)


davisking:
Cool, that's a good idea :)

But is taking that `py::array` `img` and making a copy of it, which calls python functions.  So you can't unlock the gil there.  The gil needs to only be unlocked around code that is not calling into any python code.  So the unlock needs to be moved to be around only the actual detector execution.

akshaym:
Thank you for catching that!

Reading through the code, it seems the detector is passed a `dlib::numpy_image` which is a `py::array_t` and also a pyobject. I don't think any python functions are called on it, but I'm not a 100% sure - is it OK to just go ahead and scope the release to the detector call?

davisking:
It's not clear to me either.  You need to refer to the pybind11 documentation to find out if it's ok to call that thing's methods when the GIL is not locked.  It wouldn't surprise me if it was making calls into the python runtime.  So you might have to make a temporary image (just some other struct that implements the image interface: http://dlib.net/dlib/image_processing/generic_image.h.html.  I.e. holds pointers to the image data and all that).  Then pass that struct instead since that struct would be clearly not calling python functions.

davisking:
Er, what am I saying.  The `image_view` is itself a generic image.  So you can just call `make_image_view` to get an image that doesn't have any pybind11 stuff in it and then you just use that.

akshaym:
Thanks for your help. Changed per your suggestion.

davisking:

PR #2253, opened 2020-12-07
Arbitrary sized FFTs using modified kissFFT as default backend and MKL otherwise

pfeatherstone:
This PR adds a C++ port of kissFFT as the default backend of dlib's fft routines. All the kiss code is inserted into the dlib namespace to avoid conflicts with user code that may or may not be using original (C) kissFFT. All MKL fft code is put into it's own separate header file. Now ```dlib/matrix_fft.h``` simply calls the right wrappers depending on a pre-processor macro (as before). I've removed the FFTW wrapper code as per @davisking's request as the fftw planner isn't thread safe. I've also removed the original fft backend code, again, as per @davisking's request.

Note that this PR isn't quite ready yet. Need more unit tests. Not quite sure why MKL wrappers only worked for ```matrix<complex<double>>``` before. There's no reason why it couldn't work for ```matrix<complex<float>>``` if using ```DFTI_SINGLE``` instead of ```DFTI_DOUBLE```. (maybe user code had to do a cast?) So need more units tests for MKL wrappers. 

Not quite happy with the number of copies at the moment. When passing ```const matrix_exp<EXP>& data``` as input, there are 2 copies: one to evaluate the matrix expression to ```matrix<typename EXP::type>``` and one to do either the in-place fft (copy is done in the implementation details of the fft functions) or out-of-place fft (copy is done explicitly in  ```dlib/matrix_fft.h```). For example, if you want to do the fft of std::vector<std::complex<float>>, you have to use ```dlib::mat``` which returns a matrix expression then ```dlib::fft``` will evaluate that to ```dlib::matrix<std::complex<float>>```. So there is a copy of std::vector to dlib::matrix that is unnecessary. Could add some function overloads to support std::vector. Don't know what's the best thing to do without dirtying the API. Could benefit from @davisking's advice. 

I think if we've gone this far, we should support real FFTs  so the requirement that input should be complex vanishes. Output will be complex though. Both KISS and MKL support this so there shouldn't be a lot of leg work. More unit tests.


pfeatherstone:
Also would be a good idea to SIMD the kissFFT wrappers using dlib's SIMD wrappers.

pfeatherstone:
Do we want to cache and copy MKL descriptors using ```DftiCopyDescriptor```?

pfeatherstone:
kissFFT backend could also be optionally parallelised if a ```dlib::thread_pool``` is provided

pfeatherstone:
> kissFFT backend could also be optionally parallelised if a ```dlib::thread_pool``` is provided, using ```dlib::parallel_for```

Actually that's not a good idea.

pfeatherstone:
For unit tests, it looks like a lot of libraries simply do something like :
```
assert(abs(ifft(fft(x)) - x) < tol)
```
assuming ```ifft``` performs scaling and ```tol``` is some small value. 
Is this sufficient? Doesn't feel like it. Maybe should test FFTs of sinusoids, complex exponentials, standard window functions, etc. Would be great if there was some open source dataset for unit testing FFTs...

pfeatherstone:
https://dsp.stackexchange.com/a/635 gives some ideas.

pfeatherstone:
Real FFTs are fixed now. Still need to implement ```fftr``` and ```ifftr``` for MKL

pfeatherstone:
@davisking It would be great if you could do a review. I don't expect this PR to be complete. I still have to finish MKL stuff but there is enough here i think to have an intermediate review.

davisking:
> For unit tests, it looks like a lot of libraries simply do something like :
> 
> ```
> assert(abs(ifft(fft(x)) - x) < tol)
> ```
> 
> assuming `ifft` performs scaling and `tol` is some small value.
> Is this sufficient? Doesn't feel like it. Maybe should test FFTs of sinusoids, complex exponentials, standard window functions, etc. Would be great if there was some open source dataset for unit testing FFTs...

This test should be done.  But adding a single additional test that runs known data with known outputs through each codepath is also a good idea.  There is some of this already in the dlib code.  I don't think it matters what's in the test data so long as it's not trivial in some way.  I would just add in random gaussian noise as I have a hard time imagining code that somehow computes correct ffts for random noise but doesn't for other inputs.

dlib/test/fft.cpp
```diff
@@ -21,19 +21,53 @@ namespace
     using namespace std;
 
     logger dlog("test.fft");
-
+    static dlib::rand rnd;
+    
+// ----------------------------------------------------------------------------------------
+    
+    static inline int fft_next_fast_size(int n)
```
davisking, line 28: Things inside an anonymous namespace are implicitly static already, so don't declare stuff static as it's unneeded.

dlib/matrix/mkl_fft.h
```diff
@@ -0,0 +1,69 @@
...
+            strides[1] = size[1];
+            strides[2] = 1;
+
+            status = DftiSetValue(h, DFTI_INPUT_STRIDES, strides);
+            DLIB_DFTI_CHECK_STATUS(status);
+            status = DftiSetValue(h, DFTI_OUTPUT_STRIDES, strides);
+            DLIB_DFTI_CHECK_STATUS(status);
+        }
+        else
+            throw dlib::error("Need to implement MKL 3D FFT");
```
davisking, line 45: add {}

dlib/matrix/mkl_fft.h
```diff
@@ -0,0 +1,69 @@
...
+        status = DftiSetValue(h, DFTI_THREAD_LIMIT, 1);
+        DLIB_DFTI_CHECK_STATUS(status);
+
+        status = DftiCommitDescriptor(h);
+        DLIB_DFTI_CHECK_STATUS(status);
+
+        if (is_inverse)
+            status = DftiComputeBackward(h, (void*)fin, (void*)fout);
+        else
+            status = DftiComputeForward(h, (void*)fin, (void*)fout);
```
davisking, line 61: Don't add `(void*)`, it's not needed.   Also `in` and `out` might be better names.  It's pretty common for `fin` and `fout` to specifically reference files, but that isn't what these are.

dlib/matrix/matrix_utilities.h
```diff
@@ -48,6 +48,20 @@ namespace dlib
...
+        This is a template that can be used to remove std::complex from the underlying type.
+
+        For example:
+            remove_complex<float>::type == float
+            remove_complex<std::complex<float> >::type == float
+    !*/
+    template <typename T>
+    struct remove_complex {typedef T type;};
+    template <typename T>
+    struct remove_complex<std::complex<T> > {typedef T type;};
```
davisking, line 61: You should make an alias of this that's `remove_complex_t`, as is done with all the similar type traits in the standard library, then client code is much simpler.  In fact, calling it `real_t` would be nice.  Then `std::is_same<real_t<complex<float>> , float>::value == true`

dlib/matrix/mkl_fft.h
```diff
@@ -0,0 +1,69 @@
...
+#define DLIB_DFTI_CHECK_STATUS(s) \
+    if((s) != 0 && !DftiErrorClass((s), DFTI_NO_ERROR)) \
+    { \
+        throw dlib::error(DftiErrorMessage((s))); \
+    }
+
+namespace dlib
+{
+    template<typename T, typename std::enable_if<std::is_floating_point<T>::value>::type* = nullptr>
+    void mkl_fft(const std::vector<int>& dims, const std::complex<T>* fin, std::complex<T>* fout, bool is_inverse)
```
davisking, line 16: It would be good to use something other than `std::vector` for dims.   Anything that doesn't involve a memory allocation since this is just 1 or 2 numbers and an allocation is a lot of overhead for that.  I'm still looking at how this is all used, but maybe `std::array` would be good.

dlib/matrix/matrix_fft.h
```diff
@@ -9,837 +9,142 @@
...
 
-    inline void fft_inplace (matrix<std::complex<double> >& data)
-    {
-        call_mkl_fft_inplace(data, false);
-    }
-    inline void ifft_inplace(matrix<std::complex<double> >& data)
+// ----------------------------------------------------------------------------------------
+    
+    template <typename EXP>
+    matrix<typename remove_complex<typename EXP::type>::type> ifftr (const matrix_exp<EXP>& data)
```
davisking, line 130: I'm not convinced there should be `fftr` and `ifftr`.  The input type of the matrix tells us if it's real or complex, so we shouldn't force the user to tell us that twice by also putting `r` at the end of the function's name.  There can still be overloads specifically for real inputs though.  But they should just be called `fft()` and `ifft()`

dlib/matrix/matrix_fft.h
```diff
@@ -9,837 +9,142 @@
...
-    inline matrix<std::complex<double> > fft (const matrix<std::complex<double> >& data)
-    {
-        return call_mkl_fft(data, false);
-    }
-    inline matrix<std::complex<double> > ifft(const matrix<std::complex<double> >& data)
+    template < typename T, long NR, long NC, typename MM, typename L >
+    void ifft_inplace (matrix<std::complex<T>,NR,NC,MM,L>& data)
     {
-        return call_mkl_fft(data, true) / data.size();
+        kiss_fft({(int)data.nr(), (int)data.nc()}, &data(0,0), &data(0,0), true);
```
davisking, line 110: Make the type of the dims argument `long` so that these `(int)` casts are not required.

dlib/matrix/matrix_fft.h
```diff
@@ -9,837 +9,142 @@
...
-            "\t matrix ifft(data)"
-            << "\n\t The number of rows and columns must be powers of two."
-            << "\n\t data.nr(): "<< data.nr()
-            << "\n\t data.nc(): "<< data.nc()
-            << "\n\t is_power_of_two(data.nr()): " << is_power_of_two(data.nr())
-            << "\n\t is_power_of_two(data.nc()): " << is_power_of_two(data.nc())
-            );
-
+        // You have to give a complex matrix
+        COMPILE_TIME_ASSERT(is_complex<typename EXP::type>::value);
```
davisking, line 87: This can use `static_assert` now.  `COMPILE_TIME_ASSERT` is a pre-c++11 workaround.

dlib/matrix/kiss_fft.h
```diff
@@ -0,0 +1,733 @@
...
+            size_t operator()(const plan_key& key) const {return key.hash();}
+        };
+
+        template<typename plan_type>
+        class cache
+        {
+        public:
+            static cache& get()
+            {
+                static cache singleton;
```
davisking, line 613: Might be faster to use `thread_local` instead of `static` and then not use the mutex.

dlib/matrix/kiss_fft.h
```diff
@@ -0,0 +1,733 @@
...
+                std::lock_guard<std::mutex> l(m);
+                auto it = plans.find(key);
+                if (it != plans.end())
+                {
+                    plan = it->second;
+                }
+                else
+                {
+                    plans[key] = plan_type(key);
+                    plan = plans[key];
```
davisking, line 628: This is making a copy.  Why not return a reference to this?

davisking (commented):
Seems generally reasonable.  I like the tests too :)

dlib/matrix/matrix_utilities.h
pfeatherstone: I'll try using ```real``` and ```real_t```. But i'm wondering if i'm gonna get name conflicts with other parts of dlib

dlib/matrix/matrix_fft.h
pfeatherstone: ```ifft``` might be confusing. Both a complex inverse FFT and real inverse FFT require complex input...

pfeatherstone: @davisking Don't you think using ```fftr``` and ```ifftr``` make it clear to the user what the fft is actually doing in this case. It also means we can check input type, shape, etc. A lot of libraries do this, like numpy, torch, etc.

dlib/matrix/kiss_fft.h
pfeatherstone: If i return a reference, then another thread comes in and creates a new plan, then the reference to the first object may potentially be incorrect no? Or do unordered_maps not reallocate memory when there is an insertion?

dlib/matrix/matrix_utilities.h
pfeatherstone: I also quite like ```remove_complex``` as it follows the same theme as ```remove_const```, ```remove_cv```, ```remove_reference```, ```remove_pointer```, etc in ```<type_traits>```

dlib/matrix/kiss_fft.h
pfeatherstone: i've never used ```thread_local``` before. I would need to read up on it. I could do without the mutex if i used ```thread_local``` ??

dlib/matrix/kiss_fft.h
pfeatherstone: According to cppreference : ```References and pointers to either key or data stored in the container are only invalidated by erasing that element, even when the corresponding iterator is invalidated.```. So it's ok to return references?

dlib/matrix/kiss_fft.h
pfeatherstone: Ooof. Not sure i like ```thread_local```. @davisking is it ok to leave the mutex? I'm a bit nervous using ```thread_local```. Maybe using that could be another PR which does some performance measurements with and without it. I imagine there will be other PRs relating to FFTs in the near future that add SIMD optimisations for example. So it's not necessary to cram everything into this 1 PR.

dlib/matrix/mkl_fft.h
pfeatherstone: Hmm, ```std::array``` is fixed size...  It means i can't do stuff like ```dims.erase(std::remove(dims.begin(), dims.end(), 1), dims.end());```

dlib/matrix/matrix_fft.h
pfeatherstone: > `ifft` might be confusing. Both a complex inverse FFT and real inverse FFT require complex input...

Let me clarify: a forward real FFT is r2c, while a backward real FFT is c2r. A normal FFT is c2c. So both c2r and c2c take complex input. So just having ```ifft``` would be ambiguous.

dlib/matrix/kiss_fft.h
pfeatherstone: Also, if you were using ```dlib::parallel_for``` for example, with a fresh ```dlib::thread_pool```, and each thread was doing some FFTs, and you were using ```thread_local```, you would have to recreate the plans every time. I feel like using mutex should be better. But again, i could be proven wrong in a future PR that tests this more thoroughly.

pfeatherstone:
@davisking Silly question. I don't know much about travis and appveyor, but is there a way to build the tests with MKL? At the moment all the tests pass even though i haven't done the MKL wrappers for ```fftr``` and ```ifftr```

dlib/matrix/kiss_fft.h
pfeatherstone: Or could use a spinlock instead.

dlib/matrix/kiss_fft.h
davisking: Right, so long as you don't erase the element it's fine to return a reference.  And this copy is expensive.  It's copying a bunch of stuff and doing memory allocations and all that.

dlib/matrix/kiss_fft.h
davisking: A mutex is fine. `thread_local` gives you the potential to reduce contention among many threads though.  But you should measure to see what is best.  In any case, using a mutex here is fine.

This code could also be made a bit easier to follow by using a single templated function that returned a reference to the plan.  Like:
```
template <typename plan_type>
plan_type& get_plan(const plan_key& key)
{
    static std::mutex m;
    static std::unordered_map<plan_key, plan_type, hasher> plans;
    ...
}
```

dlib/matrix/kiss_fft.h
```diff
@@ -0,0 +1,733 @@
...
+                freqdata[k]         = half * (f1k + tw);
+                freqdata[nfft_h-k]  = half * std::conj(f1k - tw);
+            }
+        }
+
+        template<typename T>
+        void kiss_fftri(const kiss_fftr_state<T>& plan, const std::complex<T>* freqdata, T* timedata)
+        {
+            if (!plan.substate.inverse)
+                throw std::runtime_error("bad fftr plan : need an inverse plan. This is a forward plan");
```
davisking, line 519: This kind of thing should be a `DLIB_ASSERT`.

dlib/matrix/kiss_fft.h
```diff
@@ -0,0 +1,733 @@
...
+        void kiss_fftr(const kiss_fftr_state<T>& plan, const T* timedata, std::complex<T>* freqdata)
+        {
+            if (plan.substate.inverse)
+                throw std::runtime_error("bad fftr plan : need a forward plan. This is an inverse plan");
+
+            const int nfft_h = plan.substate.nfft; //recall that the FFT size is actually half the original requested FFT size, i.e. the size of timedata
+
+            /*perform the parallel fft of two real signals packed in real,imag*/
+            std::vector<std::complex<T>> tmpbuf(nfft_h);
+            kiss_fft_stride(plan.substate, (const std::complex<T>*)timedata, &tmpbuf[0], 1);
```
davisking, line 487: Nothing requires timedata to have a number of elements that is a multiple of two.  However, `(const std::complex<T>*)timedata` will fail when the number of elements pointed to by `timedata` is not a multiple of two. 

It would also be better to use `reinterpret_cast` for this rather than a c-style cast to make it clear that this is doing that kind of strong casting.

dlib/matrix/matrix_utilities.h
davisking: There is a `real`, so you can't use that name.  But there is no `real_t`.  But sure, `remove_complex` is nice for the reason you suggested .  But it's still better to have a `remove_const_t` to reduce the client boilerplate.

dlib/matrix/matrix_fft.h
```diff
@@ -9,837 +9,142 @@
...
+    template <typename EXP>
+    matrix<typename remove_complex<typename EXP::type>::type> ifftr (const matrix_exp<EXP>& data)
     {
-        call_mkl_fft_inplace(data, true);
+        // You have to give a complex matrix
+        COMPILE_TIME_ASSERT(is_complex<typename EXP::type>::value);        
+        matrix<typename remove_complex<typename EXP::type>::type> out;
+        if (data.size() == 0)
+            return out;
+        matrix<typename EXP::type> in = data;
```
davisking, line 137: This copy is not needd when the input type is already this type of matrix.  So there should be some overload or something to avoid this when it's not needed.

dlib/matrix/kiss_fft.h
```diff
@@ -0,0 +1,733 @@
...
+             * The sum of tdc.r and tdc.i is the sum of the input time sequence.
+             *      yielding DC of input time sequence
+             * The difference of tdc.r - tdc.i is the sum of the input (dot product) [1,-1,1,-1...
+             *      yielding Nyquist bin of input time sequence
+             */
+
+            freqdata[0]         = std::complex<T>(tmpbuf[0].real() + tmpbuf[0].imag(), 0);
+            freqdata[nfft_h]    = std::complex<T>(tmpbuf[0].real() - tmpbuf[0].imag(), 0);
+
+            static constexpr T half = 0.5;
```
davisking, line 501: Remove `static`.

dlib/matrix/matrix_fft.h
davisking: Sorry, I wrote that very unclearly.  Yes, the inverse fft is going to have complex inputs.  But the forward fft can take real inputs just fine.  There isn't any need to make the user write `ffti(whatever)` when we know the type of `whatever`.  There shouldn't be anything to check, as we do the checking by having the appropriate overload.

Then `ifft()` is always a function that maps from a complex input to a complex output.  Someone can just do ```
matrix<double> m = real(ifft(some_complex_matrix));
```
if they want just the real part.  It's possible to make `real(ifft(some_complex_matrix))` map to a different inverse fft function than just `ifft(some_complex_matrix)`.  But it is simpler to write `ifftr` and just document that it's logically equivalent to `real(ifft())`, so it's got that going for it and is maybe fine.  But `fftr` has no such complications.

File: dlib/matrix/mkl_fft.h (hunk 5 above)
Comment (davisking) [reply dlib/matrix/mkl_fft.h]: Memory allocations are expensive.  They do something like lock a global mutex, make system calls, and all that.  If you are going to all the trouble to make the ffts faster as you are it's almost certainly avoiding memory allocations as well.

File: dlib/test/rand.cpp [hunk 15]
```diff
@@ -432,7 +432,7 @@ namespace
         print_spinner();
         dlib::rand rnd(0);
         
-        const size_t N = 1024*1024*128;
+        const size_t N = 1024*1024*16;
```
davisking, line 435: These changes don't seem to be needed in this PR

dlib/test/fft.cpp
```diff
@@ -171,6 +359,14 @@ namespace
...
             test_random_real_ffts();
+            test_linearity_real<float>();
+            test_linearity_real<double>();
+            test_linearity_complex<float>();
+            test_linearity_complex<double>();
+            test_kronecker_delta_impulse_response<float>();
+            test_kronecker_delta_impulse_response<double>();
+            test_time_shift<float>();
+            test_time_shift<double>();
         }
```
davisking, line 370: Be sure to test FFTs that have sizes that are not even.  Like check ffts of size 1, 3, 123, and other oddball sizes since this new stuff is supposed to support ffts of any size.

davisking (changes requested):

davisking:
> @davisking Silly question. I don't know much about travis and appveyor, but is there a way to build the tests with MKL? At the moment all the tests pass even though i haven't done the MKL wrappers for `fftr` and `ifftr`

Not that I know of.  I assume you are testing it locally with the Intel MKL though?

dlib/matrix/kiss_fft.h
pfeatherstone: I hadn't thought of doing it that way. Very nice. Makes it a bit more functional.

dlib/matrix/mkl_fft.h
pfeatherstone: ok cool. Will have a think. Presumably then we use std::array with a templated size? I imagine we can still pass initialiser lists as arguments?

dlib/matrix/matrix_fft.h
pfeatherstone: ok i will go with having ```fft``` which can be complex to complex or real to complex, ```ifft``` which is complex to complex only and ```ifftr``` which is complex to real. That's my personal preference. And the API is then crystal clear as to what is happening

pfeatherstone: you make a good point that there should be a unit test that checks ```real(ifft(x)) == ifftr(x)```

dlib/matrix/kiss_fft.h
pfeatherstone: yep

dlib/matrix/kiss_fft.h
pfeatherstone: ok will add a requirement that ```timedata``` needs to have size that is a multiple of 2

dlib/matrix/matrix_fft.h
pfeatherstone: yeah, i was going to add some overloads for ```dlib::matrix``` and ```std::vector```. For ```std::vector```, i really don't want to do unecessary copies of the form ```std::vector - > dlib::matrix_exp -> dlib::matrix```

pfeatherstone: Can you think of any other useful overloads?

pfeatherstone: Also, what do you think the return type of the overload for ```std::vector``` should be? ```std::vector``` or ```dlib::matrix``` ?

dlib/matrix/kiss_fft.h
pfeatherstone: yep

dlib/test/rand.cpp
pfeatherstone: Yes that's true. I knew you would spot that. I realised that test takes ages... Using 1024*1024*16 was sufficient. I will undo that commit. Maybe that can be another tinee tiny PR.

dlib/test/fft.cpp
pfeatherstone: Yep. I think in general there should be more unit tests. Like the DFTs of standard window functions could be computed and checked analytically. And then add for-loops around them for different FFT sizes.

pfeatherstone:
> > @davisking Silly question. I don't know much about travis and appveyor, but is there a way to build the tests with MKL? At the moment all the tests pass even though i haven't done the MKL wrappers for `fftr` and `ifftr`
> 
> Not that I know of. I assume you are testing it locally with the Intel MKL though?

I haven't done the MKL stuff yet. I wanted to test the MKL stuff standalone in a Netbeans project (i'm not that good with cmake yet) but was struggling to link to all the MKL libraries manually. So will go back to just testing with the dlib build. For real FFTs, I'm a bit confused by the CCE and CCS format + all the other configurations that MKL supports for real FFTs. Will get there when i have time to work on this again.

pfeatherstone:
Hmm, test failed with :
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

dlib/matrix/mkl_fft.h
pfeatherstone: std::array doesn't work when passing initialiser lists. C-style arrays are fine though. I'll use that.

pfeatherstone: The best thing would be a container that has the small size optimisation, a bit like what std::vector and std::string had in C++98 i think. So that i could do something like:
```dims.erase(std::remove(dims.begin(), dims.end(), 1), dims.end());```

Comment (pfeatherstone) [reply dlib/matrix/mkl_fft.h]: Actually, I'm gonna use std::vector for now because that's what ```plan_key``` uses. All FFT plans use the same ```plan_key``` type and i don't want to template it for different sizes of ```dims```.  I think something like [SmallVector](https://llvm.org/doxygen/SmallVector_8h_source.html) would be ideal and tick all the boxes. That could be another PR adding something like SmallVector to dlib.

Comment (pfeatherstone) [reply dlib/matrix/mkl_fft.h]: In the grand scheme of things, i don't think doing a small allocation for ```dims``` is going to be the bottleneck when computing FFTs.

Comment (davisking) [reply dlib/matrix/mkl_fft.h]: Depends on the fft size. Someone doing billions of small 1D ffts is going to care.

Comment (pfeatherstone) [reply dlib/matrix/mkl_fft.h]: For now, do you mind if i focus on getting everything working including MKL bindings, ```fftr``` getting merged into ```fft``` and adding all the overloads for normal ```matrix``` type. Little details i will do at the end, or ideally in another PR which adds performance enhancements here and there.

Comment (davisking) [reply dlib/matrix/mkl_fft.h]: Yeah that’s fine :)

**Comment by pfeatherstone** [comment]
@davisking If you have time, do you mind reviewing the overloads I added ? Also i amended find_blas.cmake to check for the oneAPI install directory. Hope that's ok.

**Comment by pfeatherstone** [comment]
@davisking I think i'm gonna have to recommit that change to the ```rand``` tests. The continuous-integration tests fail with error:
```
[100%] Linking CXX executable dtest
[100%] Built target dtest
+./dtest --runall
Running test_rand - 
The job exceeded the maximum time limit for jobs, and has been terminated.
```

**Comment by davisking** [comment]
> @davisking I think i'm gonna have to recommit that change to the `rand` tests. The continuous-integration tests fail with error:
> 
> ```
> [100%] Linking CXX executable dtest
> [100%] Built target dtest
> +./dtest --runall
> Running test_rand - 
> The job exceeded the maximum time limit for jobs, and has been terminated.
> ```

Huh, yeah I didn't realize how long these tests were taking.  I'm going to push a change that makes them run much faster.

**Comment by pfeatherstone** [comment]
@davisking Ok, so i've replaced std::vector<long> with a new type ```fft_size```. Do you mind doing a quick review and checking you're happy with that? It wraps std::array so there aren't any dynamic memory allocations. Hopefully it's a tad faster.

**Comment by pfeatherstone** [comment]
@davisking I think this is nearly finished. Not sure if we want to do caching with MKL descriptors? In theory why not, but i think that could be in a different PR. The purpose of this PR is to generalize the default backend.

**Comment by pfeatherstone** [comment]
Weird, an SVM test failed.

**Comment by davisking** [comment]
> @davisking I think this is nearly finished. Not sure if we want to do caching with MKL descriptors? In theory why not, but i think that could be in a different PR. The purpose of this PR is to generalize the default backend.

Yeah, don't worry about that for now.

File: dlib/matrix/fft_size.h [hunk 17]
```diff
@@ -0,0 +1,97 @@
...
+
+#include <array>
+#include <algorithm>
+#include <numeric>
+#include "../assert.h"
+#include "../hash.h"
+
+namespace dlib
+{
+    class fft_size
```
davisking, line 12: I find this class confusing to reason about.  Like it's not clear what it represents at any moment and what the valid states are when I look at the code that uses it.  I mean, I can follow what's happening, but it's subtle.

What I meant was something like this:
```
class fft_size
{
public:
    explicit fft_size(long x_dim) : x_(x_dim){}
    /*! 
        ensures
            - x() == x_dim
            - y() == 1
    !*/
    fft_size(long x_dim, long y_dim) : x_(x_dim), y_(y_dim) {}
    /*! 
        ensures
            - x() == x_dim
            - y() == y_dim
    !*/

    // get size of dimensions.
    long x() const { return x_; }
    long y() const { return y_; }

    long num_elements() const { return x_*y_; }

    bool operator==(const fft_size& other) const;
    uint32 hash() const;
private:
    
    long x_ = 1;
    long y_ = 1;
};
```

That object is really simple and immutable.  It always represents the same thing.  There isn't optionality about being 1D sometimes and 2D other times.  A 1D FFT just means y==1.  If 3D ffts are needed later someone can add a z (and even a w) dimension.

You could add global helper functions for the higher level stuff you are doing with it.  For example, you might say ```
if (is_1d(some_fft_size))
    // do whatever
```
and that would make the code that uses it more readable.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: Ok. Honestly i don't mind. If I need to go 3D and beyond at some point, I can always submit another PR. Whatever you think is best.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: Though ```x_``` and ```y_``` may be confusing... By ```x_``` do you mean the fft size along the column dimension and by ```y_``` do you mean the fft size along the row dimension? In which case, the order of FFTs is ```y_``` then ```x_```.  Lets go with that, but i find this class more confusing. Whereas a simple vector-like class that describes arbitrary dimensions seems more appropriate in the KISS and MKL backends. Indeed the fft sizes are dims[0], dims[1], ..., dims[N-1]. The interpretation of what the dimensions mean seems more appropriately dealt with in ```matrix_fft.h```. But that's my opinion. I'll go with what you want since dlib is your repo at the end of the day.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: In other words, the ordering of the dimensions is not at all obvious now.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: And going with the ordering of the alphabet is still not clear because, typically, matrix layout is row major, so ordering goes ```y``` then ```x```.

Comment (davisking) [reply dlib/matrix/fft_size.h:12]: x is the first dimension, y the next.  But yeah, I see what you mean.  `dims[0]` and `dims[1]` is better. The main thing is to make `fft_size` a simple immutable container that always has the same number of integers in it.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: So what do you suggest? i had ```std::array<long,5>``` do cater for up to 5. But i could have ```std::array<long,2>``` and not allow it to be empty.

Comment (davisking) [reply dlib/matrix/fft_size.h:12]: Unless it simplifies the code that uses it names like `dim0` and `dim1` would be best, since code that does `dim2` is a compile error while `dim[2]` is not.  It doesn't look like there needs to be runtime indexing into these things.  If that's so then `dim0` and `dim1` seem good.

To your question, put 2 longs in there since that's what's actually used right now.  If it needs to be more later it can be changed.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: My problem with this is that we are simplifying backend code when there is no need to. I understand making the code in ```matrix_fft.h``` as full-proof as possible but i don't see any reason to in the implementations details. The user should have no visibility of this. If we go down the rabbit hole of specialising stuff, we could go much further and replace the ND case for 2D only. Then the 2D plan could have 2 1D plans instead of a std::vector of 1D plans. The real ND plan currently has 1 real 1D plan and 1 (complex) ND plan. We could change that to 1 complex 1D plan and 1 real 1D plan. And so forth... This would be undoing all the hard work of supporting ND FFTs. I know what you suggest is simply refactoring ```fft_size``` object. But it makes no sense to simplify a tiny skin depth layer of the backend. Either simplify everything or leave it general as it is now.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: The original point of ```fft_size``` was to remove dynamic allocations, which was achieved...

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: For me, ```fft_size``` is just a much simpler version of boost's ```static_vector``` to replace ```std::vector``` with something that doesn't perform dynamic memory allocations. And that's it. It's just a container for FFT dimensions (with a hard upper bound of 5)

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: All of this to say, i find the proposed amendment a bit itchy. It's like, "Why am I restricting all the KISS FFT stuff to 1d and 2d but all the rest of the code seems to support, or is set up to support, ND case. That's weird... Something has gone wrong somewhere along the way...."

Comment (davisking) [reply dlib/matrix/fft_size.h:12]: Yeah, don't undo work in kissfft.  What's in here is just hard for me to follow.  Like it's not clear what exactly is supported and supposed to work and what isn't at first.  Like it's not obvious what kinds of `fft_size` objects are valid and meaningful and which aren't. E.g. Are size 0 ones generally ok?  Setting it up so the full power of kissfft is part of the kissfft interface (so the kissfft functions, the plans, and the `fft_size` object) is preferable though.  So making the number of dimensions in `fft_size` a runtime option and letting it be anything is fine.  But make the interface clear.  Just adding comments that say what these interfaces claim to support would maybe be all that's really required.  Although making `fft_size` immutable would be a good thing regardless.

I guess I'm saying that it's hard to follow sine most of these things in the diff have no comments.  So it's unclear what the exact contracts for some of these functions are.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: By making ```fft_size``` immutable, do you mean it can't be copyable or moveable, or do you mean all of its member functions must be const ?

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: For example, i can make a function ```remove_ones()``` a member function that isn't const, or i can make a static function that takes a ```const fft_size&``` and returns a ```fft_size```. The latter is just functional style programming. The same applies to a function like ```pop_back()```. I can make it a non-const member function or create a functional style equivalent.

Comment (davisking) [reply dlib/matrix/fft_size.h:12]: I mean all the members are const.

Comment (davisking) [reply dlib/matrix/fft_size.h:12]: Don't make any class static methods.  Just make global methods.

Comment (pfeatherstone) [reply dlib/matrix/fft_size.h:12]: Cool

**Comment by pfeatherstone** [comment]
@davisking sorry for the delay. I think this is ready for another round of light roasting. Also, happy holidays

File: dlib/matrix/fft_size.h [hunk 18]
```diff
@@ -0,0 +1,215 @@
...
+        }
+        /*!
+            requires
+                - size.size() > 0
+            ensures
+                - returns a copy of size with the last element removed.
+        !*/
+        
+        /*global helpers*/
+    };
```
davisking, line 212: Don't use friend functions for this.  The utility of having these things as global functions is that they don't complicate the interface of `fft_size`.  Classes should be minimal and complete. Then higher level functions can be layered on top of them.  This way reasoning about the class invariants is easy as little code is in scope of the class's member variables.  More generally, classes should be open for extension but closed for modification.  So the interface should be sufficient to allow functions like this to be written as non-friend globals.

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+        }
+        /*!
+            ensures
+                - returns true if two fft_size objects have same size and same elements.
+        !*/
+        
+        bool is_valid() const
+        {
+            return _size > 0 && std::find_if(begin(), end(), [](long dim) {return dim <= 0;}) == end();
+        }
```
davisking, line 152: This function isn't needed.  The condition can be checked in the constructor.  Then client code can assert `size() != 0`.  Having `is_valid()` and `size()` gives two ways for clients to do the same thing.  But there should be one unambiguous way.

davisking: Also, negative dims can never happen, because the constructor's preconditions forbid it.  The logic here is making negative dimensions a proper state of this class.  But such `fft_size` objects should never exist at any point.

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+                - dlib::fatal_error if requirements aren't satisfied.
+        !*/
+        
+        size_t size() const
+        {
+            return _size;
+        }
+        /*!
+            ensures
+                - returns the number of elements in *this
```
davisking, line 99: Returns the number of dimensions.

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+        {
+            DLIB_ASSERT(index < _size, "index " << index << " out of range [0," << _size << ")");
+            return _dims[index];
+        }
+        /*!
+            requires
+                - size() > 0
+                - index < size()
+            ensures
+                - returns a const reference to the element at position index
```
davisking, line 112: Say dimension instead of element since otherwise it's ambiguous with respect to text that talks about fft sizes.  E.g. `dimprod` talks about number of elements too but it means something entirely different.

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+        !*/
+        
+        long dimprod() const
+        {
+            DLIB_ASSERT(_size > 0, "object is empty");
+            return std::accumulate(begin(), end(), 1, std::multiplies<long>());
+        }
+        /*!
+            requires
+                - size() > 0
```
davisking, line 134: This should probably just return 0 when `size()==0`

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+            return _dims[_size-1];
+        }
+        /*!
+            requires
+                - size() > 0
+            ensures
+                - returns a const reference to (*this)[size()-1]
+        !*/
+        
+        long dimprod() const
```
davisking, line 127: It would be more clear if `size()` and `dimprod()` were renamed to `num_dims()` and `num_elements()` or something like that.

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+                  of elements.
+        !*/
+        
+        bool operator==(const fft_size& other) const
+        {
+            return this->_size == other._size && std::equal(begin(), end(), other.begin());
+        }
+        /*!
+            ensures
+                - returns true if two fft_size objects have same size and same elements.
```
davisking, line 146: I.e. returns true if they have identical states.

dlib/matrix/fft_size.h
```diff
@@ -0,0 +1,215 @@
...
+                const auto newend = std::remove(newsize.begin(), newsize.end(), 1);
+                const long nremoved = std::distance(newend, newsize.end());
+                newsize._size -= nremoved;
+            }
+            return newsize;
+        }
+        /*!
+            ensures
+                - returns a copy of size with all but 1 elements equal to 1 removed
+        !*/
```
davisking, line 195: The dlib style is to put the comment above the code, so like this
```
void foo()
/*!
   comments
!*/
{
...
}
```

Also, `returns a copy of size with all but 1 elements equal to 1 removed` is ungrammatical.   How about something like: removes dimensions with values equal to 1, yielding a new size with the same dimprod() but fewer dimensions.

dlib/test/rand.cpp
```diff
@@ -413,7 +413,7 @@ namespace
         print_spinner();
         dlib::rand rnd(0);
         
-        const size_t N = 1024*1024*4;
+        const size_t N = 1024*1024*16;
```
davisking, line 416: Does this need to be increased?

dlib/matrix/mkl_fft.h
```diff
@@ -0,0 +1,210 @@
...
+#define DLIB_DFTI_CHECK_STATUS(s) \
+    if((s) != 0 && !DftiErrorClass((s), DFTI_NO_ERROR)) \
+    { \
+        throw dlib::error(DftiErrorMessage((s))); \
+    }
+
+namespace dlib
+{
+    template<typename T>
+    void mkl_fft(const fft_size& dims, const std::complex<T>* in, std::complex<T>* out, bool is_inverse)
```
davisking, line 17: Add docs for these `mkl_fft` and `mkl_fftr` functions too.

dlib/matrix/fft_size.h
pfeatherstone: okidok

dlib/matrix/matrix_fft.h
```diff
@@ -9,836 +9,135 @@
...
-#endif // DLIB_USE_FFTW
+#include "mkl_fft.h"
+#define impl_fft    mkl_fft
+#define impl_fftr   mkl_fftr
+#define impl_fftri  mkl_fftri
+#else
+#include "kiss_fft.h"
+#define impl_fft    kiss_fft
+#define impl_fftr   kiss_fftr
+#define impl_fftri  kiss_fftri
```
davisking, line 20: Don't use `#defines`.  It makes the code harder to follow and might clash with symbols in other people's code.  It's fine to just give these functions the same names in both the mkl_fft.h and kiss_fft.h files.  That's what most dlib code that does this kind of switching does.  Then this kind of name switching here is not needed.

dlib/matrix/matrix_fft.h
```diff
@@ -9,836 +9,135 @@
...
-        else
-            return count_bits(value) == 1;
+        return nc == 0 ? 0 : nc/2+1;
     }
-
-// ----------------------------------------------------------------------------------------
-
-    namespace impl
+    
+    constexpr long ifftr_nc_size(long nc)
```
davisking, line 32: Add some docs for what these are about.

dlib/matrix/matrix_fft.h
```diff
@@ -9,836 +9,135 @@
...
-            << "\n\t The number of rows and columns must be powers of two."
-            << "\n\t data.nr(): "<< data.nr()
-            << "\n\t data.nc(): "<< data.nc()
-            << "\n\t is_power_of_two(data.nr()): " << is_power_of_two(data.nr())
-            << "\n\t is_power_of_two(data.nc()): " << is_power_of_two(data.nc())
-            );
-
-        impl::twiddles<T> cs;
-        impl::fft1d_inplace(data, false, cs);
+        /*complex FFT*/
```
davisking, line 42: Use C++ style comments for this kind of thing: `//complex FFT`

dlib/matrix/fft_size.h
pfeatherstone: Yep. another good point

dlib/matrix/fft_size.h
pfeatherstone: yep

dlib/matrix/fft_size.h
pfeatherstone: yep

dlib/matrix/fft_size.h
pfeatherstone: yep

dlib/matrix/fft_size.h
pfeatherstone: okidok

dlib/test/rand.cpp
pfeatherstone: I thought i reverted that back to what it was before. To answer your question, if it passes the test then no. Unfortunately, measuring kurtosis and skewness requires a lot of samples. Maybe there could be other, equally meaningful yet faster unit tests for rand. But that should be dealt with in another PR maybe.

dlib/matrix/matrix_fft.h
pfeatherstone: So what names should i give everyone? ```impl_fft```, ```impl_fftr``` and ```impl_fftri``` ?

dlib/matrix/matrix_fft.h
pfeatherstone: I agree. All libraries i've seen use ```fftr```. So users doing FFTs will know exactly what ```fftr``` is doing.

davisking:
This all looks pretty good overall.  Just some relatively minor stuff.  Run a little benchmark too to check that this is faster or at least the same as the current code too.  Looks like it ought to be though.

pfeatherstone:
I added some unit tests that compared the kiss backend to mkl backend, provided mkl is available. So naming the functions the same won't allow for this test

dlib/matrix/matrix_fft.h
davisking: Yeah `impl_*` is fine.  Although probably better to put them in a namespace, like `impl::fft()`.

pfeatherstone:
I wanted a test that checked that all backends agreed on outputs. That's how i found the bug with mkl_fftr and mkl_fftri having incorrect strides. So this kind of test is important i think.

dlib/matrix/matrix_fft.h
davisking: Yeah.  Me too.  It's just always bothered me that it wasn't captured in the type system :|

But me being bothered doesn't really matter.  What's least confusing to users is always right.

davisking:
> I added some unit tests that compared the kiss backend to mkl backend, provided mkl is available. So naming the functions the same won't allow for this test

Oh right.  Yeah, well, just don't use #define :) .  Put #ifdef blocks inside `dlib::fft()` instead or something like that then.

dlib/matrix/matrix_fft.h
pfeatherstone: In the end, i did capture it in the type system right? I just used operator overloading. So functionality was there at least.

davisking:
> I wanted a test that checked that all backends agreed on outputs. That's how i found the bug with mkl_fftr and mkl_fftri having incorrect strides. So this kind of test is important i think.

Yeah, definitely.

pfeatherstone:
Right, will correct everything tomorrow. If i spend more time on this today, my partner won't be in a pleasant state.

dlib/matrix/matrix_fft.h
davisking: I mean that the returned type from `fft(some_real_matrix)` would encode the fact that it holds a packed representation.  So that then `ifftr()` wouldn't be required.  You would just do `ifft()` and it would know it was a packed representation.  It's complicated though because someone is going to do stuff to the complex packed matrix, like transform it somehow and the fact that it's packed would need to be propagated all the way through all that too.  I can think of a bunch of ways to go at that problem in the typesystem. I'm not sure what would actually not just end up irritating users though.  

The simplest thing I can think of is to just have a new complex type like `dlib::packed_complex<T>` or something like that.  Then the output of a real fft would be something like `dlib::matrix<dlib::packed_complex<double>>` and `ifft(dlib::matrix<dlib::packed_complex<double>>)` would know to do the packed version.  But now there is this other complex type which is definitely going to be irritating to users.

Anyway, none of the ways to do it are worth it and probably all are confusing to users.

pfeatherstone:
Struggling to re-implement ```squeeze_ones``` without it being friend specified, and therefore without the non-const versions of ```begin()``` and ```end()``` available. I've removed them otherwise ```fft_size``` wouldn't be truly immutable.

pfeatherstone:
I might put the code from ```squeeze_ones``` inside the constructor of ```fft_size```, and update the contract so users know exactly what's going on. I.e. the constructor will remove values equal to 1 such that the product of all dimensions stays unchanged. If the initializer_list contains a single element and it is equal to 1, then nothing is removed.

dlib/matrix/mkl_fft.h
pfeatherstone: yep

davisking:
> Struggling to re-implement `squeeze_ones` without it being friend specified, and therefore without the non-const versions of `begin()` and `end()` available. I've removed them otherwise `fft_size` wouldn't be truly immutable.

Add a constructor that takes a std::array. Then you can make whatever array and pass that.

pfeatherstone:
> > Struggling to re-implement `squeeze_ones` without it being friend specified, and therefore without the non-const versions of `begin()` and `end()` available. I've removed them otherwise `fft_size` wouldn't be truly immutable.
> 
> Add a constructor that takes a std::array. Then you can make whatever array and pass that.

I just added the functionality of ```squeeze_ones``` in the constructor

davisking:
> > > Struggling to re-implement `squeeze_ones` without it being friend specified, and therefore without the non-const versions of `begin()` and `end()` available. I've removed them otherwise `fft_size` wouldn't be truly immutable.
> > 
> > 
> > Add a constructor that takes a std::array. Then you can make whatever array and pass that.
> 
> I just added the functionality of `squeeze_ones` in the constructor

Isn’t that kinda weird though?  I need to look at the code again, maybe it’s not. But it seems kinda funky.

pfeatherstone:
MSVC complaining about casting ```const long*``` to ```const_iterator```. Argh

pfeatherstone:
> > > > Struggling to re-implement `squeeze_ones` without it being friend specified, and therefore without the non-const versions of `begin()` and `end()` available. I've removed them otherwise `fft_size` wouldn't be truly immutable.
> > > 
> > > 
> > > Add a constructor that takes a std::array. Then you can make whatever array and pass that.
> > 
> > 
> > I just added the functionality of `squeeze_ones` in the constructor
> 
> Isn’t that kinda weird though? I need to look at the code again, maybe it’s not. But it seems kinda funky.

it is a bit, but the contract stipulates that it does that. And fft_size doesn't even entertain the fact that it resembles an STL container. So it doesn't have to behave like one

davisking:
> > > > > Struggling to re-implement `squeeze_ones` without it being friend specified, and therefore without the non-const versions of `begin()` and `end()` available. I've removed them otherwise `fft_size` wouldn't be truly immutable.
> > > > 
> > > > 
> > > > Add a constructor that takes a std::array. Then you can make whatever array and pass that.
> > > 
> > > 
> > > I just added the functionality of `squeeze_ones` in the constructor
> > 
> > 
> > Isn’t that kinda weird though? I need to look at the code again, maybe it’s not. But it seems kinda funky.
> 
> it is a bit, but the contract stipulates that it does that. (or at least it guarantees that num_elements is not affected) And fft_size doesn't even entertain the fact that it resembles an STL container. So it doesn't have to behave like one

Right. But having such a constructor makes it possible to implement things like this squeeze function. This suggests that the interface isn’t really complete as there is something you think is important but couldn’t implement without such an constructor. 

Anyway. If it really just doesn’t make sense in the code to ever have a size {1,1} then it’s cool and to make it not a possible state. I’m on my phone and haven’t looked at the code yet.

pfeatherstone:
> > > > > > Struggling to re-implement `squeeze_ones` without it being friend specified, and therefore without the non-const versions of `begin()` and `end()` available. I've removed them otherwise `fft_size` wouldn't be truly immutable.
> > > > > 
> > > > > 
> > > > > Add a constructor that takes a std::array. Then you can make whatever array and pass that.
> > > > 
> > > > 
> > > > I just added the functionality of `squeeze_ones` in the constructor
> > > 
> > > 
> > > Isn’t that kinda weird though? I need to look at the code again, maybe it’s not. But it seems kinda funky.
> > 
> > 
> > it is a bit, but the contract stipulates that it does that. (or at least it guarantees that num_elements is not affected) And fft_size doesn't even entertain the fact that it resembles an STL container. So it doesn't have to behave like one
> 
> Right. But having such a constructor makes it possible to implement things like this squeeze function. This suggests that the interface isn’t really complete as there is something you think is important but couldn’t implement without such an constructor.
> 
> Anyway. If it really just doesn’t make sense in the code to ever have a size {1,1} then it’s cool and to make it not a possible state. I’m on my phone and haven’t looked at the code yet.

I see your point. Having {1,1} is definitely a valid state. So i will re-implement ```squeeze_ones```. Then it is clear that it's kind of like an optimisation step used in the backend FFT code, whereby we remove "dummy" dimensions.

pfeatherstone:
Really don't like the fact visual studio compiler doesn't allow a cast from std::initializer_list<long>::const_iterator to std::array<long,5>::const_iterator. Oh well

pfeatherstone:
@davisking So i think the business with ```squeeze_ones``` is resolved. i.e. constructors don't implement ```squeeze_ones```. Backend FFT code calls that explicitly. So there shouldn't be any confusions as to what is valid state inside ```fft_size```.

pfeatherstone:
@davisking this is ready for another (hopefully final) review

davisking (changes requested):
Lookings pretty good. 

Aside from my other comments, can `fft_next_fast_size` be removed from the unit tests?  Like those tests that use it aren't testing the full range of valid input sizes.  It would be better if they checked them all.  Most of those tests are also wrapped in an outer loop that runs the whole block 10 times.  That's probably excessive.  We don't want travis CI to time out the testing jobs and this all adds up.

Also add more tests for `fftr()` and `ifftr()`.  They only appear in one place in the unit tests right now with a check for linearity.  A `test_against_saved_good_ffts()` but for real ffts would be good.

dlib/test/rand.cpp
davisking: Don't forget to revert this.

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -63,7 +131,97 @@ namespace dlib
...
+// ----------------------------------------------------------------------------------------
+    
+    template <typename EXP>
+    matrix<add_complex_t<typename EXP::type>> fftr (
+        const matrix_exp<EXP>& data
+    );  
+    /*!
+        requires
+            - data contains elements of type double, float, or long double.
+            - data.nc() is even
```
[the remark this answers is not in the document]
pfeatherstone: I agree, i will leave that for another PR. I anticipate more than 1 PR in the future relating to FFTs. Like SIMD optimisations, possibly some more overloads for std::vector as input for example, ... I'm certainly going to start using this a lot. Hopefully other people will find this useful if their target architecture doesn't support MKL.

dlib/matrix/kiss_fft.h
```diff
@@ -675,15 +699,27 @@ namespace dlib
      */
     template<typename T>
     void kiss_fftri(const fft_size& dims, const std::complex<T>* in, T* out)
```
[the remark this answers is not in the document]
pfeatherstone: yep

dlib/matrix/fft_size.h
```diff
@@ -119,97 +138,90 @@ namespace dlib
...
-        !*/
-        
-        friend inline fft_size pop_back(const fft_size& size)
+        else
         {
-            DLIB_ASSERT(size.size() > 0);
-            fft_size newsize = size;
-            newsize._size--;
-            return newsize;
+            std::array<long,5> tmp;
```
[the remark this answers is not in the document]
pfeatherstone: i've done one better i think. i have an alias for std::array<long,5> then i use that alias.

pfeatherstone:
@davisking For ```test_against_saved_good_ffts()```, what did you use as an authority? numpy ?

pfeatherstone:
Hmm, replacing ```nr = fft_next_fast_size(nr + 1)``` with ```nr++```, and same for ```nc``` in the tests, has dramatically increased the time it takes to run the tests, even when removing the ```iter``` for-loops. Oh well

davisking:
> @davisking For `test_against_saved_good_ffts()`, what did you use as an authority? numpy ? I could use MKL, but the dlib MKL wrappers are subject to testing too, so by definition, they cannot be used as an authority

I just grabbed it out of octave IIRC.

davisking:
> Hmm, replacing `nr = fft_next_fast_size(nr + 1)` with `nr++`, and same for `nc` in the tests, has dramatically increased the time it takes to run the tests, even when removing the `iter` for-loops. Oh well

How about doing fewer sizes then.  Like do sizes in the range 1 to 12 instead of 1 to 128.  It's fairly important that these tests not take a long time.

pfeatherstone:
@davisking Presumably, the contents of ```get_decoded_string()``` was machine generated, correct ? If so, how did you print it out? I can compress and base64 encode it easily enough. But when printing the contents to stdout or to file, lets say 80 characters per line, i get something like this:
```
const std::string str = "gREgHkVPSRDT5z8rk175NS624DOfWDQcPg2MOBWP6zKp5jxHZcD2U1VxamzdjaOSqfLISu"
"O5XxmI
wNbs2oCQ05Jnba2XSMe0/fI4D3HwzE2Py5EXfucAyO7xzaN+5OQO/vFwu9K7ocE"
"hrFnCIIrELaOE
F3Ocpy9CMf1bs5fhz2LHnfXO/hYboIhIiT7tDmny1OI/diOkdfSvEOBA"
"RrBx72j+beMHFoAuH4/L
/V76B7hRdlR2O9G6w7NcgpqkAMav7L6NshODMCnm3a+oCMOb9"
"k+x20CupDgjh5GrUk4TS98a+6vF
n3jhOLiW08ydsHJN7L15AKc1IFXITHLlHJzReNVI/v"
"lsJ/bfpahh/gJ1ToHm8OVsSQnBX2SBqv9Z
IxDea8Gbkjsxo2Ptvewv5MTm7d9K13T8ACo"
"IP8rhyhd/J7d1n7/qb7dwIquGMnd4ugbX2UdXGzhA
lQjtEbulmCpTfGvR2owFYhcViTmz"
"h1Q4nMYfT1JFAZ9bWIVMi//QdBKASkhcepD/FS8eCKcNDqzJ
4oVpYHvJNauFx2visid1/"
"uiRVfuI3hzOkVFQtxzNywqU/nJ4TgY6cGk+Fc74pXezjNQt653LDEQz
0m2Tzq+IjGymYR"
"KU+PqcQbxzO+7A47wg2wRO5bHCxI1DZ9rU45jrxkyzgYrmx6B8R4wXLJlqB2Jc
4LmVQXi"
"xuu7Y6ria6Dmdq91/P1QW5F/GPirFEhV+1k1FBg3gj+wDdWwVaEnFP/sSWdNfsu8ysBlS
"
"sroFquFysayDxtmxu4SA5u17bOhpRvWIkEm8MciQaTYNYvQcgfShlFSBrmdnbGGZyO4vTC"
...
```

Did you have to use ```dlib::ustring``` or something?

pfeatherstone:
> > @davisking For `test_against_saved_good_ffts()`, what did you use as an authority? numpy ? I could use MKL, but the dlib MKL wrappers are subject to testing too, so by definition, they cannot be used as an authority
> 
> I just grabbed it out of octave IIRC.

Cool, i just used numpy in the end. It was an absolute PAIN, getting python data, to dlib::matrix types, compressed and base64 encoded (those 2 bits were simple actually), then printed to a C++ file that can be included in ```fft.cpp```.

pfeatherstone:
Python code was:
```
def save_good_fftr():
    import numpy as np
    import msgpack
    np.random.seed(10000)

    ret = []

    x = np.random.randn(64,64)
    y = np.fft.rfftn(x)
    y2 = np.stack((y.real,y.imag),-1)
    ret.append([x.tolist(),y2.tolist()])

    x = np.random.randn(131,48)
    y = np.fft.rfftn(x)
    y2 = np.stack((y.real, y.imag), -1)
    ret.append([x.tolist(),y2.tolist()])

    with open("good_real_ffts.msgpack", "wb") as f:
        msgpack.pack(ret,f)
```

And C++ code to generate a header file was:
```
   dlib::matrix<double> from_msgpack(const std::vector<std::vector<double>>& x_v)
   {
    dlib::matrix<double> x(x_v.size(),x_v[0].size());
    
    int offset = 0;
    for (const auto& v : x_v)
    {
        assert(v.size() == x_v[0].size());
        std::copy(v.begin(), v.end(), x.begin() + offset);
        offset += v.size();
    }
    
    return x;
}

dlib::matrix<complex<double>> from_msgpack(const std::vector<std::vector<std::vector<double>>>& y_v)
{
    dlib::matrix<complex<double>> y(y_v.size(), y_v[0].size());
    
    int offset = 0;
    for (const auto& v : y_v)
    {
        assert(v.size() == y_v[0].size());
        for (const auto& x : v)
        {
            assert(x.size() == 2);
            *(y.begin() + offset++) = std::complex<double>(x[0],x[1]);
        }
    }
    
    return y;
}

void save_good_fftr()
{
    std::ifstream in("good_real_ffts.msgpack", std::ios::binary);
    assert(in.good());
    nlohmann::json j = nlohmann::json::from_msgpack(in);
    
    std::vector<std::vector<double>> x_v;
    std::vector<std::vector<std::vector<double>>> y_v;
    
    dlib::matrix<double> x1,x2;
    dlib::matrix<complex<double>> y1,y2;
    
    j[0][0].get_to(x_v);
    j[0][1].get_to(y_v);
    x1 = from_msgpack(x_v);
    y1 = from_msgpack(y_v);
    
    j[1][0].get_to(x_v);
    j[1][1].get_to(y_v);
    x2 = from_msgpack(x_v);
    y2 = from_msgpack(y_v);
    
    
    std::stringstream compress_in, compress_out, base64_out;
    dlib::serialize(compress_in) << x1 << y1 << x2 << y2;
    
    dlib::compress_stream::kernel_1ea compressor;
    compressor.compress(compress_in, compress_out);
    
    dlib::base64 base64_coder;
    base64_coder.encode(compress_out, base64_out);
    
    string out1 = base64_out.str().insert(0, "\"");
    out1.push_back('\"');
    const string out = std::regex_replace(out1, std::regex("\n"), "\"\"");

    std::ofstream fout("compressed_fftrs.txt");
    fout << "static const std::string GOOD_FFTRs_COMPRESSED_BASE64_ENCODED =";
    
    const int step = 78;
    for (int i = 0 ; i < out.size() ; i += step)
    {
        fout << endl;
        const int end = min<int>(i + step, out.size());
        for (int j = i ; j < end ; j++)
            fout << out[j];
    }
    fout << ";" << endl;
}
```

there probably was a much simpler way of doing all this...

pfeatherstone:
> > Hmm, replacing `nr = fft_next_fast_size(nr + 1)` with `nr++`, and same for `nc` in the tests, has dramatically increased the time it takes to run the tests, even when removing the `iter` for-loops. Oh well
> 
> How about doing fewer sizes then. Like do sizes in the range 1 to 12 instead of 1 to 128. It's fairly important that these tests not take a long time.

If you look at the last commit, i've replaced large primes like 1021 and 1031 to smaller primes like 131. That helped a lot with speed. And i now check for all states up to 64, not 128. that helps a lot. And ```test_against_saved_good_fftrs()``` works as expected.

pfeatherstone:
@davisking Sorry for the last commits being such a mess. But the tests all run under 30s on my machine now, and all successful. So hopefully we are all good now.

pfeatherstone:
```fatal error C1091: compiler limit: string exceeds 65535 bytes in length``` 
MSVC is such a pile of shit.

**Comment by pfeatherstone** [comment]
Ok and i've added another test for ```fftr```. It's only for 1D stuff though. But still good to have.

**Comment by davisking** [comment]
> > > @davisking For `test_against_saved_good_ffts()`, what did you use as an authority? numpy ? I could use MKL, but the dlib MKL wrappers are subject to testing too, so by definition, they cannot be used as an authority
> > 
> > 
> > I just grabbed it out of octave IIRC.
> 
> Cool, i just used numpy in the end. It was an absolute PAIN, getting python data, to dlib::matrix types, compressed and base64 encoded (those 2 bits were simple actually), then printed to a C++ file that can be included in `fft.cpp`.

Ah, sorry, I thought you knew about the http://dlib.net/file_to_code_ex.cpp.html example.  It does all that for you :|

**Comment by davisking** [comment]
> `fatal error C1091: compiler limit: string exceeds 65535 bytes in length`
> MSVC is such a pile of shit.

Ha, yeah.  Use http://dlib.net/file_to_code_ex.cpp.html.  it will format it all so it works on any compiler (where any means works around MSVC nonsense like this)

**Comment by pfeatherstone** [comment]
Yeah got there in the end. Hopefully all ready now

**Comment by pfeatherstone** [comment]
Morning @davisking. Please let me know if there are any additional changes required. Otherwise, I'm looking forward to this getting merged.

File: dlib/matrix/matrix_fft.h [hunk 34]
```diff
@@ -9,836 +9,166 @@
...
+    
+    template < typename T, long NR, long NC, typename MM, typename L >
+    void fft_inplace (matrix<std::complex<T>,NR,NC,MM,L>& data)
     {
-        return call_mkl_fft(data, true) / data.size();
+#ifdef DLIB_USE_MKL_FFT
+        mkl_fft({data.nr(),data.nc()}, &data(0,0), &data(0,0), false);
+#else
+        kiss_fft({data.nr(),data.nc()}, &data(0,0), &data(0,0), false);
+#endif
```
davisking, line 157: Needs to be guarded by `if(data.size() != 0)`

dlib/matrix/matrix_fft.h
```diff
@@ -9,836 +9,166 @@
...
-    inline void ifft_inplace(matrix<std::complex<double> >& data)
+    template < typename T, long NR, long NC, typename MM, typename L >
+    void ifft_inplace (matrix<std::complex<T>,NR,NC,MM,L>& data)
     {
-        call_mkl_fft_inplace(data, true);
+#ifdef DLIB_USE_MKL_FFT
+        mkl_fft({data.nr(),data.nc()}, &data(0,0), &data(0,0), true);
+#else
+        kiss_fft({data.nr(),data.nc()}, &data(0,0), &data(0,0), true);
+#endif
```
davisking, line 169: Needs to be guarded by `if(data.size() != 0)`

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -21,16 +21,36 @@ namespace dlib
...
 // ----------------------------------------------------------------------------------------
+   
+    constexpr long fftr_nc_size(long nc)
+    {
+        return nc == 0 ? 0 : nc/2+1;
+    }
+    /*!
+        ensures
+            - returns the output dimension of a 1D real FFT
+    !*/
```
davisking, line 32: Please move all the `/*! !*/` comments before the `{}` so it looks like the rest of dlib.

davisking (changes requested):
Looks good, aside from the minor things I just noted.  I'll look it over a bit more later today and merge it once those bits are fixed up.

davisking:
urg, wtf.  I just pulled this PR, added a commit, and pushed it back but somehow this PR's branch had these other commits in it :|

davisking:

pfeatherstone:
> urg, wtf. I just pulled this PR, added a commit, and pushed it back but somehow this PR's branch had these other commits in it :|

I don't quite understand what the problem was. Maybe i should have rebased this PR's branch? Please let me know what the issue was so i know not to do it next time. Thanks again @davisking

davisking:
> > urg, wtf. I just pulled this PR, added a commit, and pushed it back but somehow this PR's branch had these other commits in it :|
> 
> I don't quite understand what the problem was. I never pulled master into my branch, so when you pulled and added a commit, i don't see where you would have got conflicts. Maybe i should have rebased this PR's branch yesterday just to be sure. Please let me know what the issue was so i know not to do it next time. Thanks again @davisking

It wasn't rebased.  I'm surprised github itself doesn't complain that it fails to rebase.  It must fall back on some more aggressive merge/squash when it checks if a PR applies cleanly.

Issue #2264, opened 2020-12-17
Feature request: draw_string() on image types


pfeatherstone:
It would be great if there was a version of ```draw_string()``` that worked on image types like ```dlib::matrix<dlib::rgb_pixel>``` for example.
Might submit a PR at some point. This issue could serve as a reminder.

arrufat:
Yes, I agree! I've also needed that feature some times! (to be honest I have to use OpenCV just for that...)

I've thought about it many times, but never got into it. Using the builtin [bdf](http://sofia.nmsu.edu/~mleisher/Software/otf2bdf/) font support would be a way to do it, but the main issue is that you need a font for each size you want (unless you don't care about proper scaling).

pfeatherstone:
Looking at the code in ```gui_widget.cpp``` and ```font.cpp```, i see no reason why bits of code couldn't be ripped out and refactored for image types rather than canvas. Unless i've missed something...

arrufat:
Yes, as I said, I never looked into it... Maybe it's straightforward :)

pfeatherstone:
Might give it a quick go now.

pfeatherstone:
[main.cpp.txt](https://github.com/davisking/dlib/files/5710327/main.cpp.txt)

pfeatherstone:
That worked for me. It's a basis. Probs need to template to support different image types, pixel types and probably other stuff

pfeatherstone:
Also support different fonts. Here i just use the default font

pfeatherstone:
Yeah so this is the function that works for me:
```
void draw_string(matrix<rgb_pixel>& img, dlib::point start_point, std::string str, dlib::rgb_pixel color)
{
    const dlib::rectangle rect(start_point,start_point);
    
    auto f_ptr = default_font::get_font();   
    const font& f = *f_ptr;
    
    long y_offset = rect.top() + f.ascender() - 1;

    long pos = rect.left()+f.left_overflow();
    for (string::size_type i = 0; i < str.size(); ++i)
    {
        // ignore the '\r' character
        if (str[i] == '\r')
            continue;
        // A combining character should be applied to the previous character, and we
        // therefore make one step back. If a combining comes right after a newline, 
        // then there must be some kind of error in the string, and we don't combine.
        if(is_combining_char(str[i]) && 
           pos > rect.left() + static_cast<long>(f.left_overflow()))
        {
            pos -= f[str[i]].width();
        }

        if (str[i] == '\n')
        {
[...]

PR #2270, opened 2020-12-27
Draw string on images

arrufat:
As discussed in #2264 with @pfeatherstone.

It's almost the same as the one defined in `fonts.h` for the gui widgets.

### DIFF
--- a/dlib/image_transforms/draw.h
+++ b/dlib/image_transforms/draw.h
@@ -7,6 +7,7 @@
 #include "../algs.h"
 #include "../pixel.h"
 #include "../matrix.h"
+#include "../gui_widgets.h"
 #include <cmath>
 
 namespace dlib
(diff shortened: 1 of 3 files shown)


pfeatherstone:
Awesome! Looking forward to this. I imagine dlib users will be very happy with this and will probably be able to remove opencv as a massive dependency.

davisking:
Nice, looks good.  Can you add a small test that does something like draws "cat" into a small image and checks that the expected result is obtained?  You could use `file_to_code_ex.cpp` to embed the correct image into the test code.

arrufat:
Yes, I was wondering how to test it systematically.

Let me make sure I understand what you mean:
- embed a target image with the string already drawn (a small black image with cat written in white)
- generate a black image of the same size, write "cat" at the same position in white
- compare both images

davisking:
> Yes, I was wondering how to test it systematically.
> 
> Let me make sure I understand what you mean:
> 
> * embed a target image with the string already drawn (a small black image with cat written in white)
> * generate a black image of the same size, write "cat" at the same position in white
> * compare both images

Yep.

davisking:

PR #2271, opened 2020-12-27
Fixed CUDA 11.1 compilation.

stoperro:
Fixes:
- ::floor() -> ::floorf() which should be same as float is used, yet otherwise compiler complains host function can't be used from global.

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -1549,12 +1549,12 @@ namespace dlib
                 const int c = idx%dnc;
 
                 const float y = r*y_scale;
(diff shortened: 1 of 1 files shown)


arrufat:
Which compiler are you using? Just wondering because I've been using CUDA 11.1 with GCC 10.2 without any issues.

stoperro:
Good point, I assumed this affects every compiler on top of CUDA. I use MSVC 16.9.0 Preview 2.0, 19.28.29617.
Also VCPKG version won't compile with now (with e.g. dlib[core,cuda]:x64-windows).

davisking:
Yeah, different compilers and platforms are the bane of cross-platform code :)

PR #2278, opened 2021-01-11
Fix build with clang on ppc64le

pkubaj:
Clang doesn't implement vec_rsqrt.
Code taken from https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/vsx_utils.hpp#L360

### DIFF
--- a/dlib/simd/simd4f.h
+++ b/dlib/simd/simd4f.h
@@ -524,7 +524,7 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-#ifndef vec_rsqrt
+#if defined(DLIB_HAVE_VSX) && !defined(vec_rsqrt)
     extern inline __vector float __attribute__((always_inline)) vec_rsqrt(const __vector float& a)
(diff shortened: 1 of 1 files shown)


dlib/simd/simd4f.h
```diff
@@ -524,6 +524,11 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+#if defined(DLIB_HAVE_VSX) && !defined(vec_rsqrt)
+    extern inline __vector float __attribute__((always_inline)) vec_rsqrt(const __vector float& a)
+    { return vec_div((__vector float){1, 1, 1, 1}, vec_sqrt(a)); }
```
davisking, line 529: Why isn't this `vector float` instead of `__vector vector`?

pkubaj: Same reason as in e.g. https://ffmpeg.org/pipermail/ffmpeg-devel/2019-August/247682.html - a namespace collision.

davisking:
Sweet, thanks for the PR :)

PR #2280, opened 2021-01-12
dlib::static_vector

pfeatherstone:
This is a stack-based, fixed capacity, dynamic size container. It is inspired by boost's static_vector. 
It provides hashing capability using both dlib::hash and std::hash and dlib serialization. Nice.
This is not ready yet.

### DIFF
--- a/dlib/algs.h
+++ b/dlib/algs.h
@@ -111,7 +111,7 @@ namespace std
 #include "assert.h"
 #include "error.h"
 #include "noncopyable.h"
-#include "enable_if.h"
+#include "traits.h"
 #include "uintn.h"
 #include "numeric_constants.h"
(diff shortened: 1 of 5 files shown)


pfeatherstone:
This PR contains some tidy up of type traits in dlib. This is because ```dlib::static_vector``` introduced a couple more and it made sense to put them in their own header as they were general. And it therefore made sense to put all other dlib type traits there too. A few of the dlib type traits were pre-C++11 implementations. Now, since dlib requires C++11, these have been refactored to aliases in the standard library. There is an exception: ```dlib::is_convertible```, which is functionally exactly the same as ```std::is_convertible``` has a specialisation in ```dlib/type_safe_union.h```. You cannot specialise a template alias. So i had to leave the old implementation of ```dlib::is_convertible``` as-is.

pfeatherstone:
Goodbye

PR #2281, opened 2021-01-12
Early termination for find_{min,max}_global

### LINKED ISSUE #2250: [Feature Improvement] add `should_stop` callback to dlib::find_max_global and friends
@davisking Thanks for your great C++ library!

My colleagues and I wrote a modified version of dlib's find_max_global that allowed the user to pass a callback `should_stop` which accepts a function-like object which has a method `bool operator()(double y)` where y is the result of the function evaluation, and it returns true if the search should stop.

We use this to provide coordinated early termination in two cases:

1. When we know that the user is willing to accept an solution with an objective that is past a certain threshold.  In this case we can stop as soon we find this objective.
2. We invoke find_max_global from multiple MPI ranks (read separate processes, so memory is not shared).  We pass different subdomains to each rank, and want to terminate early if one of the subdomains finds a solution that meets the user's requirements, we can stop all of the searches at their next iteration.

We used this technique to great success in our paper [FRaZ: A Generic High-Fidelity Fixed-Ratio Lossy Compression Framework for ScientificFloating-point Data](https://arxiv.org/pdf/2001.06139.pdf).  We would like to upstream this feature if possible, but wanted to see if there was interest before we go through the effort to update our fork.

robertu94:
This patch adds a callback to allow the user to request cancellation of a
search using find_{min,max}_global.  This enables users to cancel
searches when they are no-longer relevent, or when the user has some
special knowledge of the solution that they can use the stop the search
early.

closes  #2250

### DIFF
--- a/dlib/global_optimization/find_max_global.h
+++ b/dlib/global_optimization/find_max_global.h
@@ -130,7 +130,8 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
             const max_function_calls num,
             const std::chrono::nanoseconds max_runtime = FOREVER,
             double solver_epsilon = 0,
-            std::vector<std::vector<function_evaluation>> initial_function_evals = {}
+            std::vector<std::vector<function_evaluation>> initial_function_evals = {},
+            stop_condition& should_stop = never_stop_early
         ) 
         {
             // Decide which parameters should be searched on a log scale.  Basically, it's
@@ -176,17 +177,19 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
             using namespace std::chrono;
 
             const auto time_to_stop = steady_clock::now() + max_runtime;
+            //atomic<bool> doesn't support .fetch_or, use std::atomic<int> instead
+            std::atomic<int> this_should_stop{false};
 
             double max_solver_overhead_time = 0;
 
             // Now run the main solver loop.
-            for (size_t i = 0; i < num.max_calls && steady_clock::now() < time_to_stop; ++i)
+            for (size_t i = 0; i < num.max_calls && steady_clock::now() < time_to_stop && !this_should_stop.load(); ++i)
             {
                 const auto get_next_x_start_time = steady_clock::now();
                 auto next = std::make_shared<function_evaluation_request>(opt.get_next_x());
                 const auto get_next_x_runtime = steady_clock::now() - get_next_x_start_time;
 
-                auto execute_call = [&functions,&ymult,&log_scale,&eval_time_mutex,&objective_funct_eval_time,next]() {
+                auto execute_call = [&functions,&ymult,&log_scale,&eval_time_mutex,&objective_funct_eval_time,next,&should_stop,&this_should_stop]() {
                     matrix<double,0,1> x = next->x();
                     // Undo any log-scaling that was applied to the variables before we pass them
                     // to the functions being optimized.
@@ -198,6 +201,7 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
                     const auto funct_eval_start = steady_clock::now();
                     double y = ymult*call_function_and_expand_args(functions[next->function_idx()], x);
                     const double funct_eval_runtime = duration_cast<nanoseconds>(steady_clock::now() - funct_eval_start).count();
+                    this_should_stop.fetch_or(should_stop(y*ymult));
                     next->set(y);
                     
                     std::lock_guard<std::mutex> lock(eval_time_mutex);
@@ -366,6 +370,10 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
         template <typename T>
         const T& normalize(const T& item) {
             return item;
+        }
+
+        inline stop_condition& normalize(stop_condition& condition) {
+          return condition;
         }
 
         inline std::vector<std::vector<function_evaluation>> normalize(
--- a/dlib/global_optimization/find_max_global.h
+++ b/dlib/global_optimization/find_max_global.h
@@ -10,6 +10,7 @@
 #include <chrono>
 #include <memory>
 #include <thread>
+#include <functional>
 #include "../threads/thread_pool_extension.h"
 #include "../statistics/statistics.h"
 #include "../enable_if.h"
@@ -114,6 +115,8 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
 // ----------------------------------------------------------------------------------------
 
     const auto FOREVER = std::chrono::hours(24*365*290); // 290 years
+    using stop_condition = std::function<bool(double)>;
+    const stop_condition never_stop_early = [](double) { return false; };
 
 // ----------------------------------------------------------------------------------------
 
@@ -131,7 +134,7 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
             const std::chrono::nanoseconds max_runtime = FOREVER,
             double solver_epsilon = 0,
             std::vector<std::vector<function_evaluation>> initial_function_evals = {},
-            stop_condition& should_stop = never_stop_early
+            stop_condition should_stop = never_stop_early
         ) 
         {
             // Decide which parameters should be searched on a log scale.  Basically, it's
@@ -368,12 +371,9 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
         // find_max_global() instances below and turn them into the argument types expected by
         // find_max_global() above.
         template <typename T>
-        const T& normalize(const T& item) {
+        const T& normalize(const T& item) 
+        {
             return item;
-        }
-
-        inline stop_condition& normalize(stop_condition& condition) {
-          return condition;
         }
 
         inline std::vector<std::vector<function_evaluation>> normalize(
--- a/dlib/global_optimization/find_max_global_abstract.h
+++ b/dlib/global_optimization/find_max_global_abstract.h
@@ -10,10 +10,21 @@
 #include "../threads/thread_pool_extension_abstract.h"
 #include <utility>
 #include <chrono>
+#include <functional>
 
 namespace dlib
 {
 
+    /*!
+       WHAT THIS OBJECT REPRESENTS
+          A call-back that returns true when the search should stop
+
+          It is useful when the user either wants to terminate the search based on either special knowledge
+          of the function, the user's preferences regarding what is a "good-enough" solution, or
+          based on the results of some external process which may have found a solution and this search
+          is no longer required.
+     !*/
+    using stop_condition = std::function<bool(double)>;
 // ----------------------------------------------------------------------------------------
 
     template <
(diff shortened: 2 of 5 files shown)


robertu94:
@davisking I assume there are probably some things that you want updated like documentation, but I wasn't sure what that list of things was or how you would like it done.  I'd be happy to make the changes necessary to get this ready for merge.

dlib/global_optimization/global_function_search.h
```diff
@@ -13,6 +13,19 @@
...
+      explicit stop_condition(Function&& f): check(std::forward<Function>(f)) {}
+      // returns true if the search should terminate early
+      bool operator()(double v) {
+        return check(v);
+      }
+      std::function<bool(double)> check;
+    };
+
+    // a default stop condition that never terminates early
+    extern stop_condition default_stop_condition;
```
davisking, line 28: This shouldn't be needed.  You can make the functions just take a generic lambda.  See my other comment for more details.

dlib/global_optimization/find_max_global.h
```diff
@@ -949,10 +995,11 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
         const std::vector<bool>& is_integer_variable,
         const std::chrono::nanoseconds max_runtime,
         double solver_epsilon = 0,
-        const std::vector<function_evaluation>& initial_function_evals = {}
+        const std::vector<function_evaluation>& initial_function_evals = {},
+        stop_condition should_stop = default_stop_condition
```
davisking, line 999: So instead of adding more parameters here it would be better to do this:
```
    template <
        typename funct,
        typename... Args
        >
    function_evaluation find_min_global (
        thread_pool& tp, 
        funct f,
        const matrix<double,0,1>& bound1,
        const matrix<double,0,1>& bound2,
        const std::vector<bool>& is_integer_variable,
        const std::chrono::nanoseconds max_runtime,
        Args&&... args
    )   
    {   
        return find_min_global(tp, std::move(f), bound1, bound2, is_integer_variable, max_function_calls(), max_runtime, solver_epsilon, std::forward<Args>(args)...);
    } 
```

Then the function that actually implements all this can take these things as parameters and the `stop_condition` can just be a `std::function`.  I would otherwise suggest a universal reference instead of a `std::function` but if someone wants to add more arguments to this function later it will work out better if the type is something more narrow than universal reference.

davisking (changes requested):
Cool.  I left some comments.  Basically you can use a variadic template to make this simpler.

As for docs, yeah, definitely update the docs too.  See the instructions http://dlib.net/howto_contribute.html.  But here you just need to update `dlib/global_optimization/find_max_global_abstract.h`.

dlib/global_optimization/find_max_global.h
robertu94: How broadly do you want this applied?  There are 20+ overloads of either find_{min,max}_global each with subtly different arguments, and replacing these calls with a variadic template that maintains all of the old usages could be tricky depending on how many I need to move to the variadic template.  I also don't know how well default arguments work for function parameters that have template type on older compilers like gcc 4.X.

dlib/global_optimization/global_function_search.h
robertu94: I thought about a generic lambda, but I favored a separate class for a C++11 compatible way to enforce the type signature on the function-object and because it makes the call site more self documenting.  I also wasn't sure how well function arguments of template type worked with defaults in older compilers that dlib supports.

I can change this if needed.

dlib/global_optimization/find_max_global.h
davisking: It will only work for the two arguments I suggested: initial_function_evals and should_stop. So do it only for those two. But do it to all the overloads.

dlib/global_optimization/global_function_search.h
davisking: A generic lambda is the idiomatic way to do this in c++11 or later.  And using std::function when extra type emphasis is called for, which I think is the case here.

dlib/global_optimization/find_max_global.h
robertu94: I just finished trying this, and unfortunately this would be a breaking change in the API to apply this change to all of the overloads, so I applied the change to as many as I could without breaking the API.  This allowed me to update all the calls that don't invoke `impl::find_max_global`directly.

dlib/global_optimization/find_max_global_abstract.h
```diff
@@ -81,7 +92,8 @@ namespace dlib
         const max_function_calls num,
         const std::chrono::nanoseconds max_runtime = FOREVER,
         double solver_epsilon = 0,
-        const std::vector<std::vector<function_evaluation>>& initial_function_evals = {}
+        const std::vector<std::vector<function_evaluation>>& initial_function_evals = {},
+        stop_condition should_stop
```
davisking, line 96: This needs some kind of default, like `[](double){ return false; }`

dlib/global_optimization/find_max_global_abstract.h
```diff
@@ -134,6 +146,7 @@ namespace dlib
             - find_max_global() runs until one of the following is true:
                 - The total number of calls to the provided functions is == num.max_calls
                 - More than max_runtime time has elapsed since the start of this function.
+                - should_stop(f(x)) returns true, the default never returns true
```
davisking, line 149: Only say `should_stop(f(x)) returns true`.  The identity of the default should be communicated to the reader above in the signature.

dlib/global_optimization/find_max_global_abstract.h
davisking: Actually, `stop_condition should_stop = never_stop_early` is probably better.  And renaming `default_stop_condition` to `never_stop_early` and also noting it in the docs (i.e in the abstract.h file).  The word default doesn't communicate to someone what it does, but using a name like `never_stop_early` tells the reader something important about what it does.

dlib/global_optimization/find_max_global.h
davisking: What part would be a breaking change?  I just looked over the code and I don't see it.  Like if `impl::find_max_global()` is the only place that declares `initial_function_evals` and `should_stop` it sure looks like using a parameter pack in all other locations would work and and not break the API.

robertu94: The compile fails when building the python bindings for some reason when passing an internalizer list for initial_function_evals to impl::find_max_global when using perfect forwarding because the compliler deduces an incomplete parameter pack and discards correct the template candidate.  I'm guessing we hit some obscure portion of the standard relating to perfect forwarding and defaulted arguments.

Here are the relevant portions of the compiler error

```console
/home/runderwood/git/contrib/dlib/dlib/../dlib/global_optimization/find_max_global.h:378:16: error: no matching function for call to 'find_max_global'
        return find_max_global(functions, std::move(specs), num, max_runtime, solver_epsilon, {initial_function_evals}).second;
               ^~~~~~~~~~~~~~~
/home/runderwood/git/contrib/dlib/dlib/../dlib/global_optimization/find_max_global.h:460:16: note: in instantiation of function template specialization 'dlib::find_max_global<(lambda at /home/runderwood/git/contrib/dlib/tools/python/src/global_optimization.cpp:124:17)>' requested here
        return find_max_global(std::move(f), bound1, bound2, is_integer_variable, num, FOREVER, solver_epsilon, std::forward<Args>(args)...);
               ^
/home/runderwood/git/contrib/dlib/tools/python/src/global_optimization.cpp:129:19: note: in instantiation of function template specialization 'dlib::find_max_global<(lambda at /home/runderwood/git/contrib/dlib/tools/python/src/global_optimization.cpp:124:17)>' requested here
    auto result = find_max_global(func, list_to_mat(bound1), list_to_mat(bound2),
                  ^
[...snip...]

/home/runderwood/git/contrib/dlib/dlib/../dlib/global_optimization/find_max_global.h:297:43: note: candidate template ignored: substitution failure [with funct = (lambda at /home/runderwood/git/contrib/dlib/tools/python/src/global_optimization.cpp:124:17)]: deduced incomplete pack <(no value)> for template parameter 'Args'
    std::pair<size_t,function_evaluation> find_max_global (
                                          ^

```

davisking: Oh yeah, duh.  I wasn't paying attention.  It's because some of the `initial_function_evals` are `std::vector<function_evaluation>` while others are `std::vector<std::vector<function_evaluation>>`.   Let me see if I can refactor this a little bit so there aren't so many overloads.  It's kinda gotten out of hand.  I would have done that to begin with but I didn't want to conflict your pull request.  Anyway, I'll see if I can simplify this.

davisking: I just pushed a change that simplifies this a lot.  So you should now only have to add a parameter to the main top function and all the overloads will just automatically incorporate it.

robertu94:
@davisking I just rebased this code on master; I think it is ready for review.

davisking:
Urg.  I really wish github would automatically do some kind of pull request rebasing.

davisking:
Anyway, I pushed some small changes.  If this works for what you are doing (I would be surprised if it doesn't) can you add a little test to `dlib/test/global_optimization.cpp` that exercizes a non-default stopping condition?

robertu94:
@davisking added the test case.

dlib/test/global_optimization.cpp
```diff
@@ -232,6 +233,15 @@ namespace
         DLIB_TEST(std::abs(result.x - 1) < 1e-9);
         print_spinner();
 
+        constexpr auto close_enough = -16.0;
+        auto early_result = find_max_global([&early_evals](double x){ early_evals++; return -std::pow(x-2,2.0); }, -10, 1, max_function_calls(10), 0.0, std::vector<function_evaluation>{}, [](double y){ return (y >= close_enough);});
+        dlog << LINFO << "(x-2)^2, bound at 1: " <<  trans(early_result.x);
+        DLIB_TEST(early_result.x.size()==1);
+        DLIB_TEST(std::abs(early_result.y) <= std::abs(close_enough));
+        DLIB_TEST(std::abs(early_result.x - 1) <= 4);
+        DLIB_TEST(normal_evals >= early_evals);
```
davisking, line 242: This needs to be `DLIB_TEST(normal_evals > early_evals);` or we don't know if the stopping condition did anything.

dlib/test/global_optimization.cpp
```diff
@@ -232,6 +233,15 @@ namespace
         DLIB_TEST(std::abs(result.x - 1) < 1e-9);
         print_spinner();
 
+        constexpr auto close_enough = -16.0;
+        auto early_result = find_max_global([&early_evals](double x){ early_evals++; return -std::pow(x-2,2.0); }, -10, 1, max_function_calls(10), 0.0, std::vector<function_evaluation>{}, [](double y){ return (y >= close_enough);});
```
davisking, line 237: `close_enough` needs to be captured in the lambda.  I don't know why the tests passed.  But I got an email from appveyor with an error saying visual studio failed to compile because `close_enough` wasn't captured.  But somehow this diff isn't marked as failing any tests.

robertu94: Lambda expressions don't need to explicitly capture constant expressions according to [expr.prim.lambda]/12 because constant expressions are not ODR-used in a lambda according to [basic.def.odr]/3.

Apparently this is an known issue with MSVC fixed in a recent release. https://developercommunity.visualstudio.com/content/problem/367326/problems-with-capturing-constexpr-in-lambda.html

I pushed a fix for this to support older MSVC.

davisking: Yeah, MSVC has lots of standards compliance problems of this nature :|

davisking: Huh, I don't see any change from the last time I looked. But appveyor also ran MSVC on this PR again and this time didn't complain, so ¯\_(ツ)_/¯

davisking:

PR #2284, opened 2021-01-13
fix find_min_global finding wrong answers

### LINKED ISSUE #2283: find_min_global does not behave as expected when inital_function_evaluations are passed
## Expected Behavior

For this code,

```cpp
#include <dlib/global_optimization.h>
#include <cmath>
#include <vector>
#include <utility>

/// has a known min at 31/20 = 1.55
double f(double v) {
  return std::pow(v,2) - 3.1*v + 10.1;
}

void print_solution(std::string_view const& name, dlib::function_evaluation const& result) {
  std::cout << name << " (" << *result.x.begin() << ", " << result.y << ')' << std::endl;
}

void run_search(std::string_view name, std::vector<dlib::function_evaluation> const& evaluations) {
  auto result = dlib::find_min_global(f, -10.0, 10.0,
      dlib::max_function_calls(1000),
      std::chrono::seconds(3),
      1.0,
      evaluations
      );
    print_solution(name, result);
}

int main()
{
  const auto min = -10.0;
  const auto max = 10.0;
  const auto count = 20;
  std::vector<dlib::function_evaluation> evaluations;
  for (auto i = min; i < max; i += ((max-min)/count)) {
    evaluations.emplace_back(dlib::matrix<double,0,1>{i}, f(i));
  }

  auto best_inital = std::min_element(std::begin(evaluations), std::end(evaluations), []
      (dlib::function_evaluation const& lhs, dlib::function_evaluation const& rhs) {
        return lhs.y < rhs.y;
      });
  print_solution("initial", *best_inital);

  run_search("with_eval", evaluations);
  run_search("wo_eval", {});
  
  return 0;
}
```

I expect output something like the following:

```console
initial (2, 7.9)
with_eval (1.55, 7.6975)
wo_eval (1.55, 7.6975)
```

## Current Behavior

What I actually get is:

```
initial (2, 7.9)
with_eval (-10, -141.1)
wo_eval (1.55, 7.6975)
```

Note that with_eval no longer found a minimum. I believe this is happening because find_min_global internally multiplies the results of function evaluations by -1 (ymult) in order to minimize instead of maximize.  The documentation states that 

> initial_function_evals are incorporated into the solver state startup.  This is useful if you have information from a previous 
> optimization attempt  of f(x) **or just know some good initial x values that should be attempted as a baseline**. (emphasis added)

From my perspective, this would be resolved by either updating the documentation to say that it should be an "opaque" state returned from a prior use of a `global_function_search` for a maximum (or a minimum if the function evaluations .y values are multiplied by -1), or if find_global_min preprocessed the elements of initial_points so that you can use invocations of the function directly for find_min_global.

My use case is that my team writes a data compression analysis tool called z-checker, and we have already evaluated compressors on a particular dataset a several times with different settings for other purposes before we use dlib to find a particular set of compressors settings that have some property that we want (i.e. a particular peak-signal-to-noise-ratio, compression ratio, etc...).  We're happy to make the transformation ourselves before passing these initial evaluations to dlib, but would like either the documentation clarified that is the intended behaivor or the behavior changed to be more intuitive.

## Steps to Reproduce

* **Version**: `9e5c556740dc846160271b4f4aa39444af9d993b`
* I got dlib from `github.com/davisking/dlib`
* **Platform**:  cat `/etc/os-release`

```bash
NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo/Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
```

* **Compiler**: `gcc --version`

```console
gcc (Gentoo 8.4.0-r2 p3) 8.4.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

robertu94:
Previously, find_min_global would produce wrong output when passed
a collection of initial evaluations because the solver expected the y values to
be multiplied by -1.  This fix does that when minimizing.

closes #2283

### DIFF
--- a/dlib/global_optimization/find_max_global.h
+++ b/dlib/global_optimization/find_max_global.h
@@ -161,6 +161,11 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
             {
                 initial_function_evals.resize(specs.size());
             }
+            for (auto& evals : initial_function_evals) {
+              for (auto& eval  : evals) {
+                 eval.y *= ymult;
+              }
+            }
 
             global_function_search opt(specs, {initial_function_evals});
             opt.set_solver_epsilon(solver_epsilon);
--- a/dlib/global_optimization/find_max_global.h
+++ b/dlib/global_optimization/find_max_global.h
@@ -162,9 +162,9 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
                 initial_function_evals.resize(specs.size());
             }
             for (auto& evals : initial_function_evals) {
-              for (auto& eval  : evals) {
-                 eval.y *= ymult;
-              }
+                for (auto& eval : evals) {
+                    eval.y *= ymult;
+                }
             }
 
             global_function_search opt(specs, {initial_function_evals});


davisking:

PR #2285, opened 2021-01-14
Add Clipped ReLU and ELU activations

arrufat:
This PR will add the following activation layers, already present in CUDNN:
- clipped_relu
- elu

Progress:

- [x] cudnn dlib api
- [x] clipped_relu cuda
- [x] clipped_relu cpu
- [x] clipped_relu layer
- [x] elu cuda
- [x] elu cpu
- [x] elu layer
- [x] documentation
- [x] tests

### DIFF
--- a/dlib/cuda/cudnn_dlibapi.cpp
+++ b/dlib/cuda/cudnn_dlibapi.cpp
@@ -1759,34 +1759,34 @@ namespace dlib
         // The CUDNN implementation seems to be bugged, so the elu_gradient is implemented
         // in dlib/cuda/cuda_dlib.cu
 
-        void elu_gradient (
-            tensor& grad,
-            const tensor& dest,
-            const tensor& gradient_input,
-            const float coef
-        )
-        {
-            DLIB_CASSERT(
-                  have_same_dimensions(dest,gradient_input) == true &&
-                  have_same_dimensions(dest,grad) == true);
-            if (dest.size() == 0)
-                return;
+        // void elu_gradient (
+        //     tensor& grad,
+        //     const tensor& dest,
+        //     const tensor& gradient_input,
+        //     const float coef
+        // )
+        // {
+        //     DLIB_CASSERT(
+        //           have_same_dimensions(dest,gradient_input) == true &&
+        //           have_same_dimensions(dest,grad) == true);
+        //     if (dest.size() == 0)
+        //         return;
 
-            const float alpha = 1;
-            const float beta = is_same_object(grad,gradient_input) ? 0 : 1;
-            CHECK_CUDNN(cudnnActivationBackward(context(),
-                                          clipped_relu_activation_descriptor(coef),
-                                          &alpha,
-                                          descriptor(dest),
-                                          dest.device(),
-                                          descriptor(gradient_input),
-                                          gradient_input.device(),
-                                          descriptor(dest),
-                                          dest.device(),
-                                          &beta,
-                                          descriptor(grad),
-                                          grad.device()));
-        }
+        //     const float alpha = 1;
+        //     const float beta = is_same_object(grad,gradient_input) ? 0 : 1;
+        //     CHECK_CUDNN(cudnnActivationBackward(context(),
+        //                                   clipped_relu_activation_descriptor(coef),
+        //                                   &alpha,
+        //                                   descriptor(dest),
+        //                                   dest.device(),
+        //                                   descriptor(gradient_input),
+        //                                   gradient_input.device(),
+        //                                   descriptor(dest),
+        //                                   dest.device(),
+        //                                   &beta,
+        //                                   descriptor(grad),
+        //                                   grad.device()));
+        // }
 
     // ------------------------------------------------------------------------------------
     }
--- a/dlib/cuda/cudnn_dlibapi.cpp
+++ b/dlib/cuda/cudnn_dlibapi.cpp
@@ -1759,34 +1759,34 @@ namespace dlib
         // The CUDNN implementation seems to be bugged, so the elu_gradient is implemented
         // in dlib/cuda/cuda_dlib.cu
 
-        // void elu_gradient (
-        //     tensor& grad,
-        //     const tensor& dest,
-        //     const tensor& gradient_input,
-        //     const float coef
-        // )
-        // {
-        //     DLIB_CASSERT(
-        //           have_same_dimensions(dest,gradient_input) == true &&
-        //           have_same_dimensions(dest,grad) == true);
-        //     if (dest.size() == 0)
-        //         return;
+        void elu_gradient (
+            tensor& grad,
+            const tensor& dest,
+            const tensor& gradient_input,
+            const float coef
+        )
+        {
+            DLIB_CASSERT(
+                  have_same_dimensions(dest,gradient_input) == true &&
+                  have_same_dimensions(dest,grad) == true);
+            if (dest.size() == 0)
+                return;
 
-        //     const float alpha = 1;
-        //     const float beta = is_same_object(grad,gradient_input) ? 0 : 1;
-        //     CHECK_CUDNN(cudnnActivationBackward(context(),
-        //                                   elu_activation_descriptor(coef),
-        //                                   &alpha,
-        //                                   descriptor(dest),
-        //                                   dest.device(),
-        //                                   descriptor(gradient_input),
-        //                                   gradient_input.device(),
-        //                                   descriptor(dest),
-        //                                   dest.device(),
-        //                                   &beta,
-        //                                   descriptor(grad),
-        //                                   grad.device()));
-        // }
+            const float alpha = 1;
+            const float beta = is_same_object(grad,gradient_input) ? 0 : 1;
+            CHECK_CUDNN(cudnnActivationBackward(context(),
+                                          clipped_relu_activation_descriptor(coef),
+                                          &alpha,
+                                          descriptor(dest),
+                                          dest.device(),
+                                          descriptor(gradient_input),
+                                          gradient_input.device(),
+                                          descriptor(dest),
+                                          dest.device(),
+                                          &beta,
+                                          descriptor(grad),
+                                          grad.device()));
+        }
 
     // ------------------------------------------------------------------------------------
     }
--- a/dlib/cuda/cudnn_dlibapi.cpp
+++ b/dlib/cuda/cudnn_dlibapi.cpp
@@ -170,18 +170,6 @@ namespace dlib
             return des.get_handle();
         }
 
-        static cudnnActivationDescriptor_t clipped_relu_activation_descriptor(float coef)
-        {
-            thread_local cudnn_activation_descriptor des(CUDNN_ACTIVATION_CLIPPED_RELU, CUDNN_PROPAGATE_NAN, coef);
-            return des.get_handle();
-        }
-
-        static cudnnActivationDescriptor_t elu_activation_descriptor(float coef)
-        {
-            thread_local cudnn_activation_descriptor des(CUDNN_ACTIVATION_ELU, CUDNN_PROPAGATE_NAN, coef);
-            return des.get_handle();
-        }
-
     // ------------------------------------------------------------------------------------
 
         tensor_descriptor::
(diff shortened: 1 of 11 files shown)


dlib/cuda/cudnn_dlibapi.cpp
```diff
@@ -131,11 +131,11 @@ namespace dlib
             cudnn_activation_descriptor(
                 cudnnActivationMode_t mode,
                 cudnnNanPropagation_t reluNanOpt,
-                double reluCeiling
+                double coef
```
arrufat, line 134: I've renamed this to `coef`, because that's the name used in the [CUDNN API](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnSetActivationDescriptor) and it can be used for other purposes other than relu ceiling

dlib/test/dnn.cpp
```diff
@@ -2002,6 +2002,18 @@ namespace
...
+            print_spinner();
+            clipped_relu_ l;
+            auto res = test_layer(l);
+            DLIB_TEST_MSG(res, res);
+        }
+        {
+            print_spinner();
+            elu_ l;
+            auto res = test_layer(l);
+            DLIB_TEST_MSG(res, res);
```
arrufat, line 2015: I don't know why ELU is failing... I've run some tests and the CPU version works well, but not the CUDA one, which is surprising, because I'm just calling the CUDNN implementation...

arrufat:
I can't find the problem with the CUDA version of the ELU activation (which just calls CUDNN)... Other than that it's ready...

dlib/cuda/tensor_tools.cpp
```diff
@@ -999,6 +999,64 @@ namespace dlib { namespace tt
...
+
+    void elu_gradient (
+        tensor& grad,
+        const tensor& dest,
+        const tensor& gradient_input,
+        const float alpha
+    )
+    {
+#ifdef DLIB_USE_CUDA
+        cuda::elu_gradient(grad, dest, gradient_input, alpha);
```
arrufat, line 1054: Actually, changing this to `cpu::elu_gradient` works and makes the tests pass... I'll keep trying to figure out what I am doing wrong, but I might end up writing the elu gradient in CUDA myself...

arrufat:
After coding a custom gradient for the ELU layer, I realized the tests passed on the CI... So it must be a problem of my setup: I am currently using CUDA 11.2.0 and CUDNN 8.0.5.39...

davisking:
> After coding a custom gradient for the ELU layer, I realized the tests passed on the CI... So it must be a problem of my setup: I am currently using CUDA 11.2.0 and CUDNN 8.0.5.39...

CI doesn't build any of the cuda stuff (would be cool if it did but I don't think travis supports running cuda code).  So might not be your setup.  It also wouldn't be the first time someone found a bug in cuDNN :)

arrufat:
> > After coding a custom gradient for the ELU layer, I realized the tests passed on the CI... So it must be a problem of my setup: I am currently using CUDA 11.2.0 and CUDNN 8.0.5.39...
> 
> CI doesn't build any of the cuda stuff (would be cool if it did but I don't think travis supports running cuda code). So might not be your setup. It also wouldn't be the first time someone found a bug in cuDNN :)

Thank you for the reply, should we waite for a response in the [NVIDIA developer forum](https://forums.developer.nvidia.com/t/problem-with-elu-backward-in-cudnn/165999), or should I just use my one elu_gradient implementation?

Honestly, it seems odd that there's a bug in the *Exponential* Linear Unit gradient... But maybe it's so easy to compute the derivative that nobody bothered reporting the bug and implemented it themselves...

arrufat:
I think we can use the custom gradient implementation for now...

davisking:
> I think we can use the custom gradient implementation for now...

Yeah, that's what I would do :)

davisking:
Let me know when this is ready for review too :)

arrufat:
I think it's ready now, I am not sure about leaving the commented code for the elu gradient, though.

arrufat:
There has been some activity in the [forum](https://forums.developer.nvidia.com/t/problem-with-elu-backward-in-cudnn/165999).

I think there's a deeper problem (maybe my CPU implementations or API usage are wrong?)

Also, is it possible to pass the `test_layer` but still have a gradient computed wrong?
Because it's the case for me... I've created a [branch](https://github.com/arrufat/dlib/tree/elu-gradient-problem) that only contains 3 tests

- `relu`: all is good
- `clipped_relu`: `test_layer` and `forward` are good, but `backward` is wrong (CPU vs GPU)
- `elu`: only `forward` is good, rest fails.

https://github.com/arrufat/dlib/blob/elu-gradient-problem/dlib/test/dnn.cpp

It's weird because I think I am doing it well, but I might have overlooked something. I certainly don't think there are bugs in both CUDNN layers, so it's definitely me.

dlib/cuda/cudnn_dlibapi.cpp
```diff
@@ -170,6 +170,18 @@ namespace dlib
             return des.get_handle();
         }
 
+        static cudnnActivationDescriptor_t clipped_relu_activation_descriptor(float coef)
+        {
+            thread_local cudnn_activation_descriptor des(CUDNN_ACTIVATION_CLIPPED_RELU, CUDNN_PROPAGATE_NAN, coef);
```
davisking, line 175: So this doesn't work because `coef` is used only on the first call to `clipped_relu_activation_descriptor()`.  All subsequent calls (from the same thread anyway) are going to ignore whatever `coef` is given.  That's probably why you are seeing the odd behavior you noted.

davisking: Anyway, I don't know what costs there are in descriptor creation in cudnn.  The API is pretty suggestive of there at least being a dynamic allocation, which sucks.  But maybe it's not doing that and it's just a stack local variable through and through.  Maybe just moving `cudnnSetActivationDescriptor` calls out so you can set `coef` is fine.  You should check if there is any notable runtime cost to calling those cudnn functions that we currently avoid calling more than one via use of `thread_local`.

arrufat: OK, I will test it soon (if I understood it correctly, thanks for looking into it :))

Honestly, I don't need these activations myself, but I thought it was a shame that they were already implemented in CUDNN and not in dlib. This was supposed to be a simple PR, haha.

arrufat:
I am not sure I understood perfectly your previous message, but I've tried calling the `clipped_relu` like this, by creating an activation description every time I call these functions:

``` c++
void clipped_relu (
    tensor& dest,
    const tensor& src,
    const float coef
)
{
    DLIB_CASSERT(have_same_dimensions(dest,src));
    if (src.size() == 0)
        return;

    const float alpha = 1;
    const float beta = 0;
    cudnn_activation_descriptor des(CUDNN_ACTIVATION_CLIPPED_RELU, CUDNN_PROPAGATE_NAN, coef);
    cudnnActivationDescriptor_t handle;
    CHECK_CUDNN(cudnnCreateActivationDescriptor(&handle));
    CHECK_CUDNN(cudnnSetActivationDescriptor(handle, CUDNN_ACTIVATION_CLIPPED_RELU, CUDNN_PROPAGATE_NAN, coef));
    CHECK_CUDNN(cudnnActivationForward(context(),
                                 handle,
                                 &alpha,
                                 descriptor(src),
                                 src.device(),
                                 &beta,
                                 descriptor(dest),
                                 dest.device()));
}

void clipped_relu_gradient (
    tensor& grad,
    const tensor& dest,
    const tensor& gradient_input,
    const float coef
)
{
    DLIB_CASSERT(
          have_same_dimensions(dest,gradient_input) == true &&
          have_same_dimensions(dest,grad) == true);
    if (dest.size() == 0)
        return;

    const float alpha = 1;
    const float beta = is_same_object(grad,gradient_input) ? 0 : 1;
    cudnnActivationDescriptor_t handle;
    CHECK_CUDNN(cudnnCreateActivationDescriptor(&handle));
    CHECK_CUDNN(cudnnSetActivationDescriptor(handle, CUDNN_ACTIVATION_CLIPPED_RELU, CUDNN_PROPAGATE_NAN, coef));
    CHECK_CUDNN(cudnnActivationBackward(context(),
                                  handle,
                                  &alpha,
                                  descriptor(dest),
                                  dest.device(),
                                  descriptor(gradient_input),
                                  gradient_input.device(),
                                  descriptor(dest),
                                  dest.device(),
                                  &beta,
                                  descriptor(grad),
                                  grad.device()));
}
```

And this test is still failing (the forward pass is OK, but not the gradient, the difference is `1`, same as before this change):

``` c++
void test_clipped_relu()
{
    using namespace dlib::tt;
    print_spinner();
    const long n = 5;
    const long k = 5;
    const long nr = 3;
    const long nc = 3;
    const float ceiling = 6.0f;
    resizable_tensor src(n, k, nr, nc);
    tt::tensor_rand rnd;
    rnd.fill_uniform(src);
    resizable_tensor dest_cuda, dest_cpu;
    dest_cuda.copy_size(src);
    dest_cpu.copy_size(src);
    // initialize to different values in order to make sure the output is actually changed
    dest_cuda = 1;
    dest_cpu = 2;
    cuda::clipped_relu(dest_cuda, src, ceiling);
    cpu::clipped_relu(dest_cpu, src, ceiling);

    DLIB_TEST_MSG(max(abs(mat(dest_cuda) - mat(dest_cpu))) < 1e-7, max(abs(mat(dest_cuda) - mat(dest_cpu))));
    // test gradients
    resizable_tensor grad_cuda, grad_cpu, grad_input;
    grad_cuda.copy_size(src);
    grad_cpu.copy_size(src);
    grad_input.copy_size(src);
    rnd.fill_uniform(grad_input);
    grad_cuda = 1;
    grad_cpu = 2;
    cuda::clipped_relu_gradient(grad_cuda, dest_cuda, grad_input, ceiling);
    cpu::clipped_relu_gradient(grad_cpu, dest_cpu, grad_input, ceiling);
    const auto error = max(abs(mat(grad_cuda) - mat(grad_cpu)));
    DLIB_TEST_MSG(error < 1e-7, "error: " << error);
}
```

I am not sure what to do next:
- implement these layers in CUDA
- close this PR, since I don't really care about these activations, the whole point was to add wrappers for the missing cuDNN activations.

What's your call?

arrufat:
But weirdly enough, these two functions are pass the test:
``` c++
{
    print_spinner();
    clipped_relu_ l;
    auto res = test_layer(l);
    DLIB_TEST_MSG(res, res);
}
{
    print_spinner();
    elu_ l;
    auto res = test_layer(l);
    DLIB_TEST_MSG(res, res);
}
```

arrufat:
Oh, I found something weird, I was bothered by the fact that difference is exactly one, so I printed the gradients, and it seems that the values are added to the initial values (1 and 2) hence the difference, even though it shouldn't do that, because `grad` and `grad_input` are not the same object...

arrufat:
Oh, I found the problem, since `grad` and `grad_input` are not the same object, the backward pass is adding the values to the existing values in `grad_cpu` and `grad_cuda`, hence the difference of _**exactly**_ 1. 😅

arrufat:
So, after giving it a second look, I think all is good now (except for the ELU gradient, which seems to be bugged in cuDNN, so I implemented a custom gradient function, and I will test from time to time if they fix it... they didn't seem very responsive in the forum...)

dlib/cuda/cudnn_dlibapi.cpp
davisking: This can't be `thread_local` since only the first `coeff` will be used.   Like if you add a unit test that tests `clipped_relu()` that first calls it with a small coeff and then with a bigger one you should see the second call return the wrong result.

dlib/cuda/cudnn_dlibapi.cpp
```diff
@@ -170,6 +170,18 @@ namespace dlib
...
+            return des.get_handle();
+        }
+
+        static cudnnActivationDescriptor_t elu_activation_descriptor(float coef)
+        {
+            thread_local cudnn_activation_descriptor des(CUDNN_ACTIVATION_ELU, CUDNN_PROPAGATE_NAN, coef);
```
davisking, line 181: Same problem here.  It just can't be `thread_local` since `coef` is ignored on subsequent calls when it's `thread_local`.  

Either the descriptor needs to be created on every call or held as a class member if descriptor creation is too expensive.

davisking: It's really wacky to me that cuDNN works like this too.  Like WTF, it's just a single scalar.  I don't know why there is this heavy handed descriptor.  So like, unless cuDNN's version of these is somehow faster than the cuda code you wrote I would just not bother with cuDNN for these activations.

davisking: In fact, it wouldn't surprise me if the cuDNN implementation is slower.

arrufat: Ok, I will try again by putting the scalar as a class member and benchmark them.
I also thought about writing the CUDA activations myself, but I thought that the cuDNN versions would be faster.

Also, I am not sure these activations _actually_ useful, so if you think it's not worth it, we can close this PR.
It helped me better understand how the dlib cuDNN API works, so it's already a win for me.

davisking: ¯\_(ツ)_/¯   Nothing wrong with more activations.  I would honestly just write the cuda version yourself rather than deal with cuDNN for these.

arrufat: Yes, I get your point, I guess the only advantage of having the cuDNN version is to be able to fuse them with the convolution and the bias in a single cuDNN call.
I will try to make it work with cuDNN (just because I'm curious about the performance)

Also, off-topic, there's one thing that bothers me, your shruggie is missing one forearm... ¯\_(ツ)_/¯

arrufat: Ah, it's not missing! GitHub removes it :O

arrufat: You were right, after comparing both cuDNN and my CUDA implementation, there are no noticeable differences in speed.

arrufat:
I think it's finally done :)

davisking:
Sweet, looks good :)

PR #2286, opened 2021-01-15
FFT: added std::vector overloads for fft, ifft, fft_inplace and ifft_inplace

pfeatherstone:
This is ready.

### DIFF
--- a/dlib/matrix/matrix_fft.h
+++ b/dlib/matrix/matrix_fft.h
@@ -39,6 +39,25 @@ namespace dlib
   
 // ----------------------------------------------------------------------------------------
     
+    template < typename T, typename Alloc >
+    matrix<std::complex<T>,0,1> fft (const std::vector<std::complex<T>, Alloc>& in)
+    {
+        //complex FFT
+        static_assert(std::is_floating_point<T>::value, "only support floating point types");
+        matrix<std::complex<T>,0,1> out(in.size());
+        if (in.size() != 0)
+        {
+#ifdef DLIB_USE_MKL_FFT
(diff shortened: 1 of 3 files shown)


pfeatherstone:
@davisking In case you hadn't spotted this. This is ready for review/merge.

davisking:
> @davisking In case you hadn't spotted this. This is ready for review/merge.

 Nice. I’ll take a look in a bit.

davisking:
Looks good, thanks for another PR.  Just update the _abstract.h doc file to note these overloads and it's all good to go.

pfeatherstone:
> Looks good, thanks for another PR. Just update the _abstract.h doc file to note these overloads and it's all good to go.

In the last FFT PR, you mentioned that the _abstract.h file didn't require mentioning overloads.

pfeatherstone:
Or is it because, this time, the overloads aren't matrix-like?

davisking:
The _abstract.h file is the user facing documentation.  So it's got to be
the case that someone can read that and only that and understand fully how
to use the API.  The other overloads were all just providing optimizations
for different kinds of matrices.  The API was just "give any dlib matrix".
The API supports all dlib matrix objects so also noting that "some sub-type
of matrix also has fft functions for it" doesn't tell the user something
new since they already know all matrices are supported. But with this PR
someone can also give a std::vector, a distinctly different type.  They
have to be able to find out this is possible by reading the docs.

Conversely, anything someone doesn't need to know to correctly use the API
should not be in the abstract file.

pfeatherstone:
Got it. Thanks

pfeatherstone:
@davisking Hopefully the abstracts are good now.

davisking:
Looks good :)

PR #2294, opened 2021-02-02
Add support for fused convolutions

arrufat:
I've been playing a bit with the idea of having fused convolutions (convolution + batch_norm) in dlib.
I think the first step would be to move all the operations that are done by the `affine_` layer into the convolution, that is, update the bias of the convolution and re-scale the filters.

This PR adds some helper methods that allow doing this. The next step could be adding a new layer that can be constructed from an `affine_` layer and it's a no-op, like the tag layers, or add a version of the affine layer that does nothing (just outputs its input, without copying or anything). How would you approach this?

Finally, here's an example that uses a visitor to update the convolutions that are below an affine layer.
It can be build from by putting the file into the examples folder and loading the pretrained resnet 50 from the [`dnn_introduction3_ex.cpp`](http://dlib.net/dnn_introduction3_ex.cpp.html). If we manage to make something interesting out of it, maybe it would be interesting to have this visitor, too.

``` cpp
#include "resnet.h"

#include <dlib/dnn.h>
#include <dlib/image_io.h>

using namespace std;
using namespace dlib;

class visitor_fuse_convolutions
{
    public:
    template <typename T> void fuse_convolutions(T&) const
    {
        // disable other layer types
    }

    // handle the standard case (convolutional layer followed by affine;
    template <long nf, long nr, long nc, int sy, int sx, int py, int px, typename U, typename E>
    void fuse_convolutions(add_layer<affine_, add_layer<con_<nf, nr, nc, sy, sx, py, px>, U>, E>& l)
    {
        // get the parameters from the affine layer as alias_tensor_instance
        auto gamma = l.layer_details().get_gamma();
        auto beta = l.layer_details().get_beta();

        // get the convolution below the affine layer and its paramaters
        auto& conv = l.subnet().layer_details();
        const long num_filters_out = conv.num_filters();
        const long num_rows = conv.nr();
        const long num_cols = conv.nc();
        tensor& params = conv.get_layer_params();
        // guess the number of input filters
        long num_filters_in;
        if (conv.bias_is_disabled())
            num_filters_in = params.size() / num_filters_out / num_rows / num_cols;
        else
            num_filters_in = (params.size() - num_filters_out) / num_filters_out / num_rows / num_cols;

        // set the new number of parameters for this convolution
        const size_t num_params = num_filters_in * num_filters_out * num_rows * num_cols + num_filters_out;
        alias_tensor filters(num_filters_out, num_filters_in, num_rows, num_cols);
        alias_tensor biases(1, num_filters_out);
        if (conv.bias_is_disabled())
        {
            conv.enable_bias();
            resizable_tensor new_params = params;
            new_params.set_size(num_params);
            biases(new_params, filters.size()) = 0;
            params = new_params;
        }

        // update the biases
        auto b = biases(params, filters.size());
        b+= mat(beta);

        // rescale the filters
        DLIB_CASSERT(filters.num_samples() == gamma.k());
        auto t = filters(params, 0);
        float* f = t.host();
        const float* g = gamma.host();
        for (long n = 0; n < filters.num_samples(); ++n)
        {
            for (long k = 0; k < filters.k(); ++k)
            {
                for (long r = 0; r < filters.nr(); ++r)
                {
                    for (long c = 0; c < filters.nc(); ++c)
                    {
                        f[tensor_index(t, n, k, r, c)] *= g[n];
                    }
                }
            }
        }

        // reset the affine layer
        gamma = 1;
        beta = 0;
    }

    template <typename input_layer_type>
    void operator()(size_t , input_layer_type& l) const
    {
        // ignore other layers
    }

    template <typename T, typename U, typename E>
    void operator()(size_t , add_layer<T, U, E>& l)
    {
        fuse_convolutions(l);
    }
};

try
{
    resnet::infer_50 net1, net2;
    std::vector<std::string> labels;
    deserialize("resnet50_1000_imagenet_classifier.dnn") >> net1 >> labels;
    net2 = net1;
    matrix<rgb_pixel> image;
    load_image(image, "elephant.jpg");

    const auto& label1 = labels[net1(image)];
    const auto& out1 = net1.subnet().get_output();
    resizable_tensor probs(out1);
    tt::softmax(probs, out1);
    cout << "pred1: " << label1 << " (" << max(mat(probs)) << ")" << endl;


    // fuse the convolutions in the network
    dlib::visit_layers_backwards(net2, visitor_fuse_convolutions());
    const auto& label2 = labels[net2(image)];
    const auto& out2 = net2.subnet().get_output();
    tt::softmax(probs, out2);
    cout << "pred2: " << label2 << " (" << max(mat(probs)) << ")" << endl;

    cout << "max abs difference: " << max(abs(mat(out1) - mat(out2))) << endl;
    DLIB_CASSERT(max(abs(mat(out1) - mat(out2))) < 1e-2);
}
catch (const exception& e)
{
    cout << e.what() << endl;
    return EXIT_FAILURE;
}
```

output with this image (`elephant.jpg`):
elephant

```
pred1: African_elephant (0.962677)
pred2: African_elephant (0.962623)
max abs difference: 0.00436211
```

UPDATE: make visitor more generic and show a results with a real image

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -108,6 +108,7 @@ namespace dlib
         void set_bias_learning_rate_multiplier(double val) { bias_learning_rate_multiplier = val; }
         void set_bias_weight_decay_multiplier(double val)  { bias_weight_decay_multiplier  = val; }
         void disable_bias() { use_bias = false; }
+        void enable_bias() { use_bias = true; }
         bool bias_is_disabled() const { return !use_bias; }
 
         inline dpoint map_input_to_output (
@@ -2390,6 +2391,12 @@ namespace dlib
                     tt::multiply_conv(true, data_grad, gradient_input, g);
             }
         }
+
+        alias_tensor_instance get_gamma() { return gamma(params, 0); };
+        alias_tensor_const_instance get_gamma() const { return gamma(params, 0); };
+
+        alias_tensor_instance get_beta() { return beta(params, gamma.size()); };
+        alias_tensor_const_instance get_beta() const { return beta(params, gamma.size()); };
 
         const tensor& get_layer_params() const { return empty_params; }
         tensor& get_layer_params() { return empty_params; }
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2569,7 +2569,7 @@ namespace dlib
             }
 
             template <typename input_layer_type>
-            void operator()(size_t , input_layer_type& l) const
+            void operator()(size_t , input_layer_type& ) const
             {
                 // ignore other layers
             }
@@ -2587,6 +2587,7 @@ namespace dlib
         net_type& net
     )
     {
+        DLIB_CASSERT(count_parameters(net) > 0, "The network has to be allocated before fusing the convolutions.");
         visit_layers_backwards(net, impl::visitor_fuse_convolutions());
     }
 
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -21,11 +21,6 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-    namespace impl
-    {
-        class visitor_fuse_convolutions;
-    }
-
     struct num_con_outputs
     {
         num_con_outputs(unsigned long n) : num_outputs(n) {}
@@ -113,8 +108,23 @@ namespace dlib
         void set_bias_learning_rate_multiplier(double val) { bias_learning_rate_multiplier = val; }
         void set_bias_weight_decay_multiplier(double val)  { bias_weight_decay_multiplier  = val; }
         void disable_bias() { use_bias = false; }
-        void enable_bias() { use_bias = true; }
         bool bias_is_disabled() const { return !use_bias; }
+        void enable_bias()
+        {
+            if (use_bias == true)
+                return;
+            else
+                use_bias = true;
+            // setup biases if network has been allocated
+            if (params.size() > 0)
+            {
+                auto temp = params;
+                params.set_size(params.size() + num_filters_);
+                std::copy(temp.begin(), temp.end(), params.begin());
+                biases = alias_tensor(1, num_filters_);
+                biases(params, filters.size()) = 0;
+            }
+        }
 
         inline dpoint map_input_to_output (
             dpoint p
@@ -342,8 +352,6 @@ namespace dlib
             out << mat(item.params);
             out << "</con>";
         }
-
-        friend class impl::visitor_fuse_convolutions;
 
     private:
 
@@ -2526,28 +2534,19 @@ namespace dlib
                 // get the convolution below the affine layer
                 auto& conv = l.subnet().layer_details();
 
-                // guess the number of input channels
-                long k_in;
-                if (conv.bias_is_disabled())
-                    k_in = conv.params.size() / conv.num_filters() / conv.nr() / conv.nc();
-                else
-                    k_in = (conv.params.size() - conv.num_filters()) / conv.num_filters() / conv.nr() / conv.nc();
-
-                // enable bias and update the parameters size
                 if (conv.bias_is_disabled())
                 {
-                    const size_t num_params = k_in * conv.num_filters() * conv.nr() * conv.nc() + conv.num_filters();
-                    DLIB_CASSERT(num_params == conv.params.size() + conv.num_filters());
                     conv.enable_bias();
-                    resizable_tensor filters = conv.params;
-                    conv.params.set_size(num_params);
-                    std::copy(filters.begin(), filters.end(), conv.params.begin());
-                    conv.biases = alias_tensor(1, conv.num_filters());
-                    conv.biases(conv.params, conv.filters.size()) = 0;
                 }
 
+                tensor& params = conv.get_layer_params();
+
                 // update the biases
-                conv.biases(conv.params, conv.filters.size()) += mat(beta);
+                auto biases = alias_tensor(1, conv.num_filters());
+                biases(params, params.size() - conv.num_filters()) += mat(beta);
+
+                // guess the number of input channels
+                const long k_in = (params.size() - conv.num_filters()) / conv.num_filters() / conv.nr() / conv.nc();
 
                 // rescale the filters
                 DLIB_CASSERT(conv.num_filters() == gamma.k());
@@ -2555,7 +2554,7 @@ namespace dlib
                 const float* g = gamma.host();
                 for (long n = 0; n < conv.num_filters(); ++n)
                 {
-                    filter(conv.params, n * filter.size()) *= g[n];
+                    filter(params, n * filter.size()) *= g[n];
                 }
 
                 // disable the affine layer
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2532,13 +2532,13 @@ namespace dlib
                 // set the new number of parameters for this convolution and enable bias if needed
                 const size_t num_params = num_filters_in * num_filters_out * num_rows * num_cols + num_filters_out;
                 resizable_tensor new_params(num_params);
+                std::copy(params.begin(), params.end(), new_params.begin());
                 alias_tensor filters(num_filters_out, num_filters_in, num_rows, num_cols);
                 alias_tensor biases(1, num_filters_out);
                 if (conv.bias_is_disabled())
                 {
                     conv.enable_bias();
                     new_params.set_size(num_params);
-                    std::copy(params.begin(), params.end(), new_params.begin());
                     biases(new_params, filters.size()) = 0;
                 }
 
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -21,6 +21,11 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    namespace impl
+    {
+        class visitor_fuse_convolutions;
+    }
+
     struct num_con_outputs
     {
         num_con_outputs(unsigned long n) : num_outputs(n) {}
@@ -337,6 +342,8 @@ namespace dlib
             out << mat(item.params);
             out << "</con>";
         }
+
+        friend class impl::visitor_fuse_convolutions;
 
     private:
 
@@ -2521,34 +2528,31 @@ namespace dlib
                 const long num_filters_out = conv.num_filters();
                 const long num_rows = conv.nr();
                 const long num_cols = conv.nc();
-                tensor& params = conv.get_layer_params();
+                // tensor& params = conv.get_layer_params();
                 // guess the number of input filters
                 long num_filters_in;
                 if (conv.bias_is_disabled())
-                    num_filters_in = params.size() / num_filters_out / num_rows / num_cols;
+                    num_filters_in = conv.params.size() / num_filters_out / num_rows / num_cols;
                 else
-                    num_filters_in = (params.size() - num_filters_out) / num_filters_out / num_rows / num_cols;
+                    num_filters_in = (conv.params.size() - num_filters_out) / num_filters_out / num_rows / num_cols;
 
                 // set the new number of parameters for this convolution and enable bias if needed
                 const size_t num_params = num_filters_in * num_filters_out * num_rows * num_cols + num_filters_out;
-                resizable_tensor new_params(num_params);
-                std::copy(params.begin(), params.end(), new_params.begin());
                 alias_tensor filters(num_filters_out, num_filters_in, num_rows, num_cols);
                 alias_tensor biases(1, num_filters_out);
                 if (conv.bias_is_disabled())
                 {
                     conv.enable_bias();
-                    new_params.set_size(num_params);
-                    biases(new_params, filters.size()) = 0;
+                    conv.params.set_size(num_params);
+                    biases(conv.params, filters.size()) = 0;
                 }
 
                 // update the biases
-                auto b = biases(new_params, filters.size());
-                b+= mat(beta);
+                biases(conv.params, filters.size()) += mat(beta);
 
                 // rescale the filters
                 DLIB_CASSERT(filters.num_samples() == gamma.k());
-                auto t = filters(new_params, 0);
+                auto t = filters(conv.params, 0);
                 float* f = t.host();
                 const float* g = gamma.host();
                 for (long n = 0; n < filters.num_samples(); ++n)
@@ -2565,7 +2569,6 @@ namespace dlib
                     }
                 }
                 // disable the affine layer
-                params = new_params;
                 l.layer_details().disable();
             }
 
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2478,8 +2478,8 @@ namespace dlib
             else
                 out << "<affine_fc";
             if (item.is_disabled)
-                out << " disabled='"<<(item.is_disabled?"true":"false");
-            out << "'>\n";
+                out << " disabled='"<< std::boolalpha << item.is_disabled << "'";
+            out << ">\n";
             out << mat(item.params);
 
             if (item.mode==CONV_MODE)
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2331,12 +2331,21 @@ namespace dlib
 
         layer_mode get_mode() const { return mode; }
 
+        void disable()
+        {
+            params.clear();
+            is_disabled = true;
+        }
+
         inline dpoint map_input_to_output (const dpoint& p) const { return p; }
         inline dpoint map_output_to_input (const dpoint& p) const { return p; }
 
         template <typename SUBNET>
         void setup (const SUBNET& sub)
         {
+            if (is_disabled)
+                return;
+
             if (mode == FC_MODE)
             {
                 gamma = alias_tensor(1,
@@ -2358,6 +2367,9 @@ namespace dlib
 
         void forward_inplace(const tensor& input, tensor& output)
         {
+            if (is_disabled)
+                return;
+
             auto g = gamma(params,0);
             auto b = beta(params,gamma.size());
             if (mode == FC_MODE)
@@ -2372,6 +2384,9 @@ namespace dlib
             tensor& /*params_grad*/
         )
         {
+            if (is_disabled)
+                return;
+
             auto g = gamma(params,0);
             auto b = beta(params,gamma.size());
 
@@ -2403,11 +2418,12 @@ namespace dlib
 
         friend void serialize(const affine_& item, std::ostream& out)
         {
-            serialize("affine_", out);
+            serialize("affine_2", out);
             serialize(item.params, out);
             serialize(item.gamma, out);
             serialize(item.beta, out);
             serialize((int)item.mode, out);
+            serialize(item.is_disabled, out);
         }
 
         friend void deserialize(affine_& item, std::istream& in)
@@ -2435,7 +2451,7 @@ namespace dlib
                 return;
             }
 
-            if (version != "affine_")
+            if (version != "affine_" && version != "affine_2")
                 throw serialization_error("Unexpected version '"+version+"' found while deserializing dlib::affine_.");
             deserialize(item.params, in);
             deserialize(item.gamma, in);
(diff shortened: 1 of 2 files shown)


pfeatherstone:
This looks good. Does it speed up performance? CPU and GPU?

arrufat:
I didn't benchmark anything, but I guess it will be the same as without "fusing" the convolutions, since we are still calling the `affine_` layer. Now we need a layer that can be constructed from an `affine_` layer and it's a no-op. Then process could look like:

1. train with network defined with `bn_con_`
1. assign to the same network, but defined with `affine_`
1. run the `visitor_fuse_convolutions`
1. assign the modified network to another one defined with a `no-op` layer instead of `affine_`

If we could add a method like `.disable()` to the `affine_` layer and make it a no-op, that would be great, since step 4 could be done in the visitor.

pfeatherstone:
> If we could add a method like `.disable()` to the `affine_` layer and make it a no-op, that would be great, since step 4 could be done in the visitor.

I would do that. Just need to add a boolean flag to the ```affine_``` layer. Need to make that flag serializable and bump the version of the layer when serializing while still supporting the old version (i.e. without the boolean flag) when deserializing. That's the easiest thing to do from a user perspective. They don't have to create a 3rd type of network and they don't need to care about how deserialization works.

arrufat:
Yes, if that is possible, me too :)

davisking:
Yeah, I would go with a .disable() on the affine layer :)

arrufat:
I've updated the `affine_` layer to support the disabled mode.
In that example, I ran the image with both networks 1000, and these are the average timings:

|   layer  | inference | fps |   net size  |   VRAM  |
|:--------:|----------:|-----|:-----------:|:-------:|
|  affine  |  4.115 ms | 243 | 86.9994 MiB | 594 MiB |
| disabled |  3.829 ms | 261 | 86.8261 MiB | 588 MiB |

So, some improvements. that we can get for free :)

pfeatherstone:
So 7% performance gain depending on which way you look at it. Nice work!

pfeatherstone:
Darknet FPS / dlib FPS ~ 70%. So it must something else that's holding dlib back

arrufat:
Resnet 50 has only 49 batch normalizations (YOLOv3 has 75 and YOLOv4 109). So I am expecting to gain a bit more on those models :)

arrufat:
@davisking when you have time, let us know what you think about this PR, notably the naming of functions, etc.

Also, maybe that visitor could be more generic and accept the case where the batch norm has an fc layer as its input... But I am not sure it's worth it... But in that case, the visitor should be named differently.

Ah, I put the visitor there because it depends directly on the `affine_` layer (I tried to put it into the utilities header, but that didn't work.

pfeatherstone:
It does make you wonder if there is any point in the affine layer anymore. Unless people are using it during training too. My guess is they are using it for converting batchnorm layers.

arrufat:
From the [`affine_` layer documentation](http://dlib.net/dlib/dnn/layers_abstract.h.html#affine_), this layer does nothing (identity transform) unless initialized with a `bn_` layer. And it has no trainable parameters, so there's no point in using it for training.

I guess it should be disabled by default, and then enabled once it's initialized with a `bn_` layer.
That way, when the user `cout`s the net, it will see `disabled` if the `affine_` hasn't been initialized properly. Thoughts on this?

dlib/dnn/layers.h
```diff
@@ -2463,11 +2492,105 @@ namespace dlib
...
+                alias_tensor filters(num_filters_out, num_filters_in, num_rows, num_cols);
+                alias_tensor biases(1, num_filters_out);
+                if (conv.bias_is_disabled())
+                {
+                    conv.enable_bias();
+                    resizable_tensor new_params = params;
+                    new_params.set_size(num_params);
+                    biases(new_params, filters.size()) = 0;
+                    params = new_params;
+                }
```
arrufat, line 2543: I've realized this block does not work properly. I get a segfault... I guess it's not a good idea to resize the params of the layer once it has already been allocated. What would be a nice way of doing it?
Anyway, if the user is planning on fusing the convolutions after the training, it shouldn't matter that much, but still...

arrufat:
After fixing the segfault, now I get this error when trying to do inference in a fused network that had bias disabled in convolutions:

```
Error detected at line 258.
Error detected in file ../external/dlib/dlib/cuda/cudnn_dlibapi.cpp.
Error detected in function void dlib::cuda::add(float, dlib::tensor&, float, const dlib::tensor&).

Failing expression was (have_same_dimensions(src, dest) || (src.num_samples()==1 && src.k()==dest.k() && src.nr()==1 && src.nc()==1) || (src.num_samples()==1 && src.k()==dest.k() && src.nr()==dest.nr() && src.nc()==dest.nc()) || (src.num_samples()==1 && src.k()==1 && src.nr()==dest.nr() && src.nc()==dest.nc()) || (src.num_samples()==dest.num_samples() && src.k()==1 && src.nr()==1 && src.nc()==1)) && is_same_object(src,dest) == false.

         dest.num_samples(): 1
         dest.k():           64
         dest.nr():          112
         dest.nc():          112
         src.num_samples():  0
         src.k():            0
         src.nr():           0
         src.nc():           0
```

So I am guessing there are some other parameters that should be modified elsewhere in the network... I'll keep digging.

dlib/dnn/layers.h
```diff
@@ -21,6 +21,11 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    namespace impl
+    {
+        class visitor_fuse_convolutions;
+    }
+
```
arrufat, line 28: - Is it OK to put this here?
- Do you mind this class being a friend of the `con_` class?

Maybe I could move the definition of the `affine_` class here.

pfeatherstone:
So if ```affine_``` is only ever used to evaluate batchnorm layers, and since there is no reason not to fuse conv-batchnorm layers, there could also be a third type of layer, maybe called ```fuse_tag_```, that is a no-op, which when assigning networks, signifies that batchnorm layers and the preceding conv layers should be fused. So rather than do:
```
using net_train = model<bn>;
using net_infer = model<affine>;
net_train net1;
net_infer net2;
net2 = net1;
visit_layers_backwards(net2, visitor_fuse_convolutions());
```
we could have:
```
using net_train = model<bn>;
using net_infer = model<fuse_tag>;
net_train net1;
net_infer net2;
net2 = net1; //fusing is done here
```

That way there are no changes required to the ```affine_``` layer, the user does not need to call a visitor, the fusing is done implicitly when converting between ```net_train``` type to ```net_infer``` type. The user does not need to compile 3 types, still 2, because we don't have to go via model<affine> anymore.

arrufat:
@pfeatherstone I am bit confused by your message. I think affine is now a no-op when disabled, right?

pfeatherstone:
> @pfeatherstone I am bit confused by your message. I think affine is now a no-op when disabled, right?

Sorry, i've realised what i said is a bit confusing. What I mean is, does the user actually need to call a visitor? Can't the fusing be done when assigning networks, i.e. when doing : 
```
using net_train = model<bn>;
using net_infer = model<affine>;
net_train net1;
net_infer net2;
net2 = net1; // <----- HERE
```

arrufat:
If it can be done, I can't think of an easy way of doing it...

And by the way, I've tried applying this visitor to the YOLO networks and it didn't change anything... If anything, it uses a bit more VRAM and it seems a bit slower...

pfeatherstone:
> If it can be done, I can't think of an easy way of doing it...

Ok just a thought. To be honest, none of the other frameworks do automatic fusing so why should dlib. In which case, having to call a visitor manually isn't out of the norm. 

> And by the way, I've tried applying this visitor to the YOLO networks and it didn't change anything... If anything, it uses a bit more VRAM and it seems a bit slower...

Bugger

arrufat:
Can you try this fused YOLOv3? https://we.tl/t-L9N6JguB53

pfeatherstone:
> Can you try this fused YOLOv3? https://we.tl/t-L9N6JguB53

Will do later. I'm about to head to work. Thanks for the weights

dlib/dnn/layers_abstract.h
```diff
@@ -1899,6 +1899,38 @@ namespace dlib
...
+            ensures
+                - returns the beta parameter that defines the behavior of forward().
+        !*/
+
+        alias_tensor_const_instance get_beta() const;
+        /*!
+            ensures
+                - returns the beta parameter that defines the behavior of forward().
+        !*/
+
```
arrufat, line 1933: I don't know if these methods are useful... They could be removed by making the visitor a friend of `affine_`, too.

arrufat:
After looking at the CUDNN api, I am not sure this is the best we can do...

We should use [cudnnConvolutionBiasActivationForward](
https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnConvolutionBiasActivationForward), instead of [cudnnConvolutionForward](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnConvolutionForward) + bias + activation.
This way we could do everything in one call, but I don't know if it's possible without breaking too many things.

davisking:
> After looking at the CUDNN api, I am not sure this is the best we can do...
> 
> We should use [cudnnConvolutionBiasActivationForward](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnConvolutionBiasActivationForward), instead of [cudnnConvolutionForward](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnConvolutionForward) + bias + activation.
> This way we could do everything in one call, but I don't know if it's possible without breaking too many things.

You shouldn't have to break anything though.  Why not make a new layer type that is specific to this operation?  Then you don't have to figure out how to work it into existing layers.

dlib/dnn/layers.h
davisking: I'm not super jazzed about friend functions like this.  It seems like it would be a lot simpler to add new layers entirely that support these fused operations and leave existing layers alone.

arrufat:
What I was thinking was:

- if a convolution has disabled, call `cudnnConvolutionForward`, just like now
- else, call `cudnnConvolutionBiasActivationForward` with the identity transform (since, as I see it, this call only works with CUDNN activations, right?)

This way we could avoid the call that just ads the bias, since it would be handled by CUDNN. Do you think it's feasible or even a good idea? That way we don't need to introduce new layers and previously trained networks could still be "optimized" using the visitor.

davisking:
Maybe.  But the thing that bothers me about that is that it makes it less clear when these folding operations happen.  E.g. someone could put something between a conv and relu layer and that would inhibit the folding.  Maybe that's not a big deal though.

I guess so long as the visitor can work in a reasonably simple way it's fine.  But no friend functions.  Every layer needs to have a well defined interface rather than having special functions pierce the public interface and do private manipulation.  Seems like that could be accomplished by just having methods on the layers as appropriate, like one that tells conv to include a relu activation.  And other layers like relu have a .disable() method.

davisking:
But yeah, being able to activate this on already trained networks is no small thing.  I was thinking about that last night as well, and that's really the main selling point of modifying existing layers rather than adding new layers.  At least to my mind anyway.

arrufat:
Ok, then I will try first by making a visitor without friend classes and try to modify fuse the layers to be compatible with previously trained networks.

I hope I didn't misjudge the task I'm not signing up on something above my skill set level... anyway, I'll try and learn something in the worst case scenario.

davisking:
Na, you can totally handle it :)

arrufat:
Oops, I tapped on the wrong place on the Github app...

arrufat:
I am not sure about last change, I need to test it more, but at least, it seems to work... And I also don't know why it doesn't work on Travis... It works with GCC 10.2...

pfeatherstone:
Is there an update on this? Very keen to see the end result !

arrufat:
Honestly, not much has been done...
I noticed that the speed I gain here is because this ResNet50 applied bias twice (once in the convolutions and once in the batch normalization).
For models that already have bias-less convolutions, there is no improvement after fusing, which kind of got me a bit less motivated to work on this...

Then I tried to fuse the ReLU layer into the `cudnnConvolutionBiasActivationForward`, unsuccessfully.
The weird thing is that this works:
``` c++
CHECK_CUDNN(cudnnConvolutionBiasActivationForward(
        context(),
        &alpha1,
        descriptor(data),
        data.device(),
        (const cudnnFilterDescriptor_t)filter_handle,
        filters.device(),
        (const cudnnConvolutionDescriptor_t)conv_handle,
        (cudnnConvolutionFwdAlgo_t)forward_algo,
        forward_workspace,
        forward_workspace_size_in_bytes,
        &alpha2,
        (const cudnnTensorDescriptor_t)bias_handle,
        b,
        (const cudnnTensorDescriptor_t)bias_handle,
        b,
        identity_activation_descriptor(),
        descriptor(output),
        output.device()));
```

But, if I change `identity_activation_descriptor()` to `relu_activation_descriptor()`, I get:
```
Error while calling cudnnConvolutionBiasActivationForward( context(), &alpha1, descriptor(data), data.device(), (const cudnnFilterDescriptor_t)filter_handle, filters.device(), (const cudnnConvolutionDescriptor_t)conv_handle, (cudnnConvolutionFwdAlgo_t)forward_algo, forward_
workspace, forward_workspace_size_in_bytes, &alpha2, (const cudnnTensorDescriptor_t)bias_handle, b, (const cudnnTensorDescriptor_t)bias_handle, b, relu_activation_descriptor(), descriptor(output), output.device()) in file /home/adria/Projects/dlib/dlib/cuda/cudnn_dlibapi.cp
p:1225. code: 9, reason: CUDNN_STATUS_NOT_SUPPORTED
```
Of course, I modified the visitor to disable the ReLU layer and call that method. I will push the latest changes (and the example file), so you can try it for yourself. I am probably calling the CUDNN API wrong...

pfeatherstone:
Hmm, maybe it's something else causing bottlenecks then.

pfeatherstone:
This looks interesting. Alexey from darknet claims a 20% performance enhancement for detection models.
https://developer.nvidia.com/blog/cuda-graphs/

arrufat:
I'm still working on it, making slow progress :P

arrufat:
Still in my to-do list, but taking longer than expected... testing out this stuff is time-consuming :P

arrufat:
I've been side tracked with #2376. After merging, I will focus on this again, hopefully that will improve the speed of YOLO detectors.

pfeatherstone:
That would be cook. 70FPS at 416x416 on single GPU is the target :) :) :)

pfeatherstone:
Having had a look at ncnn and opencv, it looks like these repos speed things up with:
* operator fusion (like what this PR is doing)
* specialised kernels

For the latter, I noticed ncnn has special cases for 1x1, 3x3, 5x5 and 7x7 convolutions that use a combination of SIMD and parallel for-loops. This doesn't really help for CUDA though.

pfeatherstone:
Hello again. Sorry if my comments/suggestions are starting to resemble spam. I know this PR specifically deals with fused operations, but since the underlying theme concerns the improvement of inference speed, [cuda-graphs](https://developer.nvidia.com/blog/cuda-graphs/) seem to be fairly new and could be beneficial. Though, i have a feeling this could imply some substantial refactoring of dlib backend stuff.

arrufat:
I am having a weird error when deserializing a fused network and I can't see what I'm doing wrong
Steps:
``` bash
# run the example on the ResNet50 and fuse the layers
./dnn_fused_convolution_ex resnet50_1000_imagenet_classifier.dnn 1
# rerun the example on the fused network
./dnn_fused_convolution_ex resnet50_1000_imagenet_classifier-fused.dnn
```
```
An error occurred while trying to read the first object from the file 'resnet50_1000_imagenet_classifier-fused.dnn'.
ERROR: Error deserializing object of type int
   while deserializing object of type std::vector
```
The error occurs after deserializing the first convolution. I reach the end of the `deserialize` for that convolution.
It might be a stupid thing I missed/messed...

When that is done, I will proceed with the benchmarking.
I noticed some speed improvements, but the CUDA runtime gives me very different results every time I try: for YOLOv3 I get between 50 and 60 fps for the non-fused and up to 65 fps for the fused, but with high variance...

arrufat:
So, it is working now as expected. I renamed the function to `fuse_layers`, since it might be extended to other kind of layers, not just convolutions, at some point.

I was expecting more gain... I get around 6-7% improvement... Maybe that's what it is.
For YOLOv3 on NVIDIA GTX 1080 Ti, I go from 48 fps to 51 fps after fusing the layers. It's a tiny improvement, but it's free.
On the ResNet50 example, which doesn't have the duplicative biases disabled, I go 270 from to 290 fps.

I also added the `cudnnConvolutionBiasActivationForward`, so now all convolutions with bias use this cuDNN function, instead of calling two separate functions (one for the convolution and one for the biases)

There's just one thing that bothers me. The YOLOv3 model is using more VRAM after fusing the layers than before (894 MiB vs 873 MiB). It shouldn't be, since after fusing, the affine parameters are empty and I rescale the convolution weights and update the biases, so it should use a little bit less memory.

I must have missed something, please have a look if you have time :)

davisking:
Sweet. Can't look today. Might look tomorrow. Not sure.

arrufat:
No hurries :) Thank you!

arrufat:
Update, I tried fusing the weights of another network (VoVNet based) I had already trained, which already had duplicative bias disabled.
The inference speed increased by around 7% as expected, and the VRAM usage decreased as well, in contrast to the YOLOv3 model... I don't know what you're going to do with this information, but here it is anyway :)

davisking:
> From the [`affine_` layer documentation](http://dlib.net/dlib/dnn/layers_abstract.h.html#affine_), this layer does nothing (identity transform) unless initialized with a `bn_` layer. And it has no trainable parameters, so there's no point in using it for training.
> 
> I guess it should be disabled by default, and then enabled once it's initialized with a `bn_` layer.
> That way, when the user `cout`s the net, it will see `disabled` if the `affine_` hasn't been initialized properly. Thoughts on this?

Yeah, it defaults to an identity transform by default.  So nothing wrong with having it default do not actually doing anything until it gets assigned.  Not sure how I missed this comment and didn't reply until now :/

davisking:
I'm severely tardy in looking at this PR too :(  Should I look now or are you editing it again?

arrufat:
There's no hurry, don't feel any pressure :)
You can have a look now, I think it's ready (it has worked on all networks I tried)

arrufat:
The functionality is there already, I should probably add some tests for networks that have convolutions with and without biases.
I mean, it works, but there are no tests.
And also probably update an example program (of your choosing) to showcase this new functionality. Having the extra 7% speedup is quite nice :)

dlib/dnn/layers_abstract.h
```diff
@@ -1919,6 +1930,38 @@ namespace dlib
...
+        /*!
+            ensures
+                - sets the params.size() of this layer to 0.
+                - when forward_inplace and backward_inplace are called, they return immediately.
+        !*/
+
+        alias_tensor_instance get_gamma();
+        /*!
+            ensures
+                - returns the gamma parameter that defines the behavior of forward().
```
davisking, line 1944: Update the docs for the class to talk about the parameters in terms of `gamma` and `beta`.  Or rename these methods and docs to talk in terms of A and B params.  But one way or another make it so they are in sync with how they talk about them so that someone reading the docs will know for sure what is being referenced.

davisking (changes requested):
Yeah this looks cool.  I left some comments.  Totally add some tests too :)

dlib/cuda/tensor_tools.h
```diff
@@ -1042,6 +1042,64 @@ namespace dlib { namespace tt
                 - #output.nc() == 1+(data.nc() + 2*padding_x - filters.nc())/stride_x
         !*/
 
+#ifdef DLIB_USE_CUDA
```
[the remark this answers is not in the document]
arrufat: I thought about it... in CUDA it makes sense because there's a cuDNN function for that, but on the CPU it would do the same thing as calling first the convolution, and then the biases. But maybe you're suggesting that to have a uniform API, right?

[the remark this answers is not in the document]
davisking: Right. Just talking about uniformity and simplicity of the api.

arrufat:
If the tests pass, then it's ready :)

facug91:
I was having a problem in `master` when running the same example code I used in #2655, using Docker image `nvidia/cuda:10.2-cudnn7-devel-ubuntu18.04`. When I ran the example it threw:  
```
Error while calling cudnnConvolutionBiasActivationForward( context(), &alpha1, descriptor(data), data.device(), (const cudnnFilterDescriptor_t)filter_handle, filters.device(), (const cudnnConvolutionDescriptor_t)conv_handle, (cudnnConvolutionFwdAlgo_t)forward_algo, forward_workspace, forward_workspace_size_in_bytes, &alpha2, out_desc, out, descriptor(biases), biases.device(), identity_activation_descriptor(), out_desc, out) in file /home/user/dlib/dlib/cuda/cudnn_dlibapi.cpp:1219. code: 9, reason: CUDNN_STATUS_NOT_SUPPORTED
```  
I found that this specific commit is the one causing that problem. Testing it with the commit before this one, it works.

arrufat:
Oh, that's odd, I've been using this a lot, without issues. Can you give more details? Maybe it's related to CUDA? I've only ever tried it with CUDA 11, I think

facug91:
I should read the code, but it seems to me that the problem might be with cudnn 7

PR #2300, opened 2021-02-07
Enhanced serialization support to and from iostream with templated byte type.

pfeatherstone:


pfeatherstone:
Still haven't sorted out ```proxy_serialize``` and ```proxy_deserialize```. Not sure yet which is the cleanest way to do it.

pfeatherstone:
I thought this would be a 5 commit PR. Turns out not.

pfeatherstone:
Something to do with EOF. What is it for uint8_t?

pfeatherstone:
So ```int8_t``` is a bit funny. ```(int8_t)-1``` is legal of course and coincides with ```std::char_traits<int8_t>::eof()```. BUT. ```(char)-1``` wraps to 255 and therefore doesn't wrap to ```std::char_traits<char>::eof()```. Don't get it.

pfeatherstone:
I don't think c++ ```iostream``` was designed to be easily extended with ```int8_t``` and ```uint8_t```. The standard provides specialisations for ```char```, ```char16```, etc. Funny things happen when you default back to default implementation of ```std::char_traits<T>```, which is the case for ```uint8_t``` and ```int8_t```.

pfeatherstone:
To make this work i think we would need proper specialisations of ```std::char_traits<int8_t>``` and ```std::char_traits<uint8_t>```

pfeatherstone:
So char traits are dealt with. Just need to extend proxy_serialize and proxy_deserialize to accomodate for template basic_istream and basic_ostream.

pfeatherstone:
@davisking What do you think of this? I mostly need this so i can serialize to std::vector<uint8_t> without having to go via std::vector<char>. I have less than 1MB of RAM. So i need to avoid unecessary copies at all costs.

pfeatherstone:
Ok added tests and all works. @davisking I think we are ready for review.

pfeatherstone:
I think i'm only going to provide upwards compatibility for dlib::matrix and all things in dnn as i think those are the things that users use most. Nothing is broken, but if someone wants to serialize an object to ```vector<int8_t>``` for example, and that object only provides ```serialize``` and ```deserialize``` function for ```std::ostream``` and ```std::istream``` respectively, then it won't work. However, if they serialize to ```vector<char>```, then everything is good.

pfeatherstone:
I can propagate the changes to other objects but i think ```dlib::matrix``` and dnn networks should make most people happy. It's not a big deal templating ```serialize``` and ```deserialize``` for other objects as and when people complain they can't (de)serialize objects to/from ```vector<(u)int8_t>```

pfeatherstone:
```DLIB_DEFINE_DEFAULT_SERIALIZATION``` has been updated so any user code that uses that is automatically updated

davisking:
I'm not sure why so much stuff is changing here.  If the point is just to make `vectorstream` work for `uint8_t` then nothing but the `vectorstream` files should change.

pfeatherstone:
I don't think you can easily make ```iostream``` use a ```uint8_t``` buffer for example. The underlying character type is ```char```

davisking:
It's also important to not break the API.  Existing user code should not be broken.  But this change would break a wide range of user code.

pfeatherstone:
It wouldn't break anything if they kept on using normal ```iostream```

davisking:
`iostream` can read and write to anything.  It doesn't have to know the underlying buffer is `uint8_t` or any other thing.  That's the point of the `streambuf`.  It provides an interface that hides those details from the stream.

pfeatherstone:
So it doesn't break existing user code.

davisking:
All user code that uses `vectorstream` will fail to compile.

pfeatherstone:
> All user code that uses `vectorstream` will fail to compile.

Ah yes, you're right

davisking:
Not to mention you can see examples of all the other stuff you have been changing to fix these other rippling compile errors.

pfeatherstone:
Well those were mainly to update ```dlib::matrix``` and dnn stuff. i could have left it alone and everything would still work with ```iostream```

pfeatherstone:
But i see your point, you can't use ```vectorstream``` anymore without specifying a template parameter.

pfeatherstone:
And i wasn't aware i could just change streambuf inside vectorstream. Balls

pfeatherstone:
It would be much easier to just change ```vectorstream```'s ```streambuf``` class.

pfeatherstone:
doesn't ```basic_iostream``` dictate the template parameters of it's underlying ```basic_streambuf``` ?

pfeatherstone:
Like can you make ```iostream``` work with ```basic_streambuf<uint8_t>``` ?

davisking:
You don't need to use `basic_streambuf` or `basic_iostream` here.  And you definitely don't want to.  Those other streams are to handle unicode.  They aren't for handling this kind of "make stream write to some other source" problem.

pfeatherstone:
> You don't need to use `basic_streambuf` or `basic_iostream` here. And you definitely don't want to. Those other streams are to handle unicode. They aren't for handling this kind of "make stream write to some other source" problem.

Ok, Maybe i over complicated things to hastily. My initial thought was "i need to template the character type,  so i need to template the streambuf type, so i need to template the basic_iostream type". But i see your point now.

pfeatherstone:
I might start from a clean slate.

pfeatherstone:
I should stop looking at boost for inspiration.

pfeatherstone:
https://github.com/davisking/dlib/pull/2301

PR #2301, opened 2021-02-08
Serialization to and from vector<int8_t> and vector<uint8_t>

pfeatherstone:
Hopefully this time it's done properly.

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -2270,6 +2270,20 @@ namespace dlib
         }
         
         explicit proxy_serialize (
+            std::vector<int8_t>& buf
+        ) : fout_optional_owning_ptr(new vectorstream(buf)),
+            fout(*fout_optional_owning_ptr)
+        {
+        }
+        
+        explicit proxy_serialize (
+            std::vector<uint8_t>& buf
+        ) : fout_optional_owning_ptr(new vectorstream(buf)),
+            fout(*fout_optional_owning_ptr)
+        {
+        }
+        
+        explicit proxy_serialize (
             std::ostream& ss
         ) : fout_optional_owning_ptr(nullptr),
             fout(ss)
@@ -2303,6 +2317,22 @@ namespace dlib
         
         explicit proxy_deserialize (
             std::vector<char>& buf
+        ) : fin_optional_owning_ptr(new vectorstream(buf)),
+            fin(*fin_optional_owning_ptr)   
+        {
+            init();
+        }
+        
+        explicit proxy_deserialize (
+            std::vector<int8_t>& buf
+        ) : fin_optional_owning_ptr(new vectorstream(buf)),
+            fin(*fin_optional_owning_ptr)   
+        {
+            init();
+        }
+         
+        explicit proxy_deserialize (
+            std::vector<uint8_t>& buf
         ) : fin_optional_owning_ptr(new vectorstream(buf)),
             fin(*fin_optional_owning_ptr)   
         {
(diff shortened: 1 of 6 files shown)


pfeatherstone:
@davisking Hopefully this is better. I'm not overwhelmed by the fact that vectorstream has three dummy vectors and 3 streambuf's, one for char, int8_t and uint8_t. I though about using dlib::type_safe_union but i think this is fine and there isn't a huge overhead i don't think.  I've kept the tests for dnn and matrix as they are good sanity checks.

davisking:
Na, that's a good way to do it.  I'll look it over more carefully later, but looks good from an initial quick review.

pfeatherstone:
Not a huge fan that the ```iostream``` base class gets constructed with a pointer to a ```streambuf``` object that hasn't been constructed yet. Not great, but ```vectorstream``` was doing that before anyway.

pfeatherstone:
@davisking any thoughts?

pfeatherstone:
Ok it turns out you can pass a null pointer to ```iostream``` constructor so you can delay setting the ```streambuf``` pointer. That's good. Now we can use unique pointers and not need dummy buffers. I think this is better.
I don't like the fact that this implementation has the overhead of a heap allocation but maybe that's asking too much. I did think of using ```dlib::type_safe_union``` to contain all three types of streambuf objects but you can't use it because of cyclic header dependencies. So then i thought of using ```stack_based_memory_block``` and using placement new but things were getting a little complicated and cluttered and, in the end, I favored code readability over saving 1 heap allocation. At the end of the day, we're not doing embedded.

pfeatherstone:
If someone complains of the extra heap allocation in the future, we can always tell them to get a life, or we can use ```stack_based_memory_block``` and make the code look ugly.

davisking:
Yeah it's probably better to use an explicit call to `rdbuf()`.   I just looked and can't find a guarantee that it's ok to do what was here before.  So that's an improvement.  But I'm fairly sure there are people who will legitimately care about the extra `new` here now though.  Why not have the `dummy` variables and just use `rdbuf()` to set the appropriate stream?  Then there is no `new`.

pfeatherstone:
Can do. But then there are 3 vectors and 3 streambufs. There might not be any new calls but there is some object overhead.

pfeatherstone:
@davisking, if we are going to care about object overhead and heap allocations, the latest commit hopefully deals with both of those concerns. Let me know what you think.

pfeatherstone:
Hopefully it's not "double plus ungood"

davisking:
Ha, ok, that's clever but way too complicated.  The `sizeof(vectorstream)` with some dummy vectors is going to be tiny enough to be insignificant and there is no construction overhead or dynamic allocation inside std::vector's default construtor.  So just having 3 dummy std::vectors is essentially free and a lot simpler.

pfeatherstone:
@davisking Yeah i thought you might say that. It's just this way we have minimal object and heap allocation overhead. It comes at the cost of some ugly code. I will revert it back to what we had before.

pfeatherstone:
@davisking Ready for review whenever you have time.

pfeatherstone:
@davisking Just FYI, ```sizeof(vectorstream)``` is now 600 bytes whereas before, using ```std::aligned_union<storage_capacity,T0,T1,T2>::type mem;``` and placement new, ```sizeof(vectorstream)``` was 376 bytes. So there is quite an overhead there. So if people care about the size of their stack, then the latest version might be an issue. 600 bytes for vectorstream, which is a glorified wrapper around ```std::vector``` seems excessive.

pfeatherstone:
Yeah it all adds up. ```sizeof(std::iostream) == 288```, ```sizeof(std::vector<char>) == 24```, ```sizeof(vector_streambuf) == 80```. So if we use 3 dummy vectors and 3 vector_streambufs, you end up with ```sizeof(vectorstream) == 600```. If we use placement new with ```std::aligned_union``` you end up with ```sizeof(vectorstream) == 376```

So it's a matter of whether you want to sacrifice stack size for code readability. At the end of the day, I don't mind too much. L1 cache is pretty big these days. But if people care about unnecessary heap allocations, they might care about this too. 

But this has been a good experiment. And it shows that harmless objects like ```std::vector``` are actually quite big, probably because they have the "small vector" optimization with some small stack capacity. And iostream objects are enormous. No wonder why some companies don't like them.

pfeatherstone:
@davisking on another note, do you think there is user code out there that has something like:
```static_assert(sizeof(vectorstream) == 376), "poop");``` ? The latest version would break that. Whereas the one i suggested with placement new doesn't break that. I know that's an extreme example, but just another thought. Again, I don't mind which you choose, these are just things to consider.

davisking:
An extra 224 is small enough to be inconsequential for typical stack sizes.  So I'm not worried about it.

I would be really surprised if there is any code anywhere that does `static_assert(sizeof(vectorstream) == 376), "poop")`.  That would be a really bizarre and non-portable thing to have in your code.

davisking:
Anyway, this is good.  Thanks for another PR :)

pfeatherstone:
@davisking I've just realised I didn't update ```vectorstream_abstract.h```. I think it was out of date anyway from a previous PR that fixed casting issues to ```iostream&```

PR #2304, opened 2021-02-10
fix set_learning_rate_multipliers_range not working

arrufat:
I guess this was broken at some point, but it works now :)


arrufat:
It looks like it has been broken since this commit https://github.com/davisking/dlib/commit/40c3e48818b6decb97d31b902471933010bd6ff1

Tests coming up :)

davisking:
Thanks for another PR :)

PR #2306, opened 2021-02-12
[SERIALIZATION] updated vectostream_abstract

pfeatherstone:
@davisking Sorry i didn't do this in the last PR.

### DIFF
--- a/dlib/vectorstream/vectorstream_abstract.h
+++ b/dlib/vectorstream/vectorstream_abstract.h
@@ -42,18 +42,24 @@ namespace dlib
                   be used.  Therefore, any time data is written to this stream it will
                   immediately show up in the buffer.
         !*/
-
-        std::istream& seekg (
-            std::streampos pos
+        
+        vectorstream (
+            std::vector<int8_t>& buffer
         );
         /*!
             ensures
-                - The next read from this object will read from the position buffer[pos],
-                  where buffer is the std::vector given to this object's constructor.  Note
-                  that if pos >= buffer.size() then the next read will simply return EOF.
-                - returns *this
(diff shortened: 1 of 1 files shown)


davisking:
No worries, thanks for the PR :)

I thought about how this was missing early in the PR but by the time I got to merging it forgot about it too so ¯\_(ツ)_/¯

pfeatherstone:
@davisking Would you also be interested in enhancing ```dlib::vectorstream``` to support ```std::shared_ptr<char>``` (and int8_t, uint8_t) as the underlying buffer with some additional code for wrapping said buffer, carefully resizing it, etc? I have a use case where i need to serialize to ```shared_ptr<char>```, and again i have 2MB of RAM. So minimal copies and memory allocations is a priority. Now i could just serialize to ```vector<char>``` then copy at the end. That would maybe work and hopefully i wouldn't run into an "out of memory" exception. But just wondering if other people would have a need to serialize to ```shared_ptr<char>``` too, in which case adding this functionality could be useful. 

Here is an example of what i mean:

[vectorstream.h.txt](https://github.com/davisking/dlib/files/5980910/vectorstream.h.txt)

davisking:
Can't you use `ostrstream` for this then?  That's basically what it is https://en.cppreference.com/w/cpp/io/ostrstream.  It's deprecated but I have a feeling it's never actually going away.  But it's deprecated because this kind of API is pretty easy for users to screw up.

pfeatherstone:
Erh, that's the first time i've ever seen something deprecated in C++. Hmm, maybe i could use it. I've submitted another PR that gives you an idea of what i had in mind. I'm still not sure how i feel about it.

PR #2308, opened 2021-02-15
[SERIALIZATION] use std::shared_ptr<char> as a buffer

pfeatherstone:
This PR adds serialization support to and from ```std::shared_ptr<char>``` (and ```int8_t``` and ```uint8_t```). Now for most people, if they need to pass around ```std::shared_ptr<char>``` as a container for serialized data for whatever reason, they can just serialize to ```std::vector<char>``` then copy the data. Some people might argue they could do without the unecessary copy at the end. This PR addresses this. I haven't convinced myself that this is totally a good idea. It clutters ```dlib::vectorstream``` quite a bit and really ```std::shared_ptr``` is outside its scope. One could argue that another object similar to ```dlib::vectorstream``` would be more appropriate and could be generalized to support any "vector" like container, i.e. containers that satisfy a set of requirements, like has ```push_back```, has ```insert```, has ```size```, stuff like that. I don't know. These are all ideas. So please discuss. 
In my case, i need to pass around ```std::shared_ptr```, and ideally i would like to do without the copy at the end.

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -2284,6 +2284,27 @@ namespace dlib
         }
         
         explicit proxy_serialize (
+            dlib::shared_buf<char>& buf
+        ) : fout_optional_owning_ptr(new vectorstream(buf)),
+            fout(*fout_optional_owning_ptr)
+        {
+        }
+        
+        explicit proxy_serialize (
+            dlib::shared_buf<int8_t>& buf
+        ) : fout_optional_owning_ptr(new vectorstream(buf)),
+            fout(*fout_optional_owning_ptr)
+        {
+        }
+        
+        explicit proxy_serialize (
+            dlib::shared_buf<uint8_t>& buf
+        ) : fout_optional_owning_ptr(new vectorstream(buf)),
+            fout(*fout_optional_owning_ptr)
+        {
+        }
+        
+        explicit proxy_serialize (
             std::ostream& ss
         ) : fout_optional_owning_ptr(nullptr),
             fout(ss)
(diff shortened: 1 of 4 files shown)


pfeatherstone:
Also, the algorithm i've chosen for capacity growth in ```dlib::shared_buf``` is fairly simple (increase to powers of 2). Maybe ```std::vector``` uses something more clever (maybe golden ratio) and really, if i were to serialize to ```std::vector<char>``` first then do a copy at the end to ```shared_ptr<char>``` i might possibly do fewer allocations, and smaller allocations on average. I'm not sure.

pfeatherstone:
Does anybody want this? It wouldn't be an issue if some libraries didn't think it was trendy to use ```std::shared_ptr<char>``` as buffers because they can't be bothered to use move semantics or copy semantics.

pfeatherstone:
@davisking Feel free to rip this apart, i don't mind. I see this more as a proof of concept.

davisking:
I don't totally get the use-case for this.  Like `dlib::shared_buf` is basically a simple `std::vector`.  So why doesn't someone use `std::vector` instead of `dlib::shared_buf`?  It's not like `dlib::shared_buf` is using the buffer the user gives it since it might reallocate and make a new buffer at any point.

pfeatherstone:
yeah, my use case is i need an iostream object with ```shared_ptr``` as the underlying buffer since a third-party library requires a ```shared_ptr```. I don't think dlib should care about this. I will just copy paste ```dlib::vectorstream``` and repurpose is. It doesn't need to live in dlib. Or i will use ```dlib::vectorstream``` with ```std::vector``` and copy the data at the end and hope i don't run out of memory.

pfeatherstone:
And actually i don't think i'm gonna need an iostream object after all.

PR #2312, opened 2021-02-23
make mish activation an inplace layer

arrufat:
I noticed the the Mish layer, contrary to other activation layers, was not performing its computations inplace.

This PR fixes that. I didn't notice much of a speed gain, but on YOLOv4 the model uses around 200 MiB less of VRAM, so that's a win.

Also, I have a question about the `computed_output` parameter in the `backward_inplace` method: The documentation says that we can "_exclude the computed_output parameter from backward_inplace().  Doing this will allow dlib to make some layers execute in-place and therefore run a little faster and use less memory_".

I haven't seen any activation do that, only `dropout_`, `multiply_` and `affine_`. Is there a reason for that?

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -3331,24 +3331,19 @@ namespace dlib
         {
         }
 
-        template <typename SUBNET>
-        void forward(
-            const SUBNET& sub,
-            resizable_tensor& data_output
-        )
+        void forward_inplace(const tensor& input, tensor& output)
         {
-            data_output.copy_size(sub.get_output());
-            tt::mish(data_output, sub.get_output());
+            tt::mish(output, input);
         }
 
-        template <typename SUBNET>
-        void backward(
+        void backward_inplace(
+            const tensor& computed_output,
             const tensor& gradient_input,
-            SUBNET& sub,
+            tensor& data_grad,
             tensor&
         )
         {
-            tt::mish_gradient(sub.get_gradient_input(), sub.get_output(), gradient_input);
+            tt::mish_gradient(data_grad, computed_output, gradient_input);
         }
 
         inline dpoint map_input_to_output (const dpoint& p) const { return p; }
--- a/dlib/dnn/layers_abstract.h
+++ b/dlib/dnn/layers_abstract.h
@@ -2402,8 +2402,8 @@ namespace dlib
         );
 
         template <typename SUBNET> void setup (const SUBNET& sub);
-        template <typename SUBNET> void forward(const SUBNET& sub, resizable_tensor& data_output);
-        template <typename SUBNET> void backward(const tensor& gradient_input, SUBNET& sub, tensor&);
+        void forward_inplace(const tensor& input, tensor& output);
+        void backward_inplace(const tensor& computed_output, const tensor& gradient_input, tensor& data_grad, tensor& params_grad);
         dpoint map_input_to_output(dpoint p) const;
         dpoint map_output_to_input(dpoint p) const;
         const tensor& get_layer_params() const;


arrufat:
Oh, it's not passing the layer tests... Weird, maybe I missed something (gelu layer also has the same problem)

arrufat:
I've testd YOLOv4x-Mish trained on COCO with an input of 640x640 on an NVIDIA GeForce GTX 1080 Ti.
For info, this model has 134 Mish layers.

| framework       | VRAM (MiB) | fps |
|-----------------|:----------:|-----|
| darknet         |    1783    |  22 |
| dlib-master     |    2105    |  17 |
| dlib-fused_mish |    1558    |  17 |

So, non-negligible improvement there on memory requirements.
However, I still don't know why the `backward_inplace` is not passing the tests... The `forward_inplace` method works fine, though...

arrufat:
I've tested YOLOv4x-Mish trained on COCO with an input size of 640×640 on an NVIDIA GeForce GTX 1080 Ti.
For info, this model has 134 Mish layers.

| framework       | VRAM (MiB) | fps |
|-----------------|:----------:|-----|
| darknet         |    1783    |  22 |
| dlib-master     |    2105    |  17 |
| dlib-fused_mish |    1558    |  17 |

So, non-negligible improvement there on memory requirements.
However, I still don't know why the `backward_inplace` is not passing the tests... The `forward_inplace` method works fine, though...

arrufat:
I've tested YOLOv4x-Mish trained on COCO with an input size of 640×640 on an NVIDIA GeForce GTX 1080 Ti.
For info, this model has 134 Mish layers.

| framework       | VRAM (MiB) | fps |
|-----------------|:----------:|-----|
| darknet         |    1783    |  22 |
| dlib-master     |    2105    |  17 |
| dlib-mish_inplace |    1558    |  17 |

So, non-negligible improvement there on memory requirements.
However, I still don't know why the `backward_inplace` is not passing the tests... The `forward_inplace` method works fine, though...

arrufat:
as a test, I run the `dnn_introduction_ex.cpp` with the default network definition:

``` cpp
    using net_type = loss_multiclass_log<
                                fc<10,        
                                relu<fc<84,   
                                relu<fc<120,  
                                max_pool<2,2,2,2,relu<con<16,5,5,1,1,
                                max_pool<2,2,2,2,relu<con<6,5,5,1,1,
                                input<matrix<unsigned char>> 
                                >>>>>>>>>>>>;
```

And with the `relu` activations replaced with `mish` activations:

``` cpp
    using net_type = loss_multiclass_log<
                                fc<10,        
                                mish<fc<84,   
                                mish<fc<120,  
                                max_pool<2,2,2,2,mish<con<16,5,5,1,1,
                                max_pool<2,2,2,2,mish<con<6,5,5,1,1,
                                input<matrix<unsigned char>> 
                                >>>>>>>>>>>>;
```

Results with ReLU:
- training num_right: 59988
- training num_wrong: 12
- training accuracy:  0.9998
- testing num_right: 9902
- testing num_wrong: 98
- testing accuracy:  0.9902

Results with Mish:
- training num_right: 59982
- training num_wrong: 18
- training accuracy:  0.9997
- testing num_right: 9905
- testing num_wrong: 95
- testing accuracy:  0.9905

So it seems that the `backward_inplace` method is working...

davisking:
> I noticed the the Mish layer, contrary to other activation layers, was not performing its computations inplace.
> 
> This PR fixes that. I didn't notice much of a speed gain, but on YOLOv4 the model uses around 200 MiB less of VRAM, so that's a win.
> 
> Also, I have a question about the `computed_output` parameter in the `backward_inplace` method: The documentation says that we can "_exclude the computed_output parameter from backward_inplace(). Doing this will allow dlib to make some layers execute in-place and therefore run a little faster and use less memory_".
> 
> I haven't seen any activation do that, only `dropout_`, `multiply_` and `affine_`. Is there a reason for that?

Sure they do.  Like relu uses backward_inplace.

dlib/dnn/layers.h
```diff
@@ -3331,24 +3331,19 @@ namespace dlib
...
+        void backward_inplace(
+            const tensor& computed_output,
             const tensor& gradient_input,
-            SUBNET& sub,
+            tensor& data_grad,
             tensor&
         )
         {
-            tt::mish_gradient(sub.get_gradient_input(), sub.get_output(), gradient_input);
+            tt::mish_gradient(data_grad, computed_output, gradient_input);
```
davisking, line 3346: `tt::mish_gradient(data_grad, computed_output, gradient_input);`

isn't equivalent to `tt::mish_gradient(sub.get_gradient_input(), sub.get_output(), gradient_input);` since `computed_output` isn't `sub.get_output()`.  `computed_output` is the output from `tt::mish()` in this case.   You would need a different gradient function to compute it in-place.

arrufat:
> Sure they do. Like relu uses backward_inplace.

I was referring to the fact that the inplace activation layers like `relu_` don't omit the `computed_output` parameter, but other inplace layers like `dropout_` do. Does this mean that we would need to cache the output of the forward pass in the state of the layer before the forward pass? Is that what you're referring to in the documentation? That part is not clear to me.

dlib/dnn/layers.h
arrufat: Oh, I thought this was handled here:
https://github.com/davisking/dlib/blob/3162f93c5d303e593337ab8e40d79b385e025d3c/dlib/cuda/cuda_dlib.cu#L1457-L1481

At least that's how `leaky_relu` handles the inplace case... But I am clearly missing something

davisking: Right, but `computed_output` just isn't the same thing as `sub.get_output()`.  This PR changes one of the input from `sub.get_output()` to `computed_output`, so the output isn't going to be the same.

davisking:
> > Sure they do. Like relu uses backward_inplace.
> 
> I was referring to the fact that the inplace activation layers like `relu_` don't omit the `computed_output` parameter, but other inplace layers like `dropout_` do. Does this mean that we would need to cache the output of the forward pass in the state of the layer before the forward pass? Is that what you're referring to in the documentation? That part is not clear to me.

The framework holds onto `computed_output` when needed.  Layer code doesn't need to do anything, the stuff in core.h deals with it.  I'm just saying in the docs that you get to see the output of the forward layer if you want to see it.  If it's not useful to the computation then layers shouldn't ask to see it.

arrufat:
> > > Sure they do. Like relu uses backward_inplace.
> > 
> > 
> > I was referring to the fact that the inplace activation layers like `relu_` don't omit the `computed_output` parameter, but other inplace layers like `dropout_` do. Does this mean that we would need to cache the output of the forward pass in the state of the layer before the forward pass? Is that what you're referring to in the documentation? That part is not clear to me.
> 
> The framework holds onto `computed_output` when needed. Layer code doesn't need to do anything, the stuff in core.h deals with it. I'm just saying in the docs that you get to see the output of the forward layer if you want to see it. If it's not useful to the computation then layers shouldn't ask to see it.

Great, I get it now, thank you for the reply :)

dlib/dnn/layers.h
arrufat: Oh, I thought that was also handled automatically by the framework (when calling the `_inplace` variants).
So, I need to figure out how to update the gradient call now... I've been staring for an embarrassingly long time at both cuda mish implementations and leaky_relu implementations, and I am having a hard time figuring out why it doesn't work.
I get your point, that before I was acting on the input of the mish layer, and now I am acting on the ouput of the mish layer, so the gradients must obviously be different.

davisking: It’s a matter of computing the gradient. The existing mish gradient function just doesn’t compute the gradient inplace since it requires knowing the old inputs, which will have been overwritten if the layer is an inplace layer. So you need a new function.

arrufat: Thank you for the explanation, and now I see that the `leaky_relu` gradient only depends on the `gradient_input`, but I find it a bit confusing that the method is called gradient_inplace, does that mean that the current inplace implementation in dlib is wrong?

arrufat: I mean this one:
https://github.com/davisking/dlib/blob/3162f93c5d303e593337ab8e40d79b385e025d3c/dlib/cuda/cuda_dlib.cu#L1457-L1461

arrufat:
After giving it a second look with rested eyes (and your advice), I think the gradient can't be computed inplace because the Mish function is not bijective.

Sorry for the noise...

davisking:
Yeah it's not super obvious to me there is a nice way to compute it
inplace.  To be clear, let's say a layer is a function y = f(x). The thing
that allows a layer to be computed inplace in the backwards pass is that
the gradient computation doesn't need to know x.  When running inplace you
only get y (which is stored in the computed_output) argument.  If you can
compute dy/dx without knowing x (and optionally having y) then the layer
can be done in place.

arrufat:
Exactly, yesterday I was looking at the gradient of the Mish layer (f(x)), and checked if I could implement it by only knowing f(x), so I tried to factor out f(x) from the gradient, but there were always trailing x's. So I guess it's not possible, at least to the extent of my math skills. Then I realized that the function is not one-to-one, so I could never express the gradient in terms of f(x).

PR #2317, opened 2021-03-01
remove wrong inplace implementations

arrufat:
If I understood it correctly in #2312, there are some layers that have inplace implementations that are wrong.
Because of the nature of the functions, they can't be implemented inplace. This PR removes those functions.

### DIFF
--- a/dlib/cuda/cpu_dlib.cpp
+++ b/dlib/cuda/cpu_dlib.cpp
@@ -1678,16 +1678,8 @@ namespace dlib
                 return e*omega/(delta*delta);
             };
 
-            if (is_same_object(gradient_input, grad))
-            {
-                for (size_t i = 0; i < src.size(); ++i)
-                    g[i] = in[i]*calculate_gradient(s[i]);
-            }
-            else
-            {
-                for (size_t i = 0; i < src.size(); ++i)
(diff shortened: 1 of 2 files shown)


davisking:
These functions don't look wrong though.  Like they do what their contracts in tensor_tools.h says they should, or that's what it appears anyway.  What about them is incorrect?

arrufat:
Can they be used in dlib? As far as I know, we can't express the gradient of the Mish and GELU functions without referring to their inputs, so we can't implement them inplace. Maybe I misunderstood something and I'm wrong, but how can they be used inplace?

davisking:
See my comment on #2312 just now.  These function's won't help you make an inplace layer.  Like you won't be able to make backwards_inplace.  But they do exactly what their contract in tensor_tools.h says they should (seemingly).  The only thing that matters for determining if they are wrong or right is what their contracts say.  Not how they might or might not be useful in some other context.

Anyway, TLDR the contract is always the source of truth :)    Doesn't mean we can't change the contract though, and then also change the functions.  But the contract seems fine.

arrufat:
Ok I see what you mean. I got carried away without reading the contract. It's just since I can't figure out how we can use them inplace... Closing. Thank you for the explanation :)

PR #2320, opened 2021-03-06
Add softmax function for matrix type

### LINKED ISSUE #2319: Softmax for matrix class
Hi! Just some quick questions on how I would go about adding a softmax operation to the amazing matrix class in dlib.
(I know I can just use `tt::softmax(out, in)` where `out` and `in` are tensors and I can just use the `mat` function).

Here's what I got so far, which already works:

``` cpp
namespace dlib
{
    template <typename M>
    struct op_softmax
    {
        op_softmax(const M& m_) : m(m_)
        {
            static_assert((is_same_type<type, float>::value == true ||
                           is_same_type<type, double>::value == true ||
                           is_same_type<type, long double>::value == true));
            denom = sum(exp(m));
        }
        const M& m;
        const static long cost = M::cost + 9;
        const static long NR = M::NR;
        const static long NC = M::NC;
        typedef typename M::type type;
        typedef typename M::mem_manager_type mem_manager_type;
        typedef typename M::layout_type layout_type;
        typedef type const_ret_type;
        const_ret_type apply(long r, long c) const { return std::exp(m(r, c)) / denom; }
        long nr() const { return m.nr(); }
        long nc() const { return m.nc(); }
        template <typename U>
        bool aliases(const matrix_exp<U>& item) const { return m.aliases(item); }
        template <typename U>
        bool destructively_aliases(const matrix_exp<U>& item) const { return m.destructively_aliases(item); }
        private:
        type denom;
    };
    template <typename M> const matrix_op<op_softmax<M>> soft_max(const matrix_exp<M>& m)
    {
        typedef op_softmax<M> op;
        return matrix_op<op>(op(m.ref()));
    }
}
```
The big question is: am I very wrong?
The precise doubts I have are:
- will that static assert go there?
- about the cost: I saw that `exp` has a cost of 7, so here we're basically doing `exp` twice and a division, hence the 9, but there's also a sum there, what should I use as cost here?
- if you're interesting in having it in dlib, how would you name it? `softmax` clashes with the dnn layer.

EDIT: also, I would like to cache the result of applying `exp` to the input matrix, but I didn't succeed yet.

arrufat:
Closes #2319.

Still not convinced about the function name, though.

### DIFF
--- a/dlib/matrix/matrix_math_functions_abstract.h
+++ b/dlib/matrix/matrix_math_functions_abstract.h
@@ -600,7 +600,7 @@ namespace dlib
                 - R::type == the same type that was in m
                 - R has the same dimensions as m
                 - for all valid r and c:
-                  R(r,c) == std::exp(m(r,c) / sum(std::exp(m))
+                  R(r,c) == std::exp(m(r,c)) / sum(std::exp(m))
     !*/
 
 // ----------------------------------------------------------------------------------------
--- a/dlib/matrix/matrix_math_functions_abstract.h
+++ b/dlib/matrix/matrix_math_functions_abstract.h
@@ -589,6 +589,22 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    const matrix_exp soft_max (
+        const matrix_exp& m
+    );
+    /*!
+        requires
+            - matrix_exp::type == float, double, or long double
+        ensures
+            - returns a matrix R such that:
+                - R::type == the same type that was in m
+                - R has the same dimensions as m
+                - for all valid r and c:
+                  R(r,c) == std::exp(m(r,c) / sum(std::exp(m))
+    !*/
+
+// ----------------------------------------------------------------------------------------
+
 }
 
 #endif // DLIB_MATRIx_MATH_FUNCTIONS_ABSTRACT_
--- a/dlib/matrix/matrix_math_functions.h
+++ b/dlib/matrix/matrix_math_functions.h
@@ -445,15 +445,16 @@ namespace dlib
     template <typename M>
     struct op_softmax : basic_op_m<M>
     {
-        op_softmax(const M& m_) : basic_op_m<M>(m_), exps(exp(this->m - max(this->m))), denom(sum(exps)) {}
+        typedef typename M::type type;
+
+        op_softmax(const M& m_, const type& d_, const type& v_) : basic_op_m<M>(m_), d(d_), v(v_){}
+
+        const type d;
+        const type v;
 
         const static long cost = M::cost + 9;
-        typedef typename M::type type;
         typedef type const_ret_type;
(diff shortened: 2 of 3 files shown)


pfeatherstone:
This will be useful! Might need to update the unit tests.

davisking:
This looks fine, but definitely just make it a normal function since that's simpler.  Like I said in the issue, since the expression is doing everything and caching the result it's not doing anything more efficient that a normal function would do.  And a normal function's code is way easier to read.  There is also more opportunity for RVO or move construction with a normal function.  The expression stuff is only what you want when you can do it without needing to cache the result up front.

arrufat:
OK, I see your point. However, if I make it a function, would I be able to pass matrix expressions as parameters? and return matrix expressions?

To be honest, I was interested in this feature, because I was calibrating a multiclass classifier using [temperature scaling](https://arxiv.org/abs/1706.04599). In the paper, they divide the logits by K before applying softmax. I tried it and it works well.

I started implementing it as a templated function, but I couldn't do something like:
```
out = softmax(logits / k)
```
Where `k` is a scalar. That's why I decided to implement it as a matrix expression.
I just pushed a change that makes it behave more like the `normalize_op`. 

Let me know if that would be OK, or how would I go with implementing softmax as a function that can take matrix expressions as inputs.

arrufat:
Converted division into multiplication after reading [this](https://cppbenchmarks.wordpress.com/2020/11/10/float-division-vs-multiplication-speed/).

dlib/matrix/matrix_math_functions_abstract.h
```diff
@@ -587,6 +587,22 @@ namespace dlib
...
+    );
+    /*!
+        requires
+            - matrix_exp::type == float, double, or long double
+        ensures
+            - returns a matrix R such that:
+                - R::type == the same type that was in m
+                - R has the same dimensions as m
+                - for all valid r and c:
+                  R(r,c) == std::exp(m(r,c) / sum(std::exp(m))
```
davisking, line 603: There is a typo here.  It should be std::exp(m(r,c)) / sum(std::exp(m))

arrufat: fixed! :)

davisking:
What I meant before was just to do this kind of thing 
```
template <
    typename EXP
    >
typename matrix_exp<EXP>::matrix_type soft_max (
    const matrix_exp<EXP>& m
)
{
    const auto largest = max(m);
    return exp(m - largest) / sum(exp(m - largest));
}
```
Which takes a matrix expression and returns a concrete matrix instance.   But yeah, I guess what you have does return an expression which could potentially be faster.  I don't normally make functions like this into expressions because if feels over the top.  But hey, it's potentially faster for some contexts so sure :)

arrufat:
Ah, I see, now I know for the next time. Anyway I learned a bit on how this matrix expressions work. Thank you again :)

davisking:
No problem :)

PR #2335, opened 2021-03-30
Add letterbox image

arrufat:
Would you be interested in this kind of image transformation?
I have some doubts about this implementation:
- it's named `letterbox_image` but it also scales the image to fit a square of `size` × `size`
- I did this because it's the more common case and the user can always set it to `std::max(img_in.nr(), img_in.nc())`
- but maybe there should be an overload?

I have the (long term) goal of implementing YOLO training in dlib at some point in my life... This would come in handy, since YOLO trains with square images, and this would allow to convert the bounding boxes between the original and the letterboxed images.
I've already tried it locally initializing a `rectangle_transform` with the output of this function, and it's pretty convenient.

### DIFF
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -996,13 +996,11 @@ namespace dlib
         image_type temp;
         image_view<image_type> vtemp(temp);
         vtemp.set_size(std::round(scale * vimg_in.nr()), std::round(scale * vimg_in.nc()));
-        resize_image(vimg_in, temp);
         dpoint offset((size - vtemp.nc()) / 2.0, (size - vtemp.nr()) / 2.0);
-        const auto tform = point_transform_affine(identity_matrix<double>(2) * scale, offset);
         const auto r = rectangle(offset.x(), offset.y(), offset.x() + vtemp.nc() - 1, offset.y() + vtemp.nr() - 1);
         auto si = sub_image(img_out, r);
-        assign_image(si, vtemp);
-        return tform;
+        resize_image(vimg_in, si);
+        return point_transform_affine(identity_matrix<double>(2) * scale, offset);
     }
 
 // ----------------------------------------------------------------------------------------
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -967,6 +967,46 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <typename image_type>
+    point_transform_affine letterbox_image (
+        const image_type& img_in,
+        image_type& img_out,
+        long size
+    )
+    {
+        DLIB_CASSERT(size > 0, "size must be bigger than zero, but was " << size);
+        // the scaling factor of the image
+        const auto scale = size / std::max<double>(img_in.nr(), img_in.nc());
+
+        // early return if the image has already the requested size and no padding is needed
+        if (scale == 1 and img_in.nr() == img_in.nc())
+        {
+            img_out = img_in;
+            return point_transform_affine();
+        }
+
+        // black background
+        img_out.set_size(size, size);
+        assign_all_pixels(img_out, 0);
+
+        // resize the image so that it fits into a size x size image
+        image_type temp = img_in;
+        resize_image(scale, temp);
+
+        // get the row and column offsets (the padding size)
+        const point offset((size - temp.nc()) / 2, (size - temp.nr()) / 2);
+        for (long r = 0; r < temp.nr(); ++r)
+        {
+            for (long c = 0; c < temp.nc(); ++c)
+            {
+                img_out(offset.y() + r, offset.x() + c) = temp(r, c);
+            }
+        }
+        return point_transform_affine(identity_matrix<double>(2) * scale, offset);
+    }
+
+// ----------------------------------------------------------------------------------------
+
     template <
         typename image_type1,
         typename image_type2
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -979,7 +979,7 @@ namespace dlib
         const auto scale = size / std::max<double>(img_in.nr(), img_in.nc());
 
         // early return if the image has already the requested size and no padding is needed
-        if (scale == 1 and img_in.nr() == img_in.nc())
+        if (scale == 1 && img_in.nr() == img_in.nc())
         {
             img_out = img_in;
             return point_transform_affine();
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -975,34 +975,34 @@ namespace dlib
     )
     {
         DLIB_CASSERT(size > 0, "size must be bigger than zero, but was " << size);
+        const_image_view<image_type> vimg_in(img_in);
+        image_view<image_type> vimg_out(img_out);
+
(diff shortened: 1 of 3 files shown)


dlib/image_transforms/interpolation.h
```diff
@@ -965,6 +965,46 @@ namespace dlib
...
+        image_type& img_out,
+        long size
+    )
+    {
+        DLIB_CASSERT(size > 0, "size must be bigger than zero, but was " << size);
+        // the scaling factor of the image
+        const auto scale = size / std::max<double>(img_in.nr(), img_in.nc());
+
+        // early return if the image has already the requested size and no padding is needed
+        if (scale == 1 and img_in.nr() == img_in.nc())
```
davisking, line 982: Ha, `and`.  :)

Use `&&` instead of `and` though.  `&&` is way more idiomatic in C++ than `and`.  I'm pretty sure like 95% of C++ developers don't even know you can use `and`.

dlib/image_transforms/interpolation.h
```diff
@@ -965,6 +965,46 @@ namespace dlib
...
+        {
+            img_out = img_in;
```
davisking, line 984: This won't work for `array2d` or other non-copyable images.  You want to use `assign_image()`

davisking: Yeah I guess the code is assuming the image is a matrix.  Make this use the generic image interface and it's all good.  Like use `const_image_view` and `image_view` to access the image contents.

dlib/image_transforms/interpolation.h
```diff
@@ -965,6 +965,46 @@ namespace dlib
...
+        // get the row and column offsets (the padding size)
+        const point offset((size - temp.nc()) / 2, (size - temp.nr()) / 2);
+        for (long r = 0; r < temp.nr(); ++r)
+        {
+            for (long c = 0; c < temp.nc(); ++c)
+            {
+                img_out(offset.y() + r, offset.x() + c) = temp(r, c);
+            }
+        }
+        return point_transform_affine(identity_matrix<double>(2) * scale, offset);
```
davisking, line 1005: This whole function could be done with one call to `transform_image()` if you supplied the appropriate `point_transform_affine` as the `map_point`.  Are you doing it this way because it's faster?  Like because it will hit the simd optimized `resize_image()`?  What's here is making multiple passes at the image, which is not good for performance, but maybe `resize_image()` is so fast it wins out?

Whatever is faster is fine.  But you should at least replace the for loops here with an `assign_image()` statement that uses `sub_image()`.  Like `assign_image(sub_image(img_out, some_rect), temp)`.

davisking:
Yeah it's cool with me.  You can do this with one call to `transform_image()` but nothing wrong with having a convenience function for doing it.  Add some tests for it though and make it work with the generic image interface though :)

dlib/image_transforms/interpolation.h
arrufat: Yes, I usually use `and` in my code, I know it's not that idiomatic, but I kind of like it :P.

arrufat:
Thank you so much again for the detailed feedback.

I've implemented both versions and tried on two images and averaged 1000 runs:
- `faces/bald_guys.jpg`: 1986 × 1545
- `faces/2008_002470.jpg`: 500 × 332

with `resize_image`:
- `faces/bald_guys.jpg`: 0.356763 ± 0.0888909 ms
- `faces/2008_002470.jpg`: 0.30611 ± 0.0883218 ms

with `transform_image`:
- `faces/bald_guys.jpg`: 0.809438 ± 0.144507 ms
- `faces/2008_002470.jpg`: 0.717556 ± 0.288066 ms

So, in my tests, `resize_image` is over than 2 times faster.
Also, notice that I am not copying the image any more and doing the in place resize, instead I allocate an image with the target size and resize to that function.

For reference, here's the code with `transform_image`:

``` c++
const auto nr = static_cast<long>(std::round(scale * vimg_in.nr()));
const auto nc = static_cast<long>(std::round(scale * vimg_in.nc()));
const point b((size - nc) / 2, (size - nr) / 2);
const auto r = rectangle(b.x(), b.y(), b.x() + nc - 1, b.y() + nr - 1);
const auto tform = point_transform_affine(identity_matrix<double>(2) * scale, b);
const auto invt = inv(tform);
transform_image(img_in, img_out, interpolate_bilinear(), invt, white_background(), r);
return tform;
```

dlib/image_transforms/interpolation.h
```diff
@@ -965,6 +965,46 @@ namespace dlib
...
+
+        // black background
+        vimg_out.set_size(size, size);
+        assign_all_pixels(vimg_out, 0);
+
+        // resize the image so that it fits into a size x size image
+        image_type temp;
+        image_view<image_type> vtemp(temp);
+        vtemp.set_size(std::round(scale * vimg_in.nr()), std::round(scale * vimg_in.nc()));
+        resize_image(vimg_in, temp);
```
davisking, line 999: To make this even faster you could ask resize_image to write directly into the output image by using sub_image.  Then you won't have to do that extra image copy at the end.

arrufat: I will try that later! We might end up with a function that is nicer than I expected.
At the beginning I was hesitant to send this PR because the function is somewhat trivial, but I ended up learning a lot, as always. Thank you so much! :D

davisking: No problem :)

arrufat: I've removed the extra copy, but I didn't notice any performance improvement. However, I feel better with myself.
I wonder if I also should do the same with the background assignment, just assign the padding, instead of the whole image...

davisking: Cool.  Yeah, using zero_border_pixels() to zero just the border pixels might be faster.

arrufat: Oh, I never noticed the existence of `zero_border_pixels`, I knew about `assign_border_pixels`. The fact that `zero_border_pixels`takes a rectangle is very convenient!

arrufat:
I think it's ready for review, despite the Travis failure (unrelated to this PR).

dlib/image_transforms/interpolation_abstract.h
```diff
@@ -436,6 +436,86 @@ namespace dlib
...
+            - image_type1 == an image object that implements the interface defined in
+              dlib/image_processing/generic_image.h
+            - image_type2 == an image object that implements the interface defined in
+              dlib/image_processing/generic_image.h
+            - interpolation_type == interpolate_nearest_neighbor, interpolate_bilinear,
+              interpolate_quadratic, or a type with a compatible interface.
+            - size > 0
+            - is_same_object(in_img, out_img) == false
+        ensures
+            - scales in_img so that it fits into a size * size square.
```
davisking, line 463: Add a line that specifically says img_out.nr() == size and img_out.nc() == size.  Otherwise it's not totally obvious that this is so.

davisking: Same for the other functions below too.  Like be very explicit about the output image sizes.

arrufat: done :)

davisking:
Sweet.  Thanks for another PR :)

arrufat:
Thank you for the guidance!

davisking:
No problem :)

PR #2337, opened 2021-03-30
Fix input/output mappings with repeat layers

### LINKED ISSUE #2336: Question regarding repeat layer and input_tensor_to_output_tensor visitor
So, I am trying to get a correspondence between input and output in a network.
It works well when there are no `repeat` layers in the network, otherwise, the visitor does nothing.

Is there an easy way to fix this?

Here's an example that showcases the problem:

``` c++
#include <dlib/dnn.h>

using namespace std;
using namespace dlib;

template <typename SUBNET>
using conblock = relu<bn_con<add_layer<con_<16, 3, 3, 2, 2, 1, 1>, SUBNET>>>;

using net_type_1 = loss_binary_log_per_pixel<con<1, 1, 1, 1, 1,conblock<conblock<conblock<tag1<input_rgb_image>>>>>>;
using net_type_2 = loss_binary_log_per_pixel<con<1, 1, 1, 1, 1,repeat<3, conblock, tag1<input_rgb_image>>>>;

{
    matrix<rgb_pixel> image(256, 256);
    net_type_1 net_1;
    net_type_2 net_2;
    net_1(image);
    net_2(image);
    point p(64, 64);
    cout << input_tensor_to_output_tensor(net_1, p) << endl;  // ouputs (8, 8), as expected
    cout << input_tensor_to_output_tensor(net_2, p) << endl;  // outputs (64, 64)
}
```

arrufat:
Closes #2336

### DIFF
--- a/dlib/test/dnn.cpp
+++ b/dlib/test/dnn.cpp
@@ -3972,6 +3972,20 @@ namespace
 
 // ----------------------------------------------------------------------------------------
 
+    template <typename SUBNET>
+    using conblock = relu<bn_con<add_layer<con_<16, 3, 3, 2, 2, 1, 1>, SUBNET>>>;
+
+    void test_input_ouput_mappers()
+    {
+        using net_type = loss_binary_log_per_pixel<con<1, 1, 1, 1, 1,repeat<3, conblock, tag1<input_rgb_image>>>>;
+        net_type net;
+        point p(32, 32);
+        DLIB_TEST(input_tensor_to_output_tensor(net, p) == p / 8);
+        DLIB_TEST(output_tensor_to_input_tensor(net, p) == p * 8);
+    }
+
+// ----------------------------------------------------------------------------------------
+
     // This test really just checks if the mmod loss goes negative when a whole lot of overlapping
     // truth rectangles are given.  
     void test_loss_mmod()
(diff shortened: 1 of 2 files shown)


dlib/test/dnn.cpp
```diff
@@ -3970,6 +3970,20 @@ namespace
...
     }
 
+// ----------------------------------------------------------------------------------------
+
+    template <typename SUBNET>
+    using conblock = relu<bn_con<add_layer<con_<16, 3, 3, 2, 2, 1, 1>, SUBNET>>>;
+
+    void test_input_ouput_mappers()
+    {
+        using net_type = loss_binary_log_per_pixel<con<1, 1, 1, 1, 1,repeat<3, conblock, tag1<input_rgb_image>>>>;
```
arrufat, line 3980: I don't know why, but I if don't put a tag as the subnet of the `repeat` layer, this fails to compile.

davisking: Yeah, input layers are special :|

davisking:
Nice, thanks for the fix :)

dlib/test/dnn.cpp
arrufat: I have learned to love them :)

PR #2338, opened 2021-03-30
Automatically build and publish binary wheels upon release

alesanfra:
Hi,
with this pull request I propose to leverage the Github Actions service to automatically build binary wheels and publish them to pypi.org. This is useful for people that might want to try dlib without the burden of compiling the project from scratch.

The jobs defined in the yaml file compile dlib for linux, macOS and windows operating systems on amd64 architecture, and  other os/architectures may be easily added afterwards (e.g. linux and macOS on arm64).

The linux wheels are fully [manylinux2010](https://www.python.org/dev/peps/pep-0571/) compliant, thus they can be installed on almost every linux distro released in the past 10 years.

The jobs are automatically triggered upon every GitHub release, but there is also the option to trigger them manually at any commit if needed.

Here you can find an [example of run](https://github.com/alesanfra/dlib/actions/runs/702756170) tested on the forked repo. Of course the "publish to pypi.org" step failed because I didn't set the proper pypi.org credentials. If this PR is merged in order to have working upload to pypi.org the administrator of this repo must add the PYPI_PASSWORD to the[ repo secrets](https://docs.github.com/en/actions/reference/encrypted-secrets).

The wheels are built with the default flags, so they require GUI support to work; we can discuss whether it makes sense to set DLIB_NO_GUI_SUPPORT=ON or not.

As a final note, I'm also the maintainer of [dlib-bin](https://pypi.org/project/dlib-bin/) package, which is just a binary distribution of dlib similar to what I'm proposing to build here. If you want to take a look to the repo to get an idea on how it works this is the repo: https://github.com/alesanfra/dlib-wheels .

Hope this will help the project, I'm available for questions and comments.

### DIFF
--- a/.github/workflows/build_python_wheels.yaml
+++ b/.github/workflows/build_python_wheels.yaml
@@ -1,0 +1,38 @@
+name: Build python wheels
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15, windows-2019]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+
+      - name: Build wheels
+        uses: joerick/cibuildwheel@v1.10.0
+        env:
+          CIBW_BEFORE_BUILD: 'pip install cmake'
+          CIBW_BUILD: 'cp37-* cp38-* cp39-*'
+          CIBW_ARCHS: auto64
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./wheelhouse/*.whl
+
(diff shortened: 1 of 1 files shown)


davisking:
That's pretty sweet.  I don't want to change the `dlib` pip target though, since it will break users who don't use GPUs right now.  Or if we went the other way and made the `dlib` pip package build without GPU support that would disrupt users who expect it to support GPUs.

How do you feel about making the `dlib-bin` pip package be a non-GPU package and `dlib-bin-gpu` a GPU enabled version?

alesanfra:
Hi, sorry for the late reply. 
Yes, basically the goal of this PR was to replace the `dlib` target in pypi.org, but I didn't realize that it could cause disruption. 
I guess the safest thing to do here is what you are proposing, I can keep the build of `dlib-bin` as it is now (gpu disabled), so that we can have both source and binary distributions on pypi.org.
Regarding the gpu-enabled distribution, I think I'll create a second GitHub action in the other repo, any help would be really appreciated :)

Thanks for reviewing this issue (and thanks for your work on this great tool!), I guess we can cancel the pull request.

PR #2345, opened 2021-04-12
Make clipped-relu inplace and fix docs for elu

arrufat:
I just noticed that in #2285 I didn't implement the `clipped_relu` layer as an in place layer, so this PR fixes that.
Also, the documentation for `clipped_relu` and `elu` was lying about their in place behavior, so I fixed that, as well.

Sorry for missing it in the first PR... :(


davisking:
Ha, right.  I didn't notice either :)

PR #2346, opened 2021-04-13
Make ELU an inplace layer and fix Clipped ReLU implementation

arrufat:
Sorry for spamming, I didn't notice that we could also make ELU an inplace layer...

### DIFF
--- a/dlib/cuda/cpu_dlib.cpp
+++ b/dlib/cuda/cpu_dlib.cpp
@@ -1940,7 +1940,7 @@ namespace dlib
                     if (in[i] > 0)
                         out[i] = gi[i];
                     else
-                        out[i] = alpha * std::exp(in[i]) * gi[i];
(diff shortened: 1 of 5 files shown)


arrufat:
I noticed one weird thing... in the [`test_elu`](https://github.com/davisking/dlib/blob/7f53d7feb671d5c2da4a39bba613e14b02fb951d/dlib/test/dnn.cpp#L311-L347) function, in particular in the gradient function.
Even if I call it like this, the test passes, the error between `grad_cuda` and `grad_cpu` is 0.

``` c++
cuda::elu_gradient(grad_cuda, dest_cuda, grad_input, alpha);
cpu::elu_gradient(grad_cpu, dest_cpu, grad_input, 0.5 * alpha);
```

I am sure the functions are implemented correctly, but I am doing something stupid in that test, and I am unable to see what...

davisking:
Good catch :)

PR #2347, opened 2021-04-14
replace a unicode character to suppress warnings

chokomancarr:
Hi, I replaced a quote character in the comments of image_transforms/morphological_operations.h,
such that when compiling on Windows, it does not complain of file containing unicode character which is not in current codepage.

It is annoying because this warning pops up for every file compiled that includes this file.


davisking:
Looks good. Thanks for the PR :)

PR #2349, opened 2021-04-21
[TYPE_SAFE_UNION] use std::aligned_union instead of stack_based_memory_block

pfeatherstone:
This PR changes the stack-based memory block used by ```dlib::type_safe_union``` from ```stack_based_memory_block``` to ```std::aligned_union```
I can foresee Davis asking: "Why change something that ain't broke?" 
That would be a good point.
Because, since C++11, ```std::aligned_union``` has been around to do precisely this kind of thing. So why not use it and deprecate ```stack_based_memory_block``` ? As a rule of thumb, if there is something in the standard library that does precisely what you want to do, then use it.
As an aside, I think using ```reinterpret_cast``` instead of ```static_cast``` is more correct and guaranteed to work since proper alignment is satisfied.

### DIFF
--- a/dlib/type_safe_union/type_safe_union_kernel.h
+++ b/dlib/type_safe_union/type_safe_union_kernel.h
@@ -85,36 +85,11 @@ namespace dlib
         {
         }
 
-
-        const static size_t max_size = tmax<tmax<tmax<tmax<tmax<tmax<tmax<tmax<tmax<tmax<
-                                       tmax<tmax<tmax<tmax<tmax<tmax<tmax<tmax<tmax<sizeof(T1),
-                                                        sizeof(T2)>::value,
-                                                        sizeof(T3)>::value,
-                                                        sizeof(T4)>::value,
-                                                        sizeof(T5)>::value,
-                                                        sizeof(T6)>::value,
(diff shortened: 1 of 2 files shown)


pfeatherstone:
Hmm, appveyor seems stuck. It won't start any of the queued jobs. Weird.

dlib/type_safe_union/type_safe_union_kernel.h
```diff
@@ -111,7 +111,10 @@ namespace dlib
         // --------------------------------------------
 
         // member data
-        stack_based_memory_block<max_size> mem;
+        typename std::aligned_union<max_size, T1,T2,T3,T4,T5,
```
davisking, line 114: Yeah this is a good change.  It's been 10 years since C++11 and dlib requires C++11 anyway now.  So any C++11 features are fine to use.

Remove the `max_size` member though.  That's not needed with `std::aligned_union` anymore. Like just set the argument here to 0 and let it do it's thing.

pfeatherstone: Oh, i didn't know you could do that with ```std::aligned_union```

davisking:
Nice

Issue #2354, opened 2021-04-26
dlib v19.22 (and master) does not build with gcc 4.8.5 on CentOS 7


cweigel:
## Expected Behavior
`$ python setup.py build` shall succeed even with gcc 4.8.5 since centos 7 is still around and the compiler is not rejected in the `CMakeLists.txt `.

## Current Behavior
`$ python setup.py build` fails 

### Compile Issue 1

during CMake run there is a first hint during CUDA test compile:

```cmake
-- *** CUDA was found but your compiler failed to compile a simple CUDA program so dlib isn't going to use CUDA.
-- *** The output of the failed CUDA test compile is shown below:
-- ***
-- ***   Change Dir: /build-dlib/dlib/build/dlib/cuda_test_build
   ***
   ***   Run Build Command(s):/usr/bin/gmake && [ 50%] Building NVCC (Device) object CMakeFiles/cuda_test.dir/cuda_test_generated_cuda_test.cu.o
   ***   /build-dlib/dlib/dlib/cmake_utils/test_for_cuda/../../cuda/../matrix/../vectorstream/vectorstream.h:31:24: error: invalid use of dependent type 'dlib::vectorstream::vector_streambuf<CharType>::size_type'
   ***                size_type read_pos = 0; // buffer[read_pos] == next byte to read from buffer
   ***                           ^
   ***   CMake Error at cuda_test_generated_cuda_test.cu.o.cmake:276 (message):
   ***     Error generating file
   ***     /build-dlib/dlib/build/dlib/cuda_test_build/CMakeFiles/cuda_test.dir//./cuda_test_generated_cuda_test.cu.o
```

The  problem which gcc 4.8.5 complains about is initialization of the typedef'd `size_type` this line 31 in `dlib/vectorstream/vectorstream.h`. 

```
 size_type read_pos = 0; // buffer[read_pos] == next byte to read from buffer
```
for which gcc 4.8.5 gives this error:
```
[...]/dlib/dlib/threads/../binary_search_tree/../vectorstream/vectorstream.h:31:34: error: invalid use of dependent type 'dlib::vectorstream::vector_streambuf<CharType>::size_type'
```

This can be fixed by not initializing `read_pos`:

```
 size_type read_pos; // buffer[read_pos] == next byte to read from buffer
```
which should to no harm since it is initialized in the constructor as well.

### Compile Issue 2

However, when fixing compile issue 1 this issue shows up:

```cmake
[...]

Issue #2353, opened 2021-04-26
Unable to Install dlib for python3.6 on Centos7


mon28:
## Expected Behavior
Successful install of dlib in my python3.6 virtualenv. 

## Current Behavior
pip install dlib worked like a charm on macOS, but failing on centos7

```
-- Disabling CUDA support for dlib.  DLIB WILL NOT USE CUDA
-- C++11 activated.
-- Configuring done
-- Generating done
-- Build files have been written to: /data/mtare/ipa-v2-document-trainer/document-trainer/dlib/build/temp.linux-x86_64-3.6
Invoking CMake build: 'cmake --build . --config Release -- -j24'
Scanning dependencies of target dlib
[  1%] Building CXX object dlib_build/CMakeFiles/dlib.dir/entropy_decoder/entropy_decoder_kernel_2.cpp.o
[  2%] Building CXX object dlib_build/CMakeFiles/dlib.dir/bigint/bigint_kernel_1.cpp.o
[  2%] Building CXX object dlib_build/CMakeFiles/dlib.dir/base64/base64_kernel_1.cpp.o
[  2%] Building CXX object dlib_build/CMakeFiles/dlib.dir/entropy_decoder/entropy_decoder_kernel_1.cpp.o
[  2%] Building CXX object dlib_build/CMakeFiles/dlib.dir/entropy_encoder/entropy_encoder_kernel_1.cpp.o
[  3%] Building CXX object dlib_build/CMakeFiles/dlib.dir/bigint/bigint_kernel_2.cpp.o
[  4%] Building CXX object dlib_build/CMakeFiles/dlib.dir/bit_stream/bit_stream_kernel_1.cpp.o
[  4%] Building CXX object dlib_build/CMakeFiles/dlib.dir/dir_nav/dir_nav_kernel_2.cpp.o
[  5%] Building CXX object dlib_build/CMakeFiles/dlib.dir/entropy_encoder/entropy_encoder_kernel_2.cpp.o
[  5%] Building CXX object dlib_build/CMakeFiles/dlib.dir/unicode/unicode.cpp.o
[  6%] Building CXX object dlib_build/CMakeFiles/dlib.dir/gui_widgets/fonts.cpp.o
[  7%] Building CXX object dlib_build/CMakeFiles/dlib.dir/md5/md5_kernel_1.cpp.o
[  7%] Building CXX object dlib_build/CMakeFiles/dlib.dir/tokenizer/tokenizer_kernel_1.cpp.o
[...]

PR #2357, opened 2021-04-27
do not attempt to build with gcc 4.8.5 or older

arrufat:
As seen in #2353 and #2354, CentOS 7 ships with gcc 4.8.5, which is no longer supported in dlib since v19.22.

I think rejecting to build with that GCC version and pointing the users to an older version of dlib might be useful.


davisking:
Yeah, that’s a good idea.

PR #2362, opened 2021-05-08
[SERIALIZATION] added support for std::variant

pfeatherstone:
Does what it says on the tin.

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -89,6 +89,7 @@
         - std::complex
         - std::unique_ptr
         - std::shared_ptr
+        - std::variant (if C++17 is used)
         - dlib::uint64
         - dlib::int64
         - float_details
@@ -119,6 +120,7 @@
         - std::complex
         - std::unique_ptr
         - std::shared_ptr
+        - std::variant (if C++17 is used)
         - dlib::uint64
         - dlib::int64
         - float_details
@@ -233,6 +235,9 @@
 #include <limits>
 #include <type_traits>
 #include <utility>
+#if __cplusplus >= 201703L
+#include <variant>
+#endif
 #include "uintn.h"
 #include "interfaces/enumerable.h"
 #include "interfaces/map_pair.h"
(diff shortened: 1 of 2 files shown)


pfeatherstone:
I could add std::optional to this PR. That's pretty trivial and would complete the extra c++17 containers I think. (std::any is impossible to serialise/deserialise in a portable way)

davisking:
I'm not sure this is a super good idea.  Like I would in general expect people to sometimes add or remove types from a variant.  But once you have started saving an object out via serialize you often need to maintain the ability to load old serialized data in later.  But the setup here makes that impossible, since there is no way for someone to load in an old variant that has a different set of types in it.

So I think this is something users should do on their own.  And hopefully when they do it, if they think they could ever possibly care about this kind of backwards compatibility, they will add version information or something (there are lots of options) so their serialization code to accommodate supporting loading old data.  But if it's in dlib then it's basically immutable for users, which isn't great in this specific instance.

pfeatherstone:
Can't we just add a version number? 
Also i see no reason why we can't serialize ```std::variant```, but we can ```dlib::type_safe_union```

pfeatherstone:
OK i see your point. But you could argue type versioning for any type, even ```float```. I could argue the way you save ```float``` is not how I would save ```float```, but i work with how dlib saves ```float```.

pfeatherstone:
With regards to adding and removing types, you can also do that with ```std::tuple``` using ```std::tuple_cat```, yet we support serializing ```std::tuple```. It sounds like what you describe is something 1 or 2 users have done and having a built-in serializer for ```std::variant``` would break their code. I think if you're going to do something wacky like add and remove types from ```std::variant``` and still want serialization to work in a backwards compatible way, then call your serializer something different, like ```void serialize_variant_special(const std::variant<...>& item, std::ostream& out)```

davisking:
Yeah, you are right. We should have this and it’s on the user to layer whatever versioning they might need around it in some outer type. 

Can you make it not require the type be movable though?

pfeatherstone:
Yes sure. Do you have an opinion on how to do that? 
I think the compromise is that the type needs to be default constructible.

pfeatherstone:
And it might need to be swappable.

pfeatherstone:
@davisking OK, i think the only requirement is that the type be default constructible. :)

davisking:
Yep, looks good :)

pfeatherstone:
Ready

PR #2364, opened 2021-05-10
[SERIALIZATION] support for std::optional

pfeatherstone:
Does what it says on the tin.


pfeatherstone:
@davisking ready

davisking:
Sweet, looks good :)

PR #2365, opened 2021-05-11
Fix serialize variant with C++17

arrufat:
I got a compiler error because of a small typo.

### DIFF
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -1235,7 +1235,7 @@ namespace dlib
                 if (I == index)
                 {
                     auto& x = item.template emplace<std::variant_alternative_t<I,Variant>>();
-                    deserialize(bin, x);
+                    deserialize(in, x);
                 }
                 else
                 {
--- a/dlib/serialize.h
+++ b/dlib/serialize.h
@@ -1235,7 +1235,7 @@ namespace dlib
                 if (I == index)
                 {
                     auto& x = item.template emplace<std::variant_alternative_t<I,Variant>>();
-                    deserialize(in, x);
+                    deserialize(x, in);
                 }
                 else
                 {


pfeatherstone:
It should actually be, ```deserialize(x, in)```, not ```deserialize(in, x)```
Sorry. @arrufat can you fix it?

pfeatherstone:
I think the C++17 tests on travis aren't actually working. Otherwise, this would have been picked up.

arrufat:
> It should actually be, `deserialize(x, in)`, not `deserialize(in, x)`
> Sorry. @arrufat can you fix it?

Done!

> I think the C++17 tests on travis aren't actually working. Otherwise, this would have been picked up.

Yes, I think the tests only run C++11.

pfeatherstone:
So there is this test [here](https://github.com/davisking/dlib/blob/master/.travis.yml#L71) on travis, which is what i was doing locally, but i don't think that's actually working.

pfeatherstone:
I don't know cmake very well at all, but does ```set (CMAKE_CXX_STANDARD 17)``` work with [this](https://github.com/davisking/dlib/blob/master/dlib/cmake_utils/use_cpp_11.cmake) in dlib?

pfeatherstone:
Maybe the [travis file](https://github.com/davisking/dlib/blob/master/.travis.yml#L71) needs to be amended to add cmake options, like the ones you mentioned above.

pfeatherstone:
I'm not trying to shift the blame by the way, i should have checked more carefully when submitting the previous PR. It was a copy-paste error.

arrufat:
Yes, I can do that later, but if I run that on a branch that does not have this patch, I will be able to see the build fail with C++17, that's why I made a separate one.

If we decide it's better to address both issues in this PR, I will cherry-pick it.
However, I think it's better to have different PR for different issues.

PR #2366, opened 2021-05-11
Make Travis read the CXXFLAGS enviroment variable

arrufat:
If this worked, the code should not compile with C++17. See #2365.

### DIFF
--- a/dlib/travis/build-and-test.sh
+++ b/dlib/travis/build-and-test.sh
@@ -11,7 +11,7 @@ fi
 if [ "$VARIANT" = "test" ]; then
   mkdir build
   cd build
-  cmake ../dlib/test 
+  cmake ../dlib/test -DCMAKE_CXX_FLAGS="${CXXFLAGS}"
   cmake --build . --target dtest -- -j 2
   ./dtest --runall $DISABLED_TESTS
 fi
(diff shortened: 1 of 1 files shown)


pfeatherstone:
@arrufat Since the checks passed, does this mean the fix didn't work?

arrufat:
Yes :(

dlib/travis/build-and-test.sh
```diff
@@ -7,11 +7,16 @@ if [ -n "${MATRIX_EVAL+set}" ]; then
     eval "${MATRIX_EVAL}"
 fi
 
+CXX_FLAGS="-std=c++11"
+if [ ! -z ${CXXFLAGS+x} ]; then
```
davisking, line 11: Why is there that +x there?  
```suggestion
if [ ! -z ${CXXFLAGS} ]; then
```
Why not do this?

davisking: Heh, I don't recall why I did that odd `+set` thing on the one above either :|

davisking, line 11:
```suggestion
if [ ! -z ${CXXFLAGS+set} ]; then
```
Do this though, so it's the same style as the one above.

arrufat: Yes, that was my first try...

PR #2367, opened 2021-05-12
Fix disable_duplicative_biases when the input is a skip layer

arrufat:
In [my implementation](https://github.com/dlib-users/dnn/blob/master/src/classification/repvgg.h) of the [RepVGG](https://arxiv.org/abs/2101.03697) network family, I was getting a compile error when running `disable_duplicative_biases`.
The case where a normalization layer has a skip layer as an input was not handled.


davisking:
Thanks for the fix :)

PR #2376, opened 2021-06-06
YOLO loss

arrufat:
Hi, I've been spending the last few days trying to make a `loss_yolo` layer for dlib, in particular the loss presented in the YOLOv3 paper.

I think I came up with a pretty straightforward implementation but, as of now, it still does not work.

I wondered if you could have a look. I am quite confident the loss implementation is correct, however I think I might be making some assumptions about the dlib API when the loss layer takes several inputs from the network.

I tried to make the loss similar to the `loss_mmod` in the way you set the options of the layer, etc.

So, my question is, does this way of coding the loss in dlib make sense for multiple outputs? Or is dlib doing something I don't expect?

There's also a simple example program that takes a path containing a `training.xml` file (like the one from the face or vehicle detection examples).

Thanks in advance :)


dlib/dnn/loss.h
```diff
@@ -3447,6 +3447,488 @@ namespace dlib
...
+            const_label_iterator truth,
+            SUBNET& sub
+        ) const
+        {
+            DLIB_CASSERT(input_tensor.num_samples() != 0);
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+            double loss = 0;
+            for (long i = 0; i < input_tensor.num_samples(); ++i)
+            {
+                impl::yolo_helper_impl<TAG_TYPES...>::tensor_to_grad(input_tensor, truth, sub, i, options, loss);
```
arrufat, line 3863: One doubt I have comes from here.

When `back_propagate_error` will be called, I've seen that it will clear the gradient inputs. I am worried that this is happening and even if I fill the gradients for the 3 outputs, the gradients that are not at the top get zeroed out at backpropagation.

That would explain why it's not learning (but it would make it harder to fix...) I hope I just made a mistake in my implementation.

arrufat:
Honestly, after debugging the code and running many experiments, I can't see what I am doing wrong.
I have even loaded the pretrained backbone, initialized the anchors weights to 0.5 as in [here](https://github.com/pjreddie/darknet/blob/master/src/yolo_layer.c#L44-L46), used the same scales for the loss and the gradient updates as in the official YOLO code, but still couldn't make it work... maybe @pfeatherstone is right, and it's not as straightforward as I thought. I couldn't see any weird tricks in the training process in the official repository, though...


The training loss at the beginning goes down quite fast, but then it stays constant and pushes the detection confidences to 0.
I will keep checking, but I am open to suggestions. I think we're close to having this loss in dlib (unless I am making some terrible assumption about how dlib works, or a big conceptual mistake).

The most significant change that I made is having a sigmoid layer before the detector output.
As you may know, the box coordinates are computed like this:
image
And the objectness and classifier use a (multi) binary log loss, which means that a sigmoid is applied everywhere, except for the width and height outputs.
So, I decided to apply it everywhere, and then I can recover the original width and height outputs, by inverting the sigmoid:

_y = σ(x) = 1 / (1 + e^(-x)) → x = log(y / (1 - y))_

Then, when I apply the exponential to get the final width and height, it simplifies to just:

_y / (1 - y)_

Where y is the output after the sigmoid (_0 < y < 1_, so the training is stable). This allows me to apply the sigmoid everywhere using the built-in cuDNN function, which happens in place, while still being mathematically equivalent. I even made that change to Darknet and reproduced the results on VOC2012, so the problem is not there, I am a bit surprised that nobody is doing it this way...

So, any help or insights would be appreciated :)

pfeatherstone:
Thank you for all this hard work. Unfortunately I don't have time to spend on this at the moment. Maybe you could try a very simple backbone with a single yolo layer and therefore a single yolo loss function. Maybe that will help to debug. Maybe the yolo layer could be at stride 16 and you could train on VOC which is easier than COCO.

pfeatherstone:
Maybe a VGG backbone. You could even train it on darknet first. Then port the weights using your visitor. Then check the forward pass works correctly in dlib. Then go from there trying to restart the training in dlib. Maybe you could re-randomise the last couple conv layers before the yolo layer and train those.

pfeatherstone:
This is why autograd is so awesome. There are no headaches when writing custom modules and custom loss functions.

arrufat:
Thank you for those suggestions :)
I didn't try a simpler backbone, but I tried with a single YOLO output to see if there was a problem with multiple outputs, but that didn't work. I might try that again, since I addressed several problems in the loss function since then.
Frankly, I am not too worried about the gradients, it's all MSE and binary log loss, so quite straightforward. What bothers me is that, at least the objectness confidence should be easy to learn, but it gets pushed to 0 (I tried weighing differently, even though in the official implementation all weights (obj, bbr and class) are set to 1).

I am still motivated to make this work, I hope not to lose the momentum...

arrufat:
Well, after a week of trying stuff, I could not make it work...

At first, I thought it was a problem of weighing the different parts of the loss, then I managed to get the objectness not to go to zero, but the network predicts the same bounding boxes for all images...
I will keep working on it, but @davisking, do you think you can have a look at some point?
I know it's a lot to ask, and I understand it's not a priority.

My main concern is that I am not understanding how dlib works well enough to use multiple tensors in a single loss layer, and I am making a conceptual mistake...

Thank you in advance, and please, don't feel pressured or anything :)

davisking:
Yeah I’ll look in a bit when I get a chance :)

arrufat:
Thank you :D there's no hurry

dlib/dnn/loss.h
```diff
@@ -3447,6 +3447,526 @@ namespace dlib
...
+            const_label_iterator truth,
+            SUBNET& sub
+        ) const
+        {
+            DLIB_CASSERT(input_tensor.num_samples() != 0);
+            DLIB_CASSERT(sub.sample_expansion_factor() == 1);
+            double loss = 0;
+            for (long i = 0; i < input_tensor.num_samples(); ++i)
+            {
+                impl::yolo_helper_impl<TAG_TYPES...>::tensor_to_loss(input_tensor, truth, sub, i, options, loss);
```
arrufat, line 3902: After debugging for a while, I found a problem, after this line, I should add:
```c++
++truth;
```
So that the next iteration uses the correct bounding boxes, otherwise it's using the bounding boxes of the first image in the batch for all the mini batch... I hope that is the problem, running some tests now...

arrufat: I think that was the problem, it's learning slowly, but for the first time I'm getting results that look like this:
image
🤞

pfeatherstone: Is this learning from scratch? Or finetuning an existing model?

arrufat: Fine-tuning, the training code it's a bit different from the one published here. I don't know what will be best for an example program, though... I think training from scratch is simpler, and people can refer to [`dnn_introduction3_ex`](http://dlib.net/dnn_introduction3_ex.cpp.html) for how to do fine-tuning.

pfeatherstone: Well it's a good sign if you're training completely from scratch and you're getting results like that.

pfeatherstone: Usually the backbone is initialized with a pre-trained classifier. So if you're getting results without even doing that stage, then that should give you confidence that your loss layer is working.

arrufat: The code I'm running now uses this pretrained backbone: https://pjreddie.com/media/files/darknet53.conv.74.
I converted it using the Darknet weights visitor and assigned it to the network in this PR.

pfeatherstone: Oh sorry, i misunderstood. Still, there are quite a few layers after that which are randomly initialized. So your loss layer must be working now.

arrufat:
@pfeatherstone since you seemed a bit skeptical, here are some results on this implementation trained from scratch on the COCO dataset (I didn't even initialize the `input_rgb_image` with `(0, 0, 0)`, just the default values from dlib.)
image
image
image
image
image
image
image
image

The bounding box regression is not perfect, but it's been training for about 12h only. I think it's showing promising results, even if they are from the training set. I tried it on my webcam, and it also works.

So, @davisking, no need to spend time on this now, I will run some extra tests and prepare this PR for review when I have time (maybe, for the example program, a simpler backbone would be better) :)
I am thrilled to have YOLO detectors working in dlib! As always, thank you for creating this outstanding library :)

pfeatherstone:
That looks awesome. Great work! Do you have some idea of training benchmarks vs darknet or pytorch (ultralytics repos)? It would be interesting to know memory usage when training at say 416x416 and the number of epochs required before plateau'ing.

pfeatherstone:
I'm quite excited by this as this might draw more dlib users and maybe more PRs

arrufat:
> That looks awesome. Great work! Do you have some idea of training benchmarks vs darknet or pytorch (ultralytics repos)? It would be interesting to know memory usage when training at say 416x416 and the number of epochs required before plateau'ing.

Thank you!

Frankly, I've never used YOLOv5... I've only used Darknet and recently YOLOR (which is based on Scaled-YOLOv4, which in turn is based on YOLOv5, I think).

I've never liked the experience of those toolkits/repos. Everything is overly complicated to bypass the shortcomings of Python (performance and multi-threading). That is the main reason why I was motivated to add this to dlib, as it has the best user experience, for me... and I won't have to use those anymore.
In terms of memory, I think dlib uses slightly more memory than Darknet, but I can't tell for the others, since I didn't use them with the same backbones...

Regarding the plateauing, it's hard to compare, since each toolkit might normalize the gradient differently (e.g. using the batch size, the layer dimensions, etc.), so the learning rate has different meanings in each toolkit/repo.

Anyway, using the default scheduler in dlib makes me have one less thing to worry about.
I never liked training with fixed number of steps/epochs and decreasing the learning rate at arbitrary positions, which have to be decided beforehand.
That's why, in this PR, I use burn-in at the beginning (not sure if it's needed, though) and then fallback to the default learning rate scheduler.

pfeatherstone:
You might get away without burn-in, but typically with yolo models, gradients are unstable at the beginning. That's why burn-in is typically used. However, in my experience, when using CIOU loss, the gradients are way more stable and I can use a default scheduler. I wouldn't implement CIOU now though. Maybe a future PR.

arrufat:
I think it's ready for review! It's already working, I've tried in on some datasets (dlib cars, VOC2012, COCO2017) and it seems to work reasonably well. So, @davisking, you can have a look when you have time :)

davisking:
Sweet.  I'll take a look in a little bit.  About adding the image width and height as a metadata field to `dlib::image_dataset_metadata::image`, yeah, that seems like a good idea given the use case you described.  Which is a pretty important use case :)

arrufat:
Done, I've added a simple example on how one would recompute the anchor boxes. :)

dlib/dnn/loss_abstract.h
```diff
@@ -1852,6 +1852,181 @@ namespace dlib
...
+        );
+        /*!
+            ensures
+                - #get_options().overlaps_nms == test_box_overlap(iou_thresh, percent_covered_thresh)
+        !*/
+
+    };
+
+    template <typename SUBNET>
+    using loss_yolo = add_loss_layer<loss_yolo_, SUBNET>;
```
arrufat, line 2028: I am not sure which name should I use for this alias template, or if I should provide more.
I thought about following the `concat_` convention: `concat2`, `concat3` etc., where the numbers specify the number of layers to concatenate.
However, that would lead to `yolo1`, `yolo2`, `yolo3`, `yolo4`, `yolo5`, etc. Which can be rather confusing...
That's why I only defined one as `yolo`, that uses 3 outputs, which is the most common case...

pfeatherstone: If someone needs more or less than 3 outputs, that could be for a future PR. And it looks like that wouldn't be such a big piece of work right?

davisking:
I'll update the other dlib datasets to have width and height annotations so this stuff will work on them too.

arrufat:
> I'll update the other dlib datasets to have width and height annotations so this stuff will work on them too.

Great! I made a small program to do that already... I am sure you don't need it, but here it is anyway
https://gist.github.com/arrufat/d5fac2340d2a165d40627cdf338aac76

davisking:
> > I'll update the other dlib datasets to have width and height annotations so this stuff will work on them too.
> 
> Great! I made a small program to do that already... I am sure you don't need it, but here it is anyway
> https://gist.github.com/arrufat/d5fac2340d2a165d40627cdf338aac76
> EDIT: Oops, I didn't see that you added that already to imglab...

Yeah no worries :)

I'm reuploading the last of the datasets.  Should be done in a few minutes.

dlib/image_processing/full_object_detection.h
```diff
@@ -183,6 +183,60 @@ namespace dlib
...
+
+    struct yolo_rect
+    {
+        yolo_rect() = default;
+        yolo_rect(const drectangle& r) : rect(r) {}
+        yolo_rect(const drectangle& r, double score) : rect(r),detection_confidence(score) {}
+        yolo_rect(const drectangle& r, double score, const std::string& label) : rect(r),detection_confidence(score), label(label) {}
+        yolo_rect(const mmod_rect& r) : rect(r.rect), detection_confidence(r.detection_confidence), ignore(r.ignore), label(r.label) {}
+
+        drectangle rect;
```
arrufat, line 196: I made this a `drectangle` for no particular reason... it was originally a `rectangle`, but when the loss was not working, I thought it might be due to some missing cast or something (I was a bit desperate) so I made it a `drectangle` to make sure that was not causing any problems.

However, I don't know if it makes a lot of sense to have sub-pixel accuracy for the bounding box. If you think It should be a `rectangle`, I will change it back and add the proper casts in the loss function.

davisking: Na `drectangle` seems like the right call.

arrufat:
Also, after seeing how you computed the anchor boxes in the example program, maybe we don't need an `anchor_box_details` struct, and the same can be achieved with `point` or `dpoint`.

davisking:
Maybe.  I haven't gone over the code yet really.  I ran the example and
played around a bit (and it worked great).

But there is something to be said for having explicit types for specific
ideas.  Since it makes it much more clear what the arguments to functions
that use them are about.

arrufat:
> Are you sure about this? Reading the paper:
> 
> ```
> YOLOv3 predicts an objectness score for each bounding box using logistic regression. 
> This should be 1 if the bounding box prior overlaps a ground truth object by **more** than any  
> other  bounding  box  prior.   
> If  the  bounding  box  prior is  **not**  the  best  but  does  overlap  a  ground  truth  object
>  by more than some threshold we **ignore** the prediction. We use the threshold of.5. Our system
> only assigns one bounding box prior for each ground truth object.  
> If a bounding box prior is not assigned to a ground truth object
> it incurs **no** loss for coordinate or class predic-tions, only objectness.
> ```
> 
> So the best match is always used, regardless of IOU : agreed.
> I think those matches whose IOU is greater than `iou_anchor_threshold`, are NOT used in the bbox loss and class loss.

Yes, I am quite confident about this. This part is a new addition from the Alexey's fork:
https://github.com/AlexeyAB/darknet/blob/1be96802a030ae2216c7e19b7523e1e2f0948aaf/src/gaussian_yolo_layer.c#L590-L622

Here @AlexeyAB explains it: https://github.com/AlexeyAB/darknet/issues/7087#issuecomment-753706790.

As I said this part was not present in YOLOv3, but all the new models that achieve state-of-the-art results: Scaled YOLOv4, YOLOv4-CSP, YOLOR, use this setting with a value close to 0.2.

pfeatherstone:
Oh right, I'd never noticed that. I've always just implemented yolo loss as the paper described and it worked fine. Thanks for bringing this to my attention.

pfeatherstone:
As a general discussion, nothing to do with this implementation, I find yolo loss particularly faffy given all the hyper parameters and the enormous number of candidate predictions. I'd had high hopes for DeTR and other direct set prediction models using bipartite matching loss, but they are far too slow to train. It's a real shame coz they are so elegant, easy to write and intuitive. Anyone else dissatisfied with the state of SOTA detection models?

pfeatherstone:
I'd wondered if you could combine bipartite matching with yolo layers. So currently each cell has tree "slots", each one assigned to an anchor. Then matching is done based on best IOU with anchors. Instead of that, make each cell have N "slots", maybe 5 of so, and the matching is done using bipartite matching, which you could do using ```dlib::max_cost_assignment()```. Maybe there would be a lot of overhead calling ```dlib::max_cost_assignment()``` thousands of times per batch, but at least you wouldn't have to use anchors anymore and you wouldn't have hyper parameters like ```trush_threshold``` and stuff like that. Just a thought. Bye the way, this discussion is not really relevant to the PR, but i'm interested to read other people's thoughts.

arrufat:
Honestly, I haven't given a deep thought to this, I'd like to do some research at some point on how to improve detectors, and being able to implement YOLO from scratch is an essential step for me. I agree on DeTR paper, the approach seems so simple, and with less engineering, but I never got to try it for myself.

pfeatherstone:
@arrufat If you're collating ideas for other detection heads, i've been experimenting with [fcos](https://arxiv.org/pdf/1904.01355.pdf) in pytorch and you get pretty decent results and it's super easy to implement. That could be a cool future PR if you're interested.

Compaile:
Hi guys 2 quick questions

1) Is there any timeline when this will be reviewed/merged?
2) since its a long read, are thoe yolov4 features already implemented or is it "only" yolov3 ?

Huge pr !

davisking:
I'll get to it next week.

arrufat:
@Compaile,
Hi! Regarding the second question, it started as YOLOv3 (actually the loss function is almost identical in YOLOv4 and others).

I just added an extra setting that the new YOLO-based state-of-the-art models use, the `iou_anchor_threshold`, called just `iou_threshold` in the YOLOv4 repository.

It's disabled by default (set to 1), so you get YOLOv3 if you don't change that setting.
Let me know if what the setting does is clear from the documentation :)

arrufat:
@pfeatherstone https://arxiv.org/abs/2107.08430

I am thinking about trying this out. It has a nice ablation study on the YOLO loss and it's anchor free.
I'm not sure whether we should put this in this PR or even in dlib.

I think it's too early, dlib is more about well known stuff that has survived the test of time (YOLO certainly has).
Maybe we could add a `loss_yolox` later on, and it doesn't seem that hard to implement on top of this one.

I don't know, these things move so fast... it's hard to know what's worth adding...

pfeatherstone:
Ohh. I'll have a look. Thanks @arrufat

pfeatherstone:
The paper looks interesting. Though they claim the detection head is simpler, it still has some nitty gritty engineering rules, like the simOTA assignment they describe. I wonder if the results they present are a bit lucky. Like when comparing results with other models, +1 AP score could be attributed to a well tuned optimiser and scheduler and may well have nothing to do with the network architecture. Ultralytics repo proves this with yolov3. With no changes whatsoever to the architecture, they got substantially better results than darknet. So I'm not convinced all these improvements can be attributed to detection heads or label assignments. A really fair setup would be to also retrain yolov5 from scratch using their setup, training loops, schedulers, the lot. I still think the most elegant and simplest solution is what DETR did. It's a shame it doesn't really work...

pfeatherstone:
Also looking at the yolox code, it sets batchnorm parameters to unusual values. Like a momentum of 0.3, i have never seen. I don't think that's mentioned in the paper. It's the type of thing you do when you're mini-batching in torch... A few of the layers have specific methods for initialising parameters, rather than using pytorch defaults. So there are obvious engineering tricks which have been applied that hint that a lot of fine tuning has been done, all of that to get a couple AP points over yolov5. So it doesn't seem to be a dramatic improvement over the traditional yolo detection heads and loss layers. Maybe I'm wrong, I don't know.

pfeatherstone:
The biggest take away for me from the paper is decouple heads. The paper suggests it accelerates training which is always desirable, even if the end result is the same.

pfeatherstone:
There is already a thread on darknet about yolox. Alexey is all over it. He claims some of the yolov4 variants have equal if not better performance in terms of latency and AP score. So who knows. It seems to me you trade one faffy mechanism for another and get almost identical results. Yes it's publishable but not ground breaking

arrufat:
Thank you for your insights, it's always good having someone else's point of view on this. Either way, I want to try an anchor-less training with this loss, I'm really curious to see if that works.

pfeatherstone:
Maybe just have one detection per cell. So have a total of 3 anchors, all of them set to 1.

arrufat:
> Maybe just have one detection per cell. So have a total of 3 anchors, all of them set to 1.

I am currently trying this:
- predict center offsets x, y as in YOLO
- predict the height and width directly without any anchors
- filter the truth boxes by strides (from FCOS), each stride should be in charge of detecting different sizes:
  - stride 8: boxes with sides smaller than 64
  - stride 16: boxes with sides between 64 and 128
  - stride 32: boxes with larger sides

The modification in the code is pretty straightforward, just multiply the prediction by the stride of that feature map instead of the anchor. Which seems very similar to what you were saying: training with one anchor per stride.

So far, it seems to work

pfeatherstone:
It would be great if the example program trained on COCO and included an AP score function. That way you can directly compare with other frameworks and hopefully convince the user that using dlib isn't a bad shout. I think the COCO labels, as downloaded from the COCO website, not darknet, consists of big JSON files. So a COCO dataset loader in dlib would probably require a JSON parser. We could simply pinch nlohmann's single header JSON library and put that in a 'third party' or 'external' folder.

pfeatherstone:
Oh and a mosaic data augmentation function would also be awesome (sorry if I'm always adding stuff to the wish list). I think that's one of the big contributions from ultralytics yolov3 repo which added +4 AP points without any model architecture changes. With that, and all the other stuff you've done, the example program is pretty close to a fully fleshed object detection training framework.

arrufat:
Actually, I am using this at work, and I already implemented my version of mosaic. It's quite straightforward in dlib:
```
const dlib::rectangle r(x, y, x + tile_size, y + tile_size);
auto si = dlib::sub_image(canvas, r);
resize_image(image, si);
```
Then you put that in a loop with 4 iterations which gets one image at a time and provide the correct `x` and `y` values at each iteration.

arrufat:
Actually, I am using this at work, and I already implemented my version of mosaic. It's quite straightforward in dlib:
``` c++
const double scale = 0.5;
const dlib::rectangle r(x, y, x + tile_size, y + tile_size);
auto si = dlib::sub_image(canvas, r);
resize_image(image, si);
for (auto& b : boxes)
    b.rect = translate_rect(scale_rect(b.rect, scale), x, y);
```
Then you put that in a loop with 4 iterations which gets one image at a time and provide the correct `x` and `y` values at each iteration.

examples/dnn_yolo_train_ex.cpp
```diff
@@ -0,0 +1,526 @@
...
+        }
+        cout << "Average IoU: " << average_iou / samples.size() << endl;
+        return EXIT_SUCCESS;
+    }
+
+    // When computing the objectness loss in YOLO, predictions that do not have an IoU
+    // with any ground truth box of at least options.iou_ignore_threshold, will be
+    // treated as not capable of detecting an object, an therefore incur loss.
+    // Predictions above this threshold will be ignored, i.e. will not contribute to the
+    // loss. Good values are 0.7 or 0.5.
```
davisking, line 259:
```suggestion
    // Similarly, predictions above this threshold are considered correct predictions by the loss. Typical settings for this threshold are in the range 0.5 to 0.7.
```
Since they aren't ignored.  The best matching ones still contribute to the loss.

davisking (approved):
Looks great.  I left some comments but it's all tweaks to the docs. I'll let you merge those (or adjust them if they aren't quite correct suggestions) then we can merge this PR :)

examples/dnn_yolo_train_ex.cpp
arrufat: Yes, they are just ignored in this part of the loss (when computing the ones that should not detect an object).

But you are right, some of them (only the one that whose center falls into the same grid cell as the truth box) will still contribute to the loss in the second part (when we compute the predictions that should detect an object, location, and class).

The wording comes from the original YOLO implementation, so people familiar with the original code, can feel right at home when reading this.

dlib/dnn/loss_abstract.h
```diff
@@ -1852,6 +1852,191 @@ namespace dlib
...
+        (
+            double iou_thresh,
+            double percent_covered_thresh = 1,
+            bool classwise = true
+        );
+        /*!
+            ensures
+                - #get_options().overlaps_nms == test_box_overlap(iou_thresh, percent_covered_thresh)
+                - #get_options().classwise_nms == classwise
+        !*/
```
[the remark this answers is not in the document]
arrufat: The main issue is that `get_options()` is a `const` method, for 2 reasons:
- It is `const` in the MMOD loss
- All the options should not be easily exposed to be changed after the network is initialized.

In fact, the only thing that makes sense to be changed is the NMS... And it is actually not used by the loss function itself, only by the `to_labels` method. I was also thinking if it shouldn't be just an extra parameter there:

``` c++
template <
            typename SUB_TYPE,
            typename label_iterator
            >
        void to_label (
            const tensor& input_tensor,
            const SUB_TYPE& sub,
            label_iterator iter,
            double adjust_threshold = 0.25,
            test_box_overlap overlaps_nms(0.45, 1),
            bool classwise = true
        ) const;
```
But I didn't like constructing a new `test_box_overlap` object by default every time this thing is called, and it's also nice to have the NMS printed inside the loss (even if it's only used for inference).

So yes, I think it's convenient to have a small helper method to change all the stuff related to the NMS.

dlib/dnn/loss_abstract.h
```diff
@@ -1852,6 +1852,191 @@ namespace dlib
...
+            const_label_iterator truth, 
+            SUBNET& sub
+        ) const;
+        /*!
+            This function has the same interface as EXAMPLE_LOSS_LAYER_::compute_loss_value_and_gradient() 
+            except it has the additional calling requirements that: 
+                - layer<TAG_TYPE>(sub).get_output().k() == options.anchors.at(tag_id<TAG_TYPE>::id).size() * (5 + options.labels.size());
+                - sub.get_output().num_samples() == input_tensor.num_samples()
+                - sub.sample_expansion_factor() == 1
+            Also, the loss value returned corresponds to the L2 loss.
```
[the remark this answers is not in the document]
arrufat: Yes, I have mixed feelings about this. As you might have seen in the code, we're just computing the difference between the output of the network and the target (which will act as the gradient) and then squaring those to display the loss...

``` c++
loss += length_squared(rowm(mat(grad), n));
```
Even for the objectness and classes (I know they are the outputs of sigmoid functions) but it just seems like the L2 loss to me: the sum of the squared differences between the outputs and the targets.

dlib/dnn/loss_abstract.h
```diff
@@ -1852,6 +1852,191 @@ namespace dlib
...
+        // However, if other anchors have an IoU with a truth above iou_anchor_threshold, they
+        // will also be updated to match the same truth.  Setting iou_anchor_threshold to 1 will
+        // make the model use only the best anchor for each ground truth, so other anchors can be
+        // used for other ground truths in the same cell (useful for detecting objects in crowds).
+        // This setting is meant to be used with "high capacity" models, not small ones.
+        double iou_anchor_threshold = 1.0;
+        // When doing non-max suppression, we use overlaps_nms to decide if a box overlaps
+        // an already output detection and should therefore be thrown out.
+        test_box_overlap overlaps_nms = test_box_overlap(0.45, 1.0);
+        // Tell the non-max Suppression whether to take classes into account
```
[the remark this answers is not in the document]
arrufat: Yes, that was not the clearest way to express the meaning of that setting...

arrufat:
Thank you for spending time reviewing it, I've added most of your suggestions (I should improve my documentation-writing skills)...
I left a couple of comments, let me know what you think :)

dlib/dnn/loss_abstract.h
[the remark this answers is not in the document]
arrufat: Oh, there's a mistake in the documentation, I meant to assign the values, not to compare them, so it should be `=` instead of `==`.

dlib/dnn/loss_abstract.h
```diff
@@ -2020,16 +2020,15 @@ namespace dlib
...
             double iou_thresh,
             double percent_covered_thresh = 1,
             bool classwise = true
         );
         /*!
             ensures
-                - #get_options().overlaps_nms == test_box_overlap(iou_thresh, percent_covered_thresh)
-                - #get_options().classwise_nms == classwise
+                - #get_options().overlaps_nms = test_box_overlap(iou_thresh, percent_covered_thresh)
+                - #get_options().classwise_nms = classwise
```
davisking, line 2031: The convention in the dlib docs is to use == here, since the docs are saying "this expression will be true" rather than "this code will be executed".

examples/dnn_yolo_train_ex.cpp
```diff
@@ -255,8 +255,7 @@ try
     // When computing the objectness loss in YOLO, predictions that do not have an IoU
     // with any ground truth box of at least options.iou_ignore_threshold, will be
     // treated as not capable of detecting an object, an therefore incur loss.
-    // Predictions above this threshold will be ignored, i.e. will not contribute to the
-    // loss. Good values are 0.7 or 0.5.
+    // Similarly, predictions above this threshold are considered correct predictions by the loss. Typical settings for this threshold are in the range 0.5 to 0.7.
```
davisking, line 258: Add a new line here so this line isn't so long.

dlib/dnn/loss_abstract.h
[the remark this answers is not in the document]
davisking: Let me look at the code again, maybe I'm not interpreting what you are saying correctly.  

Like are you saying that the loss value coming out here isn't the loss being optimized?  Rather that it's only part of it?  The loss value coming out should be the value of the objective function the loss is optimizing.  Like the gradient of this value is what defines the gradient (or it should).   Otherwise it's pretty confusing. And stuff like testing when the loss has stopped improving isn't meaningful if the loss value isn't the actual loss being optimized.

dlib/dnn/loss_abstract.h
arrufat: Right, reverted, thank you

examples/dnn_yolo_train_ex.cpp
arrufat: Done!

dlib/dnn/loss_abstract.h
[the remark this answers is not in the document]
davisking: Oh right, I see.  The `loss` is being set to the norm of the loss's gradient.  It should be set to the loss value though.  Can you adjust it so `loss` is the loss value?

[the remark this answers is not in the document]
davisking: I mean, IDK, maybe it doesn't matter  It's technically wrong though as the value being output is not the loss value. It's the square of the norm of the gradient vector of that loss, which is in general a different value.  It's probably going to end up making training run fine.  But anyone looking at that value is going to be confused when it's some other number.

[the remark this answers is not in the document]
arrufat: So, do you mean computing the (multi)binary log loss for the objectness and the classification, and the L2 loss for the bounding box regression?

I did it this way because it's how it's done in the [official YOLO implementation](https://github.com/pjreddie/darknet/blob/masterhttps://github.com/pjreddie/darknet/blob/master/src/yolo_layer.c#L238):

```c
*(l.cost) = pow(mag_array(l.delta, l.outputs * l.batch), 2);
```
Where `l.delta` contains the gradients, and `mag_array` is the sum of the squared differences.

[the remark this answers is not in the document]
arrufat: But the official YOLO implementation doesn't use that value to make any decision, just for displaying purposes.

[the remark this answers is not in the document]
davisking: Right, the loss value is really just for display and for doing stuff like checking for convergence.

It's just funny because it's wrong according to the interface contract. 

So if you want to keep using this value instead of the loss I think that's fine. Update the docs for the loss layer to call out that the loss value isn't the loss but the squared gradient norm though. Do the logged loss value will be that instead of the loss. 

And you have found that the convergence testing stuff all works fine with this as the loss? Seems like it would be fine.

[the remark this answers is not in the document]
arrufat: I've been trying this loss for about 1 month, and it seems to work fine, but maybe it's not working at its full potential.
However, if I introduce that change, I will need to test it again (it might take some days, maybe a week)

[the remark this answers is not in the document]
davisking: Yeah. Don't worry about it. Do that in another PR if you want :)

For this one update the docs and I'll hit the merge button :)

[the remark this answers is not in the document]
arrufat: ok :)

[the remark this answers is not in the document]
arrufat: Done

davisking:
Sweet. Thanks for another great PR :)

arrufat:
Thank you, it's my pleasure!

carkaci:
1. Could you compare the accuracy of dnn_yolo with dnn_face (with mmod)?
2. Using a smaller learning rate works better in dnn_yolo, whereas in mmod starting with 0.1 is the best. I can't explain it.

arrufat:
1. I am not going to do it, but you can do it if you're interested. Note that the results you get from the example programs are not representative of what's achievable with either technique. They are just “executable documentation” to easily illustrate how to use a certain part of dlib, and they provide a good starting point. However, I doubt the example programs will give good results for any production-ready application, whatsoever.
2. I guess this is just how it is, YOLO-style detectors are famous for being unstable at the beginning of the training, hence the small learning rate. That being said, the example uses a small batch size, while the MMOD face detector uses a way larger batch size. Increasing the batch size might allow for a larger learning rate. So, I'd say it's hard to compare them, the network architectures and the loss functions are completely different.
3. What do you mean? If you are talking about the random cropper in the example, you should go and read its documentation. The example only focuses on explaining the newly introduced features. But if something is not clear, we can try to improve the documentation. However, when I made that example, I added (too) many options so that people could visualize the data augmentation, intermediate results, mAP, etc… I think that example is a bit too complicated, but to get sensible results with the YOLO loss, you usually need to do more complicated stuff. In that sense, I really like the MMOD loss: there's way less fidling.

carkaci:
Thank you very much for your explanations. 
3. Yes I mean random cropper. You have changed the default parameters, for example, `.set_translate_amount(0.5);` (0.5 is too high?) `set_background_crops_fraction(0)` (it is 0.5 in the default). I wonder the reasons why you changed them in such a way.

arrufat:
Well, I was trying the COCO dataset when I was preparing the example, and in COCO, you are expected to detect cropped objects, so that's why I set a relatively big translation value. Regarding the background fraction, all other training methods I've seen don't use object-empty images, so that's why I did it. I didn't check that those were the optimal parameters, they are very task-specific.

carkaci:
Hi Arrufat, I have two more questions.
1. By default options.iou_anchor_threshold = 0.2. What about setting it to
0.1?. (decreases the false positive rate?)
2. How can I interpret the "average loss: 0.232008"? Is it good learning,
or not? Does it mean that 23% of the time, it can't find the object?
Thanks.

arrufat:
Hi @carkaci,

> 1. By default options.iou_anchor_threshold = 0.2. What about setting it to 0.1?. (decreases the false positive rate?)

By decreasing the `yolo_options::iou_anchor_threshold` you are telling the network that any anchor that has an IoU higher than that value is also responsible for detecting that object. So, what you are effectively doing, is increasing the number of updates to the network (most of these detections will be suppressed by the NMS, anyway, but you are updating the network more often).
That's why this setting is usually meant to be used with high-capacity networks.

It's not all good, though. If you set this value to anything lower than 1, you are going to lose the ability to detect different things in the same grid cell. But maybe your dataset does not have many objects overlapping each other, so that's fine.

> 2. How can I interpret the "average loss: 0.232008"? Is it good learning, or not? Does it mean that 23% of the time, it can't find the object? Thanks.

I am afraid you can't interpret the loss like this, since it's just computed as the squared norm of the gradient in this code (and in the official Darknet YOLO version, too).
The loss will scale with the number of detection heads, the number of classes and also the number of grids cell in each feature map (image resolution). I've seen that YOLOv5 scales all these as if we were always using 3 outputs for with a fixed input size so that we can compare the losses between detectors. However, there's no easy way to know what it really means.

carkaci:
Thank you for your quick response.

1. ) By decreasing the yolo_options::iou_anchor_threshold you are telling
the network that any anchor that has an IoU higher than that value is also
responsible for detecting that object.
      Does it implicitly force the anchors to lower than the value of
yolo_options::iou_anchor_threshold
, that may result in lower false positives?
2.) What about setting options.iou_ignore_threshold  (by default it is 0.7) to
0.9 ?

arrufat:
My take is that, since you are using multiple anchors for the same object, if their shapes happen to be different (because the regression didn't work as expected) then you might have more false positives.

If you set the `yolo_options::iou_ignore_threshold` to 0.9, then all predictions that have an IoU with the truth below 0.9 will experience loss, forcing them not to detect the object. I like to think of it as a heatmap. The grid that contains the center will always predict the object, and the surrounding ones, only if they generate good predictions. What a good prediction means depends on the IoU ignore threshold. The lower the value, the larger the heatmap, the higher the value, the closer to a delta function. Note that if you set the IoU ignore threshold too high, maybe even the centered prediction will have an IoU below that value and will experience loss...

pfeatherstone:
Presumably everyone has seen the two new yolov7 repos. I love the naming wars. Lots of cool new stuff.

arrufat:
@pfeatherstone yes, I've already implemented the yolov7 backbone, but I am debugging it to make sure there are no mistakes.

VisionEp1:
Just to add some more feedback.
I played around a lot with your yolo implementation. First of all huge thanks to build it in dlib. 
The darknet yolo code is so hard to manage and adjust. 

I tried some samples with the base implementation, my own and also recently your dlibml/yolo-object-detector.

However i cant get it to work- it means i can, it works but it isn't as good as if i do my usual yolov5 c++ pipeline.  The training it self around the nn should be much better (better pre processing, better load balance etc). but the network itself is just always quite a bit worse and so is the map for example on coco. 

Mostly it has tho many false positives but also the conversion itself during training is way slower (which might be ok since its not deterministic and the sdg solver works different) 

I still suspect there are some minor bugs in the yolo implementation but i wasnt able to pin point them yet exactly. I have some vague ideas but those might be wrong.

Have you been able to produce map values close to the paper?

VisionEp1:
i also tried many different datasets (the dlib ones, coco, my own) results in regards of comparing with the  https://github.com/AlexeyAB/darknet seems to be the same

VisionEp1:
I also have some gpu resources i can use to help you guys testing. 
(currently the biggest setup is 4x3080ti). 

So my questions would be:

1. what is your best map on the coco data
2. could you provide your converted training and testing xml data for the coco2017 set? 
just to make sure they are identical
(i think thats the best dataset to validate it on since we can bench vs other Framework)

carkaci:
Great job. I am very excited to see yolo-v7 in dlib. Thank you for your valuable effort.

arrufat:
@VisionEp1 Here's the COCO dataset in XML format: http://dlib.net/files/data/dlib_coco_annotations_trainval2017.tar.bz2

I also haven't been able to reproduce the training results on COCO using this implementation with YOLOv5l.
In particular, using this dataset and [this code](https://github.com/dlibml/yolo-object-detector), I obtained these results:

```
overall performance
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.330
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.543
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.348
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.197
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.363
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.391
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.274
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.433
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.457
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.297
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.495
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.548
```

I used 4 V100 (32 GiB each), but I no longer have access to those :(

I can't see any significant difference between this implementation and the official one from Alexey (YOLOv4).
There are some differences between this one and YOLOv5. Notably, the scaling factors between the outputs in YOLOv5.
You are welcome to experiment with the code I shared, and suggest improvements.

arrufat:
Also, I have some improvements that I've been testing for YOLO over the last few months, you can check them here: https://github.com/arrufat/dlib/tree/yolo-improvements

Notably:
- focal loss for objectness and classification
- smoothed l1 loss for regression (to avoid having unbounded gradients)
- use the `truth.detection_confidence` as a positive weight to make up for the class imbalance in the dataset
- set `iou_anchor_threshold` to 0 to automatically find a good threshold based on [ATSS](https://arxiv.org/abs/1912.02424).

VisionEp1:
thanks!
33 to 49 is quite a margin :/..  I also didnt see to many differences but then again the bulky c code from Alexey is very hard to follow (imo). and clearly science focused :). 

Even yolov3 had higher map values then the current dlib implementation. So i guess there has to be some small weird bug somewhere.  I also tried many different dlib yolo architectures which didnt make a difference. My current guess is eigther:

a) some small bug in the yolo loss layer
b) the dlib solver has somehow a hard time with those object detection type loss functions.(i also had bad results with the previous dlib object detectors and spent many hours on it without major improvements (cnn mmod), but excellent results with the dlib classification models).


 



before i start another run with your xml files. would you recommend the default parameters in your training code. (i would put batch and workers high enough but other then that)

Feel free if you have any ideas how i could speed up my search
i will keep you updated

VisionEp1:
Since the yolov4 training tends to be super unstable i don't know if its better for me to use as many cool new features to get the best map, or start with the most base version (for example the previous more yolov3ish setup) and try from there. What do you think?

Also if i shall train a specific model for example the yolov7 feel free to tell me and i'll run the training.

pfeatherstone:
From memory, the ultralytics repos have a special SGD optimizer where weights are grouped and each group has slightly different settings. so I think biases have momentum, no weight decay and quite high learning rate. Convolutional weights have lower learning rate and some weight decay, and another group for batchnorm weights which had some other slightly different settings. And i think all groups had nerestov variant on.  Furthermore, the yolo backbones were pretrained on imagenet. Maybe the combination of those two things would make a difference. @arrufat please correct me if i'm chatting shit.

pfeatherstone:
Also the learning rate scheduler has a slow burning, which I'm pretty sure @arrufat's code already does, but the momentum is inversely cycled as well. Don't know if that makes a difference

pfeatherstone:
@arrufat you submitted PR a while back for semi-supervised pretraining, barlow loss or something. Maybe that would be useful for initializing yolo networks.

VisionEp1:
I dont think the pretraining is the issue. In yolov4 its mostly used to make training 
a) faster b) stable. But does not change a lot in the accuracy. I trained many networks in the other repo with and without pretrained weights. Of course your findings might differ. 

You can set a burn-in in @arrufat code. I cant say anything about the ultralytics repos i didnt use them.  I am completely with you that those adjustment in various ways can improve map. 
But when we discussed on the other repos those things always where like +-0.5 -2% map. not something like 20%+

arrufat:
Sorry, I gave you the wrong configuration. Here's the training setup
- 8 A100 40 GB
- parameters:
``` bash
./build/Release/train \
    --name yolov5l-coco2017 \
    --size 640 \
    --batch-gpu 24 \
    --learning-rate 0.01 \
    --min-learning-rate 0.0001 \
    --gpus 8 \
    --warmup 5 \
    --epochs 300 \
    --cosine \
    --iou-anchor 0.2 \
    --iou-ignore 0.7 \
    --gamma-obj 0.5 \
    --gamma-cls 1.0 \
    --mosaic 0.5 \
    --angle 3 \
    --min-coverage 0.5 \
    --hsi 0.3 0.2 0.1 \
    --scale 0.5 \
    --shift 0.2 \
    --perspective 0.01 \
    --mixup 0.01 \
    data/coco2017 | tee -a training.log

```

- loss:
loss

You'll need to use this branch: https://github.com/dlibml/yolo-object-detector/tree/improvements

VisionEp1:
thanks! will do! i just had another idea. 
The darknet yolo works with subdivision of batches and is very unstable with smaller batches.
You think that might be the reason? (since i will use 12gb ram gpus i will start training now with even lower batches and compare the results)-

arrufat:
It could be, however, I've also tried with smaller YOLOv5 variants so that I could increase the batch size (dlib uses a lot of VRAM, for some reason) but same tendency.

VisionEp1:
alright thanks. just another quick question, what did you use to visualize your training loss like in the last picture. 
That way i could compare it better with my results

arrufat:
> Since the yolov4 training tends to be super unstable i don't know if its better for me to use as many cool new features to get the best map, or start with the most base version (for example the previous more yolov3ish setup) and try from there. What do you think?
> 
> Also if i shall train a specific model for example the yolov7 feel free to tell me and i'll run the training.

I would not use my YOLOv7 implementation, it's not ready for prime time yet.
The YOLOv3 implementation and YOLOv5 are properly implemented (they have the exact same architectures as the original implementations) so I would go with those. Probably with YOLOv5, since you can choose a smaller variant to make the training faster :)

In case you're curious about YOLOv7, here's the current state of it (I am sure it has bugs, since I didn't test it yet)

[yolov7.zip](https://github.com/davisking/dlib/files/9079077/yolov7.zip)
[yolov7.pdf](https://github.com/davisking/dlib/files/9079082/yolov7.pdf)

VisionEp1:
thanks a lot again!. i will start with the yolo5 and report my findings.

arrufat:
> alright thanks. just another quick question, what did you use to visualize your training loss like in the last picture. That way i could compare it better with my results

I used this gnuplot script (it's in the repo as well)

```gnuplot
#!/usr/bin/env gnuplot

reset session

set term png size 1800,960
set output 'loss.png'
set grid

set style line 1 lc rgb '#C00000' lt 1 lw 2 pt 7 pi -1 ps 0.5
set style line 2 lc rgb '#00C000' lt 1 lw 2 pt 7 pi -1 ps 0.5
set style line 3 lc rgb '#0000C0' lt 1 lw 2 pt 7 pi -1 ps 0.5

set multiplot title 'YOLO'

set size 0.5,0.5
set origin 0,0.5
set yrange [0:20]
set title 'loss'
set xlabel 'step'

plot "< awk '$1==\"step#:\" {print $2, $8}' training.log" using 1:2 title 'train' w lines ls 1

set size 0.5,0.5
set origin 0.5,0.5
set title 'learning rate'
set xlabel 'step'
set autoscale y
plot "< awk '$1==\"step#:\" {print $2, $5}' training.log" using 1:2 title '' w lines ls 1

set size 0.25,0.5

set key right bottom
set origin 0,0
set title 'mAP\@0.5'
set yrange [0:1]
set xlabel 'epoch'
plot "< awk '$1==\"EPOCH\" {print $2, $3}' training.log" using 1:2 title '' w linespoints ls 1

set origin 0.25,0
set title 'macro'
plot "< awk '$1==\"EPOCH\" {print $2, $4}' training.log" using 1:2 title 'precision' w linespoints ls 3, \
     "< awk '$1==\"EPOCH\" {print $2, $5}' training.log" using 1:2 title 'recall' w linespoints ls 2, \
     "< awk '$1==\"EPOCH\" {print $2, $6}' training.log" using 1:2 title 'f1-score' w linespoints ls 1

set origin 0.5,0
set title 'micro'
plot "< awk '$1==\"EPOCH\" {print $2, $7}' training.log" using 1:2 title 'precision' w linespoints ls 3, \
     "< awk '$1==\"EPOCH\" {print $2, $8}' training.log" using 1:2 title 'recall' w linespoints ls 2, \
     "< awk '$1==\"EPOCH\" {print $2, $9}' training.log" using 1:2 title 'f1-score' w linespoints ls 1

set origin 0.75,0
set title 'weighted'
plot "< awk '$1==\"EPOCH\" {print $2, $10}' training.log" using 1:2 title 'precision' w linespoints ls 3, \
     "< awk '$1==\"EPOCH\" {print $2, $11}' training.log" using 1:2 title 'recall' w linespoints ls 2, \
     "< awk '$1==\"EPOCH\" {print $2, $12}' training.log" using 1:2 title 'f1-score' w linespoints ls 1
unset multiplot

pause 60; refresh; reread
```

It will generate the `loss.png` image every 60 seconds, then, if you put his HTML file next to it, you can have the poor man's version of Tensorboard:

```html
<!DOCTYPE html>
<html>
    <head>
        <meta http-equiv="refresh" content="60" >
        <title>YOLO Training</title>
    </head>
    <body>
        <center>
            <img src="loss.png"/>
        </center>
    </body>
</html>
```

You could even run:
```
python -m http.server <port>
```
And then connec to localhost:<port>. The port is optional and it defaults to 8000. I usually do this to monitor a training on a remote machine. As I said, the poor man's version of Tensorboard :sweat_smile:

carkaci:
Hi Arrufat. I have used your candidate yolov7.h. However, even for
"--batch-size=1" (RTX3060
<https://www.hepsiburada.com/lenovo-legion-5-intel-core-i7-11800h-16gb-1tb-ssd-rtx3060-freedos-173-fhd-tanabilir-bilgisayar-82jm0012tx--p-HBCV00000VZJFB?magaza=Hepsiburada>-6GB
VRAM), I am getting an insufficient memory error.

arrufat:
> Hi Arrufat. I have used your candidate yolov7.h. However, even for "--batch-size=1" (RTX3060 <https://www.hepsiburada.com/lenovo-legion-5-intel-core-i7-11800h-16gb-1tb-ssd-rtx3060-freedos-173-fhd-tanabilir-bilgisayar-82jm0012tx--p-HBCV00000VZJFB?magaza=Hepsiburada>-6GB VRAM), I am getting an insufficient memory error.

Yes, I told you there were bugs.

pfeatherstone:
Thank you everyone for the hard work! Not easy doing all of this in our free time, even harder to convince employer to do this at work. I'm excited for all of this. I do believe it will attract more users to dlib and therefore more PRs and more goodies.

VisionEp1:
Just a quick update i attached a part of my training loops as an example
[training.log](https://github.com/davisking/dlib/files/9121066/training.log)
 
`.\train.exe --name yolov5l-c
[4.log](https://github.com/davisking/dlib/files/9121092/4.log)
oco2017 --size 512 --batch-gpu 6 --learning-rate 0.01 --min-learning-rate 0.0001 --gpus 4 --epochs 300 --cosine --iou-anchor 0.2 --iou-ignore 0.7 --gamma-obj 0.5 --gamma-cls 1.0 --mosaic 0.5 --angle 3 --min-coverage 0.5 --hsi 0.3 0.2 0.1 --scale 0.5 --shift 0.2 --perspective 0.01 --test-period 30 --mixup 0.01 X:\Datasets\coco2017  | tee -Append -FilePath "training.log"`

I run the code on windows just fyi.
So the map is still increasing.

Whats interesting is when training yolo models you usually get to around 80% of the accuracy quite fast and then it takes a lot of time to gain the last % of map (which makes total sense). But already this area is at a much higher map for the same dataset as we have it with the dlib implementation.

Another finding is that (as u can also see in the loss).
[images.zip](https://github.com/davisking/dlib/files/9121148/images.zip)

It performs worse on the train data without augmentation in testing than on the actual validation data  (i attached some images). Which is a bit wierd and i have to check if the data augmentation is buggy?

I get that the loss value is better on the testdata (no agumentation vs with) but when both are loaded without the train data should perform way better.

it might also be an idea to get start with a very basic v3 model to mimic the results and the training and see how it compares. 
So not any solutions yet just sharing my findings in between.

pfeatherstone:
Just a thought, it might be a good idea to benchmark against other 'vanilla' repositories that haven't optimised the training loops and the image augmentation, such as frgfm/Holocron. If you get similar performance then it's likely there is nothing wrong with the loss function but more an issue with optimising training loops, augmentation, hyper-parameter tuning and things like that.

VisionEp1:
i might also run some training with very low batch sizes on the alexay repo just to make sure thats not the issue.

Another thing which i need to try  is changing stuff like: multi gpu vs 1 gpu, maybe different cuda versions etc.
I had hardware related bugs in the past on many different ml freamworks and sometimes they are hard to detect.
However that doesnt help if neither of us can reproduce map values which are close to the other repos. Did any of you try a small yolo network with 1 gpu only to test this as a possible cause?

I also considered a more "bruto" force method, like the ones used in the automl context (but not on the architecture) where i create a tiny yolo net and load in pretrained weights and run loads of different Trainings regarding the most common settings.

Obviously this might also lead to nothing but do you think the following could give some insights:

I define the most common settings mainly the yolo options, the data augmentation options and scheduler options
with a set of plausible values. Then i start training with random datapoints out of the possible configurations and use the dlib tools to get some insights what the best combinations might be. 

Do you have a ready dlib config which matches any known tiny yolo networks exactly? 
for example https://raw.githubusercontent.com/AlexeyAB/darknet/master/cfg/yolov4-tiny.cfg 
so following the @pfeatherstone approach i can stay as close as possible to one of those networks. 
i think the tiny ones make more sense(even tho currently i train the normal one) since we can get some insights faster

pfeatherstone:
@VisionEp1 So @arrufat created a whole bunch of network definitions. https://github.com/dlibml/dnn/blob/master/src/detection/yolov5.h contains yolov5_n which is small.

arrufat:
@VisionEp1 I am pretty confident about those YOLOv5 definitions, I see no need on trying with the YOLOv3 or YOLOv4 ones.

From my experience, on an internal dataset, the training code I shared achieves an mAP of almost 90%, while the test reaches only about 66%.
If I train the same YOLOv5l model in PyTorch on that dataset, I get about 74% mAP on the test set. So, for me, it was a problem of the network not generalizing well enough. Sadly, these days, I won't have much time to play with it.

pfeatherstone:
Also I would look at the optimizer. @arrufat correct me if i'm wrong but the optimizer is just vanilla SGD. The pytorch repos group weights into 3 different groups depending on what requires a weight decay and biases take higher lr I think. That could make a difference.

arrufat:
Yes, what I am trying to say is that I don't think the performance drop comes from the network architecture. But you don't have to believe me, in the official dlib example, that network is *exactly* the same as the one presented in the YOLOv3 paper: I even tried at some point to fine-tune from the official Darknet53 weights.

I would say that the main differences are:
- upscaling: dlib uses bilinear, yolos use nearest neighbors
- input image: YOLOv3 uses a 0-mean initialized input layer, in dlib you can achieve this by instantiating the network as `darknet::yolov3_train_type net(options, input_rgb_image(0, 0, 0))`;

So, you could use that network in the training code I shared with you, however, I don't think that'll make a difference.

pfeatherstone:
Oh I agree the network definitions are correct. I would use any architecture, even something that isn't official yolo model, maybe with only 1 head, and experiment with different training setups.

VisionEp1:
I completely agree I was not trying to find bugs where no bugs are.  I am very sorry if that came across wrong 


Thanks for the answers I will go ahead and try different training setups

arrufat:
@VisionEp1 no, you are right to question everything, I don't even trust the 1-year-ago version of myself :P

davisking:
> People do this already but using genetic algorithms to find optimal parameters. i don't know how `dlib::find_min_global()` fairs against genetic algorithms, but they have the same goal. 

A genetic algorithm will be much worse.

arrufat:
@VisionEp1 @carkaci I fixed the YOLOv7 version, I've started a training a couple of hours ago, and it seems to work... Feel free to try it out: https://github.com/dlibml/dnn/blob/master/src/detection/yolov7.h

carkaci:
Thank you, I will test it.

carkaci:
I have recently started a training. Everything seems Ok.

arrufat:
> I have recently started a training. Everything seems Ok.

Yes, me too. I started yesterday and after 22 epochs, I got an mAP of 36.91%... Let's see how it goes, but I have the feeling it learns faster than YOLOv5 equivalent models.

pfeatherstone:
@arrufat Do you have an update on training? I'm very interested. Do also have the same curves for yolov5, v4 and v3? It's a lot to ask I know :) ! If they're all similar, then it will probably indicate that the augmentation, training recipes and things like that are what contribute the most to performance.

arrufat:
@pfeatherstone I do not have curves for all those models. Currently, I only have them for YOLOv5m and YOLOv7 under same training settings, except for the batch size, 48×2 vs 14×2, but using the same learning rate scheduler, so YOLOv7 has way more steps per epoch.

Check out the training curves:
# YOLOv5m
image
# YOLOv7
image

# Training config

I have mild data augmentation, as you can see from here.
``` bash
./build/Release/train \
    --name yolov7-coco2017 \
    --size 512 \
    --batch-gpu 14 \  # this is set to 48 for YOLOv5m
    --gpus 2 \
    --warmup 1 \
    --epochs 100 --cosine \
    --learning-rate 0.001 \
    --min-learning-rate 0.0001 \
    --momentum 0.9 \
    --weight-decay 0.0005 \
    --iou-anchor 0.2 \
    --iou-ignore 0.7 \
    --lambda-box 1 \
    --lambda-obj 1 \
    --lambda-cls 1 \
    --gamma-obj 0 \
    --gamma-cls 0 \
    --mosaic 0 \
    --angle 3 \
    --min-coverage 0.5 \
    --hsi 0.0 0.0 0.0 \
    --scale 0.5 \
    --shift 0.2 \
    --perspective 0.00 \
    --mixup 0.00 \
    data/coco2017 | tee -a training.log
```

pfeatherstone:
Interesting v7 learns quicker. Thanks a lot

arrufat:
Maybe we can continue the discussion here: https://github.com/dlibml/yolo-object-detector/discussions/2

carkaci:
Like Arrufat, I also have the feeling it learns faster than YOLOv5.  I set
"--patience=200000", that is why training takes 5-6 days. I will let you
know the result when finished.

arrufat:
>  Like Arrufat, I also have the feeling it learns faster than YOLOv5.  I set
> "--patience=200000", that is why training takes 5-6 days. I will let you
> know the result when finished.

Oh, `--patience` uses the number of epochs, not the training steps in the YOLO repository. You can set it to a floating point number, though, but 200000 is definitely way too high.

VisionEp1:
@carkaci @arrufat 
if possible can we move it to your discussion post?

so we have all the "insights" over there as well

Issue #1742, opened 2019-04-25
How to turn DLIB_NO_GUI_SUPPORT off?


Hotpotboy:
/usr/local/include/dlib/gui_core/gui_core_kernel_2.h:11:2: error: "DLIB_NO_GUI_SUPPORT is defined so you can't use the GUI code.
      Turn DLIB_NO_GUI_SUPPORT off if you want to use it."
#error "DLIB_NO_GUI_SUPPORT is defined so you can't use the GUI code.  Turn DLIB_NO_GUI_SUPPORT off if you want to use it."
 ^
/usr/local/include/dlib/gui_core/gui_core_kernel_2.h:12:2: error: "Also make sure you have libx11-dev installed on your system"
#error "Also make sure you have libx11-dev installed on your system"
 ^

Hotpotboy:
???

Issue #579, opened 2017-05-07
DLIB_ISO_CPP_ONLY is broken


emptyVoid:
Here's what I get when trying to use `dlib::shape_predictor` with `DLIB_ISO_CPP_ONLY` defined:
```
In file included from /usr/local/include/dlib/threads/posix.h:4:0,
                 from /usr/local/include/dlib/threads/threads_kernel.h:13,
                 from /usr/local/include/dlib/threads.h:11,
                 from /usr/local/include/dlib/image_processing/shape_predictor.h:14,
                 from test.cpp:12:
/usr/local/include/dlib/threads/threads_kernel_2.h:7:2: error: #error "DLIB_ISO_CPP_ONLY is defined so you can't use this OS dependent code.  Turn DLIB_ISO_CPP_ONLY off if you want to use it."
 #error "DLIB_ISO_CPP_ONLY is defined so you can't use this OS dependent code.  Turn DLIB_ISO_CPP_ONLY off if you want to use it."
  ^
```

Please note, (at least as of 19.1) commenting out `dlib::shape_predictor_trainer` in `dlib/image_processing/shape_predictor.h` fixes the issue.

davisking:
It's not broken.  This is just what happens when you disable that part of the code.  I don't know what you would expect to happen.

Although, honestly, I'm tempted to remove the DLIB_ISO_CPP_ONLY switch since I suspect mightily that no one really uses it.

emptyVoid:
>  I don't know what you would expect to happen.

Wouldn't you expect it to work as outlined in the "documentation"?

> This is a #define directive that you can set to cause the library to exclude all non ISO C++ code (The things in the API wrappers section and any objects that depend on those wrappers).

davisking:
Yes. Some things depend on that stuff. Like part of the shape prediction tooling.

emptyVoid:
Though the predictor itself doesn't depend on threads, and the trainer's declaration could be moved to a separate header (not sure why the thread part has to be OS dependent in the first place, since DLib requires a C++11 compiler why can't it use `std::thread`?).

davisking:
Because a lot of dlib was written before there was C++11 support in compilers.  Also, yes, many things could be split up.  You can almost always split source files into smaller parts, we could make it so there are 10,000 source files instead of about 1000.  But I'm not doing that.

Do you have a use case that is somehow impacted by what you are posting about?

emptyVoid:
[...]

PR #2384, opened 2021-06-29
Fix cannot compile iso only code (#579)

frostbane:
also fixing (#1742)

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -261,7 +261,7 @@ if (NOT TARGET dlib)
    toggle_preprocessor_switch(DLIB_USE_MKL_FFT)
 
 
-   set(source_files 
+   set(source_files
       base64/base64_kernel_1.cpp
       bigint/bigint_kernel_1.cpp
       bigint/bigint_kernel_2.cpp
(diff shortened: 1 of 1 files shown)


dlib/CMakeLists.txt
```diff
@@ -412,6 +411,7 @@ if (NOT TARGET dlib)
             gui_widgets/drawable.cpp
             gui_widgets/canvas_drawing.cpp
             gui_widgets/style.cpp
+            gui_widgets/fonts.cpp
```
davisking, line 414: This was moved out of here in https://github.com/davisking/dlib/pull/2273 so that the font drawing could be used with draw_string().  Although it shouldn't have gone to the top since that's for only ISO standard C++ stuff.  It should go up at https://github.com/davisking/dlib/pull/2384/files#diff-49efe726bb1d239c7d05dd384d1373297dfac8392bd41748eb5e45fd618702ffR289.  Adding it up there works for what you are doing right?

davisking (changes requested):

dlib/CMakeLists.txt
frostbane: I think 99% of the dlib users don't use the ISO only option, but in my case I wanted to use only the base64 and random functions.

If I put back the `gui_widgets/fonts.cpp` here, I guess it would break #2273 ?

I'll create another pull request so that it would not break the previous commit.

davisking: Yep, exactly

davisking:
Thanks for the PR  :)

PR #2393, opened 2021-07-22
Add LayerNorm documentation

arrufat:
I just realized I never added the documentation to the LayerNorm layer.
I also corrected a couple of typos in the batch_normalize function documentation.


davisking:
Thanks :)

PR #2395, opened 2021-07-26
return the projective transform in extract_image_4points

arrufat:
This PR might introduce a breaking change.
Currently, `extract_image_4points` returns `void`, this PR changes it to return the `point_transform_projective` that maps points from the input image to the output image.

The main purpose of this is to be able to do "perspective" data augmentation on object detection, which need this transform to map the modified bounding boxes.

### DIFF
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -2118,7 +2118,7 @@ namespace dlib
     template <
         typename image_type
         >
-    void extract_image_4points (
+    point_transform_projective extract_image_4points (
         const image_type& img_,
         image_type& out_,
         const std::array<dpoint,4>& pts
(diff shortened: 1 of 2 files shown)


davisking:
Looks good.  Why would it be a breaking change?  It looks good and like it shouldn't bother anyone (unless they had some kind of template metaprogramming code that was asserting this thing returned void, but other than that I can't see how it could break any existing code.  And I would be surprised if anyone was doing something so odd).

arrufat:
Yes, or someone relying on a function pointer with that signature.

davisking:
Yeah. Or that. I wouldn't worry about that though :)

arrufat:
Yes, I also think almost nobody is doing that.

arrufat:
For those who are curious, with this PR, we can now do this kind of data augmentation (image from COCO)
image

PR #2408, opened 2021-08-01
Add function to compute string dimensions in pixels

arrufat:
In this PR, I added a function to compute the dimensions (in pixels) of the given string, using a particular font, similar to OpenCV's [getTextSize](https://docs.opencv.org/4.5.3/d6/d6e/group__imgproc__draw.html#ga3d2abfcb995fd2db908c8288199dba82).

It is mostly copy-paste of the `draw_string` function, but I couldn't see an easy way to reuse it.

I have some doubts:
- right now, I am returning a `std::pair<long, long>` so we can do `auto [w, h] = compute_string_dimensions("hello");`
- maybe `width` and `height` should be out parameters in the function, instead, or return a `matrix<long>`? I find structured bindings really convenient, though.
- I kept the `first` and `last` parameters, for coherence with the other methods, but I don't know if they are useful here.
- I am not sure about the name of this function, open to suggestions... like `get_string_shape` or other combinations.

With this function, we can draw boxes around text, get some pretty nice looking bounding boxes (drawn 100% using dlib):
detections

### DIFF
--- a/dlib/image_transforms/draw.h
+++ b/dlib/image_transforms/draw.h
@@ -228,6 +228,61 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 
     template <
+        typename T, typename traits,
+        typename alloc
+    >
+    std::pair<long, long> compute_string_dimensions (
+        const std::basic_string<T, traits, alloc>& str,
+        const std::shared_ptr<font>& f_ptr = default_font::get_font(),
+        typename std::basic_string<T,traits,alloc>::size_type first = 0,
+        typename std::basic_string<T,traits,alloc>::size_type last = (std::basic_string<T,traits,alloc>::npos)
+    )
+    {
+        using string = std::basic_string<T, traits, alloc>;
+        DLIB_ASSERT((last == string::npos) || (first <= last && last < str.size()),
+                    "\tvoid dlib::compute_string_dimensions()"
+                    << "\n\tlast == string::npos: " << ((last == string::npos)?"true":"false")
+                    << "\n\tfirst: " << (unsigned long)first
+                    << "\n\tlast:  " << (unsigned long)last
+                    << "\n\tstr.size():  " << (unsigned long)str.size());
+
+        if (last == string::npos)
+            last = str.size();
+
+        const font& f = *f_ptr;
+
+        long height = f.height();
+        long width = 0;
+        for (typename string::size_type i = first; i <= last; ++i)
+        {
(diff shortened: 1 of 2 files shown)


dlib/image_transforms/draw.h
```diff
@@ -225,6 +225,61 @@ namespace dlib
...
+
+    template <
+        typename T, typename traits,
+        typename alloc
+    >
+    std::pair<long, long> compute_string_dimensions (
+        const std::basic_string<T, traits, alloc>& str,
+        const std::shared_ptr<font>& f_ptr = default_font::get_font(),
+        typename std::basic_string<T,traits,alloc>::size_type first = 0,
+        typename std::basic_string<T,traits,alloc>::size_type last = (std::basic_string<T,traits,alloc>::npos)
```
davisking, line 238: I think it's fine to omit these two arguments.  I highly doubt anyone will ever have real reason to use them for this function.

dlib/image_transforms/draw.h
```diff
@@ -225,6 +225,61 @@ namespace dlib
         }
     }
 
+// ----------------------------------------------------------------------------------------
+
+    template <
+        typename T, typename traits,
+        typename alloc
+    >
+    std::pair<long, long> compute_string_dimensions (
```
davisking, line 234: It would be nicer to return a struct with explicitly named width and height fields.  You can still use structured bindings with a simple struct too

arrufat: Something like `string_dims` (following the convention of `chip_dims`?
Then, how about calling the function `compute_string_dims`?

davisking: Works for me 😁

arrufat: Done :)

dlib/image_transforms/draw_abstract.h
```diff
@@ -79,25 +79,42 @@ namespace dlib
             - The drawn rectangle will have edges that are thickness pixels wide.
     !*/
 
+// ----------------------------------------------------------------------------------------
+
+    struct string_dims
+    {
+        /*!
+            WHAT THIS OBJECT REPRESENTS
+                This is a simple structs that represents the size (width and height) of a
```
davisking, line 88:
```suggestion
                This is a simple struct that represents the size (width and height) of a
```

davisking (approved):
Nice

PR #2411, opened 2021-08-06
Add progress information to console_progress_indicator

arrufat:
This PR adds progress information (current/total and percent) to console_progress_indicator.

Before it used to look like this:
```
Time remaining: 12 seconds.
```
But now it looks like this
```
Progress: 2172/5000 (43.44%). Time remaining: 12 seconds.
```

I've found myself adding this extra information many times in my code: most of the time, I want to know the actual number of elements completed out of the total, and the actual progress.

If you don't think this is useful, either close this PR, or tell me to add an extra `display_percent` parameter to the function.

### DIFF
--- a/dlib/console_progress_indicator.h
+++ b/dlib/console_progress_indicator.h
@@ -30,7 +30,8 @@ namespace dlib
                     }
 
                 The above code will print a message to the console each iteration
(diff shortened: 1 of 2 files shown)


davisking:
Cool, works for me :)

PR #2414, opened 2021-08-15
Delete .hgtags

arrufat:
After seeing commit https://github.com/davisking/dlib/commit/0a5d5a2c686ff5e69eed8b4b19b0501f29a24d1d we might want to get rid of this as well.


davisking:
Yeah, don't need this anymore either :)

PR #2416, opened 2021-08-18
Handle tag layers as inputs to disable_duplicative_biases

arrufat:
`disable_duplicative_biases` prevents from building successfully if the input of the batch normalization layer is a tag layer, but there is no repeat layer afterwards. This PR fixes that.


davisking (approved):
Thanks :)

PR #2417, opened 2021-08-18
Show number of parameters if net is allocated

arrufat:
This PR makes the trainer print the number of parameters next to the network size if it has already been allocated (to avoid printing 0)


davisking (approved):
Nice

PR #2428, opened 2021-09-10
Add intel mkl search path

z-pc:
Add some paths in find_blas.cmake with new Intel mkl library.
From here: [https://software.intel.com/content/www/us/en/develop/tools/oneapi/components/onemkl.html](url)
Tested on dnn_imagenet_ex.


davisking:
Sweet, thanks :)

PR #2431, opened 2021-09-11
Allow setting custom cuda compute capabilities

arrufat:
As you mentioned in a previous issue, autodetecting the cuda capabilities might cause some issues, where people build dlib on their local fancy GPU, and they deploy on another system with an older GPU...
(I had this problem last week at work, but the other way around, the fancy GPU was on the deployment side, and didn't support `sm_50`)

For this reason, this PR is quite conservative: if the user specifies no `DLIB_USE_CUDA_COMPUTE_CAPABILITIES`, then it will use the `sm_50` as before.
However, setting `DLIB_USE_CUDA_COMPUTE_CAPABILITIES=61,75` will build dlib with those targets instead.
What do you think?


davisking:
Yeah, this is the right way to do it I think :)

PR #2432, opened 2021-09-15
Missing include for `dlib::loss_multiclass_log_per_pixel_`

Darker:
I was trying to compile the examples and encountered this issue after moving `rgb_label_image_to_index_label_image` to cpp file. Headers should include all symbols they mention.


davisking:
Indeed, thanks for the PR :)

PR #2435, opened 2021-09-25
Don't try to use labels in unsupervised losses

arrufat:
I hope that is the right way of fixing this...


arrufat:
It's obviously not 😅, I will check later :)

PR #2436, opened 2021-09-26
Fix trainer with unsupervised loss

arrufat:
Let's see if the tests pass on this one. It works on my machine.
I will then try to figure out a way of reusing the code between `send_job` functions

### DIFF
--- a/dlib/dnn/trainer.h
+++ b/dlib/dnn/trainer.h
@@ -1194,9 +1194,6 @@ namespace dlib
             job.have_data.resize(devs);
             job.test_only = test_only;
 
-            // check if the iterator points to anything
-            const bool have_labels = sizeof(*lbegin) != 1;
-
             // chop the data into devs blocks, each of about block_size elements.
             const double block_size = num / static_cast<double>(devs);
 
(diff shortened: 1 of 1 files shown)


dlib/dnn/trainer.h
```diff
@@ -1194,6 +1194,9 @@ namespace dlib
             job.have_data.resize(devs);
             job.test_only = test_only;
 
+            // check if the iterator points to anything
+            const bool have_labels = sizeof(*lbegin) != 1;
```
davisking, line 1198: Oof, yeah, I see the bug you are trying to fix.  This bit of code here is not right though.  You can't dereference an iterator to nothing, that's undefined behavior.  And it might not refer to a type that has a size of 1 as well.  So something like `std::optional<label_iterator>` is needed instead to indicate if we actually have a `lbegin`.  Can't use `std::optional` in dlib yet though :(

Maybe use `std::pair<bool, label_iterator>` instead.

arrufat: Yes, I wasn't convinced about this either. I will revert the change.
I thought checking the `labels[i].size()` would work (my first attempt). But it didn't, so i duplicated the `send_job`code...

I'm not sure i get what you're saying about the new iterator type, I'll check later. Thank you!

arrufat:
After reading the trainer code more carefully, I think you already figured out how to handle this case. I've just pushed a simple fix.

davisking:
Oh right, I totally forgot about `no_label_type`.  Yeah, this is the right way.  It's even noted in the documentation :)

PR #2437, opened 2021-09-26
Add support for reading webp

sandsmark:
I wasn't sure if you wanted a bundled version of libwebp, but it's easy to build and was easy to add.

The code uses libwebp's own unpacking for the common pixel formats, and falling back to a slow path using assign_pixel like the other image decoders for the rest.


arrufat:
Oh, I have been encoutering more and more webp images recently, so that would definetly come in handy for me.
However, since this format is not as common as PNG or JPEG, I would argue that there's less value in bundling it into dlib...
I have been thinking about adding support for webp for some time, but I would have gone with linking with the system version, instead. Let's see what Davis has to say :)

davisking:
Sweet.  But yeah, don't add webp to dlib/external.  I would prefer dlib/external not exist honestly but people essentially asked for it over and over and I eventually broke down :) :(.  It's way better if people link to the system copy of other libraries.  This way when there are important updates to those libraries they get them automatically as part of normal package updates.  Bundling into dlib/external inhibits that.  It also introduces the chance for someone to link in two copies of those libraries and create ODR violations depending on how they setup their builds.

Anyway, for now let's not add it to dlib/external.  If lots of people express confusion about how to link to webp then we can consider adding it.  You can put up a separate PR that adds it to dlib/external/libwebp if you want just to keep the code somewhere convenient should we decide later to include it there `¯\_(ツ)_/¯`

sandsmark:
> It also introduces the chance for someone to link in two copies of those libraries and create ODR violations depending on how they setup their builds.

The way e. g. pdfium solves that is to add a prefix to all symbols in their bundled version, but it quickly becomes a huge mess.

> Anyway, for now let's not add it to dlib/external. If lots of people express confusion about how to link to webp then we can consider adding it. You can put up a separate PR that adds it to dlib/external/libwebp if you want just to keep the code somewhere convenient should we decide later to include it there `¯\_(ツ)_/¯`

I just branched locally to keep the commit, no need for messing up with a separate PR.

Now also remembered to add all the files so appveyor hopefully succeeds...

dlib/pixel.h
```diff
@@ -474,7 +474,9 @@ namespace dlib
     struct pixel_traits<rgb_pixel>
     {
         constexpr static bool rgb  = true;
+        constexpr static bool bgr  = false;
```
sandsmark, line 477: I wasn't sure about adding these, but felt like a waste to not implement support for using the BGR decoding in libwebp, especially since opencv uses it (even if the rest of the necessary support hasn't been added, there's at least one less place to update in the future if it is implemented).

VisionEp1:
Isn't it better to implement an specific bgr pixel then to avoid confusion ?

davisking:
There is a bgr pixel in dlib already: http://dlib.net/dlib/pixel.h.html#bgr_pixel. Why are the bgr pixel traits changes needed here?

dlib/pixel.h
davisking: Er, I just looked at the code. I see why the pixel traits are updated, looked on my phone earlier :)

Anyway.  Yeah this is fine.  Although rename `bgr` to `bgr_layout`.  You then also don't need the `bgr_alpha` field. Since `bgr_layout` is a modifier tells you everything that's needed, which is that the pixel's memory layout goes BGR(A) rather than RGB(A).

davisking: And update the docs at the top of this file to explain the meaning of `bgr_layout` :)

dlib/image_loader/webp_loader.h
```diff
@@ -0,0 +1,141 @@
...
+        void get_info();
+        void read_argb( unsigned char *out, const size_t out_size, const int out_stride ) const;
+        void read_rgba( unsigned char *out, const size_t out_size, const int out_stride ) const;
+        void read_bgra( unsigned char *out, const size_t out_size, const int out_stride ) const;
+        void read_rgb( unsigned char *out, const size_t out_size, const int out_stride ) const;
+        void read_bgr( unsigned char *out, const size_t out_size, const int out_stride ) const;
+
+        int height_;
+        int width_;
+        std::vector<unsigned char> data;
```
davisking, line 91:
```suggestion
        std::vector<unsigned char> data_;
```
To be consistent with the naming of the other members.

dlib/image_loader/webp_loader.cpp
```diff
@@ -0,0 +1,130 @@
...
+          throw image_load_error("webp_loader: decoding failed");
+        }
+    }
+
+// ----------------------------------------------------------------------------------------
+
+    void webp_loader::read_bgra( unsigned char *out, const size_t out_size, const int out_stride ) const
+    {
+        if (!WebPDecodeBGRAInto(data.data(), data.size(), out, out_size, out_stride)) {
+          throw image_load_error("webp_loader: decoding failed");
```
davisking, line 100: Here and elsewhere, be sure to use 4 spaces in indents.

dlib/image_loader/webp_loader.cpp
```diff
@@ -0,0 +1,130 @@
...
+        if (!WebPDecodeRGBAInto(data.data(), data.size(), out, out_size, out_stride)) {
+          throw image_load_error("webp_loader: decoding failed");
+        }
+    }
+
+// ----------------------------------------------------------------------------------------
+
+    void webp_loader::read_bgra( unsigned char *out, const size_t out_size, const int out_stride ) const
+    {
+        if (!WebPDecodeBGRAInto(data.data(), data.size(), out, out_size, out_stride)) {
```
davisking, line 99:
```suggestion
        if (!WebPDecodeBGRAInto(data.data(), data.size(), out, out_size, out_stride)) 
        {
```
Here and elsewhere, match the brace style used in the rest of dlib.

sandsmark:
> Cool. This looks like it's ready for review again. Is that right? :)

Yup.

dlib/pixel.h
```diff
@@ -72,6 +74,29 @@ namespace dlib
                 - min() == 0 
                 - max() == 255
                 - is_unsigned == true
+            - else if (bgr_layout == true) then
```
davisking, line 77: This isn't a separate case from `rgb`.  The code sets both `rgb==true` and `bgr_layout==true` at the same time (which is what I think it should do).  But the docs should reflect that.  I also don't think there should be a `bgr_alpha` field.  A pixel with `pixel_traits::rgb_alpha == true && pixel_traits::bgr_layout == true` indicates it's a BGR layout pixel with an alpha for instance.  The docs should reflect this by updating the notes in the `rgb` and `rgb_alpha` `if` statement to indicate at such.

dlib/image_loader/webp_loader.h
```diff
@@ -0,0 +1,145 @@
...
+            if ( pixel_traits<pixel_type>::rgb )
+            {
+                read_rgb( output, output_size, stride );
+                return;
+            }
+            if ( pixel_traits<pixel_type>::bgr_layout )
+            {
+                read_bgr( output, output_size, stride );
+                return;
+            }
```
davisking, line 65: For `bgr_pixel` this code will do the wrong thing, since `pixel_traits<bgr_pixel>::rgb==true`.  Shouldn't it be this?

```suggestion
            if (pixel_traits<pixel_type>::rgb_alpha)
            {
                if (pixel_traits<pixel_type>::bgr_layout)
                    read_bgra(output, output_size, stride);
                else             
                    read_rgba(output, output_size, stride);   
            }
            if (pixel_traits<pixel_type>::rgb)
            {
                if (pixel_traits<pixel_type>::bgr_layout)
                    read_bgr(output, output_size, stride);
                else             
                    read_rgb(output, output_size, stride);                    
                return;
            }
```

davisking (changes requested):
Looks good, I see a bug in the selection of which code path to take though.  

Can you add a unit test that loads webp images for each of these types (like hit each of the code paths) to make sure it all works?  It can just be part of dlib/test/ and only built in if webp support is around.  I find that such testing is most easy if dlib/example/file_to_code_ex.cpp is used to embed a few tiny images into the code.  You write them out to disk and then test against them.

arrufat:
Is this far from being complete? I don't know if I am skilled enough to do this, but maybe I can try?

arrufat:
@sandsmark Is this far from being complete? I don't know if I am skilled enough to do this, but maybe I can try?

PR #2443, opened 2021-10-19
[TYPE_SAFE_UNION] upgrade

pfeatherstone:
This is a revamp of ```dlib::type_safe_union``` with the following features:
* variadic template parameters
* copy semantics (as well as move semantics of course)
* `emplace()`
* `visit()` : similar to `apply_to_contents()` except it can return non-void types (so a bit closer to ```std::visit```)
* Added `overloaded` as in [here](https://en.cppreference.com/w/cpp/utility/variant/visit) that works with C++11 onwards

### DIFF
--- a/dlib/type_safe_union/type_safe_union_kernel.h
+++ b/dlib/type_safe_union/type_safe_union_kernel.h
@@ -126,6 +126,9 @@ namespace dlib
         template<typename T>
         struct is_valid : internal::is_any<T,Types...> {};
 
+        template<typename T>
+        using is_valid_check = typename std::enable_if<is_valid<T>::value, bool>::type;
+
         template <size_t I>
         struct get_type : internal::variant_get_type<I,Types...> {};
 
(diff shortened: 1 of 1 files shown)


pfeatherstone:
It would be nice to refactor `apply_to_contents` to `visit`. I think it's more appropriate and users can relate that to ```std::variant```. But that would break existing user code which we can't do. Similar issue to all the rants online about "to break or not to break C++ ABI"

pfeatherstone:
Oh, I haven't done the abstract file. Will do that tomorrow

pfeatherstone:
Ok the error is due to ```operator()``` member functions being private. I tried doing something similar using ```std::variant``` and ```std::visit``` and you get similar errors. They use either ```result_of``` or ```invoke_result``` and they can't deduce return types. Don't know what to do other than split `apply_to_contents` into `apply_to_contents` and `visit` where the `apply_to_contents` always returns `void` and therefore doesn't need to deduce return type, and `visit` can return non-void types but requires the visitor to be publicly visible.

pfeatherstone:
Ok and abstract file updated. @davisking what do you think?

PR #2450, opened 2021-10-29
[INVOKE] C++11 backport of std::invoke, std::invoke_result, std:apply and std::make_from_tuple

pfeatherstone:
See title

### DIFF
--- a/dlib/invoke.h
+++ b/dlib/invoke.h
@@ -1,0 +1,96 @@
+// Copyright (C) 2016  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_INVOKE_Hh_
+#define DLIB_INVOKE_Hh_
+
+#include <functional>
+#include <type_traits>
+#include "utility.h"
+
+namespace dlib
+{
+    // ----------------------------------------------------------------------------------------
+
+#if __cplusplus < 201703L
+    namespace detail {
+        template< typename F, typename ... Args >
+        auto INVOKE(F&& fn, Args&& ... args)
+        -> typename std::enable_if<
+                std::is_member_pointer<typename std::decay<F>::type>::value,
+                decltype(std::mem_fn(fn)(std::forward<Args>(args)...))>::type
+        {
+            return std::mem_fn(fn)(std::forward<Args>(args)...) ;
+        }
+
+        template< typename F, typename ... Args >
+        auto INVOKE(F&& fn, Args&& ... args)
+        -> typename std::enable_if<
+                not std::is_member_pointer<typename std::decay<F>::type>::value,
+                decltype(std::forward<F>(fn)(std::forward<Args>(args)...))>::type
+        {
+            return std::forward<F>(fn)(std::forward<Args>(args)...);
+        }
+
+        template< typename AlwaysVoid, typename, typename...>
+        struct invoke_result {};
+
+        template< typename F, typename... Args >
+        struct invoke_result< decltype( void(INVOKE(std::declval<F>(), std::declval<Args>()...)) ), F, Args...>
+        {
+            using type = decltype( INVOKE(std::declval<F>(), std::declval<Args>()...) );
+        };
+    }
+
+    template< typename F, typename... Args>
+    auto invoke(F && f, Args &&... args)
+    -> decltype(detail::INVOKE(std::forward<F>( f ), std::forward<Args>( args )...))
+    {
+        return detail::INVOKE(std::forward<F>( f ), std::forward<Args>( args )...);
+    }
+
+    template< typename F, typename... Args >
+    struct invoke_result : detail::invoke_result< void, F, Args...> {};
+
+    template< typename > struct result_of;
+
+    template< typename F, typename... Args >
+    struct result_of< F(Args...) > : detail::invoke_result< void, F, Args...> {};
+
+#else
+    using std::invoke;
+    using std::invoke_result;
+    using std::result_of;
+#endif
+
+    // ----------------------------------------------------------------------------------------
+
+#if __cplusplus < 201703L
+    namespace detail {
+        template<typename F, typename Tuple, std::size_t... I>
+        auto apply_impl(F&& fn, Tuple&& tpl, index_sequence<I...>)
+        -> decltype(invoke(std::forward<F>(fn),
+                           std::get<I>(std::forward<Tuple>(tpl))...))
+        {
+            return invoke(std::forward<F>(fn),
+                          std::get<I>(std::forward<Tuple>(tpl))...);
+        }
+    }
+
+    template<typename F, typename Tuple>
+    auto apply(F&& fn, Tuple&& tpl)
+    -> decltype(detail::apply_impl(std::forward<F>(fn),
+                                   std::forward<Tuple>(tpl),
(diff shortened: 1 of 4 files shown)


pfeatherstone:
@davisking what do you think? More unit tests ? Any tests in particular ? Otherwise, I think it's ready. I don't have a windows machine so I tested it on godbolt using various MSVC compilers and playing with the c++ standards and everything behaved as expected (eventually...). I haven't bothered to add constexpr support as I want it to be as simple and as minimal as possible. It is still fairly optimal. When compiling with -O3 I get the same assembly as std::invoke using clang and gcc using c++11 so I didn't feel the need to make it more complicated.

dlib/invoke.h
```diff
@@ -0,0 +1,120 @@
...
+
+#if defined(_MSC_VER)
+    #if _MSC_VER >= 1900
+        #define DLIB_HAVE_INVOKE 1
+    #endif
+#else
+    #if DLIB_CPLUSPLUS >= 201703L
+        #define DLIB_HAVE_INVOKE 1
+    #endif
+#endif
```
davisking, line 22: Having backports of new C++ stuff is cool.  And at some point we should switch it to just use the one in the standard library.  But at any given time we should commit to one implementation so we don't have to have this kind of stuff.  Like just don't have any `#ifdef` stuff.  Make it always use the implementation you wrote.  Then someday we can delete it an make it just say `using std::whatever`.

But in the mean time it's better to avoid needing to fight with visual studio's nonsense.  And like, I'm not convinced this is robust since `_MSC_VER == 1900 ? 201103L : _MSVC_LANG` ought to be at least `_MSC_VER <= 1900 ? 201103L : _MSVC_LANG`.

Trying to weave through the parts of MSVC support for current C++ standards is always awful.  It would be a miracle if we didn't get people commenting later about how in some particular update of MSVC or with some setting something in here was failing.  MSVC is the main reason I wait so long to let dlib use current C++ features :(


Anyway, can you remove the #ifdef stuff and just have the one version of what's here?  Same for the other file.

pfeatherstone:
Ok but we should rename. If someone is compiling with c++17 enabled and has written 

using namespace std;
using namespace dlib;
invoke();

They are going to get a compiler error saying "ambiguous call..."

pfeatherstone:
What I like about #ifdef stuff is that you automatically use what's in the standard if it's available. We could maybe do that if compiling with gcc and clang. With MSVC, always use custom implementation ??

pfeatherstone:
It also means you automatically get constexpr support if you're compiling with C++17 for example.

pfeatherstone:
I have tried with pretty much every MSVC compiler on godbolt and with c++11, c++14, c++17 and c++20. I can check again on Sunday when I get back to my PC.

pfeatherstone:
But I can understand that you would rather remove the possibility of anyone ever complaining about MSVC build failed. If so, what names would you recommend ? Not trivial for the index sequence stuff

davisking:
Yeah that’s true and annoying. But it only happens if they try to call one of these methods though.  This is the case for many names in dlib already. It’s why there is a namespace in the first place. So 🤷‍♂️

davisking:
And trust me. You haven't scratched the surface of msvc variability :(

davisking:
Another reason to have just one implementation is to make code that uses it absolutely portable. It would be easy for someone to write something that only worked with the std version and not realize, thinking they are using dlib so they can copy it anywhere dlib works.

pfeatherstone:
Okidok. Will just remove the ifdefs and mention in the docs always to use dlib::invoke(), not invoke().

pfeatherstone:
@davisking I've taken your comments on board. Also, I find ```std::make_from_tuple``` very useful and i think it would fit in ```dlib/invoke.h``` nicely as it works with ```std::tuple```, ```std::array```, ```std::pair``` AND ```std::tuple```. The implementation uses similar technique to ```dlib::apply``` and is basically invoking a constructor, which is just another type of function. What do you think:

```
    namespace detail {
        template <class T, class Tuple, std::size_t... I>
        constexpr T make_from_tuple_impl( Tuple&& t, index_sequence<I...> )
        {
          return T(std::get<I>(std::forward<Tuple>(t))...);
        }
    } // namespace detail
    
    template <class T, class Tuple>
    constexpr T make_from_tuple( Tuple&& t )
    {
        return detail::make_from_tuple_impl<T>(std::forward<Tuple>(t),
            make_index_sequence<std::tuple_size<remove_reference_t<Tuple>>::value>{});
    }
```

Indeed ```std::tuple_size``` works with all 4 of the containers I mentioned above.

davisking:
Yeah, `make_from_tuple` is nice too :)

pfeatherstone:
Hmm, last commit simplified some stuff, or so I thought. ```dlib::invoke_result``` and ```dlib::is_invocable``` is not quite right with MSVC now. I cannot explain it. Godbolt is giving me the same error, but no helpful message. On linux, everything works as expected. This is interesting

pfeatherstone:
Does MSVC allow you to pass rvalues to a function that expects a non-const reference/lvalue ?

pfeatherstone:
Yeah, WTF. How is https://godbolt.org/z/P45hG3919 allowed to compile with MSVC ?

pfeatherstone:
> Hmm, last commit simplified some stuff, or so I thought. `dlib::invoke_result` and `dlib::is_invocable` is not quite right with MSVC now. I cannot explain it. Godbolt is giving me the same error, but no helpful message. On linux, everything works as expected. This is interesting

Note they are correct. They behave the same as ```std::invoke_result``` and ```std::is_invocable```.
So the test it's failing on, isn't supposed to fail. And indeed on linux it doesn't but MSVC weirdly allows you to pass rvalue references to functions expecting lvalue references.

pfeatherstone:
> Yeah, WTF. How is https://godbolt.org/z/P45hG3919 allowed to compile with MSVC ?

Anyone else agree this showcases a bug with MSVC?

pfeatherstone:
Someone else has spotted this:
https://stackoverflow.com/questions/53136307/msvc-binds-non-const-lvalue-reference-to-an-rvalue-temporary

pfeatherstone:
Yeah, so i think the CI pipelines should make sure to build with ```/Za``` when using MSVC. Why would you not build your code in a C++ standards conforming way ... ?

pfeatherstone:
Stupid MSVC!

davisking:
Heh, this is what I'm saying about MSVC.  It is an outrageous compiler.  I wish they would just abandon it and use clang or something.

pfeatherstone:
Yeah absolutely. I would have though Microsoft would have embraced LLVM/clang just like Apple did.

pfeatherstone:
Actually, apparently you can now get Clang for windows.

davisking:
> Actually, apparently you can now get Clang for windows.

Right. But MSVC is still the main compiler people use so we have to keep dealing with it :(

davisking:
Anyway, this PR ready? :)

pfeatherstone:
Yes it is. Shall we wait for tests to finish just in case ? I removed the test that MSVC failed on because of all this reference crap.

davisking:
Yeah definitely. I'll look after they pass :)

davisking (approved):
Sweet, looks good :)

PR #2452, opened 2021-11-01
Replace sgd-based fc classifier with svm_multiclass_linear_trainer

arrufat:
As you recommended, I replaced the sgd-based fc classifier with svm_multiclass_linear_trainer.
You were right, after setting the right C parameter, it results in higher accuracy while training orders of magnitude faster.
(that book is already ordered!)

Thank you :)

### DIFF
--- a/examples/dnn_self_supervised_learning_ex.cpp
+++ b/examples/dnn_self_supervised_learning_ex.cpp
@@ -293,10 +293,10 @@ try
     model::feats fnet(layer<5>(net));
     // And we will generate all the features for the training set to train a multiclass SVM
     // classifier.
-    std::vector<matrix<double, 0, 1>> features;
+    std::vector<matrix<float, 0, 1>> features;
     cout << "Extracting features for linear classifier..." << endl;
     features = fnet(training_images, 4 * batch_size);
-    svm_multiclass_linear_trainer<linear_kernel<matrix<double,0,1>>, unsigned long> trainer;
+    svm_multiclass_linear_trainer<linear_kernel<matrix<float,0,1>>, unsigned long> trainer;
     trainer.set_num_threads(std::thread::hardware_concurrency());
     // The most appropriate C setting could be found automatically by using find_max_global().  See the docs for
     // find_max_global() for further information and take particular note of model_selection_ex.cpp.
--- a/examples/dnn_self_supervised_learning_ex.cpp
+++ b/examples/dnn_self_supervised_learning_ex.cpp
@@ -298,6 +298,8 @@ try
     features = fnet(training_images, 4 * batch_size);
     svm_multiclass_linear_trainer<linear_kernel<matrix<float,0,1>>, unsigned long> trainer;
     trainer.set_num_threads(std::thread::hardware_concurrency());
+    // The most appropriate C setting could be found automatically by using find_max_global().  See the docs for
+    // find_max_global() for further information and take particular note of model_selection_ex.cpp.
     trainer.set_c(svm_c);
     cout << "Training Multiclass SVM..." << endl;
     const auto df = trainer.train(features, training_labels);
--- a/examples/dnn_self_supervised_learning_ex.cpp
+++ b/examples/dnn_self_supervised_learning_ex.cpp
@@ -292,14 +292,10 @@ try
     // Now, we initialize the feature extractor model with the backbone we have just learned.
     model::feats fnet(layer<5>(net));
     // And we will generate all the features for the training set to train a multiclass SVM
-    // classifier.  It's always a good idea to use double instead of float to improve the
-    // convergence speed and the precision of the optimizer.
+    // classifier.
     std::vector<matrix<double, 0, 1>> features;
     cout << "Extracting features for linear classifier..." << endl;
-    auto temp = fnet(training_images, 4 * batch_size);
-    for (auto&& f : temp)
-        features.push_back(matrix_cast<double>(f));
-    temp.clear();
+    features = fnet(training_images, 4 * batch_size);
     svm_multiclass_linear_trainer<linear_kernel<matrix<double,0,1>>, unsigned long> trainer;
     trainer.set_num_threads(std::thread::hardware_concurrency());
     // The most appropriate C setting could be found automatically by using find_max_global().  See the docs for
@@ -311,7 +307,7 @@ try
 
     // Finally, we can compute the accuracy of the model on the CIFAR-10 train and test images.
     auto compute_accuracy = [&fnet, &df, batch_size](
-        const std::vector<matrix<double, 0, 1>>& samples,
+        const std::vector<matrix<float, 0, 1>>& samples,
         const std::vector<unsigned long>& labels
     )
     {
@@ -334,10 +330,7 @@ try
     cout << "\ntraining accuracy" << endl;
     compute_accuracy(features, training_labels);
     cout << "\ntesting accuracy" << endl;
-    features.clear();
-    temp = fnet(testing_images, 4 * batch_size);
-    for (auto&& f : temp)
-        features.push_back(matrix_cast<double>(f));
+    features = fnet(testing_images, 4 * batch_size);
     compute_accuracy(features, testing_labels);
     return EXIT_SUCCESS;
 }
--- a/examples/dnn_self_supervised_learning_ex.cpp
+++ b/examples/dnn_self_supervised_learning_ex.cpp
@@ -292,10 +292,14 @@ try
     // Now, we initialize the feature extractor model with the backbone we have just learned.
     model::feats fnet(layer<5>(net));
     // And we will generate all the features for the training set to train a multiclass SVM
-    // classifier.
+    // classifier.  It's always a good idea to use double instead of float to improve the
+    // convergence speed and the precision of the optimizer.
     std::vector<matrix<double, 0, 1>> features;
     cout << "Extracting features for linear classifier..." << endl;
-    features = fnet(training_images, 4 * batch_size);
+    auto temp = fnet(training_images, 4 * batch_size);
+    for (auto&& f : temp)
+        features.push_back(matrix_cast<double>(f));
+    temp.clear();
     svm_multiclass_linear_trainer<linear_kernel<matrix<double,0,1>>, unsigned long> trainer;
     trainer.set_num_threads(std::thread::hardware_concurrency());
     // The most appropriate C setting could be found automatically by using find_max_global().  See the docs for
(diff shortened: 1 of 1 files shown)


examples/dnn_self_supervised_learning_ex.cpp
```diff
@@ -288,73 +289,47 @@ try
...
-        serialize("resnet50_cifar_10.dnn") << inet;
-    }
+    // Now, we initialize the feature extractor model with the backbone we have just learned.
+    model::feats fnet(layer<5>(net));
+    // And we will generate all the features for the training set to train a multiclass SVM
+    // classifier.
+    std::vector<matrix<float, 0, 1>> features;
+    cout << "Extracting features for linear classifier..." << endl;
+    features = fnet(training_images, 4 * batch_size);
+    svm_multiclass_linear_trainer<linear_kernel<matrix<float,0,1>>, unsigned long> trainer;
```
davisking, line 299:
```suggestion
    std::vector<matrix<double, 0, 1>> features;
    cout << "Extracting features for linear classifier..." << endl;
    features = fnet(training_images, 4 * batch_size);
    svm_multiclass_linear_trainer<linear_kernel<matrix<double,0,1>>, unsigned long> trainer;
```
You should try running this with `double`.  Like always use `double` unless you specifically need `float`.  Using `float` here is going to cause most of the optimizer to run with `float` which could lead to slower convergence since `float` has much less precision than `double`.  Might not matter here though, IDK, but you should check and the default recommendation should be to use `double` unless memory is so tight `float` is specifically needed.  Like on the CPU, `float` isn't faster than `double` (in 99% of cases, and in many cases the numeric loss of `float` can make it slower).

examples/dnn_self_supervised_learning_ex.cpp
```diff
@@ -288,73 +289,47 @@ try
...
+    trainer.set_num_threads(std::thread::hardware_concurrency());
+    trainer.set_c(svm_c);
```
davisking, line 301:
```suggestion
    // The most appropriate C setting could be found automatically by using find_max_global().  See the docs for
    // find_max_global() for further information and take particular note of model_selection_ex.cpp.
    trainer.set_c(svm_c);
```

davisking: IDK, you could add find_max_global() down below to auto select the right C.  At some point it makes the example too busy though.  Each example doesn't (and shouldn't) do everything possible.  I think leaving a comment letting someone know they can is the right thing in this case though, but I could maybe go either way.

davisking (changes requested):
Nice.  Can it use `double` though instead of `float`?

davisking:
> (that book is already ordered!)

Sweet.  It's a great book.  But as a disclaimer, it doesn't include this SVM solver in it.  But if you read that book you would be able to read the papers that explain how that solver works though :D

examples/dnn_self_supervised_learning_ex.cpp
arrufat: Good idea, I will try to find the optimal C using `find_max_global` and set it to that value by default, and add the comment.

examples/dnn_self_supervised_learning_ex.cpp
arrufat: I was just using `float` because the features extracted by the backbone are `float`, I will try with double.

davisking:
Well, just double check that it's not slower with float. If it's the same it's fine as float.  I just worry someone would get tripped up by it. Like it's likely that for high C values float causes much slower convergence.

davisking:
The other thing that we could, and really probably should do, is make the critical parts of the inner SVM optimizer just always use double.  I looked and this one does basically everything in whatever type the user sets here.  But there is tiny inner QP solver in it that's really impacting the relevant convergence properties here, and that thing could just as well run in double all the time regardless of what the user sets here.

arrufat:
OK, I'll let you know as soon as possible! :)

arrufat:
I've run a simple test:
- extract the features for train
- train the svm
- compute the accuracy for train and test.

The accuracy is almost identical for both `float` and `double`.
However, the runtime for `float` is about 4 minutes, and for `double`, 3 minutes 30 seconds. So I would say there's a huge boost in speed by using double, as around 30 seconds are spent extracting features and computing the accuracy. Again, the only change is `float` -> `double`.

davisking:
> I've run a simple test:
> 
> * extract the features for train
> * train the svm
> * compute the accuracy for train and test.
> 
> The accuracy is almost identical for both `float` and `double`. However, the runtime for `float` is about 4 minutes, and for `double`, 3 minutes 30 seconds. So I would say there's a huge boost in speed by using double, as around 30 seconds are spent extracting features and computing the accuracy. Again, the only change is `float` -> `double`.

Cool, yeah, not surprised.  Can you put the code back to using `float` and I'll just make the core solver Do The Right Thing?  Can you send me the `features` and `training_labels` (like serialize them to a single file) and I'll use them to test?  I'm assuming it takes a while for this example program to otherwise generate them.  I'm pretty sure I know what changes to make to make it not matter if float or double is given, but it will save me some time if I can start with a dataset that's known to exercise the difference when testing it.

arrufat:
Done.
[These](https://1drv.ms/u/s!Avd3wR2lFRWkgepACCy0XDID5b-QpA?e=3iyuSR) are the training features and labels, serialized as:

```c++
serialize("training_features_and_labels.dat") << features << training_labels;
```

arrufat:
Ah, and about adding `find_max_global` I think we can add it, the code is really short:

```c++
auto cross_validation_score = [&](const double c)
{
    svm_multiclass_linear_trainer<linear_kernel<matrix<float, 0, 1>>, unsigned long> trainer;
    trainer.set_num_threads(std::thread::hardware_concurrency());
    trainer.set_c(c);
    const auto cm = cross_validate_multiclass_trainer(trainer, features, training_labels, 10);
    const double accuracy = sum(diag(cm)) / sum(cm);
    cout << "C: " << c << endl;
    cout << "cross validation accuracy: " << accuracy << endl;;
    cout << "confusion matrix:\n " << cm << endl;
    return accuracy;
};
const auto result = find_max_global(cross_validation_score, 1e-5, 20000, max_function_calls(50));
cout << "Best C: " << result.x(0) << endl;
```

But it takes a while to find the optimal C…

VisionEp1:
i would add that part, since it seems good practice to use optimizers with svm

davisking:
Cool, yeah, adding that is fine too :D 

10 folds is probably overkill though.  I would go with 3.  It will run a lot faster.

davisking:
> i would add that part, since it seems good practice to use optimizers with svm

It's good for everything else too.  It's just that stuff like DNN solvers are so slow it's prohibitive most of the time :(

pfeatherstone:
@davisking Are there any other old-school-machine-learning / optimisation books you recommend ? Sorry this is unrelated to the PR but there isn't a "Discussion" section on the dlib repo and I think other people would be interested to see your suggested reading list. Keeping up to date with SOTA neural networks is great but this PR goes to show that there is some traditional ML stuff that's worth knowing about. And it looks like it's a dying topic because of NNs...

arrufat:
@pfeatherstone I'm sure you already checked it out, but just in case: http://dlib.net/books.html :)

pfeatherstone:
No I hadn't. Thank you !

pfeatherstone:
I'm gonna add some of these to my Christmas list :) As always with dlib the answer is : "Read the docs"

davisking:
Yeah that books list was going to be my answer.  The other thing here is that this stuff isn't dying.  It's just that computer science has become more insular again.  A bunch of this stuff is from statistics, operations research, and math departments.  Which before deep learning was an important part of popular ML methods.  But deep learning is so simple and at some level easy to get working a lot of people (in computer science grad programs) don't care anymore.   

Although the journal of machine learning research is still publishing a ton of non-deep-learning papers. 
 Check it out: https://www.jmlr.org/.

Anyway, people have this attitude that deep learning methods are "black box" methods.  Which is nonsense.  There is no such thing as a black box.  There is only the state of ignorance or understanding in the mind of the human user.  If they understand why something works then it's not a black box, and if they are ignorant then people often say "it's a black box".  But that's a statement about a human mind, not about the method itself.   Deep learning tooling has gotten itself into this situation where there is this sense that people can use it without having any idea why or how it works.  Which is good and bad.  Good because it often turns out to be true.  But bad because sometimes it doesn't, and then people haven't the foggiest idea why it's not working or how to make it work sensibly.

Anyway, this is all to say that it's good to read a good statistics textbook and a numerical optimization book since that will give you way more perspective about what's going on and what else you might do instead.  Or to say that another way, computer science departments are not the only community that tries to infer things from data.  Operations research and statistics departments have been doing that for even longer and in many ways do a much better job at it.

davisking (approved):
Anyway, this is cool :)

I'll modify the underlying solver code to be more sensible when `float` is used.

arrufat:
Again, thank you very much for your time and your invaluable advice.

davisking:
No problem.  Thanks for the cool PR :)

davisking:
> Done. [These](https://1drv.ms/u/s!Avd3wR2lFRWkgepACCy0XDID5b-QpA?e=3iyuSR) are the training features and labels, serialized as:
> 
> ```c++
> serialize("training_features_and_labels.dat") << features << training_labels;
> ```

You sure this is the right setup?  I just ran this a few times with float vs double and float was faster :|
```
 ~/s/d/e/svm_sparse_ex.cpp  ~/s/d/e/dnn_self_supervised_learning_ex.cpp                                                                                                           X
#include <iostream>
#include <vector>
#include <dlib/svm_threaded.h>
#include <dlib/time_this.h>

using namespace std;
using namespace dlib;

int main()
{
    using T = float;
    // using T = double;
    std::vector<matrix<T, 0, 1>> features;
    std::vector<unsigned long> training_labels, testing_labels;
    deserialize("training_features_and_labels.dat") >> features >> training_labels;
    cout << "features.size(): "<< features.size() << endl;
    cout << "dims: "<< features[0].size() << endl;

    auto train = [&](const double c)  
    {   
        svm_multiclass_linear_trainer<linear_kernel<matrix<T, 0, 1>>, unsigned long> trainer;
        trainer.set_num_threads(std::thread::hardware_concurrency());
        trainer.set_c(c);
        trainer.be_verbose();
        trainer.train(features, training_labels);
    };  

    TIME_THIS(
    train(10000);
    );  

}

```

arrufat:
Oh, I can't test it soon (maybe on Monday)... I'll run this and let you know...

arrufat:
Hmm, you're right, it's faster with `float` after testing again… Maybe I got mixed up last time and gave you the results the other way around…  This is what I tried:
```
float  10000 time: 2.67742min
double 10000 time: 3.2079min 
float   5000 time: 1.92706min
double  5000 time: 2.41691min
```

Sorry about that! :sweat_smile:

davisking:
Yeah no worries. There are problems for which the solver has the trouble I mentioned but guess this isn't one of them.

PR #2453, opened 2021-11-02
Avoid redundant computations in the YOLO loss

arrufat:
I noticed there were some computations that didn't depend on the output tag layer and the input tensor area was computed over and over inside of a loop.


davisking:
Looks good :)

PR #2454, opened 2021-11-02
Fix nvidia pragma warning

arrufat:
After updating to CUDA 11.5, I got this warning:
```
dlib/algs.h(25): warning #20236-D: pragma "diag_suppress" is deprecated, use "nv_diag_suppress" instead
```

According to this [blog post](https://developer.nvidia.com/blog/reducing-application-build-times-using-cuda-c-compilation-aids/), diagnostic pragmas without the `nv_` prefix have been deprecated, so I applied the proposed fix.


davisking (approved):
Nice

PR #2457, opened 2021-11-12
expose momentum_filter to Python

addam:
Prior to this patch, Kalman filtering has been only available in Python using `rect_filter`, which makes filtering a floating-point scalar rather clumsy.

I pretty much only copied parts of existing code. It was my first contact with pybind11 so I hope the result is appropriate.

### DIFF
--- a/tools/python/src/other.cpp
+++ b/tools/python/src/other.cpp
@@ -8,6 +8,7 @@
 #include <dlib/sparse_vector.h>
 #include <dlib/optimization.h>
 #include <dlib/statistics/running_gradient.h>
+#include <dlib/filtering.h>
 
 using namespace dlib;
 using namespace std;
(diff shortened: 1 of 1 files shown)


tools/python/src/other.cpp
```diff
@@ -264,5 +278,69 @@ ensures \n\
...
+                This allows the moving object to undergo large unmodeled accelerations, far
+                in excess of what would be suggested by get_typical_acceleration(), without
+                then experiencing a long lag time where the Kalman filter has to "catch
+                up" to the new position.  )asdf"
+        )
+        .def(py::init<double,double,double>(), py::arg("measurement_noise"), py::arg("typical_acceleration"), py::arg("max_measurement_deviation"))
+        .def("measurement_noise",   [](const momentum_filter& a){return a.get_measurement_noise();})
+        .def("typical_acceleration",   [](const momentum_filter& a){return a.get_typical_acceleration();})
+        .def("max_measurement_deviation",   [](const momentum_filter& a){return a.get_max_measurement_deviation();})
+        .def("__call__", [](momentum_filter& f, const double r){return f(r); })
```
davisking, line 316: This kind of thing can just be `.def("foo", &type::foo)`.  I'll adjust that in a separate commit though.

davisking (approved):
Yeah cool, looks good :)

PR #2458, opened 2021-11-13
[INVOKE] added is_invocable_r, invoke_r, made everything constexpr and conditionally noexcept

pfeatherstone:
See title.
This has required a refactor of `dlib::invoke` as `std::mem_fn` is not constexpr in C++11.

### DIFF
--- a/dlib/invoke.h
+++ b/dlib/invoke.h
@@ -11,32 +11,124 @@ namespace dlib
 {
     // ----------------------------------------------------------------------------------------
     namespace detail {
-        template< typename F, typename ... Args >
-        auto INVOKE(F&& fn, Args&& ... args)
+        template< typename T >
+        struct is_reference_wrapper : std::false_type {};
+        template< typename U >
+        struct is_reference_wrapper<std::reference_wrapper<U>> : std::true_type {};
+
+        template <
+            typename Base,
+            typename T,
+            typename Derived,
+            typename... Args
+        >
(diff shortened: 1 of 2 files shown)


pfeatherstone:
@davisking, is good?

dlib/invoke.h
```diff
@@ -11,32 +11,153 @@ namespace dlib
...
 
     template< typename F, typename... Args>
-    auto invoke(F && f, Args &&... args)
+    constexpr auto invoke(F && f, Args &&... args)
     /*!
         ensures
             - identical to std::invoke(std::forward<F>(f), std::forward<Args>(args)...)
             - works with C++11 onwards
     !*/
+    noexcept(noexcept(detail::INVOKE(std::forward<F>( f ), std::forward<Args>( args )...)))
```
davisking, line 160:
```suggestion
    noexcept(detail::INVOKE(std::forward<F>( f ), std::forward<Args>( args )...))
```

Do there need to be two `noexcept` keywords here?

pfeatherstone: I thought you needed 2: 1 which is the noexcept specifier and the other which is the noexcept operator. The operator is used to determine at compile-time whether the expression is noexcept or not, and the specifier marks the function conditionally noexcept based on a true or false statement.

davisking: Oh right, duh  :)

davisking:
Sweet, looks good.

PR #2460, opened 2021-11-16
Normalize samples for SVM classifier

arrufat:
I am not sure this is needed, since the accuracy doesn't change, but I think it's good practice to normalize the samples for SVM classifiers. What I noticed, though, is that the best C is no longer around 3000, but around 30 (I modified the search range afterwards).

If you think it's unnecessary, reject this :)

EDIT: I am reusing the `vector_normalizer` learned from the training samples on the testing set, I don't think it's fair to learn a new normalizer.

### DIFF
--- a/examples/dnn_self_supervised_learning_ex.cpp
+++ b/examples/dnn_self_supervised_learning_ex.cpp
@@ -295,6 +295,10 @@ try
     std::vector<matrix<float, 0, 1>> features;
     cout << "Extracting features for linear classifier..." << endl;
     features = fnet(training_images, 4 * batch_size);
+    vector_normalizer<matrix<float, 0, 1>> normalizer;
+    normalizer.train(features);
+    for (auto& feature : features)
+        feature = normalizer(feature);
 
     // Find the most appropriate C setting using find_max_global.
     auto cross_validation_score = [&](const double c)
@@ -309,7 +313,7 @@ try
         cout << "confusion matrix:\n " << cm << endl;
         return accuracy;
     };
-    const auto result = find_max_global(cross_validation_score, 1e-4, 10000, max_function_calls(50));
+    const auto result = find_max_global(cross_validation_score, 1e-3, 1000, max_function_calls(50));
     cout << "Best C: " << result.x(0) << endl;
 
     // Proceed to train the SVM classifier with the best C.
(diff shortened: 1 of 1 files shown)


davisking:
> I am not sure this is needed, since the accuracy doesn't change, but I think it's good practice to normalize the samples for SVM classifiers.

A linear SVM classifier is the same thing as any other linear layer you might stick on the end of the network.  So it needs or doesn't need normalization in just the same way.  Sometimes you don't want normalization.  Like if you already constructed your features so that things you thought mattered less had generally smaller values you wouldn't want to do it.  Here the network is making this vector that's already in a space that seems pretty appropriate.

>  What I noticed, though, is that the best C is no longer around 3000, but around 30 (I modified the search range afterwards).

Yeah, seems like it wouldn't matter here.  I wager all that's happening is the magnitude of the numbers in the vectors coming out of the network is smallish and they just get multiplied by some number > 1 by the normalizer.  That's going to change C.  Right, like if you look at the SVM objective function you can see that if you go multiply all the training samples by 10 and divide C by 10 you are left with the same objective function since the 10s cancel out.

> 
> If you think it's unnecessary, reject this :)

Probably not needed here :)

> 
> EDIT: I am reusing the `vector_normalizer` learned from the training samples on the testing set, I don't think it's fair to learn a new normalizer.

Yeah that's definitely the right way.

arrufat:
Understood,  thank you for the explanation :)

davisking:
Well, the other reason to do this would be if it speeds up SVM solver convergence by making it just numerically less susceptible to floating point rounding error.  Did you notice any speed change?

arrufat:
So, I ran some tests, and this is what I got:
- vanilla
  - C = 3000: 1.62 min
  - C = 30: 13.22 sec
- normalized
  - C = 3000: 3.93 min
  - C = 30: 20.53 sec

So it seems faster with the vanilla, unnormalized version. However, the normalized version looks for optimal values that are below 100 more often, and those tend to converge faster than higher values (in the order of 1000). So, in this particular case, normalizing makes the global minimization faster.

davisking:
Cool, yeah, that happens a lot (it being faster when scaled near 1).

At some point I (or whoever) should just expand out dlib/svm/auto.h to contain a function to automatically do this stuff for linear models.  Like 99% of people want the same thing, which is just normalize, run cross validation to find the best setting, then train the final linear model.

arrufat:
I am not familiar with that part of the library. It might be a good chance to start when I have time :)

PR #2465, opened 2021-11-21
Add dlib::rdrand kernel to support digital random numbers generation with RDRND instructions

joselucasoliveira:
This adds the dlib::rdrand kernel extension to support the generation of digital random numbers with RDRND instructions. A digital random number generator uses a processor resident entropy source to repeatedly seed a hardware-implemented cascade construction RNG. It includes a high-quality entropy source implementation, being very useful for scientific applications.


dlib/rand/rand_kernel_1.h
```diff
@@ -112,25 +112,25 @@ namespace dlib
                 next_gaussian = 0;
             }
 
-            unsigned char get_random_8bit_number (
+            virtual unsigned char get_random_8bit_number (
```
davisking, line 115: This class isn't meant to be inherited from.  The destructor isn't virtual for instance and you will generate object slicing problems by doing this as well.

There is also contract violation now since functions that take a `dlib::rand` expect that object.  For instance, `.clear()` and `.set_seed()` would have the wrong behavior.

The right way to do this would be to just have a separate class.

davisking (commented):
Thanks for the PR.  The CMake setup is nice.  Inheritance isn't the right way to do this though.

More generally though, why not use `std::random_device`?  `dlib::rand` is a pre-c++11 thing.  Now that there is official support in the language for this stuff I think most people should use the standard library tooling for it.

dlib/rand/rand_kernel_1.h
joselucasoliveira: Thanks for the review. I see, I'm going to separate this feature into a new class. I was trying to find a way to reuse the helper functions of the dlib::rand class like get_random_double, any recommendations?

joselucasoliveira:
I believe that `std::random_device` doesn't need to be non-deterministic and some implementations might be an improved version of a pseudo RNG seeding over `time(NULL)`. For this reason the proposal for an implementation that queries information directly from the CPU.

joselucasoliveira:
@davisking I made some changes, thanks for your review. If you think it's not interesting to insert these features in Dlib, no problem, just let me know.

davisking:
Some googling makes it appear that common implementations of the standard library do use rdrand, which is the idea behind `std::random_device`.  I.e. that when there is actually hardware randomness to be had it gives it to you.  So I think dlib users would be better off using the C++ standard tooling for this.  I try to avoid replicating stuff in the standard library, since in general dlib can't do as well as what's in the standard library.  For example, code that uses `std::random_device` will be generally more portable.  Since say, you move to an ARM machine that has some other way of getting hardware randomness and you automatically benefit from the standard implementation on that platform which should Do The Right Thing.  I do realize that sometimes people bork up code though.  But usually the standard library implementations are pretty solid.

Anyway, I'm not saying there is anything wrong with your code here.  It looks good.  I just don't want to duplicate `std::random_device` in dlib.  Especially considering this `dlib::rdrand` is inherently non-portable, and one of the goals of dlib is to be super portable.  So it would have to have a fall back to a non-rdrand back end.  But then that's again exactly what `std::random_device` is supposed to be.

joselucasoliveira:
I see, it makes sense to me, thanks again. I will reserve this for my projects. If I develop a useful feature for the Dlib proposal, I will send you a PR. I appreciate your feedback.

davisking:
Sweet :)

PR #2466, opened 2021-11-22
Use lambda obj in the ignore case and do some refactoring

arrufat:
This PR fixes a small bug where `yolo_options::lambda_obj` was not being used in the `ignore_iou` case.
It's not a big deal, since the example uses the default value of 1, but still.
I also performed some refactoring to be more clear and reuse some computations.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -3641,7 +3641,7 @@ namespace dlib
                                                              h / (1 - h) * anchors[a].height));
                                 for (long k = 0; k < num_classes; ++k)
                                 {
-                                    const float conf = out_data[tensor_index(output_tensor, n, a * num_feats + 5 + k, r, c)] * obj;
+                                    const float conf = obj * out_data[tensor_index(output_tensor, n, a * num_feats + 5 + k, r, c)];
                                     if (conf > adjust_threshold)
                                         det.labels.emplace_back(conf, options.labels[k]);
                                 }
@@ -3716,7 +3716,7 @@ namespace dlib
 
                             // Incur loss for the boxes that are below a certain IoU threshold with any truth box
                             if (best_iou < options.iou_ignore_threshold)
-                                g[o_idx] = out_data[o_idx];
+                                g[o_idx] = options.lambda_obj * out_data[o_idx];
                         }
                     }
                 }
@@ -3780,13 +3780,13 @@ namespace dlib
                             const double th = truth_box.rect.height() / (anchors[a].height + truth_box.rect.height());
 
                             // Scale regression error according to the truth size
-                            const double scale_box = 2 - truth_box.rect.area() / input_area;
+                            const double scale_box = options.lambda_box * (2 - truth_box.rect.area() / input_area);
 
                             // Compute the gradient for the box coordinates
-                            g[x_idx] = options.lambda_box * scale_box * (out_data[x_idx] - tx);
-                            g[y_idx] = options.lambda_box * scale_box * (out_data[y_idx] - ty);
-                            g[w_idx] = options.lambda_box * scale_box * (out_data[w_idx] - tw);
-                            g[h_idx] = options.lambda_box * scale_box * (out_data[h_idx] - th);
+                            g[x_idx] = scale_box * (out_data[x_idx] - tx);
+                            g[y_idx] = scale_box * (out_data[y_idx] - ty);
+                            g[w_idx] = scale_box * (out_data[w_idx] - tw);
+                            g[h_idx] = scale_box * (out_data[h_idx] - th);
 
                             // Compute the classification error
                             for (long k = 0; k < num_classes; ++k)


dlib/dnn/loss.h
```diff
@@ -3641,7 +3641,7 @@ namespace dlib
                                                              h / (1 - h) * anchors[a].height));
                                 for (long k = 0; k < num_classes; ++k)
                                 {
-                                    const float conf = out_data[tensor_index(output_tensor, n, a * num_feats + 5 + k, r, c)] * obj;
+                                    const float conf = obj * out_data[tensor_index(output_tensor, n, a * num_feats + 5 + k, r, c)];
```
arrufat, line 3644: Putting `obj` at the beginning makes it easier to read, otherwise it might be overlooked, since the line is relatively long.

dlib/dnn/loss.h
```diff
@@ -3716,7 +3716,7 @@ namespace dlib
 
                             // Incur loss for the boxes that are below a certain IoU threshold with any truth box
                             if (best_iou < options.iou_ignore_threshold)
-                                g[o_idx] = out_data[o_idx];
+                                g[o_idx] = options.lambda_obj * out_data[o_idx];
```
arrufat, line 3719: This is the actual bug fix.

dlib/dnn/loss.h
```diff
@@ -3780,13 +3780,13 @@ namespace dlib
...
 
                             // Compute the gradient for the box coordinates
-                            g[x_idx] = options.lambda_box * scale_box * (out_data[x_idx] - tx);
-                            g[y_idx] = options.lambda_box * scale_box * (out_data[y_idx] - ty);
-                            g[w_idx] = options.lambda_box * scale_box * (out_data[w_idx] - tw);
-                            g[h_idx] = options.lambda_box * scale_box * (out_data[h_idx] - th);
+                            g[x_idx] = scale_box * (out_data[x_idx] - tx);
+                            g[y_idx] = scale_box * (out_data[y_idx] - ty);
+                            g[w_idx] = scale_box * (out_data[w_idx] - tw);
+                            g[h_idx] = scale_box * (out_data[h_idx] - th);
```
arrufat, line 3789: Just computing the scaling factor once... This might have no impact and be optimized away by the compiler though...

arrufat:
Now I'm hesitating whether I should add `lambda_obj` in the `ignore` case…
On one hand, it allows weighting the objectness differently from the regression and the classification.
But on the other hand, having this asymmetry allows for changing the weights between the positive and the negative class in the objectness outputs. And the weights between objectness and regression and classification can still be adjusted with the other lambdas… So, I am not sure if this is needed… I don't remember what I had in mind at that time, but the current code seems more versatile.

davisking:
> Now I'm hesitating whether I should add `lambda_obj` in the `ignore` case… On one hand, it allows weighting the objectness differently from the regression and the classification. But on the other hand, having this asymmetry allows for changing the weights between the positive and the negative class in the objectness outputs. And the weights between objectness and regression and classification can still be adjusted with the other lambdas… So, I am not sure if this is needed… I don't remember what I had in mind at that time, but the current code seems more versatile.

Yeah what's in this PR seems like the right thing.  You have some reservations though?  I was just about to merge it :D

arrufat:
The only problem with this PR is if the use wants to use different weights for the binary loss in the objectness for the positive and negative cases. Currently, `lambda_obj` only affects the positive case. So, we can play around if we know that we have really big images and few objects on them (we can balance it out by setting `lambda_obj` to something larger than 1). And if we want to give more importance to the objectness, we have to reduce `lambda_box` and `lambda_cls`.

What I'm proposing in this PR won't allow for that use case anymore. It treats `lambda_obj` the same for the positive and negative case. That's why I'm not sure about it. I've seen some other repos that have `lambda_obj` and `lambda_pos` to handle this case, but I think it can already be done as it is now, without applying this PR... I don't know...

arrufat:
If we decide not to merge, I should update the docs about what `lambda_obj` means, then.

davisking:
> The only problem with this PR is if the use wants to use different weights for the binary loss in the objectness for the positive and negative cases. Currently, `lambda_obj` only affects the positive case. So, we can play around if we know that we have really big images and few objects on them (we can balance it out by setting `lambda_obj` to something larger than 1). And if we want to give more importance to the objectness, we have to reduce `lambda_box` and `lambda_cls`.
> 
> What I'm proposing in this PR won't allow for that use case anymore. It treats `lambda_obj` the same for the positive and negative case. That's why I'm not sure about it. I've seen some other repos that have `lambda_obj` and `lambda_pos` to handle this case, but I think it can already be done as it is now, without applying this PR... I don't know...


Sure, sometimes having that lever is useful.  Not sure what the maximal usability goodness is here though.  You have been using this the most so I'll defer to your opinion about what's nicest :)

arrufat:
Honestly, I've never touched that parameter in my life, haha, I've always left all `lambda` set to 1. YOLOv3 does not even have them. I added them to have a generic loss and not to deal with different versions of the `yolo_options` in the serialization, if we decide to add them at some point.

With these large PR, I sometimes go back and read them again with a new set of eyes because I want to make sure there are no bugs before there's a new stable release of dlib.

Let me think a bit…

arrufat:
I think we can merge it. It is consistent with the canonical YOLOv4 implementation, which also doesn't distinguish between the positive and the negative case for objectness.
The main goal of the objectness is similar to the focal loss: compensate the huge imbalance of cells with objects vs cells without objects, and YOLOv4 does it like this (same weight for positive and negative case), and it's quite battle tested, so I would say it's already good.
Sorry for the confusion.

davisking:
Na, all good details to be concerned about :)

pfeatherstone:
I love the term **maximal usability goodness**

PR #2471, opened 2021-12-07
Fix a crash when the truth center is outside the image

arrufat:
This bug has been bothering me for quite some time… It took me a really long time to figure out why the training was randomly crashing (when I used strong data augmentation).

I don't know if we should warn the user that we're ignoring these truth boxes… but we shouldn't try to back propagate on them, for sure.


davisking:
Ha, yeah, that would do it :)

I wouldn't warn about it.  I mean, other times I've done this I didn't warn about it since I found it spammy.

Anyway, nice find and thanks for another PR :)

PR #2475, opened 2021-12-17
[TYPE_SAFE_UNION] simplified some type traits and added for_each().

pfeatherstone:
This PR simplifies some type traits and mainly adds `type_safe_union::for_each()`.
This simplifies deserialization and is generally useful for iterating over the alternative types.

### DIFF
--- a/dlib/type_safe_union/type_safe_union_kernel.h
+++ b/dlib/type_safe_union/type_safe_union_kernel.h
@@ -33,52 +33,54 @@ namespace dlib
 
     template <typename... Types> class type_safe_union;
 
-    template<typename Variant>
-    struct variant_size;
-
-    template<typename Variant>
-    struct variant_size<const Variant> : variant_size<Variant> {};
-
-    template<typename Variant>
-    struct variant_size<volatile Variant> : variant_size<Variant> {};
-
-    template<typename Variant>
-    struct variant_size<const volatile Variant> : variant_size<Variant> {};
+    template<typename Tsu>
+    struct type_safe_union_size;
 
     template<typename... Types>
-    struct variant_size<type_safe_union<Types...>>
-    : std::integral_constant<size_t, sizeof...(Types)> {};
+    struct type_safe_union_size<type_safe_union<Types...>> : std::integral_constant<size_t, sizeof...(Types)> {};
+
+    template<typename Tsu> struct type_safe_union_size<const Tsu>           : type_safe_union_size<Tsu> {};
+    template<typename Tsu> struct type_safe_union_size<volatile Tsu>        : type_safe_union_size<Tsu> {};
+    template<typename Tsu> struct type_safe_union_size<const volatile Tsu>  : type_safe_union_size<Tsu> {};
 
     // ---------------------------------------------------------------------
 
+    namespace detail
+    {
+        template<size_t I, typename... Ts>
+        struct nth_type;
+
+        template<size_t I, typename T0, typename... Ts>
+        struct nth_type<I, T0, Ts...> : nth_type<I-1, Ts...> {};
+
+        template<typename T0, typename... Ts>
+        struct nth_type<0, T0, Ts...> { using type = T0; };
+    }
+
     template <size_t I, typename Variant>
-    struct variant_alternative;
+    struct type_safe_union_alternative;
 
-    template <size_t I, typename T0, typename... Ts>
-    struct variant_alternative<I, type_safe_union<T0, Ts...>> : variant_alternative<I-1, type_safe_union<Ts...>> {};
-
-    template <typename T0, typename... Ts>
-    struct variant_alternative<0, type_safe_union<T0, Ts...>>
-    { using type = T0;};
+    template <size_t I, typename... Types>
+    struct type_safe_union_alternative<I, type_safe_union<Types...>> : detail::nth_type<I, Types...>{};
 
     template<size_t I, typename Variant>
-    using variant_alternative_t = typename variant_alternative<I, Variant>::type;
+    using type_safe_union_alternative_t = typename type_safe_union_alternative<I, Variant>::type;
 
     template <size_t I, typename Variant>
-    struct variant_alternative<I, const Variant>
-    { using type = typename std::add_const<variant_alternative_t<I, Variant>>::type; };
+    struct type_safe_union_alternative<I, const Variant>
+    { using type = typename std::add_const<type_safe_union_alternative_t<I, Variant>>::type; };
 
     template <size_t I, typename Variant>
-    struct variant_alternative<I, volatile Variant>
-    { using type = typename std::add_volatile<variant_alternative_t<I, Variant>>::type; };
+    struct type_safe_union_alternative<I, volatile Variant>
+    { using type = typename std::add_volatile<type_safe_union_alternative_t<I, Variant>>::type; };
 
     template <size_t I, typename Variant>
-    struct variant_alternative<I, const volatile Variant>
-    { using type = typename std::add_cv<variant_alternative_t<I, Variant>>::type; };
+    struct type_safe_union_alternative<I, const volatile Variant>
+    { using type = typename std::add_cv<type_safe_union_alternative_t<I, Variant>>::type; };
 
     // ---------------------------------------------------------------------
 
-    namespace internal
+    namespace detail
     {
         // ---------------------------------------------------------------------
 
@@ -91,18 +93,18 @@ namespace dlib
         // ---------------------------------------------------------------------
 
         template <int nTs, typename T, typename... Ts>
-        struct variant_type_id_impl
+        struct type_safe_union_type_id_impl
                 : std::integral_constant<int, -1 - nTs> {};
 
         template <int nTs, typename T, typename T0, typename... Ts>
-        struct variant_type_id_impl<nTs, T, T0, Ts...>
-                : std::integral_constant<int, std::is_same<T,T0>::value ? 1 : variant_type_id_impl<nTs, T,Ts...>::value + 1> {};
+        struct type_safe_union_type_id_impl<nTs, T, T0, Ts...>
+                : std::integral_constant<int, std::is_same<T,T0>::value ? 1 : type_safe_union_type_id_impl<nTs, T,Ts...>::value + 1> {};
 
         template <typename T, typename... Ts>
-        struct variant_type_id : variant_type_id_impl<sizeof...(Ts),T,Ts...>{};
+        struct type_safe_union_type_id : type_safe_union_type_id_impl<sizeof...(Ts),T,Ts...>{};
 
         template <typename T, typename... Ts>
-        struct variant_type_id<in_place_tag<T>, Ts...> : variant_type_id<T,Ts...>{};
+        struct type_safe_union_type_id<in_place_tag<T>, Ts...> : type_safe_union_type_id<T,Ts...>{};
 
         // ---------------------------------------------------------------------
     }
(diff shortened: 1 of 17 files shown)


pfeatherstone:
@davisking, Is good?

dlib/type_safe_union/type_safe_union_kernel_abstract.h
```diff
@@ -192,11 +193,13 @@ namespace dlib
         );
         /*!
            ensures
-              - if (T is the same type as one of the template arguments) then
-                 - returns a number indicating which template argument it is. In particular,
-                   if it's the first template argument it returns 1, if the second then 2, and so on.
-               - else
-                  - returns -1
+                - if (T is the same type as one of the template arguments) then
+                    returns a number indicating which template argument it is. In particular,
```
davisking, line 197: Keep the leading `-` for each bullet since that's the style in the rest of the library.

dlib/type_safe_union/type_safe_union_kernel_abstract.h
```diff
@@ -280,6 +283,33 @@ namespace dlib
...
+            F&& f
+        );
+        /*!
+            requires
+                - f is a callable object such that the following expression is valid for
+                  all types U in Types...:
+
+                    std::forward<F>(f)(const in_place_tag<U>& tag, type_safe_union<Types...>& x)
+            ensures
+                - the state of x is unchanged
```
davisking, line 297: What does it do though? 

I don't totally get what this is for.  It's calling f() for each type in `Types...` right?  But why is that needed?  Like what does someone do with this?

pfeatherstone: So deserialize() uses it

pfeatherstone: The last test in the unit test shows a nice use case

pfeatherstone: Remember initially you had reservations about supporting (de)serialization for ```std::variant``` because you argued users could do some metaprogramming like re-ordering the alternative types? In which case, serialization wouldn't work because the type indices would be all wrong? Well users could write (de)serialization code that uses ```for_each()``` and type identifiers such that serialization was invariant under re-ordering. This is just one example. In general, someone might want to iterate over the types. Like they might want a string description of all the types. Or they might want a hash code that hashes all the underlying types. There are loads of use cases.

pfeatherstone:
Let me know if you want me to close this. I think it's useful but you are the dlib dictator for life so I'll do whatever :)

dlib/type_safe_union/type_safe_union_kernel_abstract.h
davisking: Eh, IDK.  So in the example you have in the unit test it's doing something that's not going to work.  Like check out https://en.cppreference.com/w/cpp/types/type_info/hash_code, those typeid hash codes are not stable even over reruns of the same binary.  There isn't any way in C++ to generate a stable ID like this for a type :(

If this is just about serialization stability over changes to the tuple it would be better to be able to say something like this: `type_safe_union<int, 1, foo_type, 2, other_type, 4>`.  That is, the user could set the id number for each type if they wanted.  That way adding or removing one isn't a huge deal since they can ensure the same ids always attach to the same types.

pfeatherstone: Serialization was just one example. I just found `for_each()` useful. I get the feeling you don't. But `deserialize()` was essentially doing the same thing as `for_each()` so i thought i would write `for_each()` then `deserialize()` could use that. Then I found it useful for other stuff. If you don't want it then I can close the PR.

pfeatherstone: With regards to generating hash codes for types, you can register them at runtime. So you're not relying on the type system to do this but at least it's stable since you're defining the codes manually.

pfeatherstone: But the point of `for_each()` isn't just (de)serialization. It literally just does what it says on the tin and users, like me, find this useful.

davisking: Sure, that's fine.  Can you make it be a global function though?  Like we shouldn't be adding methods to classes when they don't need access to the private details.  And call it `for_each_type()` to emphasize that it's looping over the types in a parameter pack.  Otherwise it's a little confusing since a type safe union only ever has one item in it.

pfeatherstone: Yeah sure. I thought that was probably best to be honest. I feel like visit() should be global too but that's done now.

pfeatherstone:
@davisking Do you mind having another look? Cheers

pfeatherstone:
I think with hindsight, i should have made `visit()` a global function and make it have same signature as `std::visit()`. I have a working version but don't know if you want it. The member function `type_safe_union::visit()` works fine but you're right, if it doesn't change the state of the object, then a global maybe makes more sense. If we were to add `dlib::visit()`, then I think we should remove `type_safe_union::visit()`, which would be a breaking change. But that's not too bad. It's a relatively new feature and hasn't even made it into a release yet.

davisking:
> I think with hindsight, i should have made `visit()` a global function and make it have same signature as `std::visit()`. I have a working version but don't know if you want it. The member function `type_safe_union::visit()` works fine but you're right, if it doesn't need to access private member variables/functions, then a global is fine. Also it means i don't have 2 copies of basically the same code: 1 for non-const references and 1 for const references.

Ha, that reminds me it's been kinda a while since a release.  I should probably make one.  But yeah, I think changing it is still fair game. So if you want to move visit to be a global function that's cool with me too.

pfeatherstone:
> Ha, that reminds me it's been kinda a while since a release. I should probably make one.

Could you do the release before the FFMPEG wrapper merge? It would be good to leave a little bit of wiggle-room for API changes before it makes its way into a proper release.

davisking:
> > Ha, that reminds me it's been kinda a while since a release. I should probably make one.
> 
> 
> 
> Could you do the release before the FFMPEG wrapper merge? It would be good to leave a little bit of wiggle-room for API changes before it makes its way into a proper release.

Yeah will do. Let me know when this PR is ready. I'll make a release right after it's in.

pfeatherstone:
@davisking PR ready for review. Cheers

pfeatherstone:
@davisking Hold the fort, i'm experimenting with a version of `apply_to_contents()` and `visit()` which don't use template recursion at all.

pfeatherstone:
Real shame. This was a known bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47226

pfeatherstone:
I'll revert back to template recursion in a bit.

pfeatherstone:
@davisking Right, I promise, no more commits. Ready for review.

dlib/type_safe_union/type_safe_union_kernel_abstract.h
```diff
@@ -345,6 +372,57 @@ namespace dlib
...
 
+// ----------------------------------------------------------------------------------------
+
+    template<
+        typename Variant,
+        typename F
+    >
+    void for_each_type(
+        Variant&& tsu,
+        F&& f
```
davisking, line 383:
```suggestion
        F&& f,
        Variant&& tsu
```
How about switching this around to match the order of `visit()`?

dlib/type_safe_union/type_safe_union_kernel_abstract.h
```diff
@@ -27,17 +27,49 @@ namespace dlib
...
+            - Identical behaviour to std::variant_alternative except it works with dlib::type_safe_union,
+              not std::variant. See cppreference on const, volative or cv qualification behaviour.
+    !*/
+
+    template<size_t I, typename Variant>
+    using variant_alternative_t;
+    /*!
+        ensures
+            - provides template alias for variant_alternative
     !*/
```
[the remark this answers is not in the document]
davisking: Also add some unit tests that `type_safe_union_alternative_t` turns into the expected type.

dlib/type_safe_union/type_safe_union_kernel.h
```diff
@@ -570,6 +514,116 @@ namespace dlib
...
+            std::size_t... I
+        >
+        void for_each_type_impl(
+            Variant&& tsu,
+            F&& f,
+            dlib::index_sequence<I...>
+        )
+        {
+            using Tsu = typename std::decay<Variant>::type;
+            return (void)std::initializer_list<int>{
```
davisking, line 531:
```suggestion
            int unused[] = {
```
I think this is a simpler way to write this kind of thing, since it's less text and avoids that `return` there that looks like it should be returning something but is actually in a `void` function.  Also avoids the `(void)` cast which is squirrely in general.

davisking (changes requested):
Cool.  Code looks good.  I've only got some minor comments, mostly about naming and other dlib style stuff.

dlib/type_safe_union/type_safe_union_kernel_abstract.h
```diff
@@ -27,17 +27,49 @@ namespace dlib
...
+        
+        It is also used with type_safe_union::for_each() to disambiguate types.
+    !*/
+
+// ----------------------------------------------------------------------------------------
 
-            A(int i_, int j_) : i(i_), j(j_) {}
-        };
+    template<typename Variant>
+    struct variant_size;
```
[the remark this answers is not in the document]
pfeatherstone: Okidok.

dlib/type_safe_union/type_safe_union_kernel.h
pfeatherstone: yes this does look a bit nicer but i'm worried that:
- people will get excessive warnings about unused variables
- compilers who are too smart for their own good will optimize out the entire thing because `unused` is unused.

dlib/type_safe_union/type_safe_union_kernel_abstract.h
pfeatherstone: yep, good idea

dlib/type_safe_union/type_safe_union_kernel_abstract.h
```diff
@@ -345,6 +372,57 @@ namespace dlib
...
+              all types U in Types...:
+
+                std::forward<F>(f)(const in_place_tag<U>& tag, Variant&& tsu)
+        ensures
+            - This function iterates over all types U in Types... and calls:
+
+                std::forward<F>(f)(const in_place_tag<U>& tag, Variant&& tsu)
+
+            - tsu is perfect-forwarded for all types U in Types... regardless of whether U is the current
+              object held by tsu.
```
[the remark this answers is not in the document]
pfeatherstone: Ok, i think my IDE is probably doing something funky, because it looks fine in my editor.

pfeatherstone:
@davisking I hope the documentation style is as you want it. I did manage to get the parameter pack version of `apply_to_contents()` and `visit()` working, even with gcc-5. I'm sorry for yet another change. But i do think those implementations are way nicer. I'll wait for your verdict on using `int unused[]` instead of `(void)std::initializer_list<int>`. I don't want a million compiler messages about unused parameters. `[[maybe_unused]]` is only valid in c++17 and `__attribute__((unused))` is not cross-platform.

dlib/type_safe_union/type_safe_union_kernel.h
davisking: Huh, yeah I suppose those are possabilities.  So keep what you have :+1:

pfeatherstone:
@davisking Ready for a fresh review

Issue #2292, opened 2021-01-27
Some issues with the repeat layer


arrufat:
I have been playing a bit with the repeat layer, and I am facing some problems.

Here's an example that showcases the problems I encoutered:

``` cpp
#include <dlib/dnn.h>

using namespace dlib;
using namespace std;

template <typename SUBNET>
using conblock = relu<bn_con<con<32, 3, 3, 1, 1, SUBNET>>>;

// With this network definition everything works
using net_type = loss_multiclass_log<fc<10, conblock<conblock<conblock<input<matrix<float>>>>>>>;

// If we try to use a repeat layer we face several problems
// this doesn't build, repeat layers can't have an input layer as a SUBNET
// using net_type = loss_multiclass_log<fc<10, repeat<3, conblock, input<matrix<float>>>>>;

// This does build, but then, we can't use the get_final_data_gradient method
// using net_type = loss_multiclass_log<fc<10, repeat<2, conblock, conblock<input<matrix<float>>>>>>;

int main()
{
    net_type net;
    std::vector<adam> solvers(net.num_computational_layers, adam());
    std::vector<long> labels(32, 1);
    std::vector<matrix<float>> samples(32, matrix_cast<float>(gaussian_randm(100, 100)));
    resizable_tensor x;
    net.to_tensor(samples.begin(), samples.end(), x);
    net.compute_loss(x, labels.begin());
    net.back_propagate_error(x);
    net.update_parameters(solvers, 0.1);
    const auto& grad = net.get_final_data_gradient();
    cout << grad.num_samples() << 'x' << grad.k() << 'x' << grad.nr() << 'x' << grad.nc() << endl;
}
```

Let's break it down, there are 3 network definitions (that define the same architecture)

1. the first one doesn't use repeat layers, and everything works fine
1. the second one replaces the 3 `conblock`s with a `repeat` layer. Apparently, the `SUBNET` can't be an `input` layer.
1. the third one uses a `repeat` layer to replace 2 `conblocks` and and extra `conblock` before the `input`, but then I realized that `repeat` layers don't have a `get_final_data_gradient()` method.

I am not too worried about the 2nd case, where we can't use an `input` layer as a `SUBNET` for the `repeat` layer, but I think not having the `get_final_data_gradient()` method is a problem... Is this easily solvable?

Thanks in advance :)

davisking:
Oh yeah.  The `repeat` layer is missing a 
```
        const tensor& get_final_data_gradient(                                                                                          
[...]

PR #2477, opened 2021-12-21
add option to not zero out gradients and method to do it

arrufat:
So, as I mentioned in #2292, it can be useful, sometimes, to inspect the intermediate gradients in the network.
Notable use cases include computing things like [GradCAM](https://arxiv.org/abs/1610.02391), or cumulating the gradients before back propagation to use a larger batch size.

With this PR, you can do something like:
```c++
net.compute_loss(x, labels.begin());
net.back_propagate_error(x, false);  // the false here means "don't zero out gradients automatically"
const auto& gi = net.subnet().get_gradient_input();  // before this PR this would always be zero.
// do something with the gradients...
net.zero_out_gradients();  // zero out input gradients for next iteration
```
I've updated the `back_propagate_error` and `compute_parameter_gradients` to have an extra boolean defaulted to the current behavior. Maybe it should be an enum?
```c++
enum class should_zero_out_gradients : uint8_t
{
    yes,
    no
}
```

I would like to support those cases in dlib, and I have tried not to break the current API. What do you think? Am I close to something that might make it into dlib?

Maybe the name of the method should be more explicit? Like `zero_out_gradient_inputs`?

If that's OK, I will update the documentation.

### DIFF
--- a/dlib/dnn/core.h
+++ b/dlib/dnn/core.h
@@ -1032,7 +1032,7 @@ namespace dlib
             subnetwork->back_propagate_error(x, zero_grads); 
 
             // zero out get_gradient_input()
-            gradient_input_is_stale = zero_grads == should_set_gradient_inputs_to_zero::yes ? true : false;
+            gradient_input_is_stale = zero_grads == should_set_gradient_inputs_to_zero::yes;
         }
 
         template <typename solver_type>
@@ -1419,7 +1419,7 @@ namespace dlib
                 gradient_input, wsub, static_cast<tensor&>(params_grad));
 
             // zero out get_gradient_input()
-            gradient_input_is_stale = zero_grads == should_set_gradient_inputs_to_zero::yes ? true : false;
+            gradient_input_is_stale = zero_grads == should_set_gradient_inputs_to_zero::yes;
         }
 
         template <typename solver_type>
--- a/dlib/dnn/core.h
+++ b/dlib/dnn/core.h
@@ -1003,21 +1003,21 @@ namespace dlib
         const tensor& get_final_data_gradient(
         ) const { return subnetwork->get_final_data_gradient(); }
 
-        void back_propagate_error(const tensor& x)
+        void back_propagate_error(const tensor& x, bool zero_grads = true)
         {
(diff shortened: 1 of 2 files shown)


dlib/dnn/core.h
```diff
@@ -1430,6 +1436,11 @@ namespace dlib
 
         unsigned int sample_expansion_factor() const { return _sample_expansion_factor; }
 
+        void zero_out_gradients()
+        {
+            x_grad = 0;
```
davisking, line 1441:
```suggestion
            gradient_input_is_stale = true;
```
This should do it and avoid unnecessary assignments when it's already 0.

dlib/dnn/core.h
```diff
@@ -1430,6 +1436,11 @@ namespace dlib
 
         unsigned int sample_expansion_factor() const { return _sample_expansion_factor; }
 
+        void zero_out_gradients()
```
davisking, line 1439:
```suggestion
        void set_gradient_inputs_to_zero()
```
How about this name?

davisking (changes requested):
Yeah, seems sensible.  I like your `enum` `should_zero_out_gradients` idea too.  Totally do that instead of using a `bool` :)

arrufat:
I think the `enum` name is a bit too long… maybe it's not an issue, but it would be nice to have a more compact name, and I can't think of a good one :(

davisking:
> should_zero_out_gradients

`zero_gradients::yes`?

arrufat:
If tests pass, I think it's ready for review :)

dlib/dnn/core.h
```diff
@@ -1057,6 +1072,12 @@ namespace dlib
 
         unsigned int sample_expansion_factor() const { return subnet().sample_expansion_factor(); }
 
+        void set_gradient_inputs_to_zero()
+        {
+            x_grad = 0;
```
davisking, line 1077:
```suggestion
            gradient_input_is_stale = true;
```
This works here too right?

arrufat: Yes, right! Thank you.

davisking:
Sweet, thanks for the PR :)

PR #2481, opened 2022-01-03
Avoid different kinds of compiler warnings

reunanen:
... that started to appear when upgrading my build environment

### DIFF
--- a/dlib/cuda/cpu_dlib.h
+++ b/dlib/cuda/cpu_dlib.h
@@ -423,20 +423,20 @@ namespace dlib
 
         void resize_bilinear (
             tensor& dest,
-            long dest_row_stride,
-            long dest_channel_stride,
+            long long dest_row_stride,
+            long long dest_channel_stride,
             const tensor& src,
-            long src_row_stride,
-            long src_channel_stride
+            long long src_row_stride,
+            long long src_channel_stride
         );
 
         void resize_bilinear_gradient (
             tensor& grad,
-            long grad_row_stride,
-            long grad_channel_stride,
+            long long grad_row_stride,
+            long long grad_channel_stride,
             const tensor& gradient_input,
-            long gradient_input_row_stride,
-            long gradient_input_channel_stride
+            long long gradient_input_row_stride,
+            long long gradient_input_channel_stride
         );
 
         inline void resize_bilinear (
--- a/dlib/cuda/cpu_dlib.h
+++ b/dlib/cuda/cpu_dlib.h
@@ -442,14 +442,12 @@ namespace dlib
         inline void resize_bilinear (
             tensor& dest,
             const tensor& src
-        ) { resize_bilinear(dest, static_cast<long>(dest.nc()), static_cast<long>(dest.nr()*dest.nc()),
-                            src, static_cast<long>(src.nc()), static_cast<long>(src.nr()*src.nc())); }
+        ) { resize_bilinear(dest, dest.nc(), dest.nr()*dest.nc(), src, src.nc(), src.nr()*src.nc()); }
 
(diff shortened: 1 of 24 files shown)


dlib/dnn/core.h
```diff
@@ -3206,25 +3206,25 @@ namespace dlib
             unsigned long count_outputs() const
             {
                 if (subnetwork)
-                    return subnetwork->count_outputs() + output.size();
+                    return static_cast<unsigned long>(subnetwork->count_outputs() + output.size());
                 else
-                    return output.size();
+                    return static_cast<unsigned long>(output.size());
```
reunanen, line 3211: In these kinds of cases, it might often make sense to change the function signature instead. But I wanted to keep the changes to a bare minimum, at least for the time being.

dlib/cuda/cpu_dlib.h
```diff
@@ -442,12 +442,14 @@ namespace dlib
         inline void resize_bilinear (
             tensor& dest,
             const tensor& src
-        ) { resize_bilinear(dest, dest.nc(), dest.nr()*dest.nc(), src, src.nc(), src.nr()*src.nc()); }
+        ) { resize_bilinear(dest, static_cast<long>(dest.nc()), static_cast<long>(dest.nr()*dest.nc()),
```
davisking, line 445: This is kinda funky.  What's the warning? Visual studio is complaining about loss of precision going from `long long` to `long` presumably?  That's a silly warning honestly.  A ton of code performs this kind of implicit casting.

Anyway, it would make more sense to change `resize_bilinear` to take `long long` though than add these casts.  Can you do that instead?

davisking (commented):
I'm not super jazzed about all these casts.  What compiler is causing the issue?  This feels like a visual studio thing.

dlib/cuda/cpu_dlib.h
reunanen: Hmm. For some reason, when I now reverted these changes in my local build, Visual Studio no longer complains about any of these issues. So it doesn't look like I need any of these changes after all. This is strange, because I do not admit having changed any compiler warning level settings in between. (The complaining started, when I upgraded my project from VS 2017 to VS 2019.)

So I'll at least revert these overly verbose `static_cast`s. But I guess there are other small changes here that could be considered minor improvements. And yes, maybe I'll also make `resize_bilinear` take `long long`, while at it?

reunanen:
There is also this:
https://github.com/reunanen/dlib/commit/f1a5e0c44ee4715c150c5c5e037e15b0ffe47845

I understand it's only a Visual Studio issue, so originally didn't even want to offer it to your upstream repo.

But let me know if you want to include it (others might suffer from the same issue), and I'll be happy to cherry-pick it to this PR.

dlib/cuda/cpu_dlib.h
davisking: Yeah it's sensible to make `resize_bilinear` use `long long` to match what `tensor` uses.  If you want to adjust that that's cool :)

davisking:
> There is also this: [reunanen@f1a5e0c](https://github.com/reunanen/dlib/commit/f1a5e0c44ee4715c150c5c5e037e15b0ffe47845)
> 
> I understand it's only a Visual Studio issue, so originally didn't even want to offer it to your upstream repo.
> 
> But let me know if you want to include it (others might suffer from the same issue), and I'll be happy to cherry-pick it to this PR.

Yeah that looks like a sensible thing to put into the mainline.  It's just large scale `static_casting` and stuff that makes the code harder to read that I don't like :shrug:

davisking:
Sweet, looks good. Thanks for another PR :)

PR #2483, opened 2022-01-10
Fix MSVC pragma warnings with other compilers

arrufat:
After #2481, several warnings appear when building with GCC. This PR fixes that.

### DIFF
--- a/dlib/geometry/rectangle.h
+++ b/dlib/geometry/rectangle.h
@@ -750,8 +750,10 @@ namespace dlib
 
     // Circumvent what appears to be a bug in Visual Studio 2019's optimizer
     // (see: https://forum.juce.com/t/warning-in-the-lastest-vs2019/38267)
+#if defined (_MSVC_VER)
 #pragma warning ( push )
 #pragma warning ( disable: 4723 )
+#endif
     inline rectangle set_rect_area (
         const rectangle& rect,
         unsigned long area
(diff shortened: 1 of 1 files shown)


davisking:
Thanks :)

reunanen:
Yeah, my bad, sorry – and thanks @arrufat for fixing!

For the future, would it make sense to add something like this in CMakeLists.txt?

`target_compile_options(dlib PRIVATE -Werror)`

This would make all warnings be treated as errors, so issues like this ought to be caught already by Travis, pre-merge.

(And while at it, why not also try `-Wall -Wextra -Wpedantic`...?)

arrufat:
@reunanen no worries :)

I am all in for the `-Wall -Wextra -pedantic`, I have them in all my builds.
However, I have to add `-Wno-deprecated-copy` as well, or I get a weird error for which I haven't spent enough time trying to fix.

davisking:
`-Werror` would invariably end up with us building fine but then users with other compiler versions we don't test on posting question like "Why do I get this dlib build error?".  We go through periods where there are no warnings from any dlib code with the highest warning levels, then new warnings start to appear with new/different compilers.  Just part of life :shrug: 

Although a `-Werror` that was only on specifically for travis builds would be fine with me.  Someone should send a PR that enables that :)

dlib/geometry/rectangle.h
```diff
@@ -750,8 +750,10 @@ namespace dlib
 
     // Circumvent what appears to be a bug in Visual Studio 2019's optimizer
     // (see: https://forum.juce.com/t/warning-in-the-lastest-vs2019/38267)
+#if defined (_MSVC_VER)
```
reunanen, line 753: Actually it seems that this should be `_MSC_VER`, i.e. no `V` there.

I will include a fix in #2490.

PR #2487, opened 2022-01-14
Fix warning about unused zero_gradients parameter

arrufat:


dlib/dnn/core.h
```diff
@@ -2251,7 +2251,7 @@ namespace dlib
 
         void back_propagate_error(
             const tensor& /*x*/,
-            zero_gradients zero_grads = zero_gradients::yes
+            zero_gradients /*zero_grads*/
```
davisking, line 2254: How about `zero_gradients /*zero_grads*/ = zero_gradients::yes`?  So the signature matches all the others still?

arrufat: Fixed, thank you!

PR #2488, opened 2022-01-15
Eliminate grid sensitivity in YOLO

arrufat:
I was re-reading the YOLOv4 paper, and came across the concept if grid sensitivity, which I somehow overlooked before.
This issue contains a good explanation: https://github.com/AlexeyAB/darknet/issues/3293.

The main idea is that, if an object has the center of its bounding box exactly on the grid lines, the center regression will not be able to detect it.
The reason is that the network will have to output either 0 or 1 for x and y, respectively, which is impossible, being the output of a sigmoid.

To solve this problem, we can just scale and shift the sigmoid function to allow the x and y outputs to range between -0.5 and 1.5 like this: output = sigmoid(input) * 2.0 – 0.5.

According to the experiments run by @AlexeyAB and @glenn-jocher, it mitigates this issue, while improving the AP a bit.
I've also noticed that, even if this parameter is configurable in the Darknet framework, state-of-the-art models like Scaled-YOLO and YOLOR have it hard-coded.

I've run some experiments on my side, and it seems to work just fine.

Anyway, I just wanted to make this change before the next dlib release, which I read it's going to be soon :)

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -3625,25 +3625,26 @@ namespace dlib
 
                 for (size_t a = 0; a < anchors.size(); ++a)
                 {
+                    const long k = a * num_feats;
                     for (long r = 0; r < output_tensor.nr(); ++r)
                     {
                         for (long c = 0; c < output_tensor.nc(); ++c)
                         {
-                            const float obj = out_data[tensor_index(output_tensor, n, a * num_feats + 4, r, c)];
+                            const float obj = out_data[tensor_index(output_tensor, n, k + 4, r, c)];
                             if (obj > adjust_threshold)
                             {
-                                const auto x = out_data[tensor_index(output_tensor, n, a * num_feats + 0, r, c)];
-                                const auto y = out_data[tensor_index(output_tensor, n, a * num_feats + 1, r, c)];
-                                const auto w = out_data[tensor_index(output_tensor, n, a * num_feats + 2, r, c)];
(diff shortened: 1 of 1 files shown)


glenn-jocher:
@arrufat yes this applies a larger 2x sigmoid area that allows for easier detection at the grid boundaries like in YOLOv5.

BUT be careful applying this during inference of already-trained models. This grid update must naturally be learned during training to work during deployment, so if you apply this update it may result in all previously trained YOLO models losing compatibility with the current code.

arrufat:
Hi, @glenn-jocher thank you for your remark, I really appreciate it, and will take the chance to congratulate you on your work on YOLO.

dlib hasn't seen a stable release since the YOLO loss made it into the master branch. That is why I think it's going to be fine if we just add the change before a tagged version is released.

davisking:
Yeah I was just about to say "Oh man, this will break existing users!" but remembered the yolo loss is new and as yet unreleased.  So a breaking change is fine.  The silent part is undesirable, but my impression is that no one but @arrufat is probably using this code yet so we are probably fine there as well.

arrufat:
Yes, I was glad that you didn't release a stable version of dlib since the YOLO loss when I saw this.
These days I also want to experiment with the cIOU loss for bounding box regression (but that shouldn't be a breaking change).
I also want to play a  bit with the new coordinate system for width and height (which uses a sigmoid instead of an exponential function, although that would be a breaking change...)

Issue #2486, opened 2022-01-13
Issue with the LayerNorm implementation


arrufat:
In #2485 I started testing dlib's implementation of the layer normalization layer (added in #2213).

The main idea is the same as the batch normalization, except that it normalizes along the features dimension, instead of along the batch dimension, therefore is batch-size independent:

[<img src="https://production-media.paperswithcode.com/methods/Screen_Shot_2020-05-19_at_4.24.42_PM.png" alt="BatchNorm vs LayerNorm" width=400>](https://arxiv.org/abs/1607.06450) [<img src="https://user-images.githubusercontent.com/1671644/149344911-fc5c91d9-26a5-47b9-a677-1e47af2bb3bc.png" alt="BatchNorm vs LayerNorm" width=400>](https://yonghyuc.wordpress.com/2020/03/04/batch-norm-vs-layer-norm/)

The issue I am facing is that this layer also learns a scale and a bias, which have `num_samples` elements, as it normalizes each sample in the mini batch independently.

So, when we change the number of samples in the mini batch, the sizes of the scale and bias do not match anymore.
I might have some conceptual problem, and I am trying to find what it is, but if someone can enlighten me, I would appreciate it.
Meanwhile, do not use LayerNorm.

davisking:
Oh yeah, I see what you mean. I was apparently not paying attention to that part of that PR :sweat_smile: 

But yeah, layer norm should have the same number of affine parameters as batch norm.  That is, each output feature gets its own rescaling just like in batch norm. Definitely doesn't make sense to do a "per batch sample" thing like the code is doing now.

arrufat:
I was also reading the [documentation in PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html), and apparently you can even specify the shape of the normalization.
If I understood it correctly, if the tensor we want to normalize has dimensions `k`×`h`×`w`, you can tell it to normalize using only `k`. So, instead of learning `k`×`h`×`w` parameters, layer will only learn `k`. That's what they do in the [ConvNeXt code](https://github.com/facebookresearch/ConvNeXt/blob/b2d5a2de8b04533cc288400584db756c29066109/models/convnext.py#L16-L19).

I was thinking about how I would implement in dlib. The first thing that came to mind is to add template parameters to the `layer_norm_` class.

```c++
template <bool K, bool NR, bool NC>
class layer_norm_
{
    // ...
};
```
[...]

PR #2489, opened 2022-01-16
Fix Layer Normalize

arrufat:
Solves #2486.
Fixes Layer Norm, introduced in #2213.

For this PR, I've decided to not over do it and just fix the implementation. Maybe at some point I will check if it's worth to add dimensionality options.


davisking:
Cool, thanks :D

Issue #2463, opened 2021-11-20
Dlib failed to build due to many errors: error C2065 and C2338


Agustinm28:
I tried to build Dlib on Windows with VS2017 Update2. It failed to build due to many errors: such as error C2065, C2338 or MSB6003. Could you please help take a look at this? Thank you!

## Expected Behavior
Build successfully.

## Current Behavior
running install
running bdist_egg
running egg_info
writing tools/python\dlib.egg-info\PKG-INFO
writing dependency_links to tools/python\dlib.egg-info\dependency_links.txt
writing top-level names to tools/python\dlib.egg-info\top_level.txt
package init file 'tools\python\dlib\__init__.py' not found (or not a regular file)
reading manifest file 'tools/python\dlib.egg-info\SOURCES.txt'
reading manifest template 'MANIFEST.in'
no previously-included directories found matching 'tools\python\build*'
adding license file 'LICENSE.txt'
writing manifest file 'tools/python\dlib.egg-info\SOURCES.txt'
installing library code to build\bdist.win32\egg
running install_lib
running build_py
running build_ext
Building extension for Python 3.10.0 (tags/v3.10.0:b494f59, Oct  4 2021, 19:00:18) [MSC v.1929 64 bit (AMD64)]
Invoking CMake setup: 'cmake C:\Users\agust\Downloads\dlib-19.22\tools\python -DCMAKE_LIBRARY_OUTPUT_DIRECTORY=C:\Users\agust\Downloads\dlib-19.22\build\lib.win32-3.10 -DPYTHON_EXECUTABLE=C:\Users\agust\AppData\Local\Programs\Python\Python310\python.exe -DCMAKE_LIBRARY_OUTPUT_DIRECTORY_RELEASE=C:\Users\agust\Downloads\dlib-19.22\build\lib.win32-3.10 -A x64'
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.22000.
-- pybind11 v2.2.4
-- Using CMake version: 3.22.0
-- Compiling dlib version: 19.22.0
-- SSE4 instructions can be executed by the host processor.
-- AVX instructions can be executed by the host processor.
-- Enabling AVX instructions
-- C++11 activated.
-- Configuring done
-- Generating done
-- Build files have been written to: C:/Users/agust/Downloads/dlib-19.22/build/temp.win32-3.10/Release
Invoking CMake build: 'cmake --build . --config Release -- /m'
[...]

PR #2492, opened 2022-01-27
Resolve davisking/dlib#2463

sambhare:
Fixes missing ssize_t definition on Windows

### DIFF
--- a/dlib/external/pybind11/include/pybind11/numpy.h
+++ b/dlib/external/pybind11/include/pybind11/numpy.h
@@ -26,6 +26,8 @@
 #if defined(_MSC_VER)
 #  pragma warning(push)
 #  pragma warning(disable: 4127) // warning C4127: Conditional expression is constant
+#include <BaseTsd.h>
+using ssize_t = SSIZE_T;
 #endif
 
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  This doesn't work though, see the build errors here https://ci.appveyor.com/project/davisking/dlib-1mqn9/build/job/dyfua7l48ii2gs9g.  I'm not sure why the appveyor integration says it passed, since the build failed on appveyor :|

The issue is sometimes ssize_t is defined.

sambhare:
Interesting. Looks like the 32 bit builds are failing and the 64 bit ones are ok. I'll update the PR to only do this on 64 bit.

Also, Python 3.6 was EOLed in Dec 2021. (https://devguide.python.org/#status-of-python-branches) Would you be updating your CI to remove it? I would recommend the oldest supported version (3.7) and the latest version (3.10)

davisking:
Yeah, I'll update the appveyor config.  I tried 3.10 a bit ago but seemingly appveyor doesn't actually support it, despite their docs saying they do.  Trying to switch to 3.9 instead.

Anyway, thanks for the PR.  CI looks good now :)

sambhare:
Thank you!

PR #2493, opened 2022-01-28
See what appveyor says about this

davisking:


davisking:
I looked for a while and it's not clear to me how to make appveyor do the right thing.   What's happening is there are 4 different appveyor config files in dlib/appveyor.  All 4 are actually run on every PR, but there is a race condition in the github integration that causes one of the 4 to be selected at random to decide if the build succeeded and if it blocks the PR or not.   And I couldn't find out how to resolve that integration bug.

It seems like you are supposed to have just one appveyor config file for this integration.  However, each of those config files corresponds to an appveyor "project".  And each project has a 2 hour time limit.  If we combined them all into one we would exceed the 2 hour build limit.  Which is why I split them up into separate files in the first place.

Anyway, I get emails when these fail. It's just annoying that PR authors to not also see appveyor CI failures in many cases.  But it's not clear to me if it's even possible to resolve this so I'm letting it sit.

arrufat:
I will check if I can finish the GitHub actions PR, maybe it works better?

The Linux one works fine, but for the macOS and Windows ones, it can't find the `dtest` executable... I don't have much experience with those OSes.

davisking:
Cool.  No idea.  I find the docs for how all these services are supposed to work to be kinda dicy.  I'm trying to figure out the YAML appveyor uses right now to build newer python versions together with older ones in a sensible way.  It is not exactly clear :|

arrufat:
Yes, I also find it a bit confusing, so I go with the trial and error method (that branch has far too many commits :sweat_smile:

davisking:
Yeah I'm still in a trial and error loop for this :|

PR #2495, opened 2022-01-30
Add missing docs for v19.23

arrufat:
I noticed that the YOLO and Barlow Twins losses were missing from the index, so I added them. I hope I did it right.

Why are some losses in `ml.html`, while others are under `dlib/dnn/loss_abstract.h.html`?


docs/docs/term_index.xml
```diff
@@ -147,6 +147,9 @@
          <term file="ml.html" name="loss_mean_squared_multioutput_" include="dlib/dnn.h"/>
          <term file="ml.html" name="loss_mmod_" include="dlib/dnn.h"/>
          <term file="dlib/dnn/loss_abstract.h.html" name="mmod_options" include="dlib/dnn.h"/>
```
arrufat, line 149: This is what I was referring to.

davisking: Oh yeah, I forgot some of these :|

Update your git though, I did add the loss_yolo to ml.xml.  You must be looking at a branch or something.

arrufat: Oh, I don't know how that happened! Fixed :)

davisking:
Sweet, thanks.

davisking:
It's live on dlib.net now too :D

PR #2496, opened 2022-01-30
Add Reorg Layer

arrufat:
This PR adds a simple Reorg layer.
I think it was introduced in YOLOv2 (and YOLOR uses it as well) instead of a strided convolution for down-sampling in the first layer.

### DIFF
--- a/dlib/cuda/tensor_tools.h
+++ b/dlib/cuda/tensor_tools.h
@@ -1851,7 +1851,8 @@ namespace dlib { namespace tt
             - dest.nr() == src.nr() / row_stride
             - dest.nc() == src.nc() / col_stride
         ensures
-            - Converts downsamples the spatial resolution into channel information
+            - Converts the spatial resolution into channel information.  So all the values in the input tensor 
+              appear in the output tensor, just in different positions.  
             - For all n, k, r, c in dest:
                 dest.host[tensor_index(dest, n, k, r, c)] ==
                 src.host[tensor_index(src,
--- a/dlib/cuda/tensor_tools.h
+++ b/dlib/cuda/tensor_tools.h
@@ -1835,6 +1835,58 @@ namespace dlib { namespace tt
 
 // ----------------------------------------------------------------------------------------
 
+    void reorg (
+        tensor& dest,
+        const int row_stride,
+        const int col_stride,
+        const tensor& src
+    );
+    /*!
+        requires
+            - is_same_object(dest, src)==false
+            - src.nr() % row_stride == 0
+            - src.nc() % col_stride == 0
+            - dest.num_samples() == src.num_samples()
+            - dest.k() == src.k() * row_stride * col_stride
+            - dest.nr() == src.nr() / row_stride
+            - dest.nc() == src.nc() / col_stride
+        ensures
+            - Converts downsamples the spatial resolution into channel information
+            - For all n, k, r, c in dest:
+                dest.host[tensor_index(dest, n, k, r, c)] ==
+                src.host[tensor_index(src,
+                                      n,
+                                      k % src.k(),
+                                      r * row_stride + (k / src.k()) / row_stride,
+                                      c * col_stride + (k / src.k()) % col_stride)]
+
+
+    !*/
+
+    void reorg_gradient (
+        tensor& grad,
+        const int row_stride,
+        const int col_stride,
+        const tensor& gradient_input
+    );
+    /*!
+        requires
+            - is_same_object(dest, src)==false
+            - gradient_input.nr % row_stride == 0
+            - gradient_input.nc % col_stride == 0
+            - dest.num_samples() == src.num_samples()
+            - grad.k() == gradient_input.k() / row_stride / col_stride
+            - grad.nr() == gradient_input.nr() * row_stride
+            - grad.nc() == gradient_input.nc() * col_stride
+        ensures
+            - Suppose that DEST is the output of reog(DEST, row_stride, col_stride, SRC)
+              for some SRC tensor, let f(SRC) == dot(gradient_input,DEST).  Then this
+              function computes the gradient of f() with respect to SRC and adds it to grad.
+            - It effectively reverts the reorg operation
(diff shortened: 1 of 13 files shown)


dlib/dnn/core.h
```diff
@@ -3273,8 +3273,8 @@ namespace dlib
                 // layer.
                 const long num_samples = rnd.get_random_32bit_number()%4+3;
                 const long k  = rnd.get_random_32bit_number()%4+2;
-                const long nr = rnd.get_random_32bit_number()%4+2;
-                const long nc = rnd.get_random_32bit_number()%4+2;
+                const long nr = ((rnd.get_random_32bit_number()%4)/2)*2+2;
+                const long nc = ((rnd.get_random_32bit_number()%4)/2)*2+2;
```
arrufat, line 3277: I am not sure about this. I had to add this change to make sure the number of rows and columns in the tensor is even, so that the test with `reorg_<2, 2>` passes… We can remove this change and the test, if you don't like it.

arrufat:
I am also not sure if I should make the tensor tools functions expect the destination tensor to be already allocated, or make them allocate the tensor instead. Do you have any preference? Maybe the latter is easier on the user side?

arrufat:
Just some tests, in case anyone is curious about this layer and has never seen what it does.

```c++
#include <dlib/dnn.h>
#include <dlib/image_io.h>

using namespace dlib;
using net_type = reorg<input_rgb_image>;

try
{
    if (argc != 2)
    {
        std::cout << "provide an image." << std::endl;
        return EXIT_FAILURE;
    }
    net_type net;
    matrix<rgb_pixel> image, letterbox;
    load_image(image, argv[1]);
    letterbox_image(image, letterbox, 512);
    net(letterbox);
    std::cout << net << std::endl;

    const auto& t = net.get_output();
    std::vector<matrix<unsigned char>> planes;
    std::vector<matrix<rgb_pixel>> images;
    for (long k = 0; k < t.k(); ++k)
    {
        matrix<float> plane = image_plane(t, 0, k);
        plane -= min(plane);
        plane /= max(plane);
        matrix<unsigned char> tmp;
        assign_image(tmp, plane * 255);
        save_png(tmp, "plane_" + pad_int_with_zeros(k, 2) + ".png");
        images.push_back(jet(tmp, 0, 255));
        planes.push_back(std::move(tmp));
    }
    save_png(tile_images(images), "reorg_images.png");
    save_png(tile_images(planes), "reorg_planes.png");
    return EXIT_SUCCESS;
}
{
    std::cout << e.what() << std::endl;
    return EXIT_FAILURE;
}
```
This code defines a network that only contains a `reorg` layer.
Since the input is an RGB image, its shape will be: `N` × `3` × `H` ×`W`. The output of the `reorg` layer will be `N` × `3·2·2` × `H/2` × `C/2` as, by default, it has row and column strides of 2.
Then the spatial information will be reorganized along the channel dimension.
If we take this famous [dog image](https://github.com/pjreddie/darknet/blob/master/data/dog.jpg), the program will print:
```
layer<0>        output size=(num:1, k:12,   nr:256, nc:256)     reorg    (row_stride=2, col_stride=2)
layer<1>        input_rgb_image(122.782,117.001,104.298)
```

And generate the following images:
reorg_planes
We can colorize the with `jet` to try to spot the differences.
reorg_images

Even though they look very similar, they are all different:
```bash
$ md5sum plane*png
8236be41acec5e3137d71f2135d5db8a  plane_00.png
dbd8e5e3b3ad090c5afc91700e00c9fa  plane_01.png
543caf2ccdbcf561484b90879f3ade48  plane_02.png
ee2b900b51357aa0ef67748d61e9e848  plane_03.png
f66c245c7e3e1bc8af2bfaa8a5bb8d01  plane_04.png
119af98e158f5b86b953d024376e8fa6  plane_05.png
1fc60ec594cc0dfd3c17efed73deb554  plane_06.png
434dfa412ec87cc8f8fa6363bff92104  plane_07.png
f7e8b3b0e96b94be34830ec521296c5e  plane_08.png
4cf34437afee02440dd00fc797451e84  plane_09.png
b63414b01269792bbb65db9d7e4154db  plane_10.png
f490cec7f6c1744e93467b7f11979da2  plane_11.png
```

dlib/dnn/core.h
davisking: Na, it's fine.

dlib/cuda/tensor_tools.h
```diff
@@ -1833,6 +1833,58 @@ namespace dlib { namespace tt
...
+        requires
+            - is_same_object(dest, src)==false
+            - src.nr() % row_stride == 0
+            - src.nc() % col_stride == 0
+            - dest.num_samples() == src.num_samples()
+            - dest.k() == src.k() * row_stride * col_stride
+            - dest.nr() == src.nr() / row_stride
+            - dest.nc() == src.nc() / col_stride
+        ensures
+            - Converts downsamples the spatial resolution into channel information
```
davisking, line 1854:
```suggestion
            - Converts the spatial resolution into channel information.  So all the values in the input tensor 
              appear in the output tensor, just in different positions.  
```

davisking (approved):
Nice

davisking:
> I am also not sure if I should make the tensor tools functions expect the destination tensor to be already allocated, or make them allocate the tensor instead. Do you have any preference? Maybe the latter is easier on the user side?

It's good how you have it.  Matches the rest of the code in there.

arrufat:
By the way, if I wanted to add this to the changelog of the upcoming version, where should I put it?
I've seen this: https://github.com/davisking/dlib/blob/master/docs/docs/release_notes.xml
But it has only the `current` tag, not a development one.

davisking:
> By the way, if I wanted to add this to the changelog of the upcoming version, where should I put it? I've seen this: https://github.com/davisking/dlib/blob/master/docs/docs/release_notes.xml But it has only the `current` tag, not a development one.
 
Just change the `<current>` block to an `<old>` one and make a new `<current>`

PR #2497, opened 2022-01-31
Add input_rgb_image_pair layer and new net visitors documentation

arrufat:
Oops, in my previous PR I missed the `input_rgb_image_pair` layer.

### DIFF
--- a/docs/docs/ml.xml
+++ b/docs/docs/ml.xml
@@ -127,6 +127,7 @@ Davis E. King. <a href="http://jmlr.csail.mit.edu/papers/volume10/king09a/king09
                <item>input_rgb_image_sized</item>
                <item>input_rgb_image_pyramid</item>
                <item>input_grayscale_image_pyramid</item>
+               <item>input_grayscale_image_pair</item>
                <item>
                   <name>EXAMPLE_INPUT_LAYER</name>
                   <link>dlib/dnn/input_abstract.h.html#EXAMPLE_INPUT_LAYER</link>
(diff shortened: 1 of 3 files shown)


arrufat:
I think we missed the docs because I should have split the self-supervised learning PR #2434 in 3:
- `input_rgb_image_pair`
- `loss_barlow_twins`
- `dnn_self_supervised_example_ex`

I will try to do more atomic PR :)

arrufat:
Done :)

docs/docs/ml.xml
```diff
@@ -817,6 +818,28 @@ Davis E. King. <a href="http://jmlr.csail.mit.edu/papers/volume10/king09a/king09
...
+         <name>input_rgb_image_pair</name>
+         <file>dlib/dnn.h</file>
+         <spec_file link="true">dlib/dnn/input_abstract.h</spec_file>
+         <description>
+            This is a simple input layer type for use in a deep neural network
+            which takes a pair of RGB images as input and loads it into a network.
+            It is useful when you want to input image pairs that are related to each other,
+            for instance, they are different distorted views of the same original image.
+            This input layer is meant to be used with a loss layer such as the
+            <a href="#loss_barlow_twins_">Barlow Twins loss layer</a>.  You can also
```
davisking, line 834: There would need to be a block on ml.xml for `#loss_barlow_twins_` to link to anything.  So you need to add one or link directly to the abstract file.

davisking:
Yeah, I forgot about these too.  I normally add docs as I go along adding things.  You should probably do that too :)

If you go do `dlib/docs` and run `./makedocs fast` it will regenerate the html docs. Then you can look at dlib/docs/docs/chm/docs/index.html and see if it looks right and if the links work.

docs/docs/ml.xml
arrufat: done!

arrufat:
Yes, I will start adding documentation there, too.
I use the `term_index.html` a lot, so if I find something it is not documented, I will add it and make a PR, now that I know how to do it and test it.
Thank you!

arrufat:
I think that's it for this PR :)

davisking:
Yep, looks good.

PR #2498, opened 2022-02-02
Do not round drectangle in rectangle transform

arrufat:
I was thinking that it makes more sense, if I pass a `drectangle` to the `rectangle_transform`, to get a `drectangle` whose coordinates have not been rounded. What do you think?

I noticed this with YOLO detectors, since I use `letterbox_image`, I need to transform the detections back to the original image space. I was surprised to see that all detections had integer coordinates for their bounding boxes.

### DIFF
--- a/dlib/geometry/point_transforms.h
+++ b/dlib/geometry/point_transforms.h
@@ -235,14 +235,15 @@ namespace dlib
             // the above box.
             double scale = std::sqrt(new_area/temp.area());
 
-            return centered_rect(center(temp), std::round(temp.width()*scale), std::round(temp.height()*scale));
+            return centered_drect(dcenter(temp), temp.width()*scale, temp.height()*scale);
         }
 
         rectangle operator() (
             const rectangle& r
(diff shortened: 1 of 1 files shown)


davisking:
Yeah this is how it should be.  I have no idea why I coded it the way it is right now.  I just looked over the git history for a while for a clue, since I obviously did it this way on purpose.  But I don't see why and I looked around the code base for a while.  This is definitely how it should be.  I'm having a moment of paranoia that there is some critical thing I'm not remembering but I'm fairly sure there isn't at this point.

Anyway, let's do it :)

arrufat:
Maybe if the user assigns the `drectangle` to a `rectangle` won't get the “proper” rounding, just a floor of the coordinates?

davisking:
> Maybe if the user assigns the `drectangle` to a `rectangle` won't get the “proper” rounding, just a floor of the coordinates?

Na, the conversion from `drectangle` to `rectangle` rounds, not floors.

PR #2499, opened 2022-02-03
Add get_random_beta to dlib::rand

arrufat:
I was trying to implement the [mixup paper](https://arxiv.org/abs/1710.09412) and I noticed they use the beta distribution to generate the blending parameter α.

So, I tried to implement it myself and added to the `dlib::rand`.
I got **a lot** of “inspiration” from the [NumPy code](https://github.com/numpy/numpy/blob/1995e2cdfd17ec75e37cc6e69a42ce9000aac966/numpy/random/src/distributions/distributions.c#L406-L439).

@pfeatherstone, since you added already 2 distribution generators (Weibull and Exponential) would you mind having a look?
Specially, I am a bit puzzled about line 421 in the NumPy code. It looks like that condition will be always true inside that block. Am I missing something? If I implement it the same way as them, the code gets stuck in an infinite loop :(

### DIFF
--- a/dlib/rand/rand_kernel_abstract.h
+++ b/dlib/rand/rand_kernel_abstract.h
@@ -218,7 +218,6 @@ namespace dlib
                 requires
                     - alpha > 0
                     - beta > 0
-
                 ensures
                     - returns a random number sampled from a Beta distribution
                       with shape parameters alpha and beta.
--- a/dlib/rand/rand_kernel_abstract.h
+++ b/dlib/rand/rand_kernel_abstract.h
@@ -211,16 +211,17 @@ namespace dlib
             !*/
 
             double get_random_beta (
-                double a,
-                double b,
+                double alpha,
+                double beta,
             )
             /*!
                 requires
-                    - a > 0, b > 0
+                    - alpha > 0
+                    - beta > 0
 
                 ensures
                     - returns a random number sampled from a Beta distribution
-                      with shape parameters a and b.
+                      with shape parameters alpha and beta.
             !*/
 
             void swap (
--- a/dlib/rand/rand_kernel_abstract.h
+++ b/dlib/rand/rand_kernel_abstract.h
@@ -209,7 +209,20 @@ namespace dlib
                       with shape parameter k, scale parameter lambda and 
                       threshold parameter gamma.
             !*/
-            
+
+            double get_random_beta (
+                double a,
+                double b,
+            )
+            /*!
+                requires
+                    - a > 0, b > 0
+
+                ensures
+                    - returns a random number sampled from a Beta distribution
+                      with shape parameters a and b.
+            !*/
+
             void swap (
                 rand& item
             );
--- a/dlib/rand/rand_kernel_1.h
+++ b/dlib/rand/rand_kernel_1.h
@@ -5,7 +5,6 @@
 
 #include <string>
 #include <complex>
-#include <random>
 #include "../algs.h"
 #include "rand_kernel_abstract.h"
 #include "mersenne_twister.h"
--- a/dlib/rand/rand_kernel_1.h
+++ b/dlib/rand/rand_kernel_1.h
@@ -299,7 +299,7 @@ namespace dlib
                 DLIB_CASSERT(beta > 0, "beta must be greater than zero");
                 auto u = std::pow(get_random_double(), 1 / alpha);
                 auto v = std::pow(get_random_double(), 1 / beta);
-                while ((u + v) > 1 || u == 0 || v == 0)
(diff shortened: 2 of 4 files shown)


arrufat:
I've found a better solution :)

pfeatherstone:
I would implement the gamma distribution then use [this](https://en.wikipedia.org/wiki/Beta_distribution#Generating_beta-distributed_random_variates)

pfeatherstone:
I wonder if ```dlib::rand``` should be refactored to simply wrap ```std::mt19937_64``` and each function uses the right distribution.
For example:
```
class rand
{
public:
    rand(size_t seed = time(NULL)) : engine(seed) {}

    float get_gamma(float alpha, float beta)
    {
        std::gamma_distribution<float> d(alpha, beta);
        return d(engine);
    }

    float get_beta(float alpha, float beta)
    {
        float x = get_gamma(alpha, 1);
        float y = get_gamma(beta, 1)
        return x / (x+y);
    }

private:
    std::mt19937_64 engine;
};

**Comment by pfeatherstone** [comment]
There would also be no need to unit test ```dlib::rand``` since you would be wrapping simple standard library calls. We could template the generator too for added configuration. Like, we could write new generators which are cryptographically secure so long as they satisfy C++ random number generator requirements [ref](https://www.cplusplus.com/reference/random/). Or you could write really fast ones like [Xoroshiro128+](https://en.wikipedia.org/wiki/Xoroshiro128%2B). We would have to unit test the generators, but not the distributions since those are simple standard library calls... Ideas.

**Comment by pfeatherstone** [comment]
But this would not be backward compatible with the current implementation of ```dlib::rand``` so @davisking would probably not be keen. However, compatibility would only be an issue if users regularly serialized ```dlib::rand```. How many users do that? Probably not many. So in my view changing the implementation is not an issue. Or we create the ```dlib::rand2``` object. People object to that kind of thing but i don't mind at all.

**Comment by pfeatherstone** [comment]
Then again, what you've done @arrufat is super simple. So maybe that's best for now.

**Comment by arrufat** [comment]
Thank you for your input, definitely a lot to think about.
However, as you said, my approach is super simple, so for the purposes of this PR I think it's already good :)

**Comment by pfeatherstone** [comment]
As an aside, Github Actions looks awesome and way faster. Is there even any need for appveyor or travis now ?

**Comment by davisking** [comment]
Na `dlib::rand` can never change.  I'm quite confident there is even code out there that depends on the exact sequence of random numbers generated by `dlib::rand` for certain seeds.  We should really try not to compete with the standard library here.  Like if `<random>` had existed way back when I wouldn't have made `dlib::rand`.  That's not to say that sometimes I don't find `dlib::rand` still convenient.  But still, the burden of backwards compatibility is definitely sitting heavy on `dlib::rand` at this point.

There really should be a unit test too that asserts that the outputs are not ever changed.  I'm going to add that now.

**Comment by davisking** [comment]
Yeah github actions seems real nice though.  I want to see it run a bit more before dumping the other CI tools though.  And to maybe move over some of the other tests.

**Comment by davisking** [comment]
For instance, I am pretty enamored of the "old cmake" tests and the debug builds.  The examples should also be built.

**Comment by arrufat** [comment]
> For instance, I am pretty enamored of the "old cmake" tests and the debug builds. The examples should also be built.

I tried building in debug mode and Windows times out.
And when building the examples, the machine runs out of RAM :(

**Comment by davisking** [comment]
> > For instance, I am pretty enamored of the "old cmake" tests and the debug builds. The examples should also be built.
> 
> I tried building in debug mode and Windows times out. And when building the examples, the machine runs out of RAM :(

lol, damn.  Well maybe they get built only on linux.  IDK.  I'll fiddle with later.

Anyway, for a specific example of why it's important to never ever change the output of dlib::rand consider the locality sensitive hashing tools.  People who depend on those would be rightly upset if the hashes suddenly changed.  It would set the world on fire in a lot of applications that depend on that stuff.

**Comment by arrufat** [comment]
Thank you for the insights on `dlib::rand`, I didn't know it was so dependable! I'll be careful if I ever touch it.

And I resolved the merge conflicts :)

**Comment by arrufat** [comment]
That should fix the tests on Windows :)

**Comment by arrufat** [comment]
I think it's ready now :)

File: dlib/rand/rand_kernel_abstract.h [hunk 1]
```diff
@@ -209,7 +209,21 @@ namespace dlib
...
+
+            double get_random_beta (
+                double alpha,
+                double beta,
+            )
+            /*!
+                requires
+                    - alpha > 0
+                    - beta > 0
+
```
davisking, line 221:
```suggestion
```

davisking (approved):
Nice

PR #2502, opened 2022-02-06
Make Github Actions build more stuff (in particular, the examples)

reunanen:
Thought it might be nice to have Github Actions build the examples, too? Just to catch as many issues as soon as possible?

Apparently there's some issue with the Windows builds (think @arrufat ran into that as well?), but I don't think this needs to prevent us from doing it on the other OSes/compilers?

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -27,7 +27,7 @@ jobs:
       working-directory: ${{ github.workspace }}/dlib/test
       run: cmake . -B ${{ env.build_dir }}
 
-    - name: Build
+    - name: Build just tests
       working-directory: ${{ github.workspace }}/dlib/test
       run: cmake --build ${{ env.build_dir }} --config ${{ env.config }} --target dtest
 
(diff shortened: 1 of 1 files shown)


arrufat:
With debug builds, windows compilation was timing out (that might be solved with parallel builds) and some times I ran out of RAM... but it seems to work now 😕 🤔

davisking:
Sweet

davisking:
I just added `--parallel 8` to the builds too :)

arrufat:
Apparently, builds fail now.
On macOS because of the MPC test condition, which it's apparently too strict. I noticed that already and fixed in PR #2499.
And, I can't get the logs as per why it's failing on Linux…

davisking:
Yeah.  It seems like there is only actually 2 cores on these github machines.  I'm trying less parallelism now.

I also suspect the MPC code is failing on mac because the build in BLAS and LAPACK on mac machines is often awful.  Going to set it to not use that on mac and see what happens.

reunanen:
Yep, `--parallel 8` indeed appeared to be too much for building the examples. I'm guessing the build agent just runs out of RAM.

PR #2503, opened 2022-02-07
Fix building tests on macOS

arrufat:


davisking:
Ha, thanks.  I was just about to fix this :sweat_smile:

arrufat:
I am trying to alleviate you from some of these more tedious tasks.

PR #2507, opened 2022-02-12
Do not use deprecated `std::random_shuffle`

reunanen:
`std::random_shuffle` has been deprecated since C++14.


davisking (commented):
Thanks :D

PR #2509, opened 2022-02-12
Make GitHub actions use `gcc-11` in parallel with the default `gcc-9` that comes with `ubuntu-latest`

### LINKED ISSUE #2508: Make Github Actions run on default gcc version from ubuntu-latest, and also gcc-11
Would be cool if github actions tests ran with multiple versions of GCC.  At some point they are going to limit us though.   But it would be cool if there was an ubuntu-gcc11 mode in the build matrix in addition to the default gcc. Maybe it would all run in parallel.

_Originally posted by @davisking in https://github.com/davisking/dlib/issues/2506#issuecomment-1037239057_

reunanen:
Resolves #2508.

At some point, not using a matrix might be more convenient.

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -23,6 +23,13 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Install gcc 11
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        sudo apt update
+        sudo apt install gcc-11 g++-11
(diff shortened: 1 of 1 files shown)


reunanen:
Also, no idea if there are cleaner ways to do this...

arrufat:
I also need to read more on the GitHub actions documentation (and see more examples)

reunanen:
Fortunately the script is still quite simple, so it's not too difficult to go back and forth.

reunanen:
I mean, if you just want to list a bunch of jobs without using a matrix, for example.

Issue #2506, opened 2022-02-12
Compiler error when building `test/serialize.cpp` on gcc 11


reunanen:
```
[1/3] Building CXX object CMakeFiles/dtest.dir/serialize.cpp.o
FAILED: CMakeFiles/dtest.dir/serialize.cpp.o 
/usr/bin/c++  -I/home/adria/Projects/reunanen/dlib/.. -O3 -DNDEBUG   -W -Wall -Wextra -Werror -Wno-unused-function -Wno-strict-overflow -DDLIB_JPEG_SUPPORT -DDLIB_USE_BLAS -DDLIB_USE_LAPACK -DDLIB_USE_CUDA -DDLIB_PNG_SUPPORT -DDLIB_GIF_SUPPORT -Wreturn-type -MD -MT CMakeFiles/dtest.dir/serialize.cpp.o -MF CMakeFiles/dtest.dir/serialize.cpp.o.d -o CMakeFiles/dtest.dir/serialize.cpp.o -c /home/adria/Projects/reunanen/dlib/test/serialize.cpp
In file included from /home/adria/Projects/reunanen/dlib/../dlib/memory_manager/../binary_search_tree/../serialize.h:241,
                 from /home/adria/Projects/reunanen/dlib/../dlib/memory_manager/../binary_search_tree/binary_search_tree_kernel_2.h:11,
                 from /home/adria/Projects/reunanen/dlib/../dlib/memory_manager/memory_manager_kernel_3.h:11,
                 from /home/adria/Projects/reunanen/dlib/../dlib/memory_manager.h:8,
                 from /home/adria/Projects/reunanen/dlib/../dlib/conditioning_class.h:13,
                 from /home/adria/Projects/reunanen/dlib/../dlib/compress_stream.h:10,
                 from /home/adria/Projects/reunanen/dlib/test/serialize.cpp:8:
[...]

PR #2510, opened 2022-02-12
Make GitHub actions build on gcc-11 and clang too

reunanen:
Here's an alternative solution to #2508, but this one includes also two Clang jobs on `ubuntu-latest`: one on the current default Clang 11, but also another job explicitly on version 13.

I guess it's a matter of taste if you prefer a matrix, or just a list of jobs. A matrix is nice if you don't have too many exceptions that you need to cover by adding `if` conditions. This one gets a bit verbose, maybe – although I do believe that some deduplication could be done. (Had just a quick look at the options, but didn't find anything that would have looked particularly appealing...)

### DIFF
--- a/.github/workflows/build_cmake.yml
+++ b/.github/workflows/build_cmake.yml
@@ -28,6 +28,7 @@ jobs:
       run: cmake --build ${{ env.build_dir }} --config ${{ env.config }} --parallel 2
 
   ubuntu-latest-gcc-11:
+    if: ${{ false }}  # disabled until https://github.com/davisking/dlib/issues/2506 has been resolved
     runs-on: 'ubuntu-latest'
     steps:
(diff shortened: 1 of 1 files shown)


reunanen:
Same here as in #2509:

> As per #2506, the build now fails of course. But that is not this PR's fault. :)

reunanen:
For clarity: I do prefer this one over #2509 (which doesn't even do any Clang builds yet). But that's just my personal opinion.

I do see benefits in the matrix approach, but at the same time I think that separately handling the exceptions (using `if`s) may quickly start to make things unnecessarily cumbersome.

reunanen:
Yes, sure! Done.

reunanen:
Looks good to me now.

davisking:
Yep, thanks :D

PR #2518, opened 2022-02-19
Fix Barlow Twins loss gradient

arrufat:
It's a really minor fix (basically a typo), which I only noticed now.
A factor of 2 in the derivative of the Barlow Twins loss is missing. It's no big deal, since it can be compensated by setting a larger lambda, but still. It's even mentioned in the comment above…
I am running the example again, but I think nothing will change.


arrufat:
As expected, nothing changes, well, in this run, the test accuracy was 89% instead of 88%, so I think we can merge it :)

PR #2519, opened 2022-02-22
Speed up Barlow Twins loss

arrufat:
I noticed that, when using high dimensional features (such 8192), the Barlow Twins loss takes a really long time to compute on GPU. This issue is not noticeable with lower dimensional features, such as 128, like in the example.

That is because, at each step, many tensors are allocated, and copied to the GPU.
I've modified the computations so that only the first step allocates the tensors.

Since the `compute_loss_and_value_gradient` is a `const` method, I had to make the `resizable_tensor` member variables `mutable`… However, if the size of the input changes, the tensors will be reallocated and recomputed.

I hope you are OK with this modification.

In my case, `num_samples == 64` and `k == 8192`, it makes the network train almost 10 times faster.
I am open to some other optimization suggestions as well.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -4016,9 +4016,6 @@ namespace dlib
 
             // Normalize both batches independently across the batch dimension
             const double eps = 1e-4;
-            resizable_tensor za_norm, means_a, invstds_a;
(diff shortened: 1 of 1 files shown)


davisking:
Cool, no problem, this is a good example of when to use `mutable` :)

Issue #2522, opened 2022-02-23
Missing Threads::Threads target when using static dlib and dlibConfig*.cmake


APTX:
## Expected Behavior
No cmake error.

## Current Behavior
```
Target "my_target" links to target "Threads::Threads" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```

## Steps to Reproduce
I use `dlib` in my cmake project the following way:
```
find_package(dlib CONFIG REQUIRED)
# [...]
target_link_libraries(my_target dlib::dlib)
```

As far as I understand this is a limitation of how cmake handles "transitive dependencies" (dependencies of dependencies that become direct dependencies of a target). This is usually only an issue with static linking.

When linking with a static library, the target has to link to the dependencies of that library. When `dlib` links statically with CUDA, the [`Threads::Threads` dependency is added by `FindCUDA`](https://github.com/Kitware/CMake/blob/859241d2bbaae83f06c44bc21ab0dbd38725a3fc/Modules/FindCUDA.cmake#L1047).

When `my_target` links statically with `dlib`, it has to link `dlib`'s dependencies, including the ones inherited from CUDA. `dlib`'s cmake config file only lists the `Threads::Threads` dependency but does not find it like `FindCUDA` does.

As a workaround it's possible to fix the error by adding `find_package(Threads)` as the cmake error suggests. This way `my_target` finds the `Threads::Threads` dependency for `dlib`, even if `my_target` doesn't use `Threads::Threads` directly.

This could be a duplicate of issue #2107.

* **Version**: 19.22
* **Where did you get dlib**: `emerge dlib` (distribution package manager)
* **Platform**: Gentoo linux
* **Compiler**: gcc (Gentoo 11.2.0 p1) 11.2.0
* **Cmake version**: 3.22.2

davisking:
Gentoo isn't a platform that's officially supported (https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html) and my experience is that they are pretty serious about really not supporting other platforms.  I know people make it work sometimes anyway.  So this isn't surprising that there are issues.

Anyway, if there is some PR you can make that works around this I'm cool with it so long as it doesn't break or harm other dlib users in some way.

APTX:
[...]

PR #2526, opened 2022-02-26
Find Threads target if necessary (#2522)

APTX:
While I am not a cmake expert (and certainly not camke 2.x) I thought I'd give it a go.


davisking:
Thanks for the fix :)

APTX:
No problem.

I am not sure the checks actually test this issue. As far as I can tell, nothing actually uses the cmake config. The builds also don't seem to use CUDA, where the `Threads` dependency comes from.

davisking:
Heh, wait, but doesn't this change fix the problem for you?

davisking:
Or are you talking just about the github CI checks?  Yeah those aren't exercising the problem you are fixing here.

APTX:
The change does fix the problem for me. I was talking about github CI.

Issue #2521, opened 2022-02-23
Feature: net_to_dot visitor


arrufat:
Some time ago, I noticed that Keras has a nice functionality to [plot models](https://keras.io/api/utils/model_plotting_utils/#plotmodel-function).

So, I thought it would be really cool to have something similar in dlib, as it allows for easier debugging of the networks, aside from being cool to visualize.

Last night, I implemented a `net_to_dot` visitor which outputs a graph in the [dot language](https://graphviz.org/doc/info/lang.html) so that we can easily convert into almost any kind of graphical format. You just need to call it exactly like the `net_to_xml` visitor. Maybe `net_to_graph` would be a better name?

I guess my question is: would you be interested in adding this kind of functionality to dlib? If so, I will prepare a PR :)

Here are some examples of what this visitor can generate.

From the `resnet.h` file: [resnet18.pdf](https://github.com/davisking/dlib/files/8121295/resnet18.pdf)
From the semantic segmentation example: [unet.pdf](https://github.com/davisking/dlib/files/8121296/unet.pdf)
A classical DenseNet: [densenet121.pdf](https://github.com/davisking/dlib/files/8121298/densenet121.pdf)
Another architecture to test more layers: [vovnet39.pdf](https://github.com/davisking/dlib/files/8121297/vovnet39.pdf)

arrufat:
I've updated the style of the nodes, so that it's easier to debug a network:
- now the output of the network with `operator<<` matches better
- I print the tags as well
- I added a custom version of a YOLO model

Enjoy!

[densenet121.pdf](https://github.com/davisking/dlib/files/8124581/densenet121.pdf)
[vovnet39.pdf](https://github.com/davisking/dlib/files/8124582/vovnet39.pdf)
[resnet18.pdf](https://github.com/davisking/dlib/files/8124583/resnet18.pdf)
[unet.pdf](https://github.com/davisking/dlib/files/8124584/unet.pdf)
[yolor-p6.pdf](https://github.com/davisking/dlib/files/8124585/yolor-p6.pdf)

arrufat:
Closing as of #2528

PR #2528, opened 2022-02-27
Add visitor to draw nets in the DOT language

arrufat:
As mentioned in #2521, I've made a visitor to draw a neural network diagram in the [DOT](https://www.graphviz.org/doc/info/lang.html) language.

There's certainly room for improvement, as it requires specializing the methods for each layer manually.
However, its usage is really straightforward (exactly the same as the `net_to_xml` visitor) and it can draw arbitrarily complicated networks.
Here's an example with the [`resnet.h`](../tree/master/examples/resnet.h) file from the examples:

> `net_to_dot.cpp`
 ```c++
#include <dlib/dnn.h>
#include "resnet.h"
int main()
{
    resnet::train_18 net;
    net_to_dot(net, std::cout);
    // we can also make it write the output to a file directly, to inspect it and modify it
    net_to_dot(net, "net.gv")
}
```

Then, we just run it like this:
```bash
./net_to_dot | dot -Tpng > net.png
```

And it will generate this:
ResNet18

As you can see, I've added the visitor to a `dnn/visitors.h` file. If we proceed with this PR, maybe it would be a good idea to move other visitors here, such as:
- `disable_duplicative_biases`
- `fuse_layers`
- `set_all_bn_running_stats_window_sizes`

What do you think?
I've been using the last week, and it has proved to be very useful for debugging complicated networks, aside from generating pretty cool diagrams.

If you think it's out of the scope of dlib, at least people can find the code in the issues :)

### DIFF
--- a/dlib/dnn.h
+++ b/dlib/dnn.h
@@ -31,6 +31,7 @@
 #include "cuda/tensor_tools.h"
 #include "dnn/utilities.h"
 #include "dnn/validation.h"
+#include "dnn/visitors.h"
 
 #endif // DLIB_DNn_
 
--- a/dlib/dnn/visitors.h
+++ b/dlib/dnn/visitors.h
@@ -7,8 +7,6 @@
 #include "layers.h"
 #include "loss.h"
 
-#endif // DLIB_DNn_VISITORS_H_
-
 namespace dlib
 {
     namespace impl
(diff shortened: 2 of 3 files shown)


davisking:
Nice, that's really cool.

Yeah, putting all the visitors in their own header seems sensible to me.  If you want to do that in another PR that's cool.

arrufat:
Great, I've just added the documentation. I will add the other visitors once this is merged :)

PR #2531, opened 2022-03-01
Fix error calling `cudnnGetConvolutionForwardWorkspaceSize`

reunanen:
**Problem**: `Error while calling cudnnGetConvolutionForwardWorkspaceSize( context(), descriptor(data), (const cudnnFilterDescriptor_t)filter_handle, (const cudnnConvolutionDescriptor_t)conv_handle, descriptor(dest_desc), (cudnnConvolutionFwdAlgo_t)forward_algo, &forward_workspace_size_in_bytes) in file C:\a\2\s\3rdparty\dlib\dlib\cuda\cudnn_dlibapi.cpp:1029. code: 9, reason: CUDNN_STATUS_NOT_SUPPORTED`

**Solution**: Add remaining dimensions of `filters`, and all dimensions of `data` to cache key

---

Not sure, but it looks like otherwise the chosen algorithm may be unsupported for some input combinations (e.g., larger input?).

Also not sure if all those key components are really needed. But reproducing the problem is not very straightforward, so figuring that out would take some time. So personally would be fine to just err on the side of caution. The present solution at least seems to fix the problem. Yes, I am painfully aware that the cache would be more useful, if the key did not include data dimensions. Possibly the data dimensions could be rounded up a bit when constructing the cache key, or something like that, in order to promote more cache hits (without risking errors that is – but at the moment I'm not sure a smaller input is always safe for a particular "supported" combination).

An alternative, I guess, could have been to try out `cudnnGetConvolutionForwardWorkspaceSize`, and in the event of failure, just clean the related cache item and try again.

### DIFF
--- a/dlib/cuda/cudnn_dlibapi.cpp
+++ b/dlib/cuda/cudnn_dlibapi.cpp
@@ -792,12 +792,14 @@ namespace dlib
         {
             // Calling the cuDNN "find the best algorithm" functions are really slow.  So we keep a
             // cache that tells us what method was best for a particular configuration.
-            thread_local std::map<std::tuple<int,int,int,int,long,long>,
+            thread_local std::map<std::tuple<int,int,int,int,long,long,long,long,long,long,long,long>,
                                   std::tuple<int,int,int>> config_to_algo_cache;
 
             // If we have already found good algorithms for this setting then just pull them from
             // the cache.
-            const auto cache_key = std::make_tuple(stride_y, stride_x, padding_y, padding_x, filters_nr, filters_nc);
+            const auto cache_key = std::make_tuple(stride_y, stride_x, padding_y, padding_x,
+                                                   filters_nr, filters_nc, filters_num_samples, filters_k,
+                                                   data_nr, data_nc, data_num_samples, data_k);
             const auto iter = config_to_algo_cache.find(cache_key);
             if (iter != config_to_algo_cache.end())
             {


davisking:
Na, we can't do this.  This defeats the point of the cache.  Which was to avoid the huge delay someone gets when sending images of different sizes through a model.  There is the additional problem that this could blow up and be huge if someone is processing a ton of images (or whatever input thing) of a large number of varying sizes.  It would basically manifest as a memory leak.

If the problem is just that there are some odd non-supported values for some algorithms it ought to be possible to find those by running a loop that just tries this code with all possible (well, a large range of) values for these settings and checking when cudnn complains.  There is probably some simple logic that could be put here to handle whatever the odd case is.

reunanen:
Yes, I am painfully aware of those points (that this defeats the point of the cache – although not completely in my understanding – and also that expanding the key could lead to something that may gradually start to look like a memory leak).

reunanen:
How about this approach then?

> An alternative, I guess, could have been to try out `cudnnGetConvolutionForwardWorkspaceSize`, and in the event of failure, just clean the related cache item and try again.

I guess that way it should eventually converge to an algorithm that at least works for all input sizes (although it may not be the optimal choice for some).

davisking:
Yeah IDK. If that works then that sounds good. The only worry is that maybe that function wouldn't ever return a failure. Like this seems like it's a bug in cudnn. So maybe the only way to find the bad settings is to run the real conv code and see if it blows up and then to manually record what values are bad.

reunanen:
I think I can test that as an alternative approach. It's not very easy for me to do, though. (The coding part is easy, but getting a conclusive test result is not.)

reunanen:
Closing because #2532 was merged instead.

nairoj:
> Na, we can't do this. This defeats the point of the cache. Which was to avoid the huge delay someone gets when sending images of different sizes through a model. There is the additional problem that this could blow up and be huge if someone is processing a ton of images (or whatever input thing) of a large number of varying sizes. It would basically manifest as a memory leak.
> 
> If the problem is just that there are some odd non-supported values for some algorithms it ought to be possible to find those by running a loop that just tries this code with all possible (well, a large range of) values for these settings and checking when cudnn complains. There is probably some simple logic that could be put here to handle whatever the odd case is.

I finally found the cause of my program's memory leak!!

PR #2533, opened 2022-03-03
Refactor visitors into their own header

arrufat:
After this change, I was able to remove the recenty-introduced `dlib/dnn/misc.h`, since the `dlib/dnn/utilities.h` no longer needs `dlib/dnn/core.h`.


davisking:
Nice :D

arrufat:
Yay, 100 PRs! I never thought I'd make it!

arrufat:
Yay, 100 commits! I never thought I'd make it!

PR #2539, opened 2022-03-07
Add missing visitor implementations to visitors.h

arrufat:
Notably, `set_all_bn_running_stats_window_sizes` and `fuse_layers`.

But also I took the chance to remove the superflous separators,
change the attribute of `upsample` layers from `stride` to `scale`,
and clean up trailing white space (since it's a new file, and it's part of a refactoring)

Sorry, that should've been in the previous PR :(


davisking:
Na, thanks for doing all this :)

Issue #2540, opened 2022-03-09
Add support to release Linux(x86_64 and aarch64) wheels


odidev:
## Expected Behavior
On Linux, `pip install dlib` should download the wheels from PyPI.

## Current Behavior
On Linux, `pip install dlib` builds the wheels from source code and then installs it. It requires the user to have a development environment installed on their system. Also, it takes more time to build the wheels than downloading and extracting the wheels from PyPI.

## Steps to Reproduce
Add support to release Linux(x86_64 and aarch64) wheels

* **Version**: dlib-19.23.0
* **Where did you get dlib**: https://pypi.org/project/dlib
* **Platform**: Ubuntu 20.04.3

@davisking, Please let me know your interest in releasing Linux(x86_64 and aarch64) wheels. I can help with this.

PR #2541, opened 2022-03-09
Add Linux(x86_64 and aarch64) wheel build support

odidev:
Added Linux(x86_64 and aarch64) wheel build support.
Related to https://github.com/davisking/dlib/issues/2540, @davisking, Could you please review this PR?

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -1,0 +1,57 @@
+name: Linux Wheels
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  package-wheel:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [x86_64, aarch64]
+        python-version: [cp36, cp37, cp38, cp39, cp310]
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Set up QEMU
+        if: ${{ matrix.arch == 'aarch64' }}
+        uses: docker/setup-qemu-action@v1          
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR.  What's this going to do?  Build the wheel and try to publish it to PyPi? Where is `secrets.PYPI_PASSWORD` coming from?

So `pip install dlib` should keep building the package.  People depend on that since it configures it for their system.  E.g. it detects CUDA and enables that if they have it.  It detects their CPU arch and enables AVX and/or SSE for instance, stuff like that.  That's all important.  But a generic binary has to disable all that stuff.  So it's not super clear to me that's really very useful to very many people.  Do you use dlib with all that stuff disabled?

odidev:
> Thanks for the PR. What's this going to do? Build the wheel and try to publish it to PyPi? Where is `secrets.PYPI_PASSWORD` coming from?

secrets.PYPI_PASSWORD is the password you’ll need to set which allows you to upload your Python distribution packages in the dist/ directory to PyPI. 
For more information, please see this: https://docs.github.com/en/actions/security-guides/encrypted-secrets

> So `pip install dlib` should keep building the package. People depend on that since it configures it for their system. E.g. it detects CUDA and enables that if they have it. It detects their CPU arch and enables AVX and/or SSE for instance, stuff like that. That's all important. But a generic binary has to disable all that stuff. So it's not super clear to me that's really very useful to very many people. Do you use dlib with all that stuff disabled?

Wheels install faster than source distributions. Also it avoids arbitrary code execution (setup.py) for installation. For more information on wheels, please check this: https://pythonwheels.com/

odidev:
@davisking, Any update on this?

davisking:
We can't make `pip install dlib` disable the features I mentioned.  There could be a `pip install dlib-linux-binary-no-gpu-or-any-acceleration` I suppose, but I'm not super sure who would use that.

PR #2545, opened 2022-03-18
Add Focal Loss

arrufat:
This PR adds the Focal Loss layer introduced in the paper: [Focal Loss for Dense Object Detection](https://arxiv.org/abs/1708.02002).

This layer can be seen as an improvement to the `loss_multibinary_log` from #2141.
I actually wonder whether I should remove the `loss_multibinary_log` and just define it as a particular case of the Focal Loss when `gamma == 0`. :sweat_smile: 

I consider this layer to be the ultimate layer for classification problems using neural networks in dlib. Here's my reasoning:

- It is based on the `loss_multibinary_log`, so:
  - It works on par with `loss_multiclass_log` on single label datasets (from my and other people's experiments).
  - It can output nothing when presented with a sample that it's not from the trained categories.
  - It can output multiple labels with high confidence if they appear in the same image (even when trained with single-label datasets).
  - It's straightforward to calibrate using dlib's `learn_platt_scaling`.
  - You can change the relative weights and make up for an imbalanced dataset: in particular, I tried [this paper](https://arxiv.org/abs/1901.05555) on a highly imbalanced dataset, and it works great.
- You can set `gamma > 0` and make the loss focus on classifying hard examples (penalizing predictions with low confidence).
- You can combine both points above.

As you can see, it is a very versatile layer, able to solve multi-label classification problems with dataset imbalance and with samples that can be hard to classify. Maybe at some point, I might add a `loss_focal_per_pixel_` for segmentation problems.

Anyway, for now, what do you think? Somehow, I never paid enough attention to the Focal loss…

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -1021,7 +1021,7 @@ namespace dlib
         }
 
     private:
-        float gamma = 0;
+        float gamma = 2.0;
     };
 
     template <typename SUBNET>
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -973,14 +973,14 @@ namespace dlib
                         const float temp = log1pexp(-out_data[idx]);
                         const float focus = std::pow(1 - g[idx], gamma);
                         loss += y * scale * temp * focus;
-                        g[idx] = y * scale * focus * (g[idx] * (1 - gamma * temp) - 1);
+                        g[idx] = y * scale * focus * (g[idx] * (gamma * temp + 1) - 1);
                     }
                     else
                     {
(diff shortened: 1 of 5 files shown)


davisking:
How about adding `gamma` to `loss_multibinary_log`?  `loss_multibinary_log` is a really nice name, it's very suggestive of what it does :)

The focal loss is basically a slightly smoothed version of the hinge loss though, which I've found works fine with a lot of DNNs.  Although I admit that sometimes the hinge loss works less well than log losses, which I think has got to come down to SGD being a bad optimizer.  But it is what it is.

arrufat:
Ah, sure, I will add the gamma parameter to the `loss_multibinary_log` it makes more sense. I don't know why I didn't think about it before.

PR #2546, opened 2022-03-21
Add focal gamma to loss_multibinary_log

arrufat:
As suggested in #2545, this PR modifies the existing `loss_multibinary_log_`, introduced in #2141 to support the gamma parameter presented in the Focal loss.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -770,6 +770,13 @@ namespace dlib
         typedef std::vector<float> training_label_type;
         typedef std::vector<float> output_label_type;
 
+        loss_multibinary_log_() = default;
+
+        loss_multibinary_log_(double gamma) : gamma(gamma)
+        {
+            DLIB_CASSERT(gamma >= 0);
+        }
+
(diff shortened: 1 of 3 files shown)


davisking:
Nice :)

pfeatherstone:
Presumably this is being used in some sick YOLO training ? :)

arrufat:
> Presumably this is being used in some sick YOLO training ? :)

Haha! Nothing that fancy… I am training a classifier at work, where there is huge class imbalance in the dataset and some categories work better than others despite having less samples (they are easier to distinguish from the rest). So, I am trying to improve the baseline by using the focal loss.

I am not sure the Focal loss will give any improvement to the YOLO loss, though. The YOLOv3 paper already tried it, and they experienced a drop in mAP (I also found a similar tendency when working with Darknet).

Moreover, YOLOv5 has the focal loss gamma, but it's set to 0 by default (and they try a lot of hyperparameter tuning, so I don't think I can find a setup that works better than what they have already).

PR #2547, opened 2022-03-23
Fix deserialize logic in loss_multibinary_log

arrufat:
I could have fixed it just by replacing the `||` by `&&`, but I wanted to set the `gamma` explicitly as well, like you do in the `loss_metric_`


davisking:
Oops, yeah, I should have noticed that too when reviewing it :facepalm: 

Thanks for the fix :)

PR #2552, opened 2022-03-30
Fix focal loss

arrufat:
The loss value should reflect the impact of the gamma parameter as well.

I was too “focused” on getting the gradient right, that I forgot about the loss…


davisking:
Ha, yep, thanks for the fix :)

PR #2553, opened 2022-04-02
Add directory_exists

arrufat:
I think it can be handy, and it provides some kind of consistency with the already present `file_exists`.

### DIFF
--- a/dlib/dir_nav/dir_nav_extensions.cpp
+++ b/dlib/dir_nav/dir_nav_extensions.cpp
@@ -53,6 +53,23 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    bool directory_exists (
+        const std::string& dirname
+    )
+    {
+        try
+        {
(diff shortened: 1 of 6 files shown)


davisking:
Do you have a C++17 compiler?  If so you can use the standard library for filesystem stuff now.  Which is totally better than what's in dlib :)  Check out https://en.cppreference.com/w/cpp/filesystem

If this is a "yeah but I don't have C++17 thing" then that's cool though.

arrufat:
I do have a C++17 compiler, so I don't really need it myself.

Honestly, I was a bit lazy here, and since I was already using the dlib `dir_nav.h` to get the files in a directory recursively. I just wanted to check that the path given to perform the search was a directory, so it was natural to do it in dlib. But I can get around it with `std::filesystem`, I've used it several times :).

Feel free to close this PR :)

davisking:
Na it's cool.  Just pointing it out in case you didn't know it was there.  Thanks for the PR :)

PR #2554, opened 2022-04-03
Check positive, negative and zero truth values in loss

arrufat:
It's a tiny change, I am not sure if it's meaningful, but it could be used as a way to ignore certain samples.
What do you think?


arrufat:
I don't know what I was thinking... that case it's already taken into account, as the whole loss and gradient are multiplied by y. It might be a micro optimization, though, so we don't need to compute the whole thing when y is zero.

arrufat:
I've been thinking again, and it makes sense to have the `y==0` case, since most of the values in the truth

PR #2557, opened 2022-04-06
Add SmeLU activation

arrufat:
For a while, I've been looking for an activation to replace ReLU.
I want a smoothed version that can be computed fast and in-place. SiLU, Mish, GELU don't fit the bill in some of those points.
Today, I found about [SmeLU](https://ai.googleblog.com/2022/04/reproducibility-in-deep-learning-and.html) ([paper](https://arxiv.org/abs/2010.09931)) (it has a really weird name). And I've implemented it as an in-place operation, so it will use the same amount of memory as a ReLU layer when training and inferring.

I've also updated some tests to actually check for negative values, i.e., replacing `rnd.fill_uniform` with `rnd.fill_gaussian`.

### DIFF
--- a/dlib/cuda/cpu_dlib.cpp
+++ b/dlib/cuda/cpu_dlib.cpp
@@ -1998,6 +1998,61 @@ namespace dlib
             }
         }
 
+        void smelu (
+            tensor& dest,
+            const tensor& src,
+            const float beta
+        )
+        {
+            const float* s = src.host();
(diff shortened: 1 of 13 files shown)


davisking:
Nice, thanks for another great PR :)

arrufat:
Oh, I've noticed you're not squashing the PRs anymore. Any reason? Just curious :)

PR #2560, opened 2022-04-10
Use CUDA in LayerNorm gradient computation

arrufat:
I don't know how I could miss this.


davisking:
Thanks, no worries :)

PR #2565, opened 2022-04-13
Add webp support

arrufat:
I will try to revive #2437.


arrufat:
I've kept the `reinterpret_cast`s because otherwise I get this error:
```
error: invalid conversion from ‘void*’ to ‘unsigned char*’ [-fpermissive]
```

arrufat:
I don't know why I can't get good PSNR… I've made a mistake, but I can't see it…
Even when saving lossless, I can see the MD5 are the same, yet the images in dlib are not the same… I don't understand… Maybe it's too late here, and I should get some rest.

Other than that, it's ready

dlib/test/image.cpp
```diff
@@ -2306,6 +2306,78 @@ namespace
...
+        compressor.decompress(sin, sout);
+        sin.clear();
+        sin.str(sout.str());
+        deserialize(rgba_img, sin);
+        // perform lossless test
+        const float quality = 101;
+        save_webp(rgba_img, "test_rgba.webp", quality);
+        load_webp(rgba_dec, "test_rgba.webp");
+        save_webp(rgba_dec, "test_rgba_dec.webp", quality);
+        // If we check both files externally, they are exactly the same, however, the test fails
```
arrufat, line 2364: @davisking @sandsmark when you have time, can you check this?
Both saved images `test_rgba.webp` and `test_rgba_dec.webp` are exactly the same (have the same md5sum) but this test fails...

arrufat:
I've found the problem, and it's due to how WebP works in lossless mode: RGB pixels in transparent areas will be modified to increase compression, so I shouldn't check for the non-alpha values if they are 100% transparent when computing the PSNR.

Then, when we assign the RGBA image to an RGB image, the values in the transparent areas will have undefined values, so the PSNR decreases a lot when performing this assignment: the image in the transparent areas is very noisy. We can fix this behavior by using the `WebPEncode` function and creating a `WebPConfig::exact = 1`, instead of the simpler `WebPEncodeLossless`.

I think the default behavior is already good, prioritizing compression when the alpha channel is enabled. However, if you think otherwise, I can modify the WebP saver and use a custom configuration.

As an aside, I might add better support to WebP in a future PR, notably to create and load WebP animations… After trying this format for image storage over the last few days, I came to like it a lot: it outperforms JPEG, PNG, and GIF in every aspect. So full support would be nice to create animations in dlib directly, since we are already adding this dependency.

Ah, and one last thing, I've added lossless compression support when setting `quality > 100`. If you think it's confusing, I can make an extra function called `save_webp_lossless` and make the `save_webp` behave like `save_jpeg`, where the quality ranges between 0 and 75. I did it like this because it felt natural to me…
Then, I saw that OpenCV uses the same convention (`quality > 100 == lossless`). However, OpenCV uses a lossless setting by default, what do you think we should do?

Now it's finally ready for review, really.

dlib/test/image.cpp
arrufat: Solved, see: https://developers.google.com/speed/webp/docs/api?hl=en#simple_encoding_api

davisking:
That all sounds good and like the right way to do it.  So it's definitely ready for review?  I briefly looked over it and it looks good.

arrufat:
Yes, it's ready :D

davisking:
> Yes, it's ready :D

Sweet.  Looks good.  I'll look it over a bit more soon and merge it :)

arrufat:
Ah, I will update the website documentation and the changelog later, too.

dlib/image_saver/save_webp_abstract.h
```diff
@@ -0,0 +1,54 @@
...
+              dlib/image_processing/generic_image.h or a matrix expression
+            - image.size() != 0
+            - quality >= 0
+        ensures
+            - writes the image to the file indicated by filename in the WEBP format.
+            - image[0][0] will be in the upper left corner of the image.
+            - image[image.nr()-1][image.nc()-1] will be in the lower right corner of the
+              image.
+            - This routine can save images containing any type of pixel.  However,
+              save_webp() can only natively store rgb_pixel, bgr_pixel,  rgb_alpha_pixel and
```
davisking, line 35:
```suggestion
              save_webp() can only natively store rgb_pixel, bgr_pixel, rgb_alpha_pixel and
```

davisking (approved):
Nice this is great :)

PR #2570, opened 2022-04-18
Workaround for #2506

reunanen:
I was trying to wrap my head around #2506, mostly because I would like to enable GitHub Actions for gcc 11 as well.

However, at least so far I couldn't find a solution where we could (de)serialize `std::vector`s of `my_custom_type`, if `immutable_type` is an alternative in the `q` variant. At least cleanly and without defining a custom move assignment operator for `my_custom_type`, or doing other (presumably) unwanted things. Not even sure if pushing such variants to a vector should be possible to begin with, as per the C++ standard – maybe gcc 11 is just doing the right thing here?

So here is one workaround, where we simply don't put such variants in vectors. We can still (de)serialize _standalone_ variants that can hold an immutable alternative.

@davisking, @pfeatherstone, @arrufat – any comments?

### DIFF
--- a/.github/workflows/build_cpp.yml
+++ b/.github/workflows/build_cpp.yml
@@ -32,7 +32,6 @@ jobs:
       run: cmake --build ${{ env.build_dir }} --config ${{ env.config }} --parallel 2
 
   ubuntu-latest-gcc-11:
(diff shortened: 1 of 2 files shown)


davisking:
Yeah, I think this is sensible. My read on the standard is that it should work but it's just a bug in that version of the compiler.  Thanks for the PR :)

PR #2580, opened 2022-04-28
Add WebP support to imglab

arrufat:
It should have been included in #2565.
I just noticed imglab couldn't find a dataset with WebP images.

### DIFF
--- a/tools/imglab/src/main.cpp
+++ b/tools/imglab/src/main.cpp
@@ -567,15 +567,15 @@ int tile_dataset(const command_line_parser& parser)
     {
         save_png(tile_images(images), out_image);
     }
-    else if (ext == "jpg")
-    {
-        save_jpeg(tile_images(images), out_image, JPEG_QUALITY);
-    }
-    else
+    else if (ext == "webp")
     {
         // Lossless by default
         const float webp_quality = get_option(parser, "webp", 101.f);
         save_webp(tile_images(images), out_image, webp_quality);
+    }
+    else
+    {
+        save_jpeg(tile_images(images), out_image, JPEG_QUALITY);
     }
 
     return EXIT_SUCCESS;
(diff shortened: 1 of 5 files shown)


tools/imglab/src/main.cpp
```diff
@@ -552,9 +564,19 @@ int tile_dataset(const command_line_parser& parser)
...
     if (ext == "png")
+    {
         save_png(tile_images(images), out_image);
+    }
+    else if (ext == "jpg")
+    {
+        save_jpeg(tile_images(images), out_image, JPEG_QUALITY);
+    }
     else
-        save_jpeg(tile_images(images), out_image);
```
davisking, line 557: jpeg is a better default for this.  It's really small and more importantly dlib will pretty much always end up being built with jpeg support, but not webp support.

arrufat: I thought about it, but to reach this point, the extension must be either ".png", ".jpg", or ".webp", any other extension will cause the program to exit at line 513, (525 after this PR). So, I think it's fine, right? We will always use the extension requested by the user.

arrufat: But maybe I can change the order of the if/else statements if that makes it more clear :)

davisking: Oh right, yeah duh.  This was fine :D

davisking:
Sweet, thanks for the PR :)

arrufat:
I see that CI on Travis for windows is failing due to not having WebP support enabled. But not on GitHub Actions, though.
I think we're not building the `tools` in GitHub (imglab, htmlify, etc).

I think we have 2 options:
- either add some #ifdef guards around the WebP code (maybe not ideal, but it's nice to have WebP support)
- remove support for writing WebP from imglab, and add only support for reading by modifying the `match_endings`, which is probably what most end users will care about: finding WebP images if they exist in their dataset and therefore, they should bother enabling WebP support.

What do you think?

PR #2582, opened 2022-04-29
imglab: Do not build WebP features if WebP is not available

arrufat:
Approach 1: fix build of imglab when WebP is not available (using `ifdef` guards around WebP-specific code)
I understand it's not very clean, I can remove the WebP-writing features from imglab in a new commit and just keep the extensions in the `match_endings` object.

Anyway, now Travis CI should pass.

### DIFF
--- a/tools/imglab/src/cluster.cpp
+++ b/tools/imglab/src/cluster.cpp
@@ -302,6 +302,7 @@ int cluster_dataset(
                 temp.push_back(images[assignments[i].idx]);
         }
 
+#ifdef DLIB_WEBP_SUPPORT
         if (parser.option("webp"))
         {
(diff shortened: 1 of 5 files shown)


davisking:
Cool, yeah, didn't realize github actions wasn't building this stuff for windows.  This PR looks good.  Let me put up a change to have github actions build this stuff on windows to see that it really builds it and fails too.

davisking:
Just remembered we run out of ram on github actions when building the examples there :|

davisking:
Yeah was thinking exactly the same thing 🤞😁

arrufat:
Maybe we can build the tools on Windows only, without the examples, I think that should do it...

PR #2583, opened 2022-04-29
Build ancillary tools on Windows

arrufat:
Let's see if this works

### DIFF
--- a/.github/workflows/build_cpp.yml
+++ b/.github/workflows/build_cpp.yml
@@ -111,6 +111,8 @@ jobs:
       run: cmake --build ${{ env.build_dir }} --config ${{ env.config }} --target dtest --parallel 4
     - name: Test
       run: ${{ env.build_dir }}/${{ env.config }}/dtest.exe --runall -q
+    - name: Build ancillary tools
+      run: cmake --build ${{ env.build_dir }} --config ${{ env.config }} --target imglab htmlify convert_dlib_nets_to_caffe_build --parallel 4
 
   macos-latest:
     runs-on: 'macos-latest'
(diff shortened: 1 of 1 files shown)


davisking:
Cool, thanks :)

arrufat:
Maybe we could build all the examples, except for the dnn ones? That should not cause an OOM exception.

davisking:
Yeah, that's kinda unsatisfying too though.  Needs to be all of them or we need to keep using appveyor.  Maybe there is some way to get more ram on github actions.

davisking:
Kinda seems like that's not an option. Maybe building in debug mode uses less ram though?

davisking:
Don't worry about it.  I just pushed a commit that does it. :)

davisking:
It's worth having a debug only build anyway since it tests codes in the `ENABLE_ASSERTS` pathways.  My server in the basement tests that stuff.  I would go crazy and add all kinds of stuff but I don't want tests to take forever or get rate limited by github like we did on travis.  Maybe that won't happen though.  But I bet at some point they are like "knock it off" if we run too much stuff.

arrufat:
Oh, I see. Yes, I remember you having to remove some tests because we ran out of quota. And also tests taking over than 1 hour to start on Appveyor.
The good thing about GitHub actions is that they run pretty quickly.

davisking:
Yeah seems to be a ton of quota on github.  Travis doesn't even run anymore.  We are constantly out of quota.

arrufat:
Oh, same problem, when building the YOLO example:
```
   dnn_yolo_train_ex.cpp
D:\a\dlib\dlib\dlib\dnn\core.h(1102,1): fatal error C1060: compiler is out of heap space [D:\a\dlib\dlib\dlib\test\build\examples\examples_build\dnn_yolo_train_ex.vcxproj]
```

davisking:
Yeah, debug build doesn't help :|

arrufat:
From what I've seen, only the YOLO and the Semantic segmentation examples cause an OOM problem…

davisking:
Yeah, same.

PR #2584, opened 2022-05-03
Add SiLU activation layer

arrufat:
Simple PR to add support for the Sigmoid Linear Unit (SiLU) activation layer.
This layer is quite popular in recent state-of-the-art models such as YOLOv5 and EfficientNetV2.
It is, however, a relatively “old” activation layer, introduced in 2016 in the [GELU paper](https://arxiv.org/abs/1606.08415).
So, I thought it would be a nice addition to dlib :)


davisking:

PR #2591, opened 2022-05-29
Fix saving grayscale WebP images

arrufat:
I just noticed that there was a problem when saving grayscale images in WebP format.
WebP does not support single-channel images, so we need to save them as RGB, which we were doing.
However, the stride was wrong: I was using the stride as if it were a single-channel image, so we need to recompute it before saving.

This PR addresses this issue.

Fun fact: even though we save a lossless RGB image in WebP, it still takes less disk space than the grayscale PNG, at least for the images I tried.


davisking:
Sweet. Thanks for the fix :)

PR #2604, opened 2022-06-16
noexcept dlib::type_safe_union::swap()

pfeatherstone:


pfeatherstone:
You might complain that users are doing something like:
```
dlib::type_safe_union<...> a,b;
a.swap(b);
```
instead of 
```
swap(a,b)
```
That would be unusual.

pfeatherstone:
Also, currently, ```std::swap(tsu&, tsu&)``` and ```swap(tsu&,tsu&)``` may be doing different things right? The first one will use move semantics and the second will use the explicit declaration. The end result is the same but still weird that there are essentially two different versions of swap for tsu. So i guess this PR fixes this by only allowing the first one.

pfeatherstone:
Added (conditional) nothrow specifiers everywhere. This helps chip away at assembly code. Also, some STL algorithms are more optimal if move operators are nothrow. So if you use ```dlib::type_safe_union``` with algorithms or STL containers, you might get slightly and theoretically more optimal

davisking:
Na we can't remove the `a.swap(b)` version.  There are totally people using that I bet.  It's pretty common.  Like even STL containers like `std::vector` have `.swap()` methods like this :)

pfeatherstone:
How about I make a.swap(b) call std::swap(a,b) ?

pfeatherstone:
Yeah that works

davisking:
> How about I make a.swap(b) call std::swap(a,b) ? 

Yeah that's fine 👍

pfeatherstone:
Appveyor is complaining about python stuff again... Failing on some import. Do we still need Appveyor now that dlib uses lovely Github Actions?

pfeatherstone:
Actually, the explicit definition of `type_safe_union::swap` allowed for a couple optimizations in the cases where the underlying types were the same or one of the types was empty. In both cases, we don't create another temporary. So i've added `.swap()` and `swap(a,b)` back in, but cleaned things up a bit. I've kept all the noexcept stuff. I think that's important.

pfeatherstone:
@davisking Do you mind having another look? Also, waddup with appveyor?

pfeatherstone:
As a separate note, and this might be the subject of another PR, it turns out the O(1) jump-table style visitor pattern implemented is actually really poor in performance, despite its elegance.  Watch https://www.youtube.com/watch?v=va9I2qivBOA and read https://mpark.github.io/programming/2019/01/22/variant-visitation-v2/ for background. This might need fixing. However, if you really care about performance and you can afford to upgrade your compiler, use ```std::variant```. Otherwise use `type_safe_union` if it isn't in your hot path or you can't upgrade your compiler. Or use something else...

davisking:
> @davisking Do you mind having another look? Also, waddup with appveyor?

Looking now.  And yeah, appveyor has never been especially reliable.

davisking:
Nice, this looks good :)

PR #2616, opened 2022-06-30
Add threadpool to Python global optimization

arrufat:
Hi, I am trying to make the global optimization work with a thread pool in Python. I thought it was going to be easy.

However, I think I am hitting one of the limitations of Python and the GIL. With the proposed modification, I can't make concurrent calls to the function. Is there a way around this, or it's just the wait it is, and that's why dlib didn't have the bindings for those overloads in the first place?

Note: adding a lock makes it work (but it defeats the purpose, haha):
```c++
std::mutex mutex;
auto func = [&](const matrix<double,0,1>& x)
{
    std::lock_guard<std::mutex> lock(mutex);
    return call_func(f, x);
};
```

Here's the Python code I am trying to run:

```python
from dlib import find_min_global
from math import sin, cos, exp, sqrt, pi
from time import sleep


def holder_table(x0, x1):
    sleep(0.1)
    return -abs(sin(x0) * cos(x1) * exp(abs(1 - sqrt(x0 * x0 + x1 * x1) / pi)))


x, y = find_min_global(holder_table, [-10, -10], [10, 10], 80, num_threads=4)

print(x, y)
```

Thanks in advance.

### DIFF
--- a/tools/python/src/global_optimization.cpp
+++ b/tools/python/src/global_optimization.cpp
@@ -115,20 +115,33 @@ py::tuple py_find_max_global (
     py::list bound2,
     py::list is_integer_variable,
     unsigned long num_function_calls,
-    double solver_epsilon = 0
+    double solver_epsilon = 0,
+    unsigned long num_threads = 1
 )
 {
     DLIB_CASSERT(len(bound1) == len(bound2));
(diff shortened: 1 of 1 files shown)


arrufat:
That's what I thought... thank you so much for confirming it. :)

PR #2617, opened 2022-07-01
Use chrono in console progress indicator and add finish method

arrufat:
Hi, I've updated a little bit the `console_progress_indicator`.

Since C++11 is a minimal requirement for dlib, it now uses `std::chrono` for all time manipulations (I really like `std::chrono`).
I've also added a `finish` method that will print the completed progress and the ellapsed time.

While running (subsequent calls to `print_status`):
```
Progress: 51/1000 (5.10%). Time remaining: 1min 34s.
```

After calling `finish`:
```
Progress: 1000/1000 (100.00%). Time elapsed: 1min 39s.
```
This `finish` method is meant to be called after a loop so that you don't need to remember the estimated time the first call to `print_progress` printed.

I hope you like it :)

### DIFF
--- a/dlib/console_progress_indicator.h
+++ b/dlib/console_progress_indicator.h
@@ -93,7 +93,7 @@ namespace dlib
         ) const;
         /*!
             ensures
-                - This objects prints the completed progress and the ellapsed time to out.
+                - This objects prints the completed progress and the elapsed time to out.
                   It is meant to be called after the loop we are training the progress of.
         !*/
 
@@ -236,7 +236,7 @@ namespace dlib
             const auto hours = std::chrono::duration_cast<std::chrono::hours>(delta_t);
             const auto minutes = std::chrono::duration_cast<std::chrono::minutes>(delta_t) - hours;
             const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(delta_t) - hours - minutes;
-            out << "Time ellapsed: ";
+            out << "Time elapsed: ";
             if (delta_t >= std::chrono::hours(1))
                 out << hours.count() << "h ";
             if (delta_t >= std::chrono::minutes(1))
--- a/dlib/console_progress_indicator.h
+++ b/dlib/console_progress_indicator.h
@@ -220,32 +220,32 @@ namespace dlib
         std::ostream& out
     ) const
     {
-            const auto oldflags = out.flags();
-            out.setf(std::ios::fixed,std::ios::floatfield);
-            std::streamsize ss;
+        const auto oldflags = out.flags();
+        out.setf(std::ios::fixed,std::ios::floatfield);
+        std::streamsize ss;
 
-            // adapt the precision based on whether the target val is an integer
-            if (std::trunc(target_val) == target_val)
-                ss = out.precision(0);
-            else
-                ss = out.precision(2);
+        // adapt the precision based on whether the target val is an integer
+        if (std::trunc(target_val) == target_val)
+            ss = out.precision(0);
+        else
+            ss = out.precision(2);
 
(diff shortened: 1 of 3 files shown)


arrufat:
I've made the `print_status` use `std::clog` by default (stderr) and the `finish` to use `std::cout`.
The main reasoning is that you can redirect the output to a file (I like using `tee`), and you'll only get the elapsed time.

dlib/console_progress_indicator.h
```diff
@@ -211,6 +213,41 @@ namespace dlib
...
     }
 
+// ----------------------------------------------------------------------------------------
+
+    void console_progress_indicator::
+    finish (
+        std::ostream& out
+    ) const
+    {
+            const auto oldflags = out.flags();
```
davisking, line 223: Fix up the tabbing here.

dlib/console_progress_indicator.h
```diff
@@ -166,21 +174,15 @@ namespace dlib
...
-            }
+            const auto hours = std::chrono::duration_cast<std::chrono::hours>(rem_time);
+            const auto minutes = std::chrono::duration_cast<std::chrono::minutes>(rem_time) - hours;
+            const auto seconds = std::chrono::duration_cast<std::chrono::seconds>(rem_time) - hours - minutes;
+            out << "Time remaining: ";
+            if (rem_time >= std::chrono::hours(1))
+                out << hours.count() << "h ";
+            if (rem_time >= std::chrono::minutes(1))
+                out << minutes.count() << "min ";
+            out << seconds.count() << "s.                \r" << std::flush;
```
davisking, line 185: Wouldn't you rather have the single floating point value printed?  Rather than something like "1h 5m 8s"?
E.g. something like this
```
            using hours = duration<double, std::ratio<60 * 60>>;
            using minutes = duration<double, std::ratio<60>>;
            using seconds = duration<double>;
            if (rem_time < minutes(1))
            {
                ss = out.precision(0);
                out << "Time remaining: " << duration_cast<seconds>(rem_time).count() << " seconds.                 \r" << std::flush;
            }
            else if (rem_time < hours(1))
            {
                ss = out.precision(2);
                out << "Time remaining: " << duration_cast<minutes>(rem_time).count() << " minutes.                 \r" << std::flush;
            }
            else 
            {
                ss = out.precision(2);
                out << "Time remaining: " << duration_cast<hours>(rem_time).count() << " hours.                 \r" << std::flush;
            }   
```

arrufat: Honestly, “1h 5min 8s” speaks more to me than “1.09 hours”, that's why I made the change. But if you prefer the floating-point format, I'll revert the change :)

I was also thinking about this format “1 hour 5 minutes 8 seconds”, but it's more verbose and needs singular/plural handling. But again, if you prefer it, I'll do it. Whatever you like the most :)

dlib/console_progress_indicator.h
arrufat: Done, sorry...

dlib/console_progress_indicator.h
davisking: Na that's fine. Not sure what my preference is. Was just not sure you knew about the option to use double with a duration :)

arrufat: Ah, I got it, yes, I was aware of this thing, I did the “h min s” format on purpose :)
I remember watching this a couple of years ago and being completely fascinated by the chrono library:
[CppCon 2016: Howard Hinnant “A ＜chrono＞ Tutorial"](https://www.youtube.com/watch?v=P32hvk8b13M)

davisking:
Yeah, nice, I like it :D

PR #2619, opened 2022-07-04
Update libjpeg to version 9e

### LINKED ISSUE #2618: Security vulnerabilities in libjpeg 8d
Libjpeg version 8d (15-Jan-2012) used in dlib has multiple security issues, for example:

https://www.cve.org/CVERecord?id=CVE-2020-14152
https://www.cve.org/CVERecord?id=CVE-2020-14153

Version used: https://github.com/davisking/dlib/blob/master/dlib/external/libjpeg/jversion.h

## Expected Action

Please update libjpeg to version 9d or newer.

* **Version**: master (2022-07-04)
* **Where did you get dlib**:  github
* **Platform**: all
* **Compiler**: -

arrufat:
Fixes #2618


arrufat:
I was wondering if we could/should replace [libjpeg](https://ijg.org/) with [libjpeg-turbo](https://www.libjpeg-turbo.org/).

Arch Linux, for example, doesn't even ship with libjpeg, but libjpegturbo instead.

Citing their webside:

> _libjpeg-turbo is a JPEG image codec that uses SIMD instructions (MMX, SSE2, AVX2, Neon, AltiVec) to accelerate baseline JPEG compression and decompression on x86, x86-64, Arm, and PowerPC systems, as well as progressive JPEG compression on x86 and x86-64 systems. On such systems, libjpeg-turbo is generally 2-6x as fast as [libjpeg](http://www.ijg.org/), all else being equal. On other types of systems, libjpeg-turbo can still outperform libjpeg by a significant amount, by virtue of its highly-optimized Huffman coding routines. In many cases, the performance of libjpeg-turbo rivals that of proprietary high-speed JPEG codecs._

davisking:
> I was wondering if we could/should replace [libjpeg](https://ijg.org/) with [libjpeg-turbo](https://www.libjpeg-turbo.org/).
> 
> Arch Linux, for example, doesn't even ship with libjpeg, but libjpegturbo instead.
> 
> Citing their webside:
> 
> > _libjpeg-turbo is a JPEG image codec that uses SIMD instructions (MMX, SSE2, AVX2, Neon, AltiVec) to accelerate baseline JPEG compression and decompression on x86, x86-64, Arm, and PowerPC systems, as well as progressive JPEG compression on x86 and x86-64 systems. On such systems, libjpeg-turbo is generally 2-6x as fast as [libjpeg](http://www.ijg.org/), all else being equal. On other types of systems, libjpeg-turbo can still outperform libjpeg by a significant amount, by virtue of its highly-optimized Huffman coding routines. In many cases, the performance of libjpeg-turbo rivals that of proprietary high-speed JPEG codecs._

Yeah that's fine so long as it builds on every platform without setup or any odd complication.  The only reason this stuff is in external is to make users who don't know how to install a system copy of libjpeg.  In particular this is a common problem for windows developers given the mess it is of having libraries installed on windows.  Seems like it's probably a drop in replacement though?

arrufat:
Ok, I will try in a different PR, and we can see if that works on GH Actions.

arrufat:
By the way, in case you missed it, it's ready for review :)

davisking:
Nice, thanks, looks good.  Appveyor is complaining but that must be the recent change to type_safe_union :|

davisking:
> Nice, thanks, looks good. Appveyor is complaining but that must be the recent change to type_safe_union :|

Yeah it's just visual studio 2015.  It's probably time to stop building against that old version of the compiler anyway.

arrufat:
Where is the Travis configuration for Windows? I can only see the setup for Linux in https://github.com/davisking/dlib/blob/master/.travis.yml.

davisking:
There is no travis config for windows.  Which is why I had appveyor setup, since they had windows support.  I'm going to delete the travis config files from the repo though.  That stuff hasn't worked in months.

arrufat:
Ah, somehow I got confused by those two... 😅

davisking:
I just removed appveyor too. Now we are 100% GitHub actions :)

arrufat:
Yes, I saw that, however, it still appears in the checks, and it's still failing… I've never used appveyor, but it looks you need to deregister dlib or something.

arrufat:
Just did that, thank you :)

PR #2621, opened 2022-07-06
Adding backport of bessel functions + window functions

pfeatherstone:
This is generally useful, but also this is laying down the foundation for stuff like filter design, computing STFTs and maybe who knows, some DSP.


dlib/math/details/bessel.h
```diff
@@ -0,0 +1,470 @@
...
+#include <cmath>
+#include <limits>
+#include <stdexcept>
+
+namespace dlib
+{
+    namespace detail
+    {
+        template <typename R>
+        void bessel_ik(R nu, R x, R& Inu, R& Knu, R& Ipnu, R& Kpnu)
```
davisking, line 15: Where is this code from?  It's got some license to it presumably.

pfeatherstone: It's pretty much copy pasted from libstdc++

pfeatherstone: Hmm, the software license of libstdc++ is GPL. Should i copy-paste from Boost instead? That has friendlier license.

davisking: Yeah definitely don't copy stuff from other libraries unless those also use the boost software license.  And then in that case also preserve the copyright notice on it.

Although you are using this stuff right?  And it's unreasonable to use boost or c++17 to get the functions?

dlib/math/constants.h
```diff
@@ -0,0 +1,104 @@
...
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_MATH_CONSTANTS
+#define DLIB_MATH_CONSTANTS
+
+namespace dlib
+{
+    template<typename R>
+    struct numeric_constants
+    {
+        static R pi() noexcept
```
davisking, line 11: There is already dlib/numeric_constants.h

dlib/math/details/bessel.h
pfeatherstone: It is reasonable to use those but not if i want stuff self-contained in dlib. I want to add STFT functions to dlib. That requires window functions, which require bessel functions. Ideally I would like to do STFTs in dlib without having to add C++17 or boost as firm dependencies to dlib.

pfeatherstone: Weirdly, i don't think libc++ supports special math functions yet. They are 5 years behind

dlib/math/constants.h
pfeatherstone: Ok latest commit addresses this.

pfeatherstone:
The boost implementation is unbelievably complicated. Sod it, would it be agreeable to add PRs for filter design and STFTs and have `__cpp_lib_math_special_functions` as a contract pre-condition?

pfeatherstone:
Looked at scipy implementation, it's Fortran...

dlib/math/details/bessel.h
davisking: > Also, if you so much as breath on 1 header file in boost, you've accidentally included 42000 other headers and now need to link to at least 1 boost shared/static library.

Ha, yeah I know this pain of boost too 😅

davisking:
> The boost implementation is unbelievably complicated. Sod it, would it be agreeable to add PRs for filter design and STFTs and have `__cpp_lib_math_special_functions` as a contract pre-condition?

Eh, I don't want to have stuff that doesn't always work. Dlib is all about everything always working regardless of platform. 

Nothing stopping you from doing this in another repo though. Dlib repo isn't special at the end of the day. 

At some point I'll jump dlib to needing c++17 too. I usually make sure dlib works with compilers from the last 6 or 7 years. Which is coming up soonish for c++17.

pfeatherstone:
Not sure. Libc++ doesn't support special math functions so clang on macOS wouldn't work. Some compilers don't fully support std::filesystem which is c++17. So I don't think c++17 is universally supported yet.

pfeatherstone:
I'll crack on with stripping boost but my god they needlessly make things complicated. Like every single function is templated with a policy, like even something as trivial sign().

pfeatherstone:
There you go, stripping boost only requires 1300 lines of code to implement two math functions. Nice and succinct in typical boost fashion.

dlib/math/kaiser.h
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2006  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_MATH_KAISER
+#define DLIB_MATH_KAISER
+
+#include "bessel.h"
+
+namespace dlib
+{
```
davisking, line 9: Can you add docs to the functions in this header?

pfeatherstone:
Don't know what's going on but the Python/windows build configuration always fails with:
```
python setup.py bdist_wheel did not run successfully.
  exit code: 1
  [6 lines of output]
  usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
     or: setup.py --help [cmd1 cmd2 ...]
     or: setup.py --help-commands
     or: setup.py cmd --help
  error: invalid command 'bdist_wheel'
```

davisking:
> Don't know what's going on but the Python/windows build configuration always fails with:
> 
> ```
> python setup.py bdist_wheel did not run successfully.
>   exit code: 1
>   [6 lines of output]
>   usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
>      or: setup.py --help [cmd1 cmd2 ...]
>      or: setup.py --help-commands
>      or: setup.py cmd --help
>   error: invalid command 'bdist_wheel'
> ```

Yeah it's on master too.  Was just looking at that.  Not sure what the deal is.

pfeatherstone:
Otherwise are you happy with the code ?

pfeatherstone:
Sorry i modified the API. I've opted for using strong types and function overloading for kaiser() instead of different function signatures. I think it's clearer and the user has to stop and think a bit about how to call the function correctly using the right strong type decorators. I think it's less error-prone.

dlib/math/windows.h
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    template<typename R>
+    inline R kaiser(index_t i, window_length N, beta_t beta, periodic_t)
+    /*!
+        This computes the kaiser window function or kaiser-bessel window function.
+        See https://en.wikipedia.org/wiki/Kaiser_window
+        This variant is a short-cut for computing a window function and storing it
+        in an array of size N where 0 <= i < N is the array index.
+        This is the periodic version.
```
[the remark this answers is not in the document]
pfeatherstone: The link https://en.wikipedia.org/wiki/Window_function#Symmetry explains it quite well I think. It follows the same nomenclature as Matlab, Scipy and probably most DSP libraries.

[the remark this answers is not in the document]
pfeatherstone: I'll need to add at some point overloads of these functions for fillings arrays, vectors and matrices. But I don't think those overloads belong in the `math` directory. Otherwise you're polluting the math headers with STL container headers and the dlib matrix stuff which pulls in quite a lot of stuff.

pfeatherstone:
Ok I think I have everything I need now. So if unit tests pass, we're good to go. For those who are interested, next PR will be STFT and ISTFT. This is useful generally in DSP and audio processing. But it's also useful in blind source separation algorithms like ILRMA and MNMF which use traditional machine learning algorithms operating on STFT domain to blindly separate channels which I think @davisking will enjoy.

PR #2622, opened 2022-07-07
Replace libjpeg with libjpeg-turbo v2.1.3

### LINKED ISSUE #2618: Security vulnerabilities in libjpeg 8d

arrufat:
Let's see if this builds everywhere.

Fixes #2618 

I am uncertain whether this is worth it, though, if we don't enable the SIMD flags…


arrufat:
Yeah, I don't want to be bothered with this… sorry :sweat_smile:

arrufat:
I mean, it works on Linux, but I don't have a Windows machine to test it, which is the main intended audience for this change, so I don't think I can maintain it if there are problems. #2619 is just a version update.

davisking:
> Yeah, I don't want to be bothered with this… sorry sweat_smile

Yeah no worries.  Getting stuff to work everywhere is a pain :sweat_smile: I'm pretty fine with saying people who want something better than this fallback should use whatever it is they want that is better for them.

PR #2623, opened 2022-07-13
Fix typo in the self-supervised learning example

arrufat:
A small typo fix to trigger the CI.


arrufat:
Right, appveyor is no longer there anymore. I am not sure that comma should go there… It feels like it doesn't, but maybe I'm wrong.

davisking:
Seems like it shouldn't be there to me too 😁

PR #2627, opened 2022-07-20
stft(), istft(), stftr() and istftr()

pfeatherstone:
Voila


dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,155 @@ namespace dlib
                   inverse transformation.  
     !*/
 
+// ----------------------------------------------------------------------------------------
+
+    template <typename EXP>
+    auto stft (
```
davisking, line 237:
```suggestion
    typename EXP::matrix_type stft (
```
Or something if this isn't right to say what the type is in some way.

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,155 @@ namespace dlib
...
+        std::size_t hoplen
+    );
+    /*!
+        requires
+            - is_vector(signal) == true, i.e. signal has rank 1
+            - signal.size() >= wlen
+            - fftsize >= wlen
+            - wlen >= hoplen
+            - EXP::type is a floating point type (float, double or long double), real or complex
+
```
davisking, line 251:
```suggestion
```

dlib/matrix/matrix_fft_abstract.h
davisking: Although I guess this is possibly confusing since the other template arguments of `matrix` might be set differently.  Maybe the docs you have are fine.  At least make this return like `matrix` or something though.   This file doesn't have to compile, just be maximally informative to the reader.

dlib/test/fft.cpp
```diff
@@ -550,7 +551,140 @@ namespace
...
+
+    template<typename R1, typename R2>
+    std::complex<R1> complex_cast(const std::complex<R2> &c) {
+        return { static_cast<R1>(c.real()), static_cast<R1>(c.imag()) };
+    }
+
+    template<typename R>
+    void test_stft()
+    {
+        constexpr R tol = std::is_same<R,float>::value ? 1e-4 : 1e-7;
```
davisking, line 563: Can these tol values be smaller?  The values in `STFT_FFT_128_WLEN_64_TONE_512` are already kinda small in general.  Would be cool if the `tol` was lower.  Like this wouldn't catch if all the really small values in `STFT_FFT_128_WLEN_64_TONE_512` were being matched against similarly really small outputs from the dlib code.

davisking:
> @davisking Is good ?

Yeah looking cool :)

dlib/matrix/matrix_fft_abstract.h
pfeatherstone: Yeah no worries. Wouldn't it be better if we just added the docs to the main header and forget about the abstract files? When Ctrl+clicking through an IDE, you never see the abstract files (at least I don't), so having the docs near the code is actually more useful. But we should be consistent with the rest of the library...

dlib/test/fft.cpp
pfeatherstone: Yeah totally, I need to figure out what the smallest values are.

pfeatherstone:
@davisking I think we should probably support 2D input, for example stereo audio channel. But then the STFT is 3D and of course `matrix` is 2D. Maybe the API is fine for now. If people want to STFT multiple audio channels, just loop over the channels and concatenate the STFTs in a `std::vector`. Any opinions or advice?

davisking:
> @davisking I think we should probably support 2D input, for example stereo audio channel. But then the STFT is 3D and of course `matrix` is 2D. Maybe the API is fine for now. If people want to STFT multiple audio channels, just loop over the channels and concatenate the STFTs in a `std::vector`. Any opinions or advice?

Na, I don't any different ideas about it :shrug:

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,151 @@ namespace dlib
...
+            - wlen >= hoplen
+            - EXP::type is a floating point type (float, double or long double), real or complex
+        ensures
+            - Performs a Short-Time-Fourier-Transform (STFT) on 1D data.
+            - Returns a matrix D where first dimension correponds to time and second dimension corresponds to frequency.
+            - Dimensions of D are:
+                - D.nr() == (signal.size() + hoplen) / hoplen and corresponds to the number of time frames.
+                - D.nc() == fftsize
+            - The type of D is add_complex_t<EXP::type>
+            - Each time frame t (equivalently, each row t) is centered on t*hoplen
```
davisking, line 258:
```suggestion
            - Each time frame t (equivalently, each row t) is centered on signal(t*hoplen)
```
?

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,151 @@ namespace dlib
...
+    matrix<complex_type> stft (
+        const std::vector<T, Alloc>& signal,
+        WindowType w,
+        std::size_t fftsize,
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        ensures
+            - This is a shortcut to calling stft(dlib::mat(signal, w, fftsize, wlen, hoplen)
```
davisking, line 280:
```suggestion
            - This is a shortcut to calling stft(dlib::mat(signal, w, fftsize, wlen, hoplen))
```

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,151 @@ namespace dlib
                   inverse transformation.  
     !*/
 
+// ----------------------------------------------------------------------------------------
+
+    template <typename EXP>
+    matrix<complex_type> stft (
+        const matrix_exp<EXP>& signal,
+        WindowType w,
```
davisking, line 239: Can you put the `WindowType` enum in this file too?  The abstract docs file shouldn't refer to anything that isn't in the docs.  Since otherwise people can't know about it by reading just the docs.

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,151 @@ namespace dlib
...
+            - wlen >= hoplen
+            - EXP::type is a floating point type (float, double or long double) and must be real
+        ensures
+            - Performs a real Short-Time-Fourier-Transform (STFTr) on 1D data.
+            - Returns a matrix D where first dimension correponds to time and second dimension corresponds to frequency.
+            - Dimensions of D are:
+                - D.nr() == (signal.size() + hoplen) / hoplen and corresponds to the number of time frames.
+                - D.nc() == fftsize/2 + 1
+            - The type of D is add_complex_t<EXP::type>
+            - Each time frame t (equivalently each row t) is centered on t*hoplen
```
davisking, line 331:
```suggestion
            - Each time frame t (equivalently each row t) is centered on signal(t*hoplen)
```

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,151 @@ namespace dlib
...
+    matrix<complex_type> stftr (
+        const std::vector<T, Alloc>& signal,
+        WindowType w,
+        std::size_t fftsize,
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        ensures
+            - This is a shortcut to calling istft(dlib::mat(signal, w, wlen, hoplen)
```
davisking, line 353:
```suggestion
            - This is a shortcut to calling istft(dlib::mat(signal, w, wlen, hoplen))
```

dlib/matrix/matrix_windows.h
```diff
@@ -0,0 +1,124 @@
...
+        {DLIB_ASSERT(is_vector(m_), "matrix expression must be a vector");}
+
+        beta_t beta;
+        WindowSymmetry t;
+
+        const static long cost = M::cost + 7;
+        typedef type const_ret_type;
+        const_ret_type apply(long r, long c) const
+        {
+            return kaiser<R>(std::size_t(r*this->m.nc()+c), (std::size_t)this->m.size(), beta, t) * this->m(r,c);
```
davisking, line 71: Did you get warnings when not using these `size_t` casts?  Those shouldn't be needed.  Like this ought to do the right thing:
```suggestion
            return kaiser<R>(r*this->m.nc()+c, this->m.size(), beta, t) * this->m(r,c);
```

dlib/matrix/matrix_windows.h
```diff
@@ -0,0 +1,124 @@
...
+        WindowType w,
+        WindowSymmetry type,
+        window_args args
+    )
+    {
+        using op = op_window<EXP>;
+        return matrix_op<op>(op(m.ref(), w, type, args));
+    }
+
+// ----------------------------------------------------------------------------------------
```
davisking, line 121: Can you add an abstract docs file for the functions in this file too?

dlib/matrix/matrix_windows.h
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    template <typename EXP>
+    const matrix_op<op_window<EXP> > window (
+        const matrix_exp<EXP>& m,
+        WindowType w,
+        WindowSymmetry type,
+        window_args args
+    )
+    {
```
davisking, line 116: I'm not super sold on the generic `window()` function.  Why provide that kind of indirection?  Someone could just call `kaiser()` or `hann()` directly.  Then stuff like `window_args` isn't needed, which is only used by some but not all functions.

dlib/math/windows.h
```diff
@@ -0,0 +1,264 @@
...
+        switch(w)
+        {
+            case HANN:              return hann<R>(i, N, type);
+            case BLACKMAN:          return blackman<R>(i, N, type);
+            case BLACKMAN_NUTTALL:  return blackman_nuttall<R>(i, N, type);
+            case BLACKMAN_HARRIS:   return blackman_harris<R>(i, N, type);
+            case BLACKMAN_HARRIS7:  return blackman_harris7<R>(i, N, type);
+            case KAISER:            return kaiser<R>(i, N, args.beta, type);
+        }
+        return R{};
```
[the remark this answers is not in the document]
davisking: Although I'm not super sold on this `window()` function.  Isn't is an unnecessary level of indirection?  Like why can't code just call the function they want to call directly?  That will avoid any runtime expense from a `switch` and also makes the interface more clear since there aren't arguments that are sometimes unused and sometimes applicable.

dlib/math/windows.h
```diff
@@ -0,0 +1,264 @@
...
+               0.21812 * std::cos(2*phi) -
+               0.06592 * std::cos(3*phi) +
+               0.01081 * std::cos(4*phi) -
+               0.00077 * std::cos(5*phi) +
+               0.00001 * std::cos(6*phi);
+    }
+
+    // ----------------------------------------------------------------------------------------
+
+    enum WindowType
```
davisking, line 229:
```suggestion
    enum window_type
```
Since everything in dlib is snake_case.

dlib/matrix/matrix_fft_abstract.h
davisking: It depends.  If the main header is sufficiently trivial (i.e. there isn't that much code in it) then it's fine.  Since someone reading it is seeing mostly the docs.   But for headers that are full of code and other implementation details it's real distracting to readers.  They just scroll around and are like "what should I look at?  what are the public functions?  What functions are even provided?"

And yeah, we should be consistent with the rest of the library.  But really the main thing is documentation readability.  Like all the matrix expression code is real unreadable to people.  For example, the list of trig functions on matrices are in the docs like this
```
    const matrix_exp cos (
        const matrix_exp& m
    );
    /*!
        requires
            - matrix_exp::type == float, double, or long double 
        ensures
            - returns a matrix R such that:
                - R::type == the same type that was in m
                - R has the same dimensions as m
                - for all valid r and c:
                  R(r,c) == std::cos(m(r,c)) 
    !*/
```
rather than as the zany preprocessor macros and expression template classes that they actually are.

So for example, it's fine if there is no abstract header if the real header is sufficiently clean and easy to read as docs.  Like if the functions are just so short and simple they are not distracting.  Or if the code is organized with prototypes at the top of the header and then all the implementation details shoved down at the bottom past some barrier that clearly indicates that only implementation details users shouldn't look at are below it.  Although if you are going to do that kind of thing you might as well just put it in two files (the normal one and the abstract one) since then you don't have to bend over backwards trying to figure out how to split it up like that.  Which is sometimes easy and sometimes tricky.

You can't find the docs in your IDE?  IDK, you should configure your IDE slightly differently :) :shrug:

dlib/test/math.cpp
```diff
@@ -0,0 +1,497 @@
...
+            0.3635733 , 0.33677071, 0.31073847, 0.28554985, 0.2612707 ,
+            0.23795919, 0.21566568, 0.19443265, 0.17429461, 0.15527819,
+            0.13740224, 0.12067792, 0.10510897, 0.09069194, 0.0774165 ,
+            0.06526578, 0.05421678, 0.0442408 , 0.03530386, 0.02736726
+    };
+
+    template<typename R>
+    void test_kaiser()
+    {
+        constexpr R tol = std::is_same<R,float>::value ? 1e-3 : 1e-7;
```
davisking, line 184: Can these tolerances be way lower too?  Presumably saving out the test data with more digits is all that's needed?

dlib/test/math.cpp
```diff
@@ -0,0 +1,497 @@
...
+            2.32086603e-01, 2.06107374e-01, 1.81288005e-01, 1.57726447e-01,
+            1.35515686e-01, 1.14743379e-01, 9.54915028e-02, 7.78360372e-02,
+            6.18466600e-02, 4.75864738e-02, 3.51117571e-02, 2.44717419e-02,
+            1.57084194e-02, 8.85637464e-03, 3.94264934e-03, 9.86635786e-04
+    };
+
+    template<typename R>
+    void test_hann()
+    {
+        constexpr R tol = std::is_same<R,float>::value ? 1e-3 : 1e-7;
```
davisking, line 250: Same here and for any other `tol` values.  I bet these can all be a lot smaller :)

davisking (commented):
Nice, this is all pretty cool.

dlib/matrix/matrix_windows.h
pfeatherstone: MSVC complained I think when turning warning into errors.

dlib/matrix/matrix_windows.h
pfeatherstone: Because it's need by `stft()` and all the others

pfeatherstone: You need to tell `stft()` which window to use.

dlib/test/math.cpp
pfeatherstone: Yes I can do that

dlib/math/windows.h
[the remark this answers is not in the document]
pfeatherstone: I mention it below, but it's need by `stft()`, `istft()` and the rest.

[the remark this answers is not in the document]
pfeatherstone: Much easier to tell the STFT functions which window to use via an enum

dlib/math/windows.h
```diff
@@ -0,0 +1,264 @@
...
+    {
+        HANN,
+        BLACKMAN,
+        BLACKMAN_NUTTALL,
+        BLACKMAN_HARRIS,
+        BLACKMAN_HARRIS7,
+        KAISER
+        //GAUSSIAN
```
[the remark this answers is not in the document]
pfeatherstone: I mainly kept this as a placeholder to remind me to add the gaussian window and to illustrate the need for `window_args`

dlib/math/windows.h
```diff
@@ -0,0 +1,264 @@
...
+    };
+
+    struct window_args
+    {
+        beta_t beta;
+        //sigma_t sigma; //for GAUSSIAN
```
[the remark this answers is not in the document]
pfeatherstone: Same as above

dlib/test/math.cpp
pfeatherstone: Can't for kaiser as the bessel function implementation is a bit naive and therefore less accurate than boost. The boost implementation is unbelievably complicated and unreadable. dlib's is much more explicit and readable. That comes at a cost though.

pfeatherstone:
@davisking So the main thing is that `stft` and co require a window type. In my view, the best way to specify that is via an enum. In which case you have to wrap all the other window function with `window()`. If you have a better idea then I'm all ears. We shouldn't encourage users to use `window()`, I agree, they should use the specific window functions. However, in my view it's required for the implementation details of `stft()`. And I think other users might benefit from it too hence why it's in the API.

pfeatherstone:
The other thing is the accuracy of `cyl_bessel_i()` which determines the accuracy of `kaiser()`. I've opted for a MUCH simpler implementation but it comes at the cost of some accuracy. IF however you compile with C++17 or a compiler that defines `__cpp_lib_math_special_functions`, then you use the standard library implementation for free and you get all the accuracy you need. I think this is better than porting 1500 lines of boost code that NOBODY understands except for maybe 2 people on this planet.

dlib/math/windows.h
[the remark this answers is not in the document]
davisking: This is adding complication though.  And for instance, the Kaiser beta is just defaulted in the use of this everywhere.  Which makes sense, since you don't want to force `window_args` into the API of all the functions that use windows.

The window switch statement stuff could be avoided by having methods that make and return window lambdas.  E.g.
```
auto kaiser_window(window_symmetry type, beta_t beta) {
    return [=](std::size_t i, std::size_t N) { return kaiser(i, N, type, beta); }
```
Then instead of passing the enum values to functions you pass the output of `kaiser_window()`.  This way the functions that build on top of window functions don't need to know anything about the parameters of the window functions.  It ought to be easier for the compiler to optimize too.  Like it would not surprise me if that `switch` statement resulted in slower execution.

More generally though, why not have functions that just build the window?  Like you call `kaiser_window(N, SYMMETRIC, beta)` and it returns a `dlib::matrix` holding the values.  Isn't that going to be way faster?  The matrix expression stuff will result in the values of the window function being recomputed over and over and they are expensive to compute.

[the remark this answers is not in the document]
pfeatherstone: I love your idea of passing a lambda! I'll do that later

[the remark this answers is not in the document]
pfeatherstone: I'm not using a lambda, but a struct with a templated `operator()` function. Otherwise you would have to template the function `kaiser_window()`. Bit overkill. If dlib used  C++14 or above, I could return a generic lambda...

pfeatherstone:
STFTs using kaiser window are now MUCH slower. I don't understand. I think it has something to do with the window now being a function object as opposed to enum. Or maybe I'm doing something dumb.

dlib/math/windows.h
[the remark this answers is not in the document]
davisking: Oh right, yeah, c++14.  Soon :)   

Are you still stuck with C++11 compilers?  I should ask around and see how close we are to being able to require C++14.  Maybe now is the time to require it.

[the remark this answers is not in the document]
pfeatherstone: I'm not. I can use c++14 everywhere and I'm pretty sure c++17 too. I would upgrade to 17 personally. I bet a whole bunch of stuff would get a lot simpler, compile faster and no doubt run faster.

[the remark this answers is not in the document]
pfeatherstone: I imagine a lot of the meta programming in dlib would get a lot better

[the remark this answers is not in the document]
davisking: Right. I'll ask around a bit and see.  C++17 is too new, but I think the time for C++14 is pretty much now.

dlib/matrix/matrix_fft.h
```diff
@@ -369,31 +386,142 @@ namespace dlib
...
+                a periodic Hann window which can be passed to the STFT family of functions.
+        !*/
+
+        template<typename R>
+        R operator()(std::size_t i, std::size_t N, type_t<R>) const
+        /*!
+            requires
+                - R is float, double, or long double
+        !*/
+        { return hann<R>(i, N, PERIODIC);};
```
davisking, line 428:
```suggestion
        double operator()(std::size_t i, std::size_t N) const
        { return hann(i, N, PERIODIC);};
```

How about this?   Like in general don't template on return type and just make everything return `double`.  Consider the hann code:
```
        const R phi = (2.0 * pi * i) / size;
        return 0.5 - 0.5 * std::cos(phi);
```
`(2.0 * pi * i) / size` is a `double`, which if `R` is `float` gets downcast to `float`, then the `float` version of `cos` runs on it and gives a less accurate result than it could have.  And then it gets converted back to `double` again by `0.5 *` before being cast back to `float` implicitly.   And `float` isn't faster than `double` on most CPUs.  So if this is all just `double` internally people will get more accurate results.  They can still stick it in a `matrix<float>` if they want certainly, but letting the cast to `float` only happen one at the very end will make the resulting value more accurate.  As well as simplifying the code here too.  Might even run faster since a bunch of float/double conversions would be removed.

dlib/math/windows.h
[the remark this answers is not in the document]
davisking: Yeah, I'm calling it now.  C++14 for dlib.  I need to update the CI, some cmake stuff a little, and the docs.  But C++14 is required for the next dlib release :)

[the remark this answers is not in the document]
pfeatherstone: Nice, you're gonna see a lot of PRs relating to modernising dlib. Generic lambdas here we come!

pfeatherstone:
I'll wait until master is updated with the C++14 stuff then I'll merge and this should pass the tests I think

dlib/math/windows.h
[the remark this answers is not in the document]
pfeatherstone: I wonder if a lot of the cmake stuff would get simpler if dlib used a more modern cmake version like 3.10 or above

pfeatherstone:
I've made a basic change to the cmake scripts to enable C++14. Hopefully the tests will pass. I've noticed though that dlib is using pretty old cmake code and maybe it's time to modernise it to maybe 3.10.

dlib/math/windows.h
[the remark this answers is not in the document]
davisking: Yeah it would.  I am hesitant to mess with what isn't broken though.  And most of the complication has nothing to do with cmake versions and is just the accumulation of all the little exceptions to people's broken systems I've accumulated over the years.

dlib/CMakeLists.txt
```diff
@@ -47,7 +47,15 @@ endif()
 
 # default to a Release build (except if CMAKE_BUILD_TYPE is set)
 include(cmake_utils/release_build_by_default)
-include(cmake_utils/use_cpp_11.cmake)
+
+function(enable_cpp14_for_target target_name)
+   set_target_properties(${target_name} PROPERTIES
+      CXX_STANDARD 17
```
arrufat, line 53: Shouldn't this be 14?

pfeatherstone: Yes you're right sorry. But this totally broke everything. So I might wait for David to commit all the 14 stuff then I'll merge

pfeatherstone: I do like cmake but i can understand why modern languages laugh at the C++ community because of our rubbish build tools. I wonder if meson or other ones are actually better, and if so, if they should get standardised just so that everyone is pushed to use to the same thing.

pfeatherstone: In my opinion, build tools and documentation are more important than the language itself, and the standards community should be putting 100% of their efforts towards improving build time, build tools and inline documentation

davisking: Na it's not cmake.  It's easy to have real clean cmake. It's all complicated because there are so many people with broken systems or compilers or whatever.  You can make it so simple.  But then someone posts "hey, I'm building on $some_weird_linux_distro_or_whatever and I'm getting this linker error saying libpng is missing some symbol".  Or they are complaining about something else that is actually broken on their system.  For instance, Anaconda includes these copies of x11 header files right, which for a lot of people end up in their compiler's include search path.  And Anaconda has helpfully edited those headers to be different from the real x11 headers, so if those get included instead the world breaks.   I could go on and on.

Anyway, those people complain to me though.  "Hey, dlib isn't building?!".  So I add some logic to the dlib cmake scripts to detect whatever is wrong with their system and work around it.  Like all that logic in there for testing if libpng is *really* there or not.

The only way this whole mess could be resolved was if applications *never ever* depended on libraries they are not themselves building.  And the building of those libraries would have to be controlled by one unified build system.  But that's a tall ask because everyone then turns around and says "hey, I hate the build times.  Can't the $(os, package system, whatever) provide precompiled binaries instead?" which is a desire 100% contradictory to resolving this problem given the nature of the One Definition Rule and the diversity of compile and build settings.

Then you top this all off with proprietary vendor libraries.  Like IntelMKL or anything CUDA related.  Which you can't build.  They have to be somehow pulled in as binaries because that's the only things those vendors will give out.  And those binaries have OS dependencies and even driver level dependencies. 

So IDK.  Any language that wants to do anything with hardware acceleration has to grapple with this.  I guess in the dream world there would be BLAS, LAPACK, and CUDA built into compilers by default.  Which is just saying that we would foist the problem of that stuff onto compiler and OS vendors and they would just deal with it.

davisking: Anyway, let me see about adjusting cmake settings to use C++14.  I'm doing that right now :)  I'm going to burn away some of the oldest cmake logic for old broken compilers too and see who complains.

pfeatherstone: > Na it's not cmake. It's easy to have real clean cmake. It's all complicated because there are so many people with broken systems or compilers or whatever. You can make it so simple. But then someone posts "hey, I'm building on $some_weird_linux_distro_or_whatever and I'm getting this linker error saying libpng is missing some symbol". Or they are complaining about something else that is actually broken on their system. For instance, Anaconda includes these copies of x11 header files right, which for a lot of people end up in their compiler's include search path. And Anaconda has helpfully edited those headers to be different from the real x11 headers, so if those get included instead the world breaks. I could go on and on.
> 
> Anyway, those people complain to me though. "Hey, dlib isn't building?!". So I add some logic to the dlib cmake scripts to detect whatever is wrong with their system and work around it. Like all that logic in there for testing if libpng is _really_ there or not.
> 
> The only way this whole mess could be resolved was if applications _never ever_ depended on libraries they are not themselves building. And the building of those libraries would have to be controlled by one unified build system. But that's a tall ask because everyone then turns around and says "hey, I hate the build times. Can't the $(os, package system, whatever) provide precompiled binaries instead?" which is a desire 100% contradictory to resolving this problem given the nature of the One Definition Rule and the diversity of compile and build settings.
> 
> Then you top this all off with proprietary vendor libraries. Like IntelMKL or anything CUDA related. Which you can't build. They have to be somehow pulled in as binaries because that's the only things those vendors will give out. And those binaries have OS dependencies and even driver level dependencies.
> 
> So IDK. Any language that wants to do anything with hardware acceleration has to grapple with this. I guess in the dream world there would be BLAS, LAPACK, and CUDA built into compilers by default. Which is just saying that we would foist the problem of that stuff onto compiler and OS vendors and they would just deal with it.

This just illustrates something is wrong with the language I think.

pfeatherstone:
@davisking I've merged master should be good to go

davisking:
Sweet. I'll look in a bit.

dlib/matrix/matrix_fft.h
```diff
@@ -263,6 +264,291 @@ namespace dlib
...
+            const std::size_t total_padding = wlen;
+            const std::size_t overlap       = wlen - hoplen;
+            const std::size_t nframes       = (signal.size() + total_padding - overlap) / hoplen;
+            matrix<C> stft = zeros_matrix<C>(nframes,
+                                             FFT_FUNC::freqsize(fftsize));
+            matrix<R> win(1,wlen);
+            for (std::size_t i = 0 ; i < wlen ; ++i)
+                win(0, i) = w(i, wlen);
+
+            /*! TODO: reduce extra buffers, e.g. padded !*/
```
davisking, line 336:
```suggestion
            // TODO: reduce extra buffers, e.g. padded
```
Is fine.  Only use the `/*! !*/` style for function and class contracts.  That style is special to some of the documentation generation code and is supposed to be a "hey this is a function/class contract" tag.

dlib/matrix/matrix_fft.h
```diff
@@ -263,6 +264,291 @@ namespace dlib
...
+    >
+    auto istftr (
+        const matrix_exp<EXP>& stft,
+        const WINDOW& w,
+        std::size_t wlen,
+        std::size_t hoplen
+    )
+    {
+        return details::istft_impl<R>(stft, w, wlen, hoplen, details::ifftr_func{});
+    }
```
davisking, line 550:
```suggestion
    template <
        typename EXP,
        typename WINDOW
    >
    auto istftr (
        const matrix_exp<EXP>& stft,
        const WINDOW& w,
        std::size_t wlen,
        std::size_t hoplen
    )
    {
        using R = remove_complex_t<typename EXP::type>;
        return details::istft_impl<R>(stft, w, wlen, hoplen, details::ifftr_func{});
    }
```
These things don't need to be template arguments.  It's simpler this way since these details don't need to be scoped somewhere client code can mess with.

dlib/matrix/matrix_fft.h
```diff
@@ -263,6 +264,291 @@ namespace dlib
...
+    >
+    auto istft (
+        const matrix_exp<EXP>& stft,
+        const WINDOW& w,
+        std::size_t wlen,
+        std::size_t hoplen
+    )
+    {
+        return details::istft_impl<T>(stft, w, wlen, hoplen, details::ifft_func{});
+    }
```
davisking, line 504:
```suggestion
    template <
        typename EXP,
        typename WINDOW
    >
    auto istft (
        const matrix_exp<EXP>& stft,
        const WINDOW& w,
        std::size_t wlen,
        std::size_t hoplen
    )
    {
        using T = typename EXP::type;
        return details::istft_impl<T>(stft, w, wlen, hoplen, details::ifft_func{});
    }
```
Since the client shouldn't be messing with `T`.

davisking (commented):
Looks nice :)

dlib/matrix/matrix_fft.h
```diff
@@ -263,6 +264,291 @@ namespace dlib
...
+
+    const auto hann_window = [](std::size_t i, std::size_t N) 
+    /*!
+        WHAT THIS OBJECT REPRESENTS
+            This returns a function object with signature double(size_t i, size_t N) which computes
+            a periodic Hann window which can be passed to the STFT family of functions.
+    !*/
+    {
+        return hann(i, N, PERIODIC);
+    };
```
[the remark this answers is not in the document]
pfeatherstone: Good question, in my head, passing a function as an argument to another function wouldn't work. But of course it would. I will try.

dlib/matrix/matrix_fft.h
pfeatherstone: Yep

dlib/matrix/matrix_fft.h
```diff
@@ -263,6 +264,291 @@ namespace dlib
...
+    /*!
+        WHAT THIS OBJECT REPRESENTS
+            This returns a function object with signature double(size_t i, size_t N) which computes
+            a periodic 7th order Blackman-Harris window which can be passed to the STFT family of functions.
+    !*/
+    {
+        return blackman_harris7(i, N, PERIODIC);
+    };
+
+    inline auto kaiser_window(beta_t beta)
```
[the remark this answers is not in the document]
pfeatherstone: I don't like this one because it makes it look like it's doing something different to the other functions. They are all doing the same thing, it just so happens kaiser requires an extra parameter, and others in the future could too, e.g. gauss window

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,151 @@ namespace dlib
...
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        requires
+            - is_vector(signal) == true, i.e. signal has rank 1
+            - signal.size() >= wlen
+            - fftsize >= wlen
+            - wlen >= hoplen
+            - EXP::type is a floating point type (float, double or long double), real or complex
```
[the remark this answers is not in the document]
pfeatherstone: yep agreed

dlib/test/fft.cpp
```diff
@@ -550,7 +551,149 @@ namespace
...
+
+        print_spinner();
+    }
+
+    template<typename R, typename WINDOW>
+    void test_random_stfts_impl(const WINDOW& w, std::size_t fftsize, std::size_t wlen)
+    {
+        constexpr R tol = std::is_same<R,float>::value ? 1e-5 : 1e-12;
+
+//        printf("%s - %i\n", __PRETTY_FUNCTION__, __LINE__);
```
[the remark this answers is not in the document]
pfeatherstone: sorry forgot to remove that

dlib/matrix/matrix_fft.h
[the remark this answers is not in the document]
pfeatherstone: Actually, I quite like the lambda version. Here's why. if you have a function like you suggest, it looks like the user is supposed to call it. But no, you shouldn't do that. You pass the function, as a function pointer, to the STFT functions. If you have a lambda object, then you're dealing with an object, a bit like an enum. Maybe it's clearer than you shouldn't call it yourself ?

dlib/matrix/matrix_fft.h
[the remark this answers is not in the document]
pfeatherstone: Could you have another think about this? In my opinion, they should all be tag types, not publicly callable things. And I'm not a fan of having some of them having different names that make them sound like factory functions, even if they are... That should be hidden to the user ideally.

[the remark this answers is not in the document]
pfeatherstone: Alternatively they are all called `make_xxx_window()`, most of them taking no args, except for kaiser which takes beta. Then the name is clear, the function is returning another function.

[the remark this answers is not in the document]
pfeatherstone: Yeah lets go with that.

pfeatherstone:
@davisking I've addressed your comments. Hopefully you like it.

dlib/matrix/matrix_fft.h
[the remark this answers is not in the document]
davisking: Yeah that's a good way :)

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,171 @@ namespace dlib
...
+        std::size_t fftsize,
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        requires
+            - is_vector(signal) == true, i.e. signal has rank 1
+            - signal.size() >= wlen
+            - w is a function object with signature double(size_t i, size_t wlen) that defines a PERIODIC window,
+              e.g. hann_window.
```
davisking, line 261:
```suggestion
              e.g. the output of make_hann().
```

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,171 @@ namespace dlib
...
+        const matrix_exp<EXP>& stft,
+        const WINDOW& w,
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        requires
+            - m has rank 2 where 1st dimension corresponds to time and second dimension corresponds to frequency
+            - w is a function object with signature double(size_t i, size_t wlen) that defines a PERIODIC window,
+              e.g. hann_window.
```
davisking, line 310:
```suggestion
              e.g. the output of make_hann().
```

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,171 @@ namespace dlib
...
+        std::size_t fftsize,
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        requires
+            - is_vector(signal) == true, i.e. signal has rank 1
+            - signal.size() >= wlen
+            - w is a function object with signature double(size_t i, size_t wlen) that defines a PERIODIC window,
+              e.g. hann_window.
```
davisking, line 338:
```suggestion
              e.g. the output of make_hann().
```

dlib/matrix/matrix_fft_abstract.h
```diff
@@ -231,6 +231,171 @@ namespace dlib
...
+        const matrix_exp<EXP>& stft,
+        const WINDOW& w,
+        std::size_t wlen,
+        std::size_t hoplen
+    );
+    /*!
+        requires
+            - m has rank 2 where 1st dimension corresponds to time and second dimension corresponds to frequency
+            - w is a function object with signature double(size_t i, size_t wlen) that defines a PERIODIC window,
+              e.g. hann_window.
```
davisking, line 387:
```suggestion
              e.g. the output of make_hann().
```

davisking (commented):
Yeah nice, looks good :)

VisionEp1:
@davisking doesn't seem to work on windows. error c2275
image

VisionEp1:
but i just compiled once, i will go ahead and check if its my setup or the code. just an FYI for now

VisionEp1:
or works with c++14 but not with c++17 or c++20.
pre this change dlib with c++17 was no issue at all

pfeatherstone:
I will have a look

pfeatherstone:
Maybe we need to update CI/CD to build with 14, 17, 20

pfeatherstone:
@VisionEp1 What's your system? Tried on Godbold with MSVC and it works: https://godbolt.org/z/qzcKEacdW

pfeatherstone:
I have a fix https://godbolt.org/z/TEhqPa37e

davisking:
Maybe you are using something older than Visual Studio 2019 @VisionEp1?  I just the other day changed the CI to use only VS 2019 and newer since github actions started warning me about the older setups they have being deprecated :|

PR #2628, opened 2022-07-22
YOLO improvements: Smoothed L1 loss, Focal Loss, Weighted Loss

arrufat:
Here there are some (backwards-compatible) for the YOLO layer:

- use the Smoothed L1 norm in the bounding box regression to stabilize the training.
- optional focal losses for the objectness and classifier.
- use the `truth.detection_confidence` as the positive weight for the classification loss.  This could be used to handle imbalance in the classifier.

By doing so, on an internal dataset, I was able to gain 1 mAP point.

I also added [Adaptive Training Sample Selection (ATSS)](https://arxiv.org/abs/1912.02424): when `iou_anchor_threshold` is set to 0, the threshold will be adjusted adaptively for each ground truth, according to how it matches the anchors. The motivation is:

- if an anchor matches really well (high IoU), the threshold will be high and other anchors won't be used. 
- if no anchor has a high IoU, then the threshold will be low, and more anchors will be used.

You are welcome to reject, of course… At least the code is now visible, in case people want to use it.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -3518,6 +3518,7 @@ namespace dlib
         double lambda_obj = 1.0;
         double lambda_box = 1.0;
         double lambda_cls = 1.0;
+        double gamma_obj = 0.0;
         double gamma_cls = 0.0;
 
     };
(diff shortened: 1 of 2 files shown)


davisking:
Nice, thanks for the PR :)

PR #2630, opened 2022-07-26
constexpr numeric constants

arrufat:
Just to be able to use them in `constexpr` functions.


pfeatherstone:
This is actually included in #2627

arrufat:
Oh I didn't notice, sorry.

davisking:
> Oh I didn't notice, sorry.

No problem :)

PR #2633, opened 2022-08-04
Modernize rounding and cast statements

arrufat:
I have some time to kill, and I want to modernize some rounding statements in dlib.

I think it won't break anything (I recently learned about `std::lround` and wanted to use it in dlib)

### DIFF
--- a/dlib/bigint/bigint_kernel_2.cpp
+++ b/dlib/bigint/bigint_kernel_2.cpp
@@ -1528,7 +1528,7 @@ namespace dlib
             const uint32 len = lhs->digits_used + rhs->digits_used;
             for (unsigned long i = 0; i < len; ++i)
             {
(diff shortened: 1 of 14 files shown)


dlib/bigint/bigint_kernel_2.cpp
```diff
@@ -1528,7 +1528,7 @@ namespace dlib
             const uint32 len = lhs->digits_used + rhs->digits_used;
             for (unsigned long i = 0; i < len; ++i)
             {
-                uint64 num1 = static_cast<uint64>(std::floor(a[i*2].real()+0.5));
+                uint64 num1 = static_cast<uint64>(std::round(a[i*2].real()));
```
arrufat, line 1531: I used `std::round` instead of `std::lround` here not to lose precision (`long` vs `uint64`)

davisking:
Nice, looks good :)

PR #2637, opened 2022-08-09
Bessel fix for MSVC

pfeatherstone:
This should fix the issue raised in #2627

### DIFF
--- a/dlib/math/bessel.h
+++ b/dlib/math/bessel.h
@@ -17,7 +17,8 @@ namespace dlib
             - works with C++11 onwards
     !*/
     {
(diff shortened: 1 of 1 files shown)


davisking:
Huh, yeah, so that works for you @VisionEp1?

davisking:
Well, let's merge it and see what people say.  Thanks for the fix @pfeatherstone  :)

VisionEp1:
srry was asleep :),
can confirm it works 
thanks again!

And sorry I couldn't provide the fix myself I never worked with the common type thing we use boost instead still...

davisking:
Sweet, no worries :)

PR #2641, opened 2022-08-13
Use only a fraction of labels for the multiclass SVM in SSL example

arrufat:
This change makes the self-supervised example more close to reality:
usually, only a fraction of the dataset is labeled, but we can harness
the whole dataset by using a self-supervised method and then train the
classifier using the fraction of labeled data.

Using 10% of the labels results in a test accuracy of 87%, compared to
the 89% we got when training the multiclass SVM with all labels.

I just added an option to change the fraction of labeled data, so that
users can experiment with it.

### DIFF
--- a/examples/dnn_self_supervised_learning_ex.cpp
+++ b/examples/dnn_self_supervised_learning_ex.cpp
@@ -377,7 +377,7 @@ try
         cout << "  error rate: " << num_wrong / static_cast<double>(num_right + num_wrong) << endl;
     };
 
-    // Using 10% of the training labels should result in a training and testing accuracies of
+    // Using 10% of the training labels should result in training and testing accuracies of
     // around 92% and 87%, respectively.
     // Had we used all labels to train the multiclass SVM classifier, we would have gotten a
     // training accuracy of around 93% and a testing accuracy of around 89%, instead.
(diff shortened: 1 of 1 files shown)


examples/dnn_self_supervised_learning_ex.cpp
```diff
@@ -347,7 +377,10 @@ try
         cout << "  error rate: " << num_wrong / static_cast<double>(num_right + num_wrong) << endl;
     };
 
-    // We should get a training accuracy of around 93% and a testing accuracy of around 89%.
+    // Using 10% of the training labels should result in a training and testing accuracies of
```
davisking, line 380:
```suggestion
    // Using 10% of the training labels should result in training and testing accuracies of
```

davisking:
Yeah nice.  This remind me, you should totally try http://dlib.net/ml.html#linear_manifold_regularizer with this.  That's one of my favorite semi-supervised methods.  You just give it a graph of samples along with a weight on each edge of the graph saying how similar two samples are.  It then gives you a linear transformation (so just a matrix) that you apply to all your samples.  That transformation is constructed such that it is the same as adding another term to the SVM objective function that penalizes learning a classifier that gives different outputs on samples connected in the graph.

Might be really great with this.  Like use the dot product between each unlabeled sample as the edge weight.  Maybe threshold it to 0 below some value.  You would want to have a bias towards making that `intrinsic_regularization_strength` as large as possible in the global optimizer part too.  So that it smears the labels out across the manifold implied by the graph as much as possible.

arrufat:
Oh great, thank you! I will definitely have a look at this.
Currently, I have some time, so I was going to add the multiclass SVM training part of this example into `dlib/svm/auto.h` as you suggested in https://github.com/davisking/dlib/pull/2460#issuecomment-971569505.

PR #2642, opened 2022-08-14
Add multiclass SVM trainer to svm/auto.h

arrufat:
As suggested in https://github.com/davisking/dlib/pull/2460#issuecomment-971569505, this PR adds the common following common use case for training multiclass SVM classifiers:

- normalize
- run cross validation to find the best setting
- train the final linear model

I am sure there is room for improvement, for instance, the type of the training samples `x`, which I set it to `matrix<float>` because that's what the neural networks return...
Maybe I could be generic in the type, or add an overload that converts `matrix<float>` to `matrix<double>` and just have that implementation?

As a side note, is there any reason why you prefer `typedef` over `using` syntax? I kind of like `using` more :P

### DIFF
--- a/dlib/svm/auto.cpp
+++ b/dlib/svm/auto.cpp
@@ -104,9 +104,9 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
-    normalized_function<multiclass_linear_decision_function<linear_kernel<matrix<float,0,1>>, unsigned long>>
+    normalized_function<multiclass_linear_decision_function<linear_kernel<matrix<double,0,1>>, unsigned long>>
     auto_train_multiclass_svm_linear_classifier (
-        std::vector<matrix<float,0,1>> x,
+        std::vector<matrix<double,0,1>> x,
         std::vector<unsigned long> y,
         const std::chrono::nanoseconds max_runtime,
         bool be_verbose
@@ -125,14 +125,14 @@ namespace dlib
 
         randomize_samples(x, y);
 
-        vector_normalizer<matrix<float,0,1>> normalizer;
+        vector_normalizer<matrix<double,0,1>> normalizer;
         // let the normalizer learn the mean and standard deviation of the samples
         normalizer.train(x);
         for (auto& samp : x)
             samp = normalizer(samp);
 
 
-        typedef linear_kernel<matrix<float,0,1>> kernel_type;
+        typedef linear_kernel<matrix<double,0,1>> kernel_type;
         normalized_function<multiclass_linear_decision_function<kernel_type, unsigned long>> df;
         df.normalizer = normalizer;
 
@@ -177,6 +177,27 @@ namespace dlib
         df.function = trainer.train(x, y);
 
         return df;
+    }
+
+    normalized_function<multiclass_linear_decision_function<linear_kernel<matrix<float,0,1>>, unsigned long>>
+    auto_train_multiclass_svm_linear_classifier (
+        const std::vector<matrix<float,0,1>>& x,
+        std::vector<unsigned long> y,
+        const std::chrono::nanoseconds max_runtime,
+        bool be_verbose
+    )
+    {
+        std::vector<matrix<double,0,1>> samples;
+        for (const auto& samp : x)
+            samples.push_back(matrix_cast<double>(samp));
+
+        auto df = auto_train_multiclass_svm_linear_classifier(samples, y, max_runtime, be_verbose);
+        normalized_function<multiclass_linear_decision_function<linear_kernel<matrix<float,0,1>>, unsigned long>> dfloat;
+        dfloat.normalizer.train(x);;
+        dfloat.function.labels = df.function.labels;
+        dfloat.function.weights = matrix_cast<float>(df.function.weights);
+        dfloat.function.b = matrix_cast<float>(df.function.b);
+        return dfloat;
     }
 }
 
--- a/dlib/svm/auto.cpp
+++ b/dlib/svm/auto.cpp
@@ -43,10 +43,10 @@ namespace dlib
             samp = normalizer(samp);
 
 
-        normalized_function<decision_function<radial_basis_kernel<matrix<double,0,1>>>> df;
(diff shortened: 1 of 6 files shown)


davisking:
Totally `using` is better.  It's just that `using` is a C++11 thing so wasn't available for a long time.

examples/dnn_self_supervised_learning_ex.cpp
```diff
@@ -325,35 +325,8 @@ try
...
-    const auto result = find_max_global(cross_validation_score, 1e-3, 1000, max_function_calls(50));
-    cout << "Best C: " << result.x(0) << endl;
-
-    // Proceed to train the SVM classifier with the best C.
-    svm_multiclass_linear_trainer<linear_kernel<matrix<float, 0, 1>>, unsigned long> trainer;
-    trainer.set_num_threads(std::thread::hardware_concurrency());
-    trainer.set_c(result.x(0));
-    cout << "Training Multiclass SVM..." << endl;
-    const auto df = trainer.train(features, training_labels);
+    const auto df = auto_train_multiclass_svm_linear_classifier(features, training_labels, std::chrono::minutes(5));
```
davisking, line 329: Does it really take 5 minutes?

arrufat: Yes, when using only 10% of the samples on 16 cores... maybe I should let it run longer?

davisking: Oh no, I was wondering if maybe 1 minute was enough :shrug: :D 

5 is fine though.

arrufat:
I've made the default implementation use a `matrix<double>` and added an overload for `matrix<float>` that just casts the inputs and outputs to be able to use the `matrix<double>` implementation.

dlib/svm/auto.cpp
```diff
@@ -101,6 +101,104 @@ namespace dlib
...
+        bool be_verbose
+    )
+    {
+        std::vector<matrix<double,0,1>> samples;
+        for (const auto& samp : x)
+            samples.push_back(matrix_cast<double>(samp));
+
+        auto df = auto_train_multiclass_svm_linear_classifier(samples, y, max_runtime, be_verbose);
+        normalized_function<multiclass_linear_decision_function<linear_kernel<matrix<float,0,1>>, unsigned long>> dfloat;
+        dfloat.normalizer.train(x);;
```
arrufat, line 196: I couldn't find a simple way to cast the normalizer weights... So, I am just recomputing it here. Even though it's quite fast, it would be nice not to do this computation twice...

arrufat: I couldn't find a simple way to cast the normalizer weights... So, I am just recomputing it here. Even though it's quite fast, it would be nice not to do this computation twice...

dlib/svm/auto_abstract.h
```diff
@@ -25,7 +25,52 @@ namespace dlib
...
+        const std::chrono::nanoseconds max_runtime,
+        bool be_verbose = true
+    );
+    /*!
+        requires
+            - is_learning_problem(x,y) == true
+            - y contains at least 6 examples of each class.
+        ensures
+            - This function is just an overload of the one defined immediately above.  It casts the
+              input data from double to float, calls the above function, and casts the results back
```
davisking, line 68:
```suggestion
              input data from float to double, calls the above function, and casts the results back
```

dlib/svm/auto.cpp
```diff
@@ -101,6 +97,101 @@ namespace dlib
...
+        std::vector<unsigned long> y,
+        const std::chrono::nanoseconds max_runtime,
+        bool be_verbose
+    )
+    {
+        const auto labels = select_all_distinct_labels(y);
+        for (const auto label : labels)
+        {
+            const auto num_samples = sum(mat(y) == label);
+            DLIB_CASSERT(num_samples >= 6,
```
davisking, line 115: Shouldn't these all be 3 instead of 6?  Since there are only 3 folds?

arrufat: right... copy paste...

davisking: Oh this is copied from what I did. :confused:

davisking: Oh but that other function uses 6 fold cross validation.  So it needs 6.  This one only needs 3 though :)

arrufat: I copied the assert only...

davisking: Yeah no worries.  I thought for a moment that somehow 3 fold CV needs 6 and was like "whaaaa???" lol

arrufat: yes, that's what I meant, that I copied the assert, without modifying the value properly, my bad, I will push the change.

davisking: Yeah, no problem :)

dlib/svm/auto_abstract.h
```diff
@@ -25,7 +25,52 @@ namespace dlib
...
+    auto_train_multiclass_svm_linear_classifier (
+        std::vector<matrix<double,0,1>> x,
+        std::vector<unsigned long> y,
+        const std::chrono::nanoseconds max_runtime,
+        bool be_verbose = true
+    );
+    /*!
+        requires
+            - is_learning_problem(x,y) == true
+            - y contains at least 6 examples of each class.
```
davisking, line 45:
```suggestion
            - y contains at least 3 examples of each class.
```

dlib/svm/auto_abstract.h
```diff
@@ -25,7 +25,52 @@ namespace dlib
...
+    auto_train_multiclass_svm_linear_classifier (
+        const std::vector<matrix<float,0,1>>& x,
+        std::vector<unsigned long> y,
+        const std::chrono::nanoseconds max_runtime,
+        bool be_verbose = true
+    );
+    /*!
+        requires
+            - is_learning_problem(x,y) == true
+            - y contains at least 6 examples of each class.
```
davisking, line 65:
```suggestion
            - y contains at least 3 examples of each class.
```

davisking:
Nice, thanks for another good PR :)

PR #2653, opened 2022-08-23
Small YOLO loss improvements

arrufat:
Small improvements:
- Compute the input area and the truth area only once.
- Improve logic legibility for the `iou_anchor_threshold`.
- Make sure the input_area is a `drectangle` to avoid unwanted rounding operations:
  - in particular when checking `input_area.contains(center(truth_box.rect))`
  - the `.contains` method expects a `point` instead of a `dpoint` if `input_area` is a `rectangle`.
  - I don't think it has much effect, but still.

### DIFF
--- a/dlib/dnn/loss.h
+++ b/dlib/dnn/loss.h
@@ -3801,66 +3801,67 @@ namespace dlib
                     {
                         // Check if the best anchor is from the current stride.
                         const bool is_best_anchor = best_tag_id == tag_id<TAG_TYPE>::id && best_a == a;
-                        // Update the best anchor unconditionally, and optionally other anchors.
-                        if (is_best_anchor || iou_anchor_threshold < 1)
+
+                        // Do nothing if it's not the best anchor and iou_ignore_threshold is disabled
+                        if (!is_best_anchor && iou_anchor_threshold == 1)
+                            continue;
+
+                        // Do not update other anchors if they have low IoU
+                        if (!is_best_anchor)
                         {
-                            // Do not update other anchors if they have low IOU.
-                            if (!is_best_anchor)
+                            const auto anchor(centered_drect(t_center, anchors[a].width, anchors[a].height));
+                            if (box_intersection_over_union(truth_box.rect, anchor) < iou_anchor_threshold)
+                                continue;
+                        }
+
+                        const long c = t_center.x() / stride_x;
+                        const long r = t_center.y() / stride_y;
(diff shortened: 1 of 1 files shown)


arrufat:
I think the new logic is a bit more clear, and also removes one extra nesting level.

I kind of prefer this way, where all the conditions at the beginning are breaking conditions, and then the normal path remains at the immediate inner loop scope.

dlib/dnn/loss.h
```diff
@@ -3797,67 +3799,65 @@ namespace dlib
...
                     for (size_t a = 0; a < anchors.size(); ++a)
                     {
-                        // Update best anchor if it's from the current stride, and optionally other anchors
-                        if ((best_tag_id == tag_id<TAG_TYPE>::id && best_a == a) || iou_anchor_threshold < 1)
+                        // We will always backpropagate on the best anchor, regardless of its IOU.
+                        // For other anchors, only if they have an IOU >= iou_anchor_threshold.
+                        if (!(best_tag_id == tag_id<TAG_TYPE>::id && best_a == a))
                         {
+                            if (iou_anchor_threshold == 1)
+                                continue;
```
arrufat, line 3807: This isn't strictly necessary, but will avoid IOU computations…

dlib/dnn/loss.h
```diff
@@ -3797,67 +3799,65 @@ namespace dlib
...
                     for (size_t a = 0; a < anchors.size(); ++a)
                     {
-                        // Update best anchor if it's from the current stride, and optionally other anchors
-                        if ((best_tag_id == tag_id<TAG_TYPE>::id && best_a == a) || iou_anchor_threshold < 1)
+                        // We will always backpropagate on the best anchor, regardless of its IOU.
+                        // For other anchors, only if they have an IOU >= iou_anchor_threshold.
+                        if (!(best_tag_id == tag_id<TAG_TYPE>::id && best_a == a))
                         {
+                            if (iou_anchor_threshold >= 1)
+                                continue;
```
arrufat, line 3807: This isn't strictly necessary, but will avoid IOU computations, even if the user sets a value larger than 1.

davisking:
> I think the new logic is a bit more clear, and also removes one extra nesting level.
> 
> I kind of prefer this pattern, where all the conditions are at the beginning and are breaking conditions, then the happy path remains at the immediate inner loop scope.

Yeah, it's a good pattern :)

PR #2654, opened 2022-08-25
if constexpr() for C++14

pfeatherstone:
It's actually more like "switch constexpr" and it's awesome.
It's inspired by hana::if_.
I've inserted it in a couple places where Davis mentioned that C++17's if constexpr would have been great.

### DIFF
--- a/dlib/constexpr_if.h
+++ b/dlib/constexpr_if.h
@@ -1,0 +1,131 @@
+// Copyright (C) 2022  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_IF_CONSTEXPR_H
+#define DLIB_IF_CONSTEXPR_H
+
+#include <utility>
+
+namespace dlib
+{
+    namespace detail
+    {
+        const auto _ = [](auto&& arg) { return std::forward<decltype(arg)>(arg); }; // same as std::identity in C++20 but better
+
+        template <
+            bool     Force,
+            typename Type,
+            typename Case
+        >
+        struct case_type_and_lambda
+        {
+            static constexpr bool force = Force;
+            using type = Type;
+            Case case_;
+
+            template<typename... Args>
+            constexpr decltype(auto) operator()(Args&& ...args) { return case_(_, std::forward<Args>(args)...); }
+        };
+
+        template<
+            typename type,
+            typename Case
+        >
+        using valid_t = std::enable_if_t<std::decay_t<Case>::force || std::is_same<type, typename std::decay_t<Case>::type>::value, bool>;
+
+        template<
+            typename type,
+            typename Case
+        >
+        using not_valid_t = std::enable_if_t<!std::decay_t<Case>::force && !std::is_same<type, typename std::decay_t<Case>::type>::value, bool>;
+    }
+
+    template <
+        typename type,
+        typename Case
+    >
+    constexpr auto case_type_(
+        Case &&case_
+    )
+    {
+        return detail::case_type_and_lambda<false, type, std::decay_t<Case>>{std::forward<Case>(case_)};
+    }
+
+    template <
+        bool result,
+        typename Case
+    >
+    constexpr auto case_(
+        Case &&case_
+    )
+    {
+        using bool_t = std::integral_constant<bool,result>;
+        return case_type_<bool_t>(std::forward<Case>(case_));
+    }
+
+    template <typename Case>
+    constexpr auto default_(Case &&case_)
+    {
+        return detail::case_type_and_lambda<true, bool, std::decay_t<Case>>{std::forward<Case>(case_)};
+    }
+
+    template <
+        typename type,
+        typename Case1,
+        typename... CaseRest,
+        detail::valid_t<type, Case1> = true
+    >
+    constexpr decltype(auto) switch_type_(
+        Case1&& first,
+        CaseRest&&...
+    )
+    {
+        return first();
+    }
+
+    template <
+        typename type,
+        typename Case1,
+        typename... CaseRest,
+        detail::not_valid_t<type, Case1> = true
+    >
+    constexpr decltype(auto) switch_type_(
+        Case1&&,
+        CaseRest&&... cases
+    )
+    {
+        return switch_type_<type>(std::forward<CaseRest>(cases)...);
+    }
+
+    template <
+        typename type,
+        typename CaseLast,
+        detail::valid_t<type, CaseLast> = true
+    >
+    constexpr decltype(auto) switch_type_(
+        CaseLast&& last
+    )
+    {
+        return last();
+    }
+
+    template <
+        typename type,
+        typename CaseLast,
+        detail::not_valid_t<type, CaseLast> = true
+    >
+    constexpr void switch_type_(CaseLast&&)
+    {
+    }
+
+    template <
+        typename...Cases
+    >
+    constexpr decltype(auto) switch_(
+        Cases&&... cases
+    )
+    {
+        return switch_type_<std::true_type>(std::forward<Cases>(cases)...);
+    }
+}
+
+#endif //DLIB_IF_CONSTEXPR_H
--- a/dlib/test/constexpr_if.cpp
+++ b/dlib/test/constexpr_if.cpp
@@ -1,0 +1,126 @@
+// Copyright (C) 2022  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+
+#include <string>
+#include <dlib/constexpr_if.h>
+#include "tester.h"
+
+namespace
+{
+    using namespace test;
+    using namespace dlib;
+
+    logger dlog("test.constexpr_if");
+
+    struct A
+    {
+        int i;
+    };
+
+    struct B
+    {
+        float f;
+    };
+
+    struct C
+    {
+        std::string str;
+    };
+
+    template<typename T>
+    auto handle_type_and_return1(T obj)
+    {
+        return switch_type_<T>(
+            case_type_<A>([&](auto _) {
+                return _(obj).i;
+            }),
+            case_type_<B>([&](auto _) {
+                return _(obj).f;
+            }),
+            case_type_<C>([&](auto _) {
+                return _(obj).str;
+            }),
+            default_([&](auto _) {
+                printf("Don't know what this type is\n");
+            })
+        );
+    }
+
+    template<typename T>
+    auto handle_type_and_return2(T obj)
+    {
+        return switch_(
+            case_<std::is_same<T,A>::value>([&](auto _) {
+                return _(obj).i;
+            }),
+            case_<std::is_same<T,B>::value>([&](auto _) {
+                return _(obj).f;
+            }),
+            case_<std::is_same<T,C>::value>([&](auto _) {
+                return _(obj).str;
+            }),
+            default_([&](auto _) {
+                printf("Don't know what this type is\n");
+            })
+        );
+    }
+
+    void test_switch_type()
+    {
+        A a{1};
+        B b{2.5f};
+        C c{"hello there!"};
+
+        {
+            auto ret = handle_type_and_return1(a);
+            static_assert(std::is_same<decltype(ret), int>::value, "failed test");
+            DLIB_TEST(ret == a.i);
+        }
+
+        {
+            auto ret = handle_type_and_return2(a);
+            static_assert(std::is_same<decltype(ret), int>::value, "failed test");
+            DLIB_TEST(ret == a.i);
+        }
+
+        {
+            auto ret = handle_type_and_return1(b);
+            static_assert(std::is_same<decltype(ret), float>::value, "failed test");
+            DLIB_TEST(ret == b.f);
+        }
+
+        {
+            auto ret = handle_type_and_return2(b);
+            static_assert(std::is_same<decltype(ret), float>::value, "failed test");
+            DLIB_TEST(ret == b.f);
+        }
+
+        {
+            auto ret = handle_type_and_return1(c);
+            static_assert(std::is_same<decltype(ret), std::string>::value, "failed test");
+            DLIB_TEST(ret == c.str);
+        }
+
+        {
+            auto ret = handle_type_and_return2(c);
+            static_assert(std::is_same<decltype(ret), std::string>::value, "failed test");
+            DLIB_TEST(ret == c.str);
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


pfeatherstone:
I would like to benchmark compile time vs if constexpr() but at the same time, I can't be bothered

dlib/constexpr_if.h
```diff
@@ -0,0 +1,131 @@
+// Copyright (C) 2022  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_IF_CONSTEXPR_H
+#define DLIB_IF_CONSTEXPR_H
+
+#include <utility>
+
+namespace dlib
```
davisking, line 8: This stuff needs docs of some kind

dlib/test/constexpr_if.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    struct C
+    {
+        std::string str;
+    };
+
+    template<typename T>
+    auto handle_type_and_return1(T obj)
+    {
+        return switch_type_<T>(
```
davisking, line 34: So in C++14 you can totally do stuff like this
```
switch(some_object,
    [](int) { cout << "some_object is an int type"; },
    [](foo) { cout << "some_object is a foo type"; },
    [](double) { cout << "some_object is a double type"; },
    [](auto) { cout << "some_object is some other type!"; },
);
```
I've written stuff like that and seen it elsewhere.  How about doing that instead?

pfeatherstone: What if you want to access a member function depending on the type ? Basically, whatever argument you have for if constexpr() applies here. Hopefully the tests show this. I started applying it to the DNN stuff in places where if constexpr would have been useful

pfeatherstone: I think your method won't work if you want each case to return something different and then the calling function to return different things

pfeatherstone: Also in your case all the branches are compiled. In this code, they aren't. Exactly like it constexpr()

dlib/constexpr_if.h
pfeatherstone: I'm tempted to say we shouldn't put this in the public API. It's more of a utility for implementation details.

pfeatherstone: There are many ways you can misuse this, particularly how to use the underscore function properly, which is critical.

pfeatherstone:
@davisking So the closest i've got to your suggestion is this:
```
namespace p
{
    namespace detail
    {
        template<typename Base, typename ... BaseRest>
        struct overloaded_helper: Base, overloaded_helper<BaseRest...>
        {
            template<typename T, typename ... TRest>
            overloaded_helper(T&& t, TRest&& ...trest) :
                Base{std::forward<T>(t)},
                overloaded_helper<BaseRest...>{std::forward<TRest>(trest)...}
            {}

            using Base::operator();
            using overloaded_helper<BaseRest...>::operator();
        };

        template<typename Base>
        struct overloaded_helper<Base> : Base
        {
            template<typename T>
            overloaded_helper<Base>(T&& t) : Base{std::forward<T>(t)}
            {}

            using Base::operator();
        };

        template<typename... T>
        overloaded_helper<std::decay_t<T>...> overloaded(T&&... t)
        {
            return overloaded_helper<std::decay_t<T>...>{std::forward<T>(t)...};
        }

        const auto _ = [](auto&& arg) -> decltype(auto){ return std::forward<decltype(arg)>(arg); };
    }

    template<typename T>
    struct type_t {using type = T;};

    template<bool v>
    using bool_v = std::integral_constant<bool,v>;    

    template<typename T, typename... F>
    decltype(auto) switch_(
        type_t<T> meta_obj,
        F&&... fs
    )
    {
        using namespace detail;
        return overloaded(std::forward<F>(fs)...)(meta_obj, _);
    }

    template<bool v, typename... F>
    decltype(auto) switch_(
        bool_v<v> meta_obj,
        F&&... fs
    )
    {
        using namespace detail;
        return overloaded(std::forward<F>(fs)...)(meta_obj, _);
    }
}
```

pfeatherstone:
Then i can do something like this:
```
struct A
{
    int i;
};

struct B
{
    float f;
};

struct C
{
    std::string str;
};

template<typename T>
auto handle_type_and_return1(T obj)
{
    using namespace p;

    return switch_(type_t<T>{},
        [&](type_t<A>, auto _) {
            return _(obj).i;
        },
        [&](type_t<B>, auto _) {
            return _(obj).f;
        },
        [&](type_t<C>, auto _) {
            return _(obj).str;
        },
        [](auto, auto) {
            printf("Don't know what this type is\n");
        }
    );
}
```

pfeatherstone:
For boolean types, you have to nest it like this:
```
template<typename T>
auto handle_type_and_return2(T obj)
{
    using namespace p;

    return switch_(std::is_same<T,A>{},
        [&](std::true_type, auto _) {
            return _(obj).i;
        },
        [&](std::false_type, auto) {
            return switch_(std::is_same<T,B>{},
                [&](std::true_type, auto _) {
                    return _(obj).f;
                },
                [&](std::false_type, auto _) {
                    return switch_(std::is_same<T,C>{},
                        [&](std::true_type, auto _) {
                            return _(obj).str;
                        },
                        [&](std::false_type, auto _) {
                            printf("Don't know what this type is\n");
                        }
                    );
                }
            );
        }
    );
}
```

This is not great and the solution currently implemented is a bit better

pfeatherstone:
But you can still do stuff like this:
```
template <typename Func, typename... Args>
bool try_invoke(Func&& f, Args&&... args)
{
    using namespace p;

    return switch_(std::is_invocable<Func, Args...>{},
        [&](std::true_type, auto _) {
            _(std::forward<Func>(f))(std::forward<Args>(args)...);
            return true;
        },
        [&](std::false_type, auto) {
            return false;
        }
    );
}
```

pfeatherstone:
And you can still do stuff like :
```
template<typename T>
using requires_set_i = decltype(std::declval<T>().set_i(int{}));

template<typename T>
void try_set_i(T& obj, int i)
{
    using namespace dlib;

    switch_(is_detected<requires_set_i, T>{},
        [&](std::true_type, auto _) {
            _(obj).set_i(i);
        },
        [](auto...) {
            printf("Type does not have set_i()\n");
        }
    );
}
```

pfeatherstone:
So I don't know. From an implementation point of view, using function overloading is really useful and removes the need to do recursive case matching as is currently done. You let the compiler do that for you via overloading.
But if you're doing if(), else if (), else if(), else(), then the current implementation is better.

pfeatherstone:
Actually the nested thing is trivially fixable. I will do that tomorrow. So overloading always wins. As always, @davisking is right. Thanks

dlib/test/constexpr_if.cpp
davisking: Yeah they all get compiled this way.  I don't know if that ends up mattering though in most cases :shrug:   But yeah I could imagine cases where it adds to compile time.

dlib/constexpr_if.h
davisking: Na, totally needs a clean and clear interface.  Good software is made in layers where each layer is written with a clear interface.  Like when I'm looking at this I don't know how to use it.  I see what the tests are doing but still.

davisking: Like why is that `_` function needed?  I've been looking at this and I even just tested some of it and I still don't see why that is needed.  I haven't even found a case where something bad happens without it yet.

pfeatherstone: If you look at try_invoke() in the unit tests you'll see why it's needed. Otherwise I'll upload code to demonstrate Monday.

pfeatherstone: I'm fairly certain if you don't use the _ function in try_invoke the program won't compile

davisking: Works fine without `_` there too :shrug:

pfeatherstone: Weird. I don't have my PC. I'll check Monday. I mean I would rather not use it but I swear there were circumstances where it was required.

pfeatherstone: Excellent news if it's never required

pfeatherstone: Originally I was getting the same problem described in boost::Hana. Dionne explains the following: "Here, we capture the enclosing args... by reference using [&], and we do not need to receive any arguments. Also, hana::eval_if assumes that its arguments are branches that can be called, and it will take care of calling the branch that is selected by the condition. However, this will still cause a compilation failure, because the bodies of the lambdas are not dependent anymore, and semantic analysis will be done for both branches even though only one would end up being used. The solution to this problem is to make the bodies of the lambdas artificially dependent on something, to prevent the compiler from being able to perform semantic analysis before the lambda is actually used. To make this possible, hana::eval_if will call the selected branch with an identity function (a function that returns its argument unchanged (_ function)), if the branch accepts such an argument". Now it's possible using the function overloading technique it's no longer required. I'll check Monday. I'll clarify though, the original implementation wasn't taken from boost. It's just I was seeing the same problems. Hence why I needed the _ identify function

pfeatherstone: I don't know how you made it work. I've committed the change where `_` isn't used, and everything breaks. I imagine you were forwarding all the arguments as opposed to capturing everything by reference in the lambdas? If so, we were comparing apples and oranges.

pfeatherstone:
@davisking Do you mind having a look? I tried removing `_` in commit [04753ff](https://github.com/davisking/dlib/pull/2654/commits/04753ff174472eb2f3eba16f287280b6f86f717e) and everything broke. So i imagine you were either using different code or using this code differently to the unit tests. It does make sense though that it wouldn't work precisely for the reasons Louis Dionne gives in boost Hana. You have to make the bodies of the generic lambdas dependent on something generic for them to be compiled "later". This generic thing is the `_` identity function. I do believe that capturing everything by reference in the lambda is better than forwarding everything back in. If your lambdas are dependent on even  10 things then that's 10 things you have to forward to the `switch_()` function and the signatures of the lambdas get very long and ugly. This way you only need one object to use `_` and you're done. 

I apologize for the docs. They are pretty rubbish. I need to have another go at explaining it more clearly and succinctly.

davisking:
> @davisking Do you mind having a look? I tried removing `_` in commit [04753ff](https://github.com/davisking/dlib/pull/2654/commits/04753ff174472eb2f3eba16f287280b6f86f717e) and everything broke. So i imagine you were either using different code or using this code differently to the unit tests. It does make sense though that it wouldn't work precisely for the reasons Louis Dionne gives in boost Hana. You have to make the bodies of the generic lambdas dependent on something generic for them to be compiled "later". This generic thing is the `_` identity function. I do believe that capturing everything by reference in the lambda is better than forwarding everything back in. If your lambdas are dependent on even 10 things then that's 10 things you have to forward to the `switch_()` function and the signatures of the lambdas get very long and ugly. This way you only need one object to use `_` and you're done.
> 
> I apologize for the docs. They are pretty rubbish. I need to have another go at explaining it more clearly and succinctly.

Huh.  All I did was this (just tried it again):
```
--- a/dlib/test/constexpr_if.cpp
+++ b/dlib/test/constexpr_if.cpp
@@ -39,13 +39,13 @@ namespace
     {
         return switch_(types_<T>{},
             [&](types_<A>, auto _) {
-                return _(obj).i;
+                return obj.i;
             },
             [&](types_<B>, auto _) {
-                return _(obj).f;
+                return obj.f;
             },
             [&](types_<C>, auto _) {
-                return _(obj).str;
+                return obj.str;
             },
             [&](auto...) {
                 printf("Don't know what this type is\n");
@@ -58,13 +58,13 @@ namespace
     {
         return switch_(bools_<std::is_same<T,A>::value, std::is_same<T,B>::value, std::is_same<T,C>::value>{},
             [&](bools_<true,false,false>, auto _) {
-               return _(obj).i;
+               return obj.i;
             },
             [&](bools_<false,true,false>, auto _) {
-               return _(obj).f;
+               return obj.f;
             },
             [&](bools_<false,false,true>, auto _) {
-               return _(obj).str;
+               return obj.str;
             },
             [&](auto...) {
                printf("Don't know what this type is\n");
@@ -120,7 +120,7 @@ namespace
     {
         return switch_(bools_<is_invocable<Func, Args...>::value>{},
             [&](bools_<true>, auto _) {
-                _(std::forward<Func>(f))(std::forward<Args>(args)...);
+                std::forward<Func>(f)(std::forward<Args>(args)...);
                 return true;
             },
             [](bools_<false>, auto) {
@@ -169,7 +169,7 @@ namespace
 
         return switch_(bools_<has_set_i>{},
             [&](bools_<true>, auto _) {
-                _(obj).set_i(i);
+                obj.set_i(i);
                 return true;
             },
             [](auto...){
@@ -205,4 +205,4 @@ namespace
         }
     } a;
 
-}
\ No newline at end of file
+}
```
And on g++ 7.5.0 it built without issue.

davisking:
Which I think ought to work.  It's making these lambdas templates (because `auto` as an argument makes them templates) and then the way you have it setup is such that only the right one actually gets instantiated and called.  So the others never "happen".

pfeatherstone:
Ah ok you left the "auto _" though. That might work for gcc7 specifically but newer compilers might realise that _ isn't used and make the lambda non-generic, or in your words, non-templated

pfeatherstone:
I'll commit and check since GitHub Actions uses latest gcc and clang. But in general, I bet compilers will be smart and realise _ isn't used and ignore it.

pfeatherstone:
> Which I think ought to work. It's making these lambdas templates (because `auto` as an argument makes them templates) and then the way you have it setup is such that only the right one actually gets instantiated and called. So the others never "happen".

Here it is not working with clang  : https://godbolt.org/z/Msrs74b9M

dlib/test/constexpr_if.cpp
```diff
@@ -0,0 +1,206 @@
...
+               return _(obj).i;
+            },
+            [&](auto, true_t, auto, auto _) {
+               return _(obj).f;
+            },
+            [&](auto, auto, true_t, auto _) {
+               return _(obj).str;
+            },
+            [&](auto...) {
+               printf("Don't know what this type is\n");
```
davisking, line 70: cstdio should be included to use this.

davisking:
Yeah dang.  Same thing with newer versions of gcc.  Let me look at this more and think about it.  That `_` is just so gnarly :|  Seems like their ought to be a better way.

davisking:
Na I've come around.  It's good and this is the best way :)

PR #2656, opened 2022-08-27
Do the bias addition in the convolution in a separate call

### LINKED ISSUE #2655: Performance regression from version 19.23 onwards
I'm using dlib 19.22. When I tried to upgrade to dlib 19.24, I've noticed that my code ran ~11% slower. Then I tried to downgrade to 19.23, but the same happened.

I can't share my code, so I made a test runner modifying example [dnn_mmod_find_cars2_ex.cpp](https://github.com/davisking/dlib/blob/master/examples/dnn_mmod_find_cars2_ex.cpp). Here it's the code I use for this new tests:

<details>
<summary>Show/hide code</summary>

```cpp
// The contents of this file are in the public domain. See LICENSE_FOR_EXAMPLE_PROGRAMS.txt
/*
    This example shows how to run a CNN based vehicle detector using dlib.  The
    example loads a pretrained model and uses it to find the front and rear ends
    of cars in an image.  The model used by this example was trained by the
    dnn_mmod_train_find_cars_ex.cpp example program on this dataset:   
        http://dlib.net/files/data/dlib_front_and_rear_vehicles_v1.tar
    
    Users who are just learning about dlib's deep learning API should read
    the dnn_introduction_ex.cpp and dnn_introduction2_ex.cpp examples to learn
    how the API works.  For an introduction to the object detection method you
    should read dnn_mmod_ex.cpp.

    You can also see a video of this vehicle detector running on YouTube:
        https://www.youtube.com/watch?v=OHbJ7HhbG74
*/


#include <iostream>
#include <dlib/dnn.h>
#include <dlib/image_io.h>
#include <dlib/gui_widgets.h>
#include <dlib/image_processing.h>
#include <opencv2/opencv.hpp>

using namespace std;
using namespace dlib;



// The front and rear view vehicle detector network
template <long num_filters, typename SUBNET> using con5d = con<num_filters,5,5,2,2,SUBNET>;
template <long num_filters, typename SUBNET> using con5  = con<num_filters,5,5,1,1,SUBNET>;
template <typename SUBNET> using downsampler  = relu<affine<con5d<32, relu<affine<con5d<32, relu<affine<con5d<16,SUBNET>>>>>>>>>;
template <typename SUBNET> using rcon5  = relu<affine<con5<55,SUBNET>>>;
using net_type = loss_mmod<con<1,9,9,1,1,rcon5<rcon5<rcon5<downsampler<input_rgb_image_pyramid<pyramid_down<6>>>>>>>>;

int main() try
{
	net_type net;
	shape_predictor sp;
	// You can get this file from http://dlib.net/files/mmod_front_and_rear_end_vehicle_detector.dat.bz2
	// This network was produced by the dnn_mmod_train_find_cars_ex.cpp example program.
	// As you can see, the file also includes a separately trained shape_predictor.  To see
	// a generic example of how to train those refer to train_shape_predictor_ex.cpp.
	deserialize("mmod_front_and_rear_end_vehicle_detector.dat") >> net >> sp;

	//image_window win;

	std::vector<std::string> paths;
	cv::glob("<path_to_images_folder>", paths);
	cv::Mat image;

	matrix<rgb_pixel> img;

	int cnt = 0, sum = 0;
	for (const auto& path : paths) {
		load_image(img, path);
		cnt++;
		auto t1 = std::chrono::high_resolution_clock::now();
		// Run the detector on the image and show us the output.
		for (auto&& d: net(img)) {
			//win.set_image(img);
			// We use a shape_predictor to refine the exact shape and location of the detection
			// box.  This shape_predictor is trained to simply output the 4 corner points of
			// the box.  So all we do is make a rectangle that tightly contains those 4 points
			// and that rectangle is our refined detection position.
			auto fd = sp(img, d);
			rectangle rect;
			for (unsigned long j = 0; j < fd.num_parts(); ++j)
				rect += fd.part(j);

			//std::cout << rect << std::endl;
			//if (d.label == "rear")
			//	win.add_overlay(rect, rgb_pixel(255, 0, 0), d.label);
			//else
			//	win.add_overlay(rect, rgb_pixel(255, 255, 0), d.label);
		}
		auto t2 = std::chrono::high_resolution_clock::now();
		auto lapse = std::chrono::duration_cast<std::chrono::milliseconds>(t2 - t1).count();
		sum += lapse;
		std::cout << cnt << " : " << lapse << std::endl;
	}
	std::cout << "Avg: " << sum / cnt << std::endl;


	cout << "Hit enter to end program" << endl;
	cin.get();
}
catch(image_load_error& e)
{
	cout << e.what() << endl;
	cout << "The test image is located in the examples folder.  So you should run this program from a sub folder so that the relative path is correct." << endl;
}
catch(serialization_error& e)
{
	cout << e.what() << endl;
	cout << "The correct model file can be obtained from: http://dlib.net/files/mmod_front_and_rear_end_vehicle_detector.dat.bz2" << endl;
}
catch(std::exception& e)
{
	cout << e.what() << endl;
}
```

</details>

I've used 970 image files I have in my computer, with many different sizes, but I think any dataset of images can be used to test it.

After running this test many times, I observed performance being consistently slower with the newer version. With dlib 19.22, processing time is at an average of 36.2 ms, while in `master` it is around 39.8 ms. This is roughly the same results I get with my own software, which runs around ~9.94% slower.

To make it easier to reproduce, I've used the Docker image: `nvidia/cuda:11.6.2-cudnn8-devel-ubuntu20.04`.

If there is any extra information which I may have forgotten, please let me know so I may provide it.

arrufat:
Fixes #2655.
Somehow, doing it in a single call is slower.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -234,7 +234,7 @@ namespace dlib
                        padding_x_);
             if (use_bias)
             {
-                // For some reason, doing this is slower than a two separate calls
+                // For some reason, doing this is slower than two separate calls
                 // conv(false, output,
                 //     sub.get_output(),
                 //     filters(params,0),
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -232,24 +232,18 @@ namespace dlib
                        _stride_x,
                        padding_y_,
                        padding_x_);
+
+            conv(false, output,
+                 sub.get_output(),
+                 filters(params,0));
+
+            // For some reason, doing this is sometimes slower than two separate calls
+            // conv(false, output,
+            //     sub.get_output(),
+            //     filters(params,0),
+            //     biases(params, filters.size()));
             if (use_bias)
-            {
-                // For some reason, doing this is slower than two separate calls
-                // conv(false, output,
-                //     sub.get_output(),
-                //     filters(params,0),
-                //     biases(params, filters.size()));
-                conv(false, output,
-                     sub.get_output(),
-                     filters(params,0));
                 tt::add(1,output,1,biases(params,filters.size()));
-            }
-            else
-            {
-                conv(false, output,
-                    sub.get_output(),
-                    filters(params,0));
-            }
         }
 
         template <typename SUBNET>
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -234,10 +234,15 @@ namespace dlib
                        padding_x_);
             if (use_bias)
             {
+                // For some reason, doing this is slower than a two separate calls
+                // conv(false, output,
+                //     sub.get_output(),
+                //     filters(params,0),
+                //     biases(params, filters.size()));
                 conv(false, output,
-                    sub.get_output(),
-                    filters(params,0),
-                    biases(params, filters.size()));
+                     sub.get_output(),
+                     filters(params,0));
+                tt::add(1,output,1,biases(params,filters.size()));
             }
             else
             {


arrufat:
@facug91 can you try this?

arrufat:
@davisking maybe there's no point in keeping both versions? But maybe at some point the single-call version is faster?

davisking:
Yeah I bet some pathways in cudnn just happen to suck and this works around that. Like it ought to be better with one call. But that totally depends on what cudnn happens to do inside itself.

pfeatherstone:
Do you see the same behaviour for all versions or cudnn ?

arrufat:
I only tried with cuDNN 8.5. Let's see what other people are using and how it affects them.

facug91:
I apologize for the delay. I tested it all again, and these are the results:  
- `v19.22`: 35.0ms    
- `master` without `fuse_layers`: 37.5ms    
- `master` with `fuse_layers`: 36.2ms    
- `PR 2656` without `fuse_layers`: 37.4ms    
- `PR 2656` with `fuse_layers`: 35.2ms    

cuDNN 8 makes performance testing really hard, because it changes a lot from one execution to another. But in general terms it seems to be working better with your changes. Thank you @arrufat :)

I'm using Docker image `nvidia/cuda:11.6.2-cudnn8-devel-ubuntu20.04`, which has cuDNN 8.4.0. It would be great if more people could try it with different versions.

PR #2660, opened 2022-09-02
Update Pybind11 to v2.10.0

arrufat:
I shamelessly stole @pfeatherstone's fix for the latest version of Pybind11 (v2.10.0) from #2636.
Let's see if the tests pass…


pfeatherstone:
I don't mind at all. Piecemeal PRs are better anyway. Shame Davis won't set minimum cmake version to 3.11. Then dlib could use FetchContent to pull in pybind rather than embed the source code. In fact, cmake could be really clever and if C++17 is detected, then use [nanobind](https://github.com/wjakob/nanobind) instead which is supposed to be a lot faster to compile and run.

davisking:
> I don't mind at all. Piecemeal PRs are better anyway. Shame Davis won't set minimum cmake version to 3.11. Then dlib could use FetchContent to pull in pybind rather than embed the source code. In fact, cmake could be really clever and if C++17 is detected, then use [nanobind](https://github.com/wjakob/nanobind) instead which is supposed to be a lot faster to compile and run.

Ha, I'll update to 3.11 at some point.  I'm surprised no one has complained to me yet about the upgrade to 3.8 :p

I don't want to make people use FetchContent though.  That will go sideways.  Like someone will have the package to install dlib and then it will fail due to a network problem fetching that.  Or they are off the internet or whatever.  That will definitely throw off python users.

pfeatherstone:
I guess so. Offline builds are not uncommon.

arrufat:
Ok, I don't know how to make macOS pick up the right version, I have never used a macOS device, so I am just shooting into the dark here… Why would they default to Python2? Has it been building on Python2 the whole time??

tools/python/test/test_numpy_returns.py
```diff
@@ -10,36 +10,40 @@
...
+    assert dlib.partition_pixels(img, 3) == truth
 
     # Call all these versions of this mainly to make sure binding to
     # various image types works.
-    assert(dlib.partition_pixels(img[:,:,0]) == 125)
-    assert(dlib.partition_pixels(img[:,:,0].astype('float32')) == 125)
-    assert(dlib.partition_pixels(img[:,:,0].astype('float64')) == 125)
-    assert(dlib.partition_pixels(img[:,:,0].astype('uint16')) == 125)
-    assert(dlib.partition_pixels(img[:,:,0].astype('uint32')) == 125)
+    assert dlib.partition_pixels(img[:, :, 0].astype("uint8")) == 125
```
arrufat, line 42: I don't understand why this doesn't work anymore… When I print the type of `img[:, :, 0]`, it is an `ndarray` of `numpy.uint8`.

dlib/external/pybind11/CMakeLists.txt
```diff
@@ -5,143 +5,295 @@
 # All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
 
-cmake_minimum_required(VERSION 3.8.0)
+cmake_minimum_required(VERSION 3.4)
```
davisking, line 8: Did something bad happen with 3.8?

davisking: Oh this is just what pybind11 uses :+1:

davisking:
Nice, thanks for the update :)

PR #2661, opened 2022-09-03
Miscellaneous cleanup and refresh

pfeatherstone:
I'm currently working on a dnn2 API. As I'm working on it, I'm spotting things in the code base that could do with a refresh.
This PR proposes some of these updates:

- old type traits are now just aliases for standard library type traits
- using `is_detected` and `switch_()` in dnn core. I think it makes things clearer and trivially update-able to `if constexpr` for when the time comes.
- added an improved `for_each_in_tuple()`. It uses fold expression if possible otherwise the parameter pack expansion trick into dummy initializer list. This improves compile time. 
- Use generic lambdas where possible instead of struct helpers
- auto deduced return types

### DIFF
--- a/dlib/dnn/core.h
+++ b/dlib/dnn/core.h
@@ -261,7 +261,7 @@ namespace dlib
         auto tuple_subset(
             const Tuple& item, 
             compile_time_integer_list<indices...>
-        )
+        ) -> decltype(std::make_tuple(std::get<indices>(item)...))
         {
             return std::make_tuple(std::get<indices>(item)...);
         }
@@ -283,7 +283,7 @@ namespace dlib
         template <typename... T>
         auto tuple_flatten(
             const std::tuple<T...>& item
-        )
+        ) -> decltype(tuple_flatten(item, typename make_compile_time_integer_range<sizeof...(T)>::type()))
         {
             return tuple_flatten(item, typename make_compile_time_integer_range<sizeof...(T)>::type());
         }
@@ -292,7 +292,7 @@ namespace dlib
         auto tuple_flatten(
             const std::tuple<T...>& item, 
             compile_time_integer_list<indices...>
-        )
+        ) -> decltype(std::tuple_cat(tuple_flatten(std::get<indices-1>(item))...))
         {
             return std::tuple_cat(tuple_flatten(std::get<indices-1>(item))...);
         }
--- a/dlib/dnn/core.h
+++ b/dlib/dnn/core.h
@@ -19,6 +19,7 @@
 #include "../rand.h"
 #include "../algs.h"
 #include "../metaprogramming.h"
+#include "../utility.h"
 #include "../constexpr_if.h"
 
 #ifdef _MSC_VER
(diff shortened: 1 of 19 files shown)


pfeatherstone:
There's a whole bunch of stuff I could do to dnn/core.h, but given i'm rewriting all  this stuff in dnn2, i don't see the point.

pfeatherstone:
I don't think windows likes me using std::isfinite for dlib::is_finite

pfeatherstone:
@davisking What do you think? Updated some things to use C++14. Re-organised some things. It was about time some of the type traits stuff just became c++11 aliases and were put in their own `type_traits.h` header. Changed bits to use C++11 type traits. Basically things should phase out using the dlib type traits if possible. And dnn stuff uses `switch_()` and `is_detected` in some places. I think it makes things simpler and easier to read.

dlib/metaprogramming.h
```diff
@@ -3,46 +3,28 @@
...
-                more complex template metaprogramming constructs.
-        !*/
-
-        template <size_t m>
-        struct push_back
-        {
-            typedef compile_time_integer_list<n..., m> type;
-        };
-    };
+    using compile_time_integer_list = std::index_sequence<n...>;
```
davisking, line 27: This isn't the same thing since it misses `push_back`.  So this would break users using that.

dlib/type_traits.h
```diff
@@ -0,0 +1,185 @@
...
+/*
+    This header contains back-ports of C++14/17 type traits
+    It also contains aliases for things found in <type_traits> which
+    deprecate old dlib type traits.
+*/
+
+#include <type_traits>
+
+// ----------------------------------------------------------------------------------------
+
```
davisking, line 15: This is missing a `namespace dlib {`

dlib/dnn/core.h
```diff
@@ -35,196 +36,160 @@ namespace dlib
...
         T& obj,
         double bias_weight_decay_multiplier
     )
     {
         DLIB_CASSERT(bias_weight_decay_multiplier >= 0);
-        impl::set_bias_weight_decay_multiplier(obj, special_(), bias_weight_decay_multiplier);
+        switch_(bools(is_detected<impl::has_set_bias_weight_decay_multiplier, T>{}),
+            [&](true_t, auto _) { _(obj).set_bias_weight_decay_multiplier(bias_weight_decay_multiplier); },
+            [](auto...)         {/*no-op*/}
+        );
```
davisking, line 163: Yeah this is nice :)

dlib/type_safe_union/type_safe_union_kernel.h
```diff
@@ -127,12 +118,16 @@ namespace dlib
             return detail::type_safe_union_type_id<T,Types...>::value;
         }
 
+        template <typename T>
+        static constexpr int get_type_id (in_place_tag<T>)
```
davisking, line 122: Update the docs in the `_abstract.h` file since this is part of the public API.  Make sure all docs that need to be updated are updated. I'm still looking this over.  The PR looks nice though :)

dlib/metaprogramming.h
pfeatherstone: Yeah I was wondering if push_back was there only for implementation details. I was thinking of putting a deprecation notice on that and all the dlib type traits.

dlib/type_traits.h
pfeatherstone: Oh yeah good shout

dlib/metaprogramming.h
davisking: Na don't say it's deprecated.  Deprecating stuff just upsets the oldest and most loyal users.  Like if something needed to change due to some specific reason then sure.  Like if CUDA was dropping a feature we would necessarily also need to react to that and that might force some API change.  But this class existing isn't bothering anything.

pfeatherstone: Ok fine. "Oldest and most loyal users"? :) Sounds like dlib is specifically built for 1 or 2 key users :) Also, shouldn't we encourage people to update their code and use best practices ? Issuing a compiler warning isn't a big deal. If they build their code with -Werror they will realise that it was about time they updated their code, even more so since updating their compiler to something that supports c++14. Also I'm pretty sure they won't be too upset. Anyway, I don't I'm not too opinionated about deprecation warnings. I don't use them myself but I sure do appreciate it when I see them. They're often educational too.

pfeatherstone: I'll put push_back in. I'm away for a week so it will have to wait. Sorry

dlib/type_traits.h
davisking: No worries.

dlib/metaprogramming.h
davisking: Ha.  Na, it's that there are a lot of people using it in products and otherwise at work.  It sucks to have to spend time (i.e. incur a cost to your business) because someone else is asking you to update code that is working perfectly fine for your needs.

pfeatherstone:
@davisking Do you mind having another look?

davisking:
> @davisking Do you mind having another look?

Yeah nice, this is great :)

PR #2662, opened 2022-09-03
Fix visitor when tag layer is followed by skip layer

arrufat:
I noticed this while plotting the YOLOv7 architecture.
After this change, it works :)

[yolov7.pdf](https://github.com/davisking/dlib/files/9483227/yolov7.pdf)


davisking:
Thanks for the fix.

PR #2665, opened 2022-09-07
Modernize random_color_transform

arrufat:
Just a small PR to modernize `random_color_transform` a bit:
- use `initializer_list` for `std::max`: I really like this.
- use `const` where possible
- use `static_cast` instead of C-style cast

I have run a small benchmark, but I couldn't find any speed difference, both implementations take about 0.0128127±0.00387623 ms to create a `random_color_transform`.

### DIFF
--- a/dlib/image_transforms/random_color_transform.h
+++ b/dlib/image_transforms/random_color_transform.h
@@ -23,44 +23,44 @@ namespace dlib
             const double color_magnitude = 0.2
         )
         {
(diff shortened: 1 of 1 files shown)


arrufat:
After removing the temporary variables, the creation time went down to 0.0123832±0.00326704 ms.
But I am not sure I am measuring just noise here...

davisking:
Thanks, looks good :)

PR #2666, opened 2022-09-08
Type erasure tooling

pfeatherstone:
Still working on dnn2, was implementing storage policies for type erasure and thought it was sufficiently general and useful that it merited its own PR.
I've applied it to `dlib::any` as an example. 
We could trivially implement `dlib::basic_any` with the storage policy as a template type if we wanted. Don't care but why not.


dlib/te.h
```diff
@@ -0,0 +1,440 @@
+// Copyright (C) 2022  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_TYPE_ERASURE_H_
+#define DLIB_TYPE_ERASURE_H_
+
+#include <type_traits>
+#include <utility>
+#include <typeindex>
+
```
pfeatherstone, line 9: Forgot <new>

dlib/any/any_decision_function.h
```diff
@@ -79,118 +72,47 @@ namespace dlib
...
-            return data->evaluate(item);
+            return evaluate_func(storage.ptr, item);
         }
 
+        template<typename T>
+        bool contains() const { return storage.contains<T>();}
+        bool is_empty() const { return storage.is_empty(); }
+        void clear()          { storage.clear(); }
+        void swap (any_decision_function& item) { std::swap(storage, item.storage); }
+
```
pfeatherstone, line 83: This doesn't swap the function pointers.
I wonder if I can simply do:
```std::swap(*this,item)```
which will do the default thing of calling move constructors and move assignment operators and do the right thing

File: dlib/any/any_trainer.h [hunk 3]
```diff
@@ -83,120 +78,51 @@ namespace dlib
...
-            return data->train(samples, labels);
+            return train_func(storage.ptr, samples, labels);
         }
 
+        template<typename T>
+        bool contains() const { return storage.contains<T>();}
+        bool is_empty() const { return storage.is_empty(); }
+        void clear()          { storage.clear(); }
+        void swap (any_trainer& item) { std::swap(storage, item.storage); }
+
```
pfeatherstone, line 89: Again, use ```std::swap(*this, item)``` if possible

dlib/te.h
```diff
@@ -0,0 +1,440 @@
...
+                copy{[](const void *self) -> void * {
+                    return new T_{*reinterpret_cast<const T_*>(self)};
+                }},
+                type_id{[] {
+                    return std::type_index{typeid(T_)};
+                }}
+            {
+            }
+
+            template <
```
pfeatherstone, line 46: I would not bother defining this. Whatever uses a storage object should implement these types of constructors themselves.

dlib/te.h
```diff
@@ -0,0 +1,440 @@
...
+                if (ptr)
+                    del(ptr);
+            }
+
+            void clear()
+            {
+                storage_heap{std::move(*this)};
+            }
+
+            bool is_empty() const
```
pfeatherstone, line 107: Maybe we need some kind of `storage_base` base class that implements `is_empty()`, `contains<T>()`, `unsafe_get<T>()`. Maybe that would remove boiler plate. Will try later

pfeatherstone:
Note that with these storage policies it's pretty trivial to implement something like `function_ref` or `function_view` using `storage_view` and adding another function pointer a bit like what `any_decision_function` does.

pfeatherstone:
then `dlib::thread_pool` could be updated to use `std::function` for `add_task_by_value` and use `dlib::function_view1` for `add_task`

pfeatherstone:
I've added all the different type erased function types for fun.

pfeatherstone:
@davisking What do you think? It's pretty useful when doing any kind of type erase. The `unique_ptr<> + virtual clone()` technique is fine but that only works for heap storage. This way, you can template on the type of storage and choose pretty easily whether you want heap-only, stack-only, sbo, or view type. I've applied it to `any`, `any_decision` and `any_trainer`. In theory we could template the policies without any changes and reap the benefits. I've provided an example with `dlib::function_basic`. And now we have things like `function_view` and an equivalent of `std::function` but where you can specify the SBO size. Pretty neat. This will useful for dnn2 too when defining a generic `layer` that isn't templated and absorbs the underlying layer details.

pfeatherstone:
@davisking What do you think?  I still need to add docs.

davisking:
> @davisking What do you think? I still need to add docs.

Yeah sorry for the delay.  I'll look later today.

dlib/te.h
```diff
@@ -0,0 +1,449 @@
...
+#include <utility>
+#include <typeindex>
+#include <new>
+#include <memory>
+
+namespace dlib
+{
+    namespace te
+    {
+        template<class T, bool null_after_move>
```
davisking, line 16: why have this type but allow `null_after_move=false`?

dlib/te.h
```diff
@@ -0,0 +1,449 @@
...
+                        other.ptr = nullptr;
+                }
+                return *this;
+            }
+
+            operator bool() const       { return ptr != nullptr; }
+            const T* operator->() const { return ptr; }
+            T*       operator->()       { return ptr; }     
+
+            T* ptr = nullptr;
```
davisking, line 46: Totally don't have public member variables.  Especially with the other stuff.  Got to protect the class invariants.

dlib/function.h
```diff
@@ -0,0 +1,91 @@
...
+            return !str.is_empty() && func;
+        }
+
+        R operator()(Args... args) const {
+            return func.ptr(const_cast<void*>(str.get_ptr()), std::forward<Args>(args)...);
+        }
+
+    private:
+        Storage str;
+        te::managed_move_pointer<R(void*, Args...), true> func;
```
davisking, line 72:
```suggestion
        R (*func)(void*, Args...) func = nullptr;
```
Why not this?  Like why have `managed_move_pointer`?

dlib/any/any.h
```diff
@@ -28,146 +25,74 @@ namespace dlib
...
+        any()                               = default;
+        any(const any& other)               = default;
+        any& operator=(const any& other)    = default;
+        any(any&& other)                    = default;
+        any& operator=(any&& other)         = default;
+
+        template<
+            typename T,
+            typename T_ = std::decay_t<T>,
+            std::enable_if_t<!std::is_same<T_, any>::value, bool> = true
```
davisking, line 38:
```suggestion
            std::enable_if_t<!std::is_same<std::decay_t<T>, any>::value, bool> = true
```
That doesn't work?

dlib/function.h
```diff
@@ -0,0 +1,91 @@
...
+    };
+
+    template <class F> 
+    using function_heap  = function_basic<te::storage_heap, F>;
```
davisking, line 76: These are all nice.  We should replace the `dlib::any_function` code though.  That thing is so aggressively pre-c++11.  Like let's just put this stuff in any_function.h and name the `function_heap` one `any_function`.  That would be super great :D

dlib/te.h
```diff
@@ -0,0 +1,449 @@
+// Copyright (C) 2022  Davis E. King (davis@dlib.net)
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_TYPE_ERASURE_H_
+#define DLIB_TYPE_ERASURE_H_
+
+#include <type_traits>
```
davisking, line 6: Put this file in dlib/any/storage.h

davisking (commented):
Yeah this looks real nice.  Totally add docs and all that though.  And totally replace the existing dlib::any_function implementation.  Just make the methods named the same so it's backwards compatible and we can drop the old code.

dlib/function.h
pfeatherstone: So that a function is always empty after move even if using stack based memory. I wasn't sure about that. I'm tempted to remove it.

dlib/te.h
pfeatherstone: Oh yeah much better

dlib/te.h
pfeatherstone: yep

dlib/function.h
pfeatherstone: The idea was to have a simple type that managed what happened to the pointer after a move operation.
I thought it would make people's lives easier to control the state of their type-erased objects after move. For example a function like `is_empty()` could be defined as:
```
bool is_empty() const { return storage.is_empty(); }
``` 
But then, depending on the storage type, `is_empty()` might have different values after move. Indeed, using `storage_heap`, then after move, `is_empty() == false`, but if using `storage_stack`, then `is_empty() == true` after move (if the original object contained something).

So i thought, maybe if your custom function pointers were managed by `managed_move_pointer` and `is_empty()` was redefined as:
```
bool is_empty() const { return storage.is_empty() && func; }
```

then after a move operation, `is_empty() == false` always, regardless of storage type. It just ensures a bit more consistency across storage types.

But, i'm not sure where I stand here. In my opinion, if an object is valid, then it's valid, and you shouldn't make it artificially not valid. But that means you can't guarantee what the state after move will be, other than it's "valid". But that's precisely what move is supposed to be: "valid" but "undetermined".
So maybe we never need `managed_move_pointer`.

Then the idea of having `null_after_move=false` was to easily switch what moved did. If `null_after_move=false` then it's just the same as vanilla pointer.

So there you go, you get to configure what the state after move is a bit more.

Opinions or ideas?

dlib/any/any.h
pfeatherstone: Yes of course, i think that was a bit of copy paste

pfeatherstone: Though I'm sure MSVC will at some point complain about completely valid C++

dlib/function.h
pfeatherstone: Yeah cool! I gave up reading `dlib::any_function` and didn't attempt to change it. But it's doing the same thing as `std::function` or `function_heap` then great let's just refactor to that.

davisking: Yeah just make it have the same member functions so it's backwards compatible.  Then all that preprocessor magic used to get a pre-c++-11 kinda variadic template goes away.

dlib/any/any.h
davisking: Ha, yeah maybe.  Get the other instances of this too in there.  Looked like there were several.

dlib/function.h
davisking: That's awfully subtle though.  And it doesn't look like the code in here needs that.  Like the move assignment operator in `storage_heap` uses `std::exchange()` to set the source to `nullptr` so it's all good already.  And that's better since it's real clear from looking at that assignment operator that it is correct.  I.e. it doesn't require reasoning about `null_after_move` or any of that other stuff to figure out if it's really working. 

But yeah, it looks like `storage_stack::is_empty()` always returns false right now.  It should definitely say false when it doesn't have anything in it. Like the meaning of `is_empty() == false` is "you can get the thing out of it or call the contained function".  Which is definitely not the case for a default initialized object that's some kind of std::function equivalent.

State after move can be any valid thing so long as the interface contract clearly says what it is.  I would go with being empty though.

pfeatherstone:
Stupid MSVC

pfeatherstone:
MSVC is so stupid. Like, what is it complaining about now? I'm gonna start tweeking random things now.

pfeatherstone:
@davisking Whenever you have time, do you mind having a look? I think I've addressed all your comments. Docs could maybe be better. Not quite sure what level of detail is required.

pfeatherstone:
Since we have `storage_shared`, in theory someone might want `storage_weak`. Now, one could debate that you could just wrap one of the other storage policies in a `shared_ptr`, but i guess the benefit here is that you've abstracted the reference counting stuff. A bit like how `torch::Tensor` and `cv::Mat` are actually reference counted objects. You don't have to play around with `std::shared_ptr<torch::Tensor>` or `std::shared_ptr<cv::Mat>`. So given that abstracting away reference counting can be beneficial, we argue that `storage_shared` is indeed useful, and in which case something like `storage_weak` would also be useful.

pfeatherstone:
I think `cast_to()` and `get()` could be added to `storage_base`. Every object that implements `cast_to()` looks exactly the same so they could just call `storage_base::cast_to()`

pfeatherstone:
> I think `cast_to()` and `get()` could be added to `storage_base`. Every object that implements `cast_to()` looks exactly the same so they could just call `storage_base::cast_to()`

Committed, lets see what MSVC thinks

pfeatherstone:
Lol, `any_decision_function` can now just be:
```
template <class sample_type, class result_type = double>
using any_decision_function = any_function<result_type(const sample_type&)>;
```

pfeatherstone:
I think the last tests need to be restarted. The error `Cannot initiate the connection to apt.llvm.org:80` has nothing to do with dlib

davisking:
> @davisking Whenever you have time, do you mind having a look? I think I've addressed all your comments. Docs could maybe be better. Not quite sure what level of detail is required.

Sorry for the late reply.  Been unusually busy, even for me :(

Just the usual docs http://dlib.net/howto_contribute.html#1 :D

davisking:
> Since we have `storage_shared`, in theory someone might want `storage_weak`. Now, one could debate that you could just wrap one of the other storage policies in a `shared_ptr`, but i guess the benefit here is that you've abstracted the reference counting stuff. A bit like how `torch::Tensor` and `cv::Mat` are actually reference counted objects. You don't have to play around with `std::shared_ptr<torch::Tensor>` or `std::shared_ptr<cv::Mat>`. So given that abstracting away reference counting can be beneficial, we argue that `storage_shared` is indeed useful, and in which case something like `storage_weak` would also be useful.

What's `weak_storage`?  The analogue of `std::weak_ptr`?  `weak_ptr` is super specialized. So IDK.  Something that is trying to combine type erasure with that kind of weak ownership model seems like the kind of thing that virtually all users will use wrong.

davisking:
> Lol, `any_decision_function` can now just be:
> 
> ```
> template <class sample_type, class result_type = double>
> using any_decision_function = any_function<result_type(const sample_type&)>;
> ```

Yeah that's dope :D

Anyway, I'll look at this in a few days.  I'm feeling sick and will probably go to sleep soon.  Otherwise I would look now :|

pfeatherstone:
I think it would be good to add the following constructors:
```
StorageType1(const StorageType2&)
StorageType1(Storagetype2&&)
```
with all possible combinations.

At the moment of course we only have
```
StorageType1(const StorageType1&)
StorageType1(StorageType1&&)
```
as well as all the converting constructors.

At the moment, converting between storage types will work of course, but you will end up having double erasure. Whereas if we were to appropriately convert between storage types, you would only have 1 level of erasure. I can see the case where the hierarchy of erasure could get out of hand.

pfeatherstone:
With a little bit of care, this could be a seriously good, general-purpose, type erasure utility where you can have runtime polymorphism with value semantics and better performance than inheritance.

pfeatherstone:
I also need to appropriately add noexcept specifiers where possible

pfeatherstone:
The motivation for converting between storage types is to have views of other erased types. In the DNN2 case, as @arrufat suggested, we could have an API which returns layer views which would be views of already erased layer details.

davisking:
Yeah that sounds cool.

davisking:
Ok, I'm looking at this.  Looks good.  Before I go merging it and tweaking some of the docs or whatever you are done right?  I wasn't sure if you had another commit you wanted to add.

pfeatherstone:
Yeah it's good to go. I can add the last comments I made in a separate PR as and when I need it. Thanks a lot

pfeatherstone:
@davisking Let me know if you need me to clarify the docs

davisking:
> @davisking Let me know if you need me to clarify the docs

Looking now.

davisking:
Ok, looks real nice.  I went over the docs and made them follow the style http://dlib.net/intro.html talks about and did a few other things.  I'll merge once all the unit tests run :)

davisking:
I also made `dlib::any_function` use `storage_sbo`.  That's a good default.  Then I removed the other overloads except for `any_function_view` since I'm hard pressed to think anyone should use any but the sbo or the view versions.  I could be argued otherwise I suppose, but the SBO one is in general pretty good as a default.

davisking:
I see the bug.   Adding a test and fixing it.

pfeatherstone:
@davisking thank you so much for all your work. You went beyond the normal code review. I really appreciate you taking time out your busy schedule. 

One thing though, I added those explicit move constructors to storage_shared and storage_view because type_id() might still return something non empty even if the underlying pointer/object is empty, for example after a move since pointers aren't nulled after move. All the tests pass but to me that's a potential bug.

pfeatherstone:
That's why I originally had `null_pointer_after_move` type. It meant a few move constructors and move assignment operators could be made default, rather than use std::exchange everywhere. I mean it all works, it's just a small niggle I had.

davisking:
> @davisking thank you so much for all your work. You went beyond the normal code review. I really appreciate you taking time out your busy schedule.
> 

No problem, thanks for the PR :)

> One thing though, I added those explicit move constructors to storage_shared and storage_view because type_id() might still return something non empty even if the underlying pointer/object is empty, for example after a move since pointers aren't nulled after move. All the tests pass but to me that's a potential bug.

I'm not sure I follow.  Some of the contracts around this stuff were unclear.  In particular, none of the member variables were private.  I made them all private and added the `type_id()` method, which has  a precondition that `is_empty() == false`.  That covers this case does it not?

davisking:
> Oh yeah guess so. Maybe it's all good then. Thanks again. I'll aim to write better docs

No problem.  Yeah, contracts are the best :)   I mean, without a contract you can't even say what is and isn't a bug (except that the contract is in someone's head.  But they forget what it is, change their idea about it, or someone else thinks it's different and now there are bugs because different client code is written under incompatible assumptions about the undocumented contract).

PR #2672, opened 2022-10-08
Modernising dlib's CMakeLists.txt

pfeatherstone:
Started again as I seriously messed up the merge in https://github.com/davisking/dlib/pull/2636


pfeatherstone:
don't understand why tools/python/test/test_global_optimization.py is aborting...

pfeatherstone:
In test_global_optimization.py, it always fails on the 2nd test that's supposed to throw, regardless of the order or which you comment out. So maybe some global cleanup is messed up somehow? Don't know how this is possible. Maybe i'm building pybind differently now.

pfeatherstone:
Absolutely not the faintest idea how to fix this...

pfeatherstone:
Hmm, is this a bug with gcc-11 ? Is [this](https://stackoverflow.com/questions/67584073/gcc-11-false-array-subscript-is-partly-outside-array-bounds-warning) a similar issue with gcc-11 ?

PR #2679, opened 2022-10-31
Fix empirical cross-correlation computation in the SSL example

arrufat:
I was using the normalized features za as both matrices, instead of za and zb. I noticed this because the empirical cross-correlation matrix was symmetrical, which is not supposed to.  It does not affect anything, as it was computed properly in the loss.


davisking:
Thanks for the PR :)

PR #2680, opened 2022-10-31
Fix computation of the Barlow Twins loss gradient

arrufat:
When computing the gradient for the off diagonal terms, I forgot one transpose:

`A' * B = C`, but `B' * A = C'`


arrufat:
I am running the example again. Hopefully, it will only perform better.
I will update the accuracy results in the example if they change.

arrufat:
Regarding the build error on macOS, It seems we will have to update to Pybind11 v2.10.1. From their changelog: "_This is the first version to fully support embedding the newly released Python 3.11._"

However, I am surprised that macOS on GitHub actions is already using Python 3.11.

arrufat:
The results didn't change, but I have peace of mind now.

arrufat:
Hmm, somehow the tests pass again...

davisking:
Huh. Yeah Mac OS cloud environments have never been super reliable for me. I wouldn't worry about it.

PR #2681, opened 2022-11-01
Update Pybind11 to v2.10.1

arrufat:
Just checking if this fixes the Python build on macOS.


arrufat:
Ok, so the error is unrelated to Python 11. I don't know what's going on. I will try another time.

arrufat:
Since it's not Pybind11 related, maybe we can close this one?

davisking:
Yeah, not sure what the deal is.  Doesn't seem to be pybind11 related.

PR #2684, opened 2022-11-09
Improve the data augmentation in the SSL example

arrufat:
I was using the data augmentation recommended for the ImageNet dataset, which is not well suited for CIFAR-10.
After doing so, the test accuracy increased by 1 point.


davisking:
Nice :D

PR #2692, opened 2022-11-28
Fix GitHub actions on Windows because of CMake 3.25.0

arrufat:
I found this on Reddit: https://www.reddit.com/r/cpp/comments/z618kl/psa_cmake_325_might_break_your_github_actions/

_The Meson devs, in their infinite wisdom, decided that it was a good idea to ignore 30 years old platform conventions and produce static libraries with a filename that implies GNU ABI on Windows, even when targeting MSVC ABI._

_CMake, in an attempt to accomodate, added this name pattern to its normal MSVC ABI search patterns in 3.25: https://gitlab.kitware.com/cmake/cmake/-/issues/23975
Since GitHub Actions runners are very frequently updating tools, this update quickly made it into the virtual envs._

### DIFF
--- a/.github/workflows/build_cpp.yml
+++ b/.github/workflows/build_cpp.yml
@@ -106,7 +106,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Configure
-      run: cmake . -B ${{ env.build_dir }}
+      run: |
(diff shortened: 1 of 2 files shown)


davisking:
Nice, thanks :D  

Yeah I saw that post on reddit and thought "maybe this is why things just broke" and was going to go fix this next but you beat me to it :D

PR #2697, opened 2022-12-07
CMake: Make dependencies private

jschueller:
most of them can be made private


pfeatherstone:
I think we should not have those lists and just use `target_sources()`, `target_include_directories()`, `target_link_libraries()`, etc. I had a PR at some point that cleaned all of this up, including making things private, but got a bit stuck on the python binding things. I wanted a single CMakeLists.txt file that built the library, python bindings, and unit tests, rather than have multiple CMakeLists.txt files for each one.

pfeatherstone:
Feel free to use https://github.com/davisking/dlib/pull/2672 this if it helps. Not perfect I know and maybe i was a bit too ambitious as I'm not a cmake expert.

jschueller:
lets do it step by step

jschueller:
this closes #3694 @davisking

davisking:
Just tested this a bit, needed one tweak, but otherwise looks good.  Thanks for the PR :)

PR #2698, opened 2022-12-09
removed extra '<' in fc, added '\n' in layers/inputs/losses XML strings

mattbn:
There were errors in some XML strings returned by to_xml in dnn/layers.h (fc, layer_norm). I also added newlines to various XML strings.
I noticed XML strings that may have errors in dnn/layers.h at lines 3181 (leaky_relu) and 3699 (smelu), i don't know if those should be changed as well


davisking:
Oh yeah, thanks for fixing this.  Those others you noted aren't right either :+1:

PR #2705, opened 2023-01-02
Added bind_front() and bind_back()

pfeatherstone:
Hopefully it is fully constexpr and conditionally noexcept with C++14 onwards


pfeatherstone:
Could add bind_back() as well. Though I've never had a need for it.

pfeatherstone:
@davisking This should be ready to go

davisking:
I'm getting a build error for this when trying it with gcc 7.5.0.  Try building with that and you will see what I mean.

pfeatherstone:
Cool. I'm currently using Ubuntu22 so will figure out a way to install gcc7

pfeatherstone:
Might have to get a rusty old RPI out. I think gcc7 is the default ok raspbian

pfeatherstone:
I think this will work. But this is definitely due to a bug in gcc7. I've seen other similar reports online

davisking:
Yeah.  Totally a bug in that version of gcc.

Anyway, thanks for the PR :)

PR #2707, opened 2023-01-06
FFMPEG wrappers: dlib::ffmpeg::decoder and dlib::ffmpeg::demuxer

pfeatherstone:
I'm having another go as I'm in need of this again and would rather do it properly.
The previous PR has fallen behind and my git rebase/merge skills are pants.
In any case, it was too big.
So i've split it into 2 parts:
* decoding/demuxing
* encoding/muxing
This PR includes the first part.


pfeatherstone:
So it looks like installing ffmpeg libraries in Github Actions is causing timeout issues with Azure. Don't know how to solve this.

pfeatherstone:
> So it looks like installing ffmpeg libraries in Github Actions is causing timeout issues with Azure. Don't know how to solve this.

It seems to  be happy now. Maybe it's just a matter of restarting the tests.

pfeatherstone:
Ok, just need to:
* test with audio files
* maybe reduce the size of the test files to accelerate unit tests
* maybe an example program

pfeatherstone:
This is tested with ffmpeg v3, v4 and v5. With v5 you get loads of warnings because of API deprecation. Nothing big, just something small with audio channel layout. That can be fixed in another PR. I'm not going to bother supporting ffmpeg v2. If you're stuck with that version, then just build ffmpeg from source. Way easier.
In another PR, i'll probably add building ffmpeg from source using cmake's ExternalProjectAdd

pfeatherstone:
I would appreciate an initial review. Thanks

davisking:
> I would appreciate an initial review. Thanks

Cool.  I'll look in a bit :)

pfeatherstone:
Possibly the API for the `Frame` object which wraps `std::unique_ptr<AVFrame>` needs to be a bit better as i expect users will want to use this a lot if they're not dealing with standard RGB or BGR images. For example, if you're decoding H264 encoding streams, which is mostly what MP4 files use, then frames are actually in YUV format. If you have a use-case for using untampered frames, (which I do), then converting to dlib objects isn't an option. So the interface to `Frame` needs to be very good. Possibly we need some more helper functions. For audio, you pretty much always want to use `Frame`.

pfeatherstone:
Actually, I think what we need are "view" objects. So instead of having `convert()` functions which allocate dlib objects, we have objects like:
* `image_view<rgb_pixel>`
* `image_view<bgr_pixel>`
* `audio_view<uint8_t, /*nchannels=*/1>`
* `audio_view<int16_t, /*nchannels=*/2>`
* etc
Each of them hold a pointer to a `AVFrame` object.
For example we could ensure `image_view<rgb_pixel>` satisfies dlib's pixel traits so it can be converted to a matrix or directly plotted to a window, or whatever. More to the point, we can manipulate the image and not allocate a whole new frame.

dlib/CMakeLists.txt
```diff
@@ -830,7 +834,20 @@ if (NOT TARGET dlib)
          mark_as_advanced(fftw fftw_path)
       endif()
 
-
+      if (DLIB_USE_FFMPEG)
+         include(cmake_utils/find_ffmpeg.cmake)
+         if (FFMPEG_FOUND)
+            list (APPEND dlib_needed_private_includes ${FFMPEG_INCLUDE_DIRS})
```
pfeatherstone, line 840: Possibly this needs to be public includes and public libs

pfeatherstone:
But I'll wait for an initial review before I make any more changes.

dlib/media/ffmpeg_utils.h
```diff
@@ -0,0 +1,403 @@
...
+            std::chrono::system_clock::time_point timestamp_us
+        );
+
+        static Frame make_audio(
+            int sample_rate,
+            int nb_samples,
+            uint64_t channel_layout,
+            AVSampleFormat samplefmt,
+            std::chrono::system_clock::time_point timestamp
+        );
```
pfeatherstone, line 202: No point in making these private. The encoder/muxer PR (not this one) will likely need to make these public.

dlib/smart_pointers/observer_ptr.h
```diff
@@ -0,0 +1,62 @@
+#ifndef DLIB_OBSERVER_PTR_H
+#define DLIB_OBSERVER_PTR_H
+
+#include <utility>
+#include <type_traits>
+
+namespace dlib
+{
+    template<class T>
+    class observer_ptr
```
davisking, line 10: I'm not super sold on the idea of `observer_ptr`.  Like why not `T*` for this kind of thing?  I've seen the arguments for it, but they don't really connect with me.

dlib/media/ffmpeg_utils.h
```diff
@@ -0,0 +1,403 @@
...
+        class sw_audio_resampler;
+        class sw_audio_fifo;
+        class decoder_extractor;
+    
+    // ---------------------------------------------------------------------------------------------------
+    }
+
+    // ---------------------------------------------------------------------------------------------------
+
+    class Frame;
```
davisking, line 122:
```suggestion
    class frame;
```
Always use snake case.  Like use the style the C++ standard library uses.

This stuff should probably also all be in a namespace like dlib::ffmeg or something like that.

davisking (commented):
Yeah this looks cool.  Been looking it over for a bit and I like the way the API flows.

Only comments I have are to make sure the docs are real good.  And definitely make there be a real clear and tight file that defines the API.  Like either make an `_abstract.h` file for that defines the interface, or organize the headers so that the only headers a user has to look at contain essentially nothing but public interface defining code and the docs for that code.

dlib/smart_pointers/observer_ptr.h
pfeatherstone: For me it's the move semantics. The pointer is nulled after move.

dlib/media/ffmpeg_utils.h
```diff
@@ -0,0 +1,403 @@
...
+    {
+    
+    // ---------------------------------------------------------------------------------------------------
+
+        class sw_image_resizer
+        {
+        public:
+            void reset(
+                int src_h, int src_w, AVPixelFormat src_fmt,
+                int dst_h, int dst_w, AVPixelFormat dst_fmt
```
pfeatherstone, line 222: Make all arguments `const`

dlib/media/ffmpeg_utils.h
```diff
@@ -0,0 +1,403 @@
...
+            );
+
+            void resize(
+                const Frame &src,
+                int dst_h, int dst_w, AVPixelFormat dst_fmt,
```
pfeatherstone, line 227: make arguments `const`

dlib/media/ffmpeg_utils.h
```diff
@@ -0,0 +1,403 @@
...
+        };
+
+    // ---------------------------------------------------------------------------------------------------
+
+        class sw_audio_resampler
+        {
+        public:
+            void reset(
+                int src_sample_rate, uint64_t src_channel_layout, AVSampleFormat src_fmt,
+                int dst_sample_rate, uint64_t dst_channel_layout, AVSampleFormat dst_fmt
```
pfeatherstone, line 261: Make arguments `const`

dlib/media/ffmpeg_utils.h
```diff
@@ -0,0 +1,403 @@
...
+            );
+
+            void resize(
+                const Frame &src,
+                int dst_sample_rate, uint64_t dst_channel_layout, AVSampleFormat dst_fmt,
```
pfeatherstone, line 266: Make arguments `const`

dlib/media/ffmpeg_utils.h
pfeatherstone: Yes, `dlib::ffmpeg` or `dlib::media` is a good idea!

dlib/media/ffmpeg_utils.h
pfeatherstone: Also, these can be normal constructors, not named constructors.

dlib/media/ffmpeg_utils.cpp
```diff
@@ -0,0 +1,763 @@
...
+
+                ptr->frame->nb_samples = ret;
+                ptr->frame->pts   = tracked_samples_;
+                ptr->timestamp    = system_clock::time_point{nanoseconds{av_rescale_q(tracked_samples_,
+                                                                                    {1, dst_sample_rate_},
+                                                                                    {nanoseconds::period::num, nanoseconds::period::den})}};
+                tracked_samples_ += ptr->frame->nb_samples;
+
+                if (ptr!= std::addressof(dst))
+                    dst = std::move(tmp);
```
pfeatherstone, line 493: Maybe just replace with:
```
av_ptr<AVFrame> tmp = make_avframe();
                tmp->sample_rate    = dst_sample_rate;
                tmp->channel_layout = dst_channel_layout;
                tmp->format         = (int)dst_fmt;

                const int ret = swr_convert_frame(audioResamplerCtx.get(), tmp.get(), src.frame.get());
                if (ret < 0)
                    throw std::runtime_error("swr_convert_frame() failed : " + get_av_error(ret));

                dst.frame       = std::move(tmp);
                dst.frame->pts  = tracked_samples;
                dst.timestamp   = system_clock::time_point{nanoseconds{av_rescale_q(tracked_samples,
                                                                                    {1, dst_sample_rate},
                                                                                    {nanoseconds::period::num, nanoseconds::period::den})}};
                tracked_samples += dst.frame->nb_samples;
```

pfeatherstone: it's a slight pessimization in the case where `dst` already has the right buffer sizes allocated, but the code is a lot cleaner and easier to reason about.

dlib/media/ffmpeg_utils.cpp
```diff
@@ -0,0 +1,763 @@
...
+    }
+
+    bool Frame::is_image() const noexcept
+    {
+        return frame && frame->width > 0 && frame->height > 0 && frame->format >= 0;
+    }
+
+    bool Frame::is_audio() const noexcept
+    {
+        return frame && frame->nb_samples > 0 && frame->channel_layout > 0 && frame->sample_rate > 0;
```
pfeatherstone, line 253: `frame->format != AV_SAMPLE_FMT_NONE` !!!!!

dlib/media/ffmpeg_utils.cpp
```diff
@@ -0,0 +1,763 @@
...
+    }
+
+    bool Frame::is_empty() const noexcept
+    {
+        return !is_image() && !is_audio();
+    }
+
+    bool Frame::is_image() const noexcept
+    {
+        return frame && frame->width > 0 && frame->height > 0 && frame->format >= 0;
```
pfeatherstone, line 248: `frame->format != AV_PIX_FMT_NONE` !!!!

dlib/smart_pointers/observer_ptr.h
pfeatherstone: I think i accidentally slipped this in. I don't need it in this PR.

pfeatherstone:
`webcam_face_pose_ex` no longer needs opencv. We use ffmpeg wrappers instead!

pfeatherstone:
Added basic example of `dlib::ffmpeg::demuxer`.

pfeatherstone:
@davisking Ready for another look if you can. Cheers

pfeatherstone:
@arrufat I seem to remember you saying you were keen for this. If you feel like it and have the time, could you stress test this a bit? Any suggestions on API, documentation or whatever, I would be grateful. Cheers

pfeatherstone:
Note to everyone, when installing ffmpeg via a package manager, you typically get the full blown installation with everything enabled and a ton of dependencies. 
What I typically do in my projects is download a release of ffmpeg (v4 please as v5 is not supported yet), build it from source using `./configure --disable-everything` and only enable the things I need. That really does reduce the size of your binaries and you truly only pay for what you use.
Then when building dlib, just use `CMAKE_PREFIX_PATH` and point to your local installation of ffmpeg for that project.

pfeatherstone:
At some point I'll PR some cool examples like using `demuxer` for screen grab, reading microphone and RTSP client.

arrufat:
I will have a look soon. I've just tried the webcam example code, and it works great.
Is there a way to force a specific FPS, like `cap.get(cv::CAP_PROP_FPS);` in OpenCV? Other than that, I only used OpenCV for the capturing the webcam stream and saving it to a file, so relatively limited use of what you're proposing here.

Again, I will read the docs thoroughly and let you know what I think :)
Thank you so much for working on this again :)

pfeatherstone:
> I will have a look soon. I've just tried the webcam example code, and it works great. Is there a way to force a specific FPS, like `cap.set(cv::CAP_PROP_FPS, 30)` in OpenCV? 

You can definitely force an FPS when **creating** a video but that will be for another PR.
When it comes to reading, I'll have a look through the ffmpeg docs to see if you can set a demuxer to read frames at a particular rate. At the moment it reads all frames, and therefore the default FPS of the video.

arrufat:
Thank you, I had already found how to query the fps. Forcing the FPS is mostly useful for Webcams, since, depending on the FPS, they will use different resolutions. I have some webcams that open a 720p stream at 10 fps, but when I force them to 30 fps, I get a 480p stream.
I think you can also go the other way around and force the resolution instead.
I use the `qv4l2` test utility to check it.

pfeatherstone:
Let me have a look

pfeatherstone:
It will definitely be possible. Probs a good feature to add in. My goal is for this to get used so if there are any issues, hopefully they get spotted, fixed, and then it benefits me.

pfeatherstone:
Just noticed dlib already has something like `image_view<>` just what we need for `ffmpeg::frame` so that we can access pixels without first copying to `array2d<rgb_pixel>`

pfeatherstone:
@arrufat I've updated the `webcam_pose_face_ex.cpp` example so you can set v4l2 options such as hardware video size and framerate via the `demuxer::args::format_options` field. You can also set the normal `demuxer::args::image_options::h` and `demuxer::args::image_options::w` options which set the bilinear resizer inside `demuxer`. In an ideal world, you would reuse `demuxer::args::image_options::h` and `demuxer::args::image_options::w` to first try set the device format first, and if that isn't possible or fails, then set the bilinear resizer. But please try the example. It words for me.

pfeatherstone:
Note, setting height and width in `demuxer::args::format_options` like the example does is a v4l2 thing not a general video file thing (like MP4). In general, if you want a different size, use `demuxer::args::image_options::h` and `demuxer::args::image_options::w` which configures a resizer.

arrufat:
@pfeatherstone wow, great! I've just tried the updated example, and I manage to get 30 fps by setting the size (`--hw_video_size 640x480`), but not by requesting the fps directly (`--hw_framerate 30`).

So, if I run:
`webcam_face_pose_ex --hw_video_size 1280x720`:
- height: 720
- width:  1280
- fps:    10

`webcam_face_pose_ex --hw_framerate 30`
> `[video4linux2,v4l2 @ 0x55c79d556840] The driver changed the time per frame from 1/30 to 1/10`
- height: 720
- width:  1280
- fps:    10

But then, if I run:
`webcam_face_pose_ex --hw_video_size 640x480`
- height: 480
- width:  640
- fps:    30

And then, from now on, even if I run without arguments, I still get 640x480@30fps. It seems like the changes are persistent across runs. Maybe that's what's supposed to happen. However, it would be great if I could just request the fps, without typing the size (just a quality of life improvement: I am used to what OpenCV does). But it's already great! Thank you!

pfeatherstone:
Yeah I'll have a look at making all of this easier. I don't like how for webcam there are two sets of sizes.

pfeatherstone:
Neither dlib or ffmpeg are caching any values, I imagine that's the v4l2 driver

pfeatherstone:
I've changed it such that if all of the following is satisfied:
* `demuxer::args::image_options::h` > 0
* `demuxer::args::image_options::w` > 0
* `demuxer::args::format_options.find("video_size") == demuxer::args::format_options.end()`

then `demuxer` will first attempt to set the video size using "video_size", in format options. This is only going to work for v4l2, xcbgrab and maybe others.

Otherwise, if :
* `demuxer::args::image_options::h` > 0
* `demuxer::args::image_options::w` > 0

then the bilinear resizer will be set (as before)

davisking:
> I think i might have to make this header-only because of ffmpeg versions. They tend to break APIs and don't want to be in a situation where someone is linking to ffmpeg v3 for example when dlib was built against v4. If everything is in a header then this is a non issue.
> 
> So maybe have `ffmpeg_utils.h` and `ffmpeg_utils_impl.h` where the later is included at the bottom of `ffmpeg_utils.h` with all the impls. Likewise for `ffmpeg_demuxer.h`.
> 
> Or we keep track of the version dlib was built with and static assert with the version being included by user code. I don't like this option.

Yeah either of those is fine. Whatever is most clean you think.  I would also probably lean towards the header only version.

davisking:
> @davisking Ready for another look if you can. Cheers

Sweet.  Will look in a bit.  Been a busy week.

pfeatherstone:
> I would also probably lean towards the header only version

Yeah I think that's better. I've added a whole bunch of code for storing and comparing ffmpeg versions, but header-only removes all that pain. This is a pity. And it doesn't feel like an ffmpeg problem, just a general C++ problem.

pfeatherstone:
Also, with header-only approach, i don't know how i statically initialize ffmpeg once. I have a symbol `dlib::ffmpeg_details::FFMPEG_INITIALIZED` which is basically a static const variable which initializes all the necessary ffmpeg library bits. If that's in a header, won't it initialize in all translation units ?

pfeatherstone:
@davisking I have a header-only version working. My only worry is the statement:
```
inline static const bool FFMPEG_INITIALIZED = ffmpeg_initialize_all();
```
I want to make sure that all translation units will use the same `FFMPEG_INITIALIZED` variable and therefore `ffmpeg_initialize_all()` will get called exactly once. I'm not super hot on the `inline` rules and how they play with `static`, internal linkage and external linkage.

In any case, I'll wait for another review before committing all the header-only stuff. it's basically the same

davisking:
> @davisking I have a header-only version working. My only worry is the statement:
> 
> ```
> inline static const bool FFMPEG_INITIALIZED = ffmpeg_initialize_all();
> ```
> 
> I want to make sure that all translation units will use the same `FFMPEG_INITIALIZED` variable and therefore `ffmpeg_initialize_all()` will get called exactly once. I'm not super hot on the `inline` rules and how they play with `static`, internal linkage and external linkage.
> 
> In any case, I'll wait for another review before committing all the header-only stuff. it's basically the same

Yeah it would be easy for someone to make something like that end up weird.  It's not like everything has to be in the header. The goal is just to make sure that user code trying to use the wrong ffmpeg doesn't compile right?  And so making all the code that calls ffmpeg directly be in a header makes that happen.  But that doesn't mean that nothing can be in a .cpp file though.  Just put the init stuff in a .cpp file and make it call an init function in a header.

pfeatherstone:
Cool. I'll still wait for anymore review suggestions before fixing stuff. Just let me know when you're done. Though I appreciate everyone is busy, including myself

pfeatherstone:
Ok, I said i would wait but I couldn't help myself. The mains things i've added are more generic `convert()` functions. Hopefully the docs are clear. Basically, provided your `decoder` or `demuxer` are configured with the right output pixel formats, you can convert to any dlib object that satisfy image traits.

pfeatherstone:
@arrufat I've changed `webcam_face_pose_ex.cpp`. Check it out. You can set height, width and framerate. At the moment, if you just set framerate, you might get warning message:
`[video4linux2,v4l2 @ 0x55c79d556840] The driver changed the time per frame from 1/30 to 1/10`
If you set all 3 parameters, then it should go away. I haven't found a way to make it work by just setting framerate. But it's usable now I hope.

pfeatherstone:
There used to be a method for querying device capabilities : `avdevice_capabilities_create()` but that's now been deprecated and just returns an error code. Shame. It used to be possible get information like supported fps and frame sizes for webcams and other devices. I'm still digging through the API, but can't find anything.

arrufat:
@pfeatherstone thank you so much for your work on this… I wish I could spend some more time trying this out. As you said, it works great once I set all with, height and fps, but not when I only set the fps. It's still a massive improvement compared to having to use OpenCV just for camera access. So, I am overall happy with this functionality.

Hopefully, I'll have some time in the coming days, since it's the Korean new year.

Again, thank you so much for the effort you're putting on this.

As a side note, I am not sure whether we should get rid of the example that uses OpenCV, since it's the only one that shows how to interact with OpenCV from dlib, should you need it.

pfeatherstone:
Happy new year to you! Ok, I'll write a new example for camera access

pfeatherstone:
`list_input_devices()` and `list_output_devices()` will now not only tell you which types of input/output devices you can read/write from, but also a list of actual devices currently connected to your machine with a name and description. Handy

pfeatherstone:
@davisking sorry to bother, but will you have time at some point to review? Thanks

davisking:
> @davisking sorry to bother, but will you have time at some point to review? Thanks 

Yeah sorry for being late. I think I can look later today.

dlib/pixel.h
```diff
@@ -107,6 +107,43 @@ namespace dlib
                 - is_unsigned is true if min() == 0 and false otherwise
     !*/
 
+// ----------------------------------------------------------------------------------------
+
+    namespace details
+    {
+        template<typename...T>
+        using void_t = void;
```
davisking, line 115: This is getting used in multiple files here an I'm sure will end up in other places.  It should live somewhere more common.  There are a bunch of similar one liners like this in dlib/algs.h.  So put it there, and I think it's fine if it's just `dlib::void_t`.  Then things can just use it.  Doubtless it will appear in things unrelated to pixels eventually :)

dlib/overloaded.h
```diff
@@ -107,6 +113,7 @@ namespace dlib
 
         assert(type_ids == vector<int>({0,1,2}));
     !*/
+    noexcept(std::is_nothrow_constructible<detail::overloaded_helper<std::decay_t<T>...>, T&&...>::value)
```
davisking, line 116: Nice, this is cool.

In the future though, if you put this into its own PR and send me that it will be easier.  Like you can stack branches and then send me this one change to overloaded and I would merge it more quickly.  Bigger PRs take more of my time to review so it ends up taking longer.  It's just easier for me to find time.  Like I'm an hour into reviewing this today and it's the middle of the night and I'm still looking it all over. So it won't get merged today.  But if it was in its own PR I would totally merge this bit right now (along with some of the other bits that could be separate).

Anyway, just saying :)  You would get faster turn around with more smaller PRs rather than fewer big ones.

dlib/pixel.h
```diff
@@ -107,6 +107,43 @@ namespace dlib
...
+        template<class Pixel>
+        struct is_pixel_type<Pixel, void_t<decltype(pixel_traits<Pixel>::grayscale),
+                                           decltype(pixel_traits<Pixel>::rgb),
+                                           decltype(pixel_traits<Pixel>::rgb_alpha),
+                                           decltype(pixel_traits<Pixel>::hsi),
+                                           decltype(pixel_traits<Pixel>::lab),
+                                           decltype(pixel_traits<Pixel>::has_alpha),
+                                           decltype(pixel_traits<Pixel>::bgr_layout),
+                                           decltype(pixel_traits<Pixel>::num),
+                                           decltype(pixel_traits<Pixel>::is_unsigned)>> : std::true_type{};
```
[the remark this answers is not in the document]
davisking: Also the `details` namespace isn't needed.  Just let `is_pixel_type` be in `dlib` since there isn't anything else going on here (aside from `void_t` but totally move that to algs.h).

examples/webcam_face_pose_ex.cpp
```diff
@@ -27,61 +27,85 @@
     2011.  SSE4 is the next fastest and is supported by most current machines.  
 */
 
-#include <dlib/opencv.h>
```
davisking, line 30: This is a great example.  Put it in a separate example though.  Call it `webcam_ffmpeg_fae_pose_ex.cpp` or something like that and keep the opencv one.  That way people who want to use opencv can still have a little example showing them how to use dlib and opencv together.

examples/ffmpeg_info.cpp
```diff
@@ -0,0 +1,87 @@
...
+using namespace std;
+using namespace dlib;
+
+int main()
+{
+    // List all codecs supported by this installation of ffmpeg libraries
+    const auto codecs = ffmpeg::list_codecs();
+    printf("Supported codecs:\n");
+    for (const auto& codec : codecs)
+        printf("name : %-16s : encoding supported %i decoding supported %i\n", codec.codec_name.c_str(), codec.supports_encoding, codec.supports_decoding);
```
davisking, line 23: Use `cout` in the example instead of `printf`.  printf is so type unsafe.  Like this compiles
```
    std::string str = "asdf";
    printf("%s\n", str);
```
but just gives nonsense. I know some compilers warn about it but still.  Not all.  People should use `std::format` for libfmt.  It's just way better.

Anyway, use `cout` in the examples.

examples/ffmpeg_video_decoding.cpp
```diff
@@ -0,0 +1,116 @@
+// The contents of this file are in the public domain. See LICENSE_FOR_EXAMPLE_PROGRAMS.txt
+/*
+
+    This is an example illustrating the use of the ffmpeg wrappers, 
+    in this case the decoding API.
+
+    This is a pretty simple example. It loads a raw codec file, parses chunks of 
+    data to the decoder and plots images to a GUI window.
+
+    Background about video files, something OpenCV doesn't make entirely clear:
```
davisking, line 10:
```suggestion
    Background about video files:
```
Just don't mention opencv :)

dlib/media/ffmpeg_abstract.h
```diff
@@ -0,0 +1,955 @@
...
+        template <class ImageContainer>
+        void convert(const frame& f, ImageContainer& image)
+        /*!
+            requires
+                - f.is_image() == true
+                - f.pixfmt() == pix_traits<typename image_traits<ImageContainer>::pixel_type>::fmt
+
+            ensures
+                - converts a frame object into array2d<rgb_pixel>
+        !*/
```
davisking, line 350:
```suggestion
        template <class image_type>
        void convert(const frame& f, image_type& image)
        /*!
            requires
                - f.is_image() == true
                - f.pixfmt() == pix_traits<typename image_traits<image_type>::pixel_type>::fmt
                - image_type == an image object that implements the interface defined in
                  dlib/image_processing/generic_image.h 
            ensures
                - Copies the image content from f into the provided image.
        !*/
```
Something like that.  One of the precondition is the type contract of the image.  And this thing supports any of them, not just array2d.  And might as well call the type `image_type` since that's what the rest of the library uses for this kind of argument.

dlib/media/ffmpeg_abstract.h
```diff
@@ -0,0 +1,955 @@
...
+
+        template <class ImageContainer>
+        void convert(const ImageContainer& img, frame& f)
+        /*!
+            ensures
```
davisking, line 355: Same thing here.  Add `requires` saying what the image type must be.  Add that anywhere there is a template argument.  The `requires` part needs to say what the type is supposed to be.  At some point we will have concepts and that will make it nicer, but not today :(

dlib/media/ffmpeg_abstract.h
```diff
@@ -0,0 +1,955 @@
...
+            /*!
+                ensures
+                    - if (DECODER_EAGAIN)
+                        dst_frame.is_empty() == true
+                        need to call push_encoded() again or flush()
+                    else if (DECODER_FRAME_AVAILABLE)
+                        dst_frame.is_empty() == false
+                    else if (DECODER_CLOSED)
+                        is_open() == false
+                        dst_frame.is_empty() == true
```
davisking, line 650:
```suggestion
            /*!
                ensures
                    - Attempts to read a frame, storing the result in dst_frame.
                    - If it is successful then returns DECODER_FRAME_AVAILABLE and it
                      will dst_frame.is_empty() == false.  Otherwise, returns one of the 
                      following:
                         - DECODER_EAGAIN: this indicates (IDK, add some words here).  In particular,
                           you will need to call push_encoded() again or flush() before calling this
                           function (I'm guessing, is this right?  Add more docs about this case to
                           make it clear).
                        - DECODER_CLOSED: This indicates there aren't any more frames.  If this happens
                          then it also means that is_open() == false.

```
Something like that.  Except fill out the parts where I was like "IDK, what does this do?" :)

dlib/media/ffmpeg_abstract.h
```diff
@@ -0,0 +1,955 @@
...
+            !*/
+
+        private:
+            // Implementation details
```
davisking, line 654:
```suggestion
```

dlib/media/ffmpeg_abstract.h
```diff
@@ -0,0 +1,955 @@
...
+                    Note that a video file, e.g. MP4, can contain multiple streams: video, audio, subtitles, etc.
+                    This class can decode both video and audio at the same time.
+                    For audio files, e.g. MP3, FLAG, it only decodes audio. (obs)
+            !*/
+
+            struct args
+            {
+                /*!
+                    WHAT THIS OBJECT REPRESENTS
+                    Constructor arguments for demuxer
```
davisking, line 676:
```suggestion
                        This holds constructor arguments for demuxer.
```

dlib/media/ffmpeg_abstract.h
```diff
@@ -0,0 +1,955 @@
...
+            /*!
+                ensures 
+                    - if (is_open())
+                        returns video rotation angle from metadata if available
+                    else
+                        returns 0
+            !*/
+
+        private:
+            // Implementation details
```
davisking, line 948:
```suggestion
```

davisking (commented):
Nice, this is really awesome.  I've been looking it over for quite a while.  I'm still going to look some more.  But it looks great.  First I'm going to sleep though :)

.vscode/settings.json
```diff
@@ -0,0 +1,88 @@
+{
```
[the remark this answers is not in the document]
pfeatherstone: Oops

pfeatherstone:
Thank you so much for looking this over, particularly through the night. I'll get to it.

With regards to pixel_trait and image_trait, I was trying to be precise and consistent with the docs and its requirements. 
Then the enable_if trick which uses those traits are trying to emulate c++ concepts. 
In fact, I noticed one of the dlib pixel traits was missing a field using this so that was pretty nice. I get it that we might be breaking user code but I see it as "assisting users in fixing their code" :)

pfeatherstone:
Also I have a tendency to add small bits in PRs and they get bigger. Sorry, I need to stop that. 
In retrospect:
- overloaded 
- void_t
- pixel_trait 
- image_trait 
Could have gone in separate PRs.

dlib/pixel.h
pfeatherstone: It's already in `dlib/type_traits.h` as a backport of C++17's `void_t` which also lives in `<type_traits>`.
So I had already done the right thing it seems. I'll just remove the duplicates and just `#include <dlib/type_traits.h>`

dlib/overloaded.h
pfeatherstone: Really sorry. I'll endeavour to do this in the future.

dlib/pixel.h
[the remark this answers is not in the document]
pfeatherstone: `void_t` is already in `dlib/type_traits.h`. I'll keep it there, I think it's more appropriate, as that header contains backports of later c++ standards and currently `std::void_t` lives in `<type_traits>` in C++17.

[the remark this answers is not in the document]
pfeatherstone: Yeah my objective was to make that type trait more precise and follow the documentation for what a pixel trait should be to the letter. Then it could be used as an emulated concept using `enable_if_t<pixel_trait<>,bool>` and later as a proper concept. I didn't think it was right to have a type trait checker if it didn't check the type trait was documentation conforming. 
I guess we can do what you say but you could have an empty type trait and it would still work. I'm not convinced that's right.

dlib/overloaded.h
davisking: Na it's cool.  You do good work, don't sweat it :)

Just saying it will get your PRs merged faster.  No shame or anything like that :)

dlib/pixel.h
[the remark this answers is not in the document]
davisking: Well, it will fail though when something tries to use it.  And at that point the compiler error will mention the name of the missing field.  So it will be a more readable error message.  That's all I'm saying.  But if you wanted to put a unit test that checked that all the dlib pixel types have correct traits that's cool.  Or did something to make the error message about a bad traits type even better than whatever the compiler normally emits that's cool too.  Although I think the normal error message for that kind of thing is fine as is.  Might be hard to make it more clear. IDK.

davisking:
> Also I have a tendency to add small bits in PRs and they get bigger. Sorry, I need to stop that. In retrospect:
> 
> * overloaded
> * void_t
> * pixel_trait
> * image_trait
>   Could have gone in separate PRs.

Na no shame or anything.  It's all good.  Just saying it will get your PRs merged faster :)

dlib/pixel.h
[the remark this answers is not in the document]
pfeatherstone: Ok. We'll just leave it as it was, roughly. 
I've added `is_complete_type<T>` type trait and now you can just write:
```
template <class Pixel>
using is_pixel_type = is_complete_type<pixel_traits<Pixel>>;
```
That's really clear. It says exactly what `is_pixel_type` is doing, i.e. a `pixel_trait` specialisation exists and also behaves in the same way as before.

examples/webcam_face_pose_ex.cpp
pfeatherstone: done

dlib/media/ffmpeg_abstract.h
pfeatherstone: Done

dlib/overloaded.h
pfeatherstone: I have to say, this kind of stuff illustrates what's wrong with C++. We have to manually tag stuff as `[[nodiscard]]`, `noexcept`, `constexpr`, etc when I'm willing to bet a modern compiler can determine all of this automatically without code having to be explicit about it. I don't think it's a matter of C++ having bad defaults, it's a matter of we shouldn't have to specify this in the first place. It should be automatically deduced. It just makes code unreadable and I doubt it's changing the generated assembly code at all. however, it's "best practice"...

pfeatherstone:
I'll address the final things tomorrow, then hopefully, Monday morning I'll wake up and it will be merged.

pfeatherstone:
I think that's everything addressed

davisking:
Nice, this is really kick ass :D

pfeatherstone:
Awesome! We got there in the end!

pfeatherstone:
Also, I'm more than happy to take ownership of any potential future issues

davisking:
> Awesome! We got there in the end! 

Yep, never had any doubts about it :)

arrufat:
Yep! @pfeatherstone rocks!

pfeatherstone:
I forgot to add ```#cmakedefine DLIB_USE_FFMPEG``` to ```dlib/config.h.in```
Presumably that's required for ODR stuff.

davisking:
Yeah, it's needed for system installed dlib too.

PR #2716, opened 2023-01-26
Fixed is_nothrow_swappable<> and simplified invoke-related type traits

pfeatherstone:


pfeatherstone:
Though i've added `dlib::is_nothrow_invocable<>` it doesn't always work and [this godbolt link](https://godbolt.org/z/KPoszfnTd) demonstrates that it's a gcc bug I think. It doesn't use a full blown `invoke()` implementation, but a cut down one. For whatever reason, between gcc 10.4 and 7.5, if you compile with c++11, the `noexcept` specifier doesn't correctly.

pfeatherstone:
I think because function pointers pre C++17 didn't allow a noexcept specification, by default, when passing a function to another function, the function pointer signature didn't have noexcept specification and therefore was assumed to never be noexcept. I guess you'd erased the function at the point and the compiler didn't know whether it would throw... Something like that.
I've literally spent 3 hours trying to understand why `dlib::is_nothrow_invocable` only worked in C++17. What a massive waste of time. I'll leave it in for now and wait for someone to tell me whether or not to remove it.

davisking:
> I think because function pointers pre C++17 didn't allow a noexcept specification, by default, when passing a function to another function, the function pointer signature didn't have noexcept specification and therefore was assumed to never be noexcept. I guess you'd erased the function at that point and the compiler didn't know whether it would throw... Something like that. I've literally spent 3 hours trying to understand why `dlib::is_nothrow_invocable` only worked in C++17. What a massive waste of time. I'll leave it in for now and wait for someone to tell me whether or not to remove it.

Yeah I guess it only works in C++17.  And `std::is_nothrow_invocable` is in C++17 so having it in dlib doesn't do anyone any good. 

The other change looks good though :)

pfeatherstone:
Yeah I'll remove that. I was mainly fixing the swap stuff. It's just the method for fixing that applied to invoke() and thought may as well fix that too.

davisking:
Sweet. Thanks for the fix :)

PR #2720, opened 2023-02-03
ffmpeg wrappers - fixed config.h.in

pfeatherstone:
That thing we discussed.


pfeatherstone:
Doesn't get much more incremental than this :)

davisking:
Nice, thanks :)

I was going to do this myself later today or tomorrow :D   (been traveling all week and doing nothing but working and sleeping so haven't yet)

PR #2721, opened 2023-02-03
Strip binaries in release mode

pfeatherstone:
I think this is a good idea.
I always compile libav* libraries as static libraries.
I noticed that `dtest` was then 100MB in size.
Stripping in release mode reduces that to 20MB on my machine. That's pretty typical given a fully enabled installation of ffmpeg.
I don't want users to get scared and think that all their computer vision applications will be > 100MB in size. That's pretty crazy and easily avoidable by adding in `-s` to your link options.


davisking:
Cool yeah, good suggestion.

PR #2728, opened 2023-02-10
Add sum method to running stats

arrufat:
I know I could've gotten around by using `mean` and `current_n`, but I felt like it would've been more convenient to have it.


davisking:
Sweet, thanks for the PR and glad you are back :)

PR #2733, opened 2023-02-21
Add support for loading custom label fonts in imglab

### LINKED ISSUE #2732: I want to know how to support unicode character in box label in imglab. Could you give me some hints. Thanks!
I want to know how to support unicode character(for example:标签) in box label in imglab. Could you give me some hints. Thanks!

arrufat:
Closes #2732

### DIFF
--- a/tools/imglab/src/main.cpp
+++ b/tools/imglab/src/main.cpp
@@ -21,7 +21,7 @@
 #include <dlib/dir_nav.h>
 
 
-const char* VERSION = "1.19";
+const char* VERSION = "1.20";
 
 
 
@@ -599,6 +599,7 @@ int main(int argc, char** argv)
 
         parser.add_option("h","Displays this information.");
         parser.add_option("v","Display version.");
+        parser.add_option("font", "Set the label font in the GUI.", 1);
 
         parser.set_group_name("Creating XML files");
         parser.add_option("c","Create an XML file named <arg> listing a set of images.",1);
@@ -671,6 +672,8 @@ int main(int argc, char** argv)
 
 
         parser.parse(argc, argv);
+
+        const std::string font_path = get_option(parser, "font", "");
 
         const char* singles[] = {"h","c","r","l","files","convert","parts","rmdiff", "rmtrunc", "rmdupes", "seed", "shuffle", "split", "add", 
                                  "flip-basic", "flip", "rotate", "tile", "size", "cluster", "resample", "min-object-size", "rmempty",
@@ -1263,7 +1266,7 @@ int main(int argc, char** argv)
 
         if (parser.number_of_arguments() == 1)
         {
-            metadata_editor editor(parser[0]);
+            metadata_editor editor(parser[0], font_path);
             if (parser.option("parts"))
             {
                 std::vector<string> parts = split(parser.option("parts").argument());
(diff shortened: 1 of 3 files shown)


pfeatherstone:
Weird how Github Actions can sometimes cancel a job for no reason.

davisking:
Nice, thanks :)

arrufat:
Oh I wasn't expecting for this to get merged before getting some feedback from who reported the issue.

I couldn't test it properly, because copy and paste don't work well with Wayland in the dlib toolkit.

The font is changed though, but I might need to handle the UTF-8 strings when loading the dataset...

I will do that tonight, probably.

davisking:
Ah ok. Well still looks like a sensible thing to have in any event 🤷‍♂️😁

arrufat:
I've started working on that, and I can get imglab to load the Chinese characters from the XML files, just by changing:

```diff
diff --git a/tools/imglab/src/metadata_editor.cpp b/tools/imglab/src/metadata_editor.cpp
index f3e0e49d..c7e59cf1 100644
--- a/tools/imglab/src/metadata_editor.cpp
+++ b/tools/imglab/src/metadata_editor.cpp
@@ -656,9 +656,13 @@ on_overlay_rect_selected(
     const image_display::overlay_rect& orect
 )
 {

-    overlay_label.set_text(orect.label);
+    overlay_label.set_text(convert_utf8_to_utf32(orect.label));
     display.set_default_overlay_rect_label(orect.label);
     display.set_default_overlay_rect_color(string_to_color(orect.label));
}
```

However, I can't get the overlay rectangle to display them properly...
Maybe its `label` field needs to be a `dlib::ustring`? But that would be a breaking change.

davisking:
Yeah `draw_string()`, called in `image_display::draw()`, needs to get a `ustring` or it's just drawing ASCII.  `image_display` should probably just interpret the `label` as utf8 always.  I'm not sure where the conversion to `ustring` should live though.  Could be in `draw_string()` but IDK, that adds a memory allocation on every call along with the conversion and that will end up making the UI laggy sometimes.  Not this UI probably, since there isn't a lot of text, but others I would expect.

Although the code in `convert_utf8_to_utf32()` could be refactored into a function like this:
```
convert_utf8_to_utf32(some_utf8_string.begin(), some_utf8_string.end(), [](unichar ch) {
   visits the unicode characters in order.
});
```
That could be done without memory allocations and then `draw_string()` could use that and everything would just work with utf8 all the time.

arrufat:
Ah, after adding `convert_utf8_to_utf32()()` all over the place in `image::display::draw()`, everything works.

However, I can't see how we can make `convert_utf8_to_utf32()` allocation free, `some_utf8_string.size()` ≤ `some_utf32_string.size()`, right?

Maybe I am fundamentally wrong, I have a very shallow understanding about character encoding…

davisking:
> convert_utf8_to_utf32

Yeah that function as is can't be allocation free.  A function like this could be:
```
convert_utf8_to_utf32(some_utf8_string.begin(), some_utf8_string.end(), [](unichar ch) {
   visits the unicode characters in order.
});
```
Just needs to use something other than that `istringstream` to iterate over the input string and then instead of appending characters to that new string it would call the provided lambda.

arrufat:
At the risk of souding a bit dumb, it's not clear to me where the output uttf-32 string would be stored :sweat_smile:

arrufat:
I will give it a go later (still not sure I am understanding this correctly, maybe it's too late in here)

Issue #463, opened 2017-03-01
drectangle with the size < 1


e-fominov:
I was trying to use drectangle with some normalized coordinate system, where x/y are in range 0..1. This allows me to make processing independent from image resolution, because modern CCTV cameras have multiple output video streams of different resolution

Current drectangle contract says that its size can't be less than 1:
```C++
 drectangle centered_drect (
        const vector<double,2>& p,
        double width,
        double height
    );
    /*!
        ensures
            - returns a rectangle R such that:
                - center(R) == p
                - if (width < 1 || height < 1)
                    - R.width() == 0 
                    - R.height() == 0 
                    - R.is_empty() == true
                - else
                    - R.width() == width
                    - R.height() == height 
    !*/
```

Is there any way to change this? It would be nice to have rectangle primitive without such strong restriction and inventing the wheel with some new rectangle class does not look interesting

The main problem I see are right() and bottom() values. Now they are pointing to a positions *inside* rectangle. OpenCV's implementation makes simple : right = left + width, while dlib's: right = left + width - 1

I can imagine some other situations where this can be usefull:
1. DNN can produce drectangles (YOLO, SSD for example are working this way)
2. some affine-transform of drectangle

davisking:
Yeah, the drectangle is weird in that way.  But it's because the rectangle, that has integer coordinates, works this way.  And this makes them behave in the same way.  For example, a rectangle(1,1,1,1) is the rectangle that contains exactly the pixel point(1,1).  If it were defined the opencv way the integer rectangle object would have an odd API.  Since the dlib::rectangle has been around forever we can't change this even if we wanted to anyway.  

The first version of the drectangle I made behaved as you suggest, which is how you would expect something with continuous coordinates to work.  But then when it's converted to or from dlib::rectangle objects it gets all weird and wrong, giving bad and surprising behavior in a number of cases.  So I changed it to work the way it does now to make dlib::rectangle and dlib::drectangle compatible.

Although, the general use case for dlib::rectangle and dlib::drectangle is to represent boxes in images, where the coordinates are always integer valued and never in the range 0..1.   For that it's convenient.  But yeah, as a general rectangle the drectangle is weird.

e-fominov:
Looks like the only solution is to have a new class copied from drectangle with minor changes.
Do you think it will be useful in dlib now?

davisking:
[...]

PR #2735, opened 2023-02-28
fix wrong union calculation in box_overlap_testing.h

smoosbau:
As explained in [this article](https://medium.com/analytics-vidhya/iou-intersection-over-union-705a39e7acef) the union for calculating the IoU of two rectangles must be the unified area of both rectangles. Currently the outer rectangle in that both rectangles fit is used, which is wrong, as this area is larger than the one that has to be used.

### DIFF
--- a/dlib/image_processing/box_overlap_testing.h
+++ b/dlib/image_processing/box_overlap_testing.h
@@ -20,8 +20,8 @@ namespace dlib
         const double inner = a.intersect(b).area();
         if (inner == 0)
             return 0;
-        const double outer = (a+b).area();
-        return inner/outer;
+        const double outer = a.area() + b.area() - inner;
+        return inner / outer;
     }
 
 // ----------------------------------------------------------------------------------------


arrufat:
Oh, that means that all the YOLO trainings I ran (which use IOU thresholds for deciding whether to treat a box as positive or negative) could be improved after this, since the current approach would give slightly lower IOUs.

dlib/image_processing/box_overlap_testing.h
```diff
@@ -17,11 +17,11 @@ namespace dlib
...
     ) 
     {
-        const double inner = a.intersect(b).area();
+        const auto inner = a.intersect(b).area();
         if (inner == 0)
             return 0;
-        const double outer = (a+b).area();
-        return inner/outer;
+        const double outer = double(a.area() + b.area() - inner);
+        return double(inner)/outer;
```
arrufat, line 24: Since we are passing `drectangle`s as parameters, we don't need those casts, or the explicit type for `outer`.

smoosbau: Correct, saw my mistake a few minutes ago. Will change it asap.
Furthermore, the check if inner == 0 should be replaced as comparing doubles with 0 zero is risky.

arrufat: I think that the check with 0 is fine.

arrufat: The thing is that `rectangle` and `drectangle` are either empty or have an area larger to 1, as you can see here:

https://github.com/davisking/dlib/blob/eb5de0d534f637253a7b34b01048b655d5befc20/dlib/geometry/drectangle.h#L95-L120

So, calling the `drectangle::area()` method is always going to return either 0 exactly or a number larger than 1.

dlib/image_processing/box_overlap_testing.h
```diff
@@ -17,11 +17,11 @@ namespace dlib
         const drectangle& b
     ) 
     {
-        const double inner = a.intersect(b).area();
-        if (inner == 0)
+        const auto inner = a.intersect(b).area();
+        if (inner <= std::numeric_limits<double>::epsilon())
```
arrufat, line 21:
```suggestion
        if (inner == 0)
```

arrufat:
By the way, I just checked the map@50 of an already trained model I had and:

- before: 0.5448
- after: 0.5506

As expected, it's slightly larger :)

Thank you for pointing this out.

dlib/image_processing/box_overlap_testing.h
arrufat: See https://github.com/davisking/dlib/pull/2735#discussion_r1120203182

smoosbau: So, you insist on it?
Sadly I did not find a good source stating out why not to do it.

smoosbau: Maybe this can be considered sufficient:
https://www.learncpp.com/cpp-tutorial/relational-operators-and-floating-point-comparisons/
Have a look at "Comparison of calculated floating point values can be problematic" under "Floating point equality" it's explicitly said that == needs to be avoided if floating point values are calculated by any chance.

davisking: It really should compare to 0.  The point of this is just to avoid division by 0.   Also don't use `auto`.  In general mostly don't use `auto`. It's a convenience for the writer of the code.  But code will be read many more times than it is written, so should be organized to make it as clear as possible.  Auto is good when the type itself is really irrelevant or harmful for readability.  Like when capturing a lambda or big complicated template type.

davisking:
Yeah thanks for this.  I knew about this.  The function is just misnamed (the docs do say it does something different than classic IoU).  I didn't change it because I was paranoid about backwards compatibility.  But probably no one will be bothered by the change and I admit it is confusing :)

dlib/image_processing/box_overlap_testing.h
davisking: There isn't anything funny or dangerous about comparing to zero.

dlib/image_processing/box_overlap_testing.h
arrufat: Sorry, I know I tend to abuse auto...

smoosbau: Well, my reaction in short:
I could live with dropping auto for double, allthough I do not fully aggree with your statement, Davis. But that's fine.
What I cannot accept is comparison for equality of a potentially (and it's quite likely in this context) calculated floating point variable with zero.
So, if both of you intend on this type of comparison I will sadly have to delete my Pull Request, as I'm not willing to be git blamable anywhere close to such a line of code.

arrufat: First of all, if you don't touch that line, you'll not be git blamable, so that should not be a problem.

Secondly, even if comparing two doubles for equality can be problematic, and even if you don't agree with comparing to 0 in particular is OK, in the `drectangle::area()` method you'll see that we are returning **exactly** 0 if the rectangle is empty.

From what I understand in that method code, there's no way to compute a 0 value because the area of the rectangle will always be either ≥ 1 (computed) or 0 (exact literal value).

smoosbau: I understood the fact, that the area function will return exactly 0 if (and only if) either width or height are exactly zero. But as drectangles can be scaled by multiplication or division there are cases for which the widht or height can less or equal to double's epsilon, for which all the checks will fail and an area less 1 will be calculated.
So why activley ignoring one of programming's standard rules to **not** compare floating point variable with 0?

arrufat: Again, the width and height methods always add 1 to the difference between the edges, so even if you scale the rectangle down, you'll get 1 for the area. You can read more about this here: https://github.com/davisking/dlib/issues/463

smoosbau: I give up.
The main thing is that the function really calculates the IoU.

arrufat (approved):

dlib/image_processing/box_overlap_testing.h
davisking: > Sorry, I know I tend to abuse auto...

No worries :)

davisking: There really isn't anything wrong with comparing to 0.  People get into trouble when they write code like this though
```
void foo(double a, double b) 
{
    double val = a + b + b; // might end up being equal to just a if b is less than a*epsilon.
    double val2 = val - 2*b; // And say 2*b > a*epsilon, then this will make val2 != val.  So now val2 != a.
    if (val2 == a) 
    {
         // this might never happen depending on the value of b.  This foo function is just fundamentally numerically unsound.
    }
}
```

But that isn't the kind of situation here. And in general you definitely do not want to use the machine epsilon as a proxy for "it's zero".  For double it's only about 2e-16.  But values of double like 1e-30 are perfectly legit and meaningful.  The machine epsilon is really just a convenient way to ask "how much precision is in the mantissa?".  It's not telling us how much precision we have left in the exponent.

So for example, if you wanted to compute the ratio of two lengths `a/b` you want to do just `a/b`.  And if `b` can be 0 you have to decide what you want `a/0` to mean.  So 
```
if (b == 0) return whatever
else return a/b;
```
is what you do.  If instead it was this
```
if (b < std::numeric_limits<double>::epsilon()) return whatever
else return a/b;
```
then the wrong value would be obtained for values like 1e-20/1e-19.  Which should be 10.0, not `whatever`.

In the above foo example, you would use the epsilon in a setup like this
```
void foo(double a, double b) 
{
    double val = a + b + b;
    double val2 = val - 2*b;
    if (std::abs(val2 - a) <= std::numeric_limits<double>::epsilon() * 2*b) 
    {
         // This will execute.
    }
}
```

PR #2737, opened 2023-03-05
Unify all conversions to UTF-32

arrufat:
This PR adds the suggestion from https://github.com/davisking/dlib/pull/2733#issuecomment-1445144139.

In short, it adds a `convert_utf8_to_utf32` overload that performs the conversion without extra memory allocations and allows for calling a custom unary operator while the conversion is being done.

This allows performing `draw_string` on UTF-8 strings more efficiently.

However, I am still facing an error:
The UTF-8 labels are displayed correctly in the image canvas, but not in the "Next Label:" text field, as you can see here:

image

I can't see why this isn't working, so any insights would be helpful.


arrufat:
I figured out why it's not working, the overload for `unichar` is not called... So I added a hack in the main function... But there should be a way to select that overload at compile time.

I can't think of any now...

dlib/unicode/unicode.h
```diff
@@ -211,6 +211,112 @@ namespace dlib
             return -1;
         }
 
+        template <
```
arrufat, line 214: It would be good if we could get rid of the implementation above, but I saw it's used here: https://github.com/davisking/dlib/blob/257b81a17288edece7ffb8bbb5f00a7c2f473494/dlib/unicode/unicode.h#L259

dlib/unicode/unicode.h
```diff
@@ -211,6 +211,112 @@ namespace dlib
...
+            typename charT,
+            typename forward_iterator
+            >
+        int u8_to_u32(
+            charT& result,
+            forward_iterator ibegin,
+            forward_iterator iend
+        )
```
arrufat, line 222: It would be good if we could get rid of the implementation above, but I saw it's used here: https://github.com/davisking/dlib/blob/257b81a17288edece7ffb8bbb5f00a7c2f473494/dlib/unicode/unicode.h#L259

davisking: Should be able to make the above one call this new one by using `std::istream_iterator`.

dlib/unicode/unicode.h
```diff
@@ -499,21 +605,71 @@ namespace dlib
...
 
-        int status;
+        int status = 0;
         unichar ch;
-        while ( (status = u8_to_u32(ch,sin)) > 0)
-            temp.push_back(ch);
+
+        for (size_t i = 0; i < str.size();)
+        {
+            status = u8_to_u32(ch, str.begin() + status, str.end());
```
davisking, line 616: Totally make this function just call the new `convert_utf8_to_utf32()` and accumulate characters into `temp` :D

tools/imglab/src/metadata_editor.cpp
```diff
@@ -656,7 +656,7 @@ on_overlay_rect_selected(
     const image_display::overlay_rect& orect
 )
 {
-    overlay_label.set_text(orect.label);
+    overlay_label.set_text(convert_utf8_to_utf32(orect.label));
```
davisking, line 659: This still needed?

dlib/unicode/unicode.h
```diff
@@ -499,21 +605,71 @@ namespace dlib
...
     }
 
+    template <typename forward_iterator, typename unary_op>
+    inline void convert_utf8_to_utf32(forward_iterator ibegin, forward_iterator iend, unary_op op)
+    {
+        unichar ch;
+        int status = 0;
+
+        // ugly hack
+        if (sizeof(*ibegin) == sizeof(unichar))
```
davisking, line 641: Yeah there are lots of ways to go about this.  You can use SFINAE (https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error#:~:text=Substitution%20failure%20is%20not%20an%20error%20(SFINAE)%20is%20a%20principle,to%20describe%20related%20programming%20techniques.) to control which overload of `convert_utf8_to_utf32()` is invoked.

But seems like 
```
if (!std::is_same<std::decay_t<decltype(*ibegin)>, char>::value) 
```
here would be sufficient?

davisking: Well, probably put this at the top
```
using char_type = std::decay_t<decltype(*ibegin)>;
```

Then also do
```
static_assert (std::is_same<char_type, char>::value || std::is_same<char_type, unichar>::value);
```

Sine it's got to be one of those types or this doesn't really make sense.

dlib/unicode/unicode.h
arrufat: I had to use `std::istreambuf_iterator`, but it worked, thank you!

tools/imglab/src/metadata_editor.cpp
arrufat: Apparently, maybe I can call `convert_utf8_to_utf32` in this method, too.

dlib/unicode/unicode.h
arrufat: Ah, yes that worked, but having put this method inside the `font::draw_string` method, sometimes it's called with `wchar_t`, so that wouldn't compile.

Maybe I should add a conversion from wchar_t to UTF-32 in this method, too… and just name it `convert_to_utf32`?

dlib/unicode/unicode.h
```diff
@@ -493,23 +499,63 @@ namespace dlib
...
+        {
+            while (ibegin != iend)
+                op(*(ibegin++));
+            return;
+        }
+
+        if (std::is_same<char_type, wchar_t>::value)
+        {
+            return;
+        }
```
arrufat, line 526: Here, we should handle this case properly, right?
Or there's another way?

dlib/unicode/unicode.h
arrufat: Right, obvious… :sweat_smile:

dlib/unicode/unicode.h
davisking: Ah, yeah I guess so.  I need to go look at this code more.  I forgot there were 3 char types floating around in there :confused:

tools/imglab/src/metadata_editor.cpp
davisking: Oh right, `set_text()` for `std::string` calls `convert_mbstring_to_wstring()`.  I didn't write that unicode stuff.  When dlib switched away from subversion in 2008 all the version control history was lost :|

Anyway, yeah, I'm not super familiar with all the varied ways characters can be represented.  My impression is that everyone uses unicode, but :shrug:  There are several different versions of unicode and there is this whole multibyte/wchar_t thing which isn't unicode.  It's just some vague seemingly not well defined type.

Anyway, yeah `convert_utf8_to_utf32()` here seems very reasonable. I would say make that be what's inside `set_text()` but I don't know if that's going to break some existing user.

arrufat: Actually, I wanted to suggest dropping that wchar_t stuff and just support Unicode, haha... But yeah, it might break some people's workflow

dlib/unicode/unicode.h
davisking: Ha yeah totally have to handle this somehow.

tools/imglab/src/metadata_editor.cpp
davisking: Yeah I am super confused about who uses `wchar_t`.  I think it's used a lot.  It's the standard type for this kind of thing.  But it's also under-specified, forcing people to do other stuff too.  I definitely don't want to drop it.  I bet people are using it.

dlib/unicode/unicode.h
davisking: Oh look at `convert_wstring_to_utf32()`.  That's got the `wchar_t` to `unichar` conversion.  Stick that in here and then this can just be the one glorious `convert_to_utf32()` function that converts any character sequence to utf32.

tools/imglab/src/metadata_editor.cpp
arrufat: Ok, I will figure out a way to make it work when I get some time :)

dlib/unicode/unicode.h
arrufat: I will have to make that function also take a range and a lambda so that it doesn't allocate, just like the `u8_to_u32`.

arrufat:
I thought this was going to be a small/simple PR, haha :sweat_smile:

tools/imglab/src/metadata_editor.cpp
davisking: I was trying to figure out if Keita Mochizuki (the person who contributed all the nice unicode stuff) is on github or even what his email address is.  I apparently don't have that information anymore and can't find him :(

It was way back on sourceforge and there isn't any useful information there.  Back in the dark ages.

davisking:
> I thought this was going to be a small/simple PR, haha sweat_smile

Yeah non-ASCII is a pain :)

tools/imglab/src/metadata_editor.cpp
arrufat: Oh, that's too bad :(

arrufat:
I think it's done. Let me know what you think when you have time :)

arrufat:
Maybe I am introducing breaking changes for no reason, and I still should provide `convert_utf8_to_utf32` and `convert_wstring_to_utf32`.
I've replaced all instances in the code with `convert_to_utf32`, but maybe some users are using these other two functions?

arrufat:
If tests pass now, it's ready for review :)

dlib/unicode/unicode_abstract.h
```diff
@@ -79,6 +79,31 @@ namespace dlib
...
+            forward_iterator ibegin,
+            forward_iterator iend,
+            unary_op op
+    );
+    /*!
+        requires
+            - forward_iterator points to either char, wchar_t or unichar types
+            - ibegin == iterator pointing to the start of the range
+            - iend == iterator pointing to the end of the range
+            - unary_op == a callable object that takes one parameter
```
davisking, line 93:
```suggestion
            - unary_op == a callable object that takes one unichar parameter
```

arrufat:
If you don't want to change the formatting of the `is_combining_char` function, I put that on a separate commit on purpose so that it's easy to revert :)

davisking:
> If you don't want to change the formatting of the `is_combining_char` function, I put that on a separate commit on purpose so that it's easy to revert :)

Na this is all great, thanks for cleaning it up :)

Issue #2711, opened 2023-01-20
does not build on ppc64le


hramrach:
## Expected Behavior
builds

## Current Behavior
does not build

## Steps to Reproduce

Build dlib after 600923359f504bde8e1747b00be3a257d0470128 on ppc64(le) platform
It contains code such as `inline vector bool int to_bool() const { return x.b; }` which is unintelligible to the compiler.

export 'CFLAGS=-O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g'
/usr/bin/python3.8 setup.py build '--executable=/usr/bin/python3.8 -s'

[  179s] In file included from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io/../image_transforms/../simd/simd4f.h:7,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io/../image_transforms/../simd.h:6,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io/../image_transforms/spatial_filtering.h:13,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io/../image_transforms/image_pyramid.h:10,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io/load_image_dataset.h:19,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io.h:12,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/tools/python/src/opaque_types.h:14,
[  179s]                  from /home/abuild/rpmbuild/BUILD/dlib-19.24.0/tools/python/src/other.cpp:4:
[  179s] /home/abuild/rpmbuild/BUILD/dlib-19.24.0/dlib/../dlib/data_io/../image_transforms/../simd/simd4i.h:53:13: error: invalid use of template-name ‘dlib::vector’ without an argument list
[  179s]    53 |             vector signed int v;
[  179s]       |             ^~~~~~
[...]

PR #2740, opened 2023-03-10
powerpc/altivec: use __vector instead of vector

hramrach:
The gcc altivec extension defines keywords for vector, bool, and pixel, both with underscores and without. However, unless another extension is used the keywords without underscores conflict with standard types.

To resolve the conflict use the keywords with underscores for the altivec vector types.

Fixes: #2711

Note: with this dlib can be compiled but crashes during tests. This crash is not specific to powerpc. There is probably something missing when neither the AMD nor the Arm vector implementation is available.

### DIFF
--- a/dlib/simd/simd4f.h
+++ b/dlib/simd/simd4f.h
@@ -89,7 +89,7 @@ namespace dlib
     class simd4f
     {    
         typedef union {
-            vector float v;
+            __vector float v;
             float x[4];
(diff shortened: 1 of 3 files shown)


davisking:
> __vector is fine, it generally should not make any difference. It's specific to avx/vsx where-as vector is a generic extension.

Yeah I'm surprised it's crashing.

PR #2741, opened 2023-03-12
Improve issue template

arrufat:
Since we already have discussions, and I've noticed that many people are opening bugs without filling the template, or haven't read the documentation, I've decided to add more strict templates.

So far, there are only two
- bug reports
- feature requests

Each one has fills that are required to open an issue.

If people want to talk about random stuff, or ask for help, maybe discussions is a better place for that.

Open to suggestions/improvements.

As always, feel free to close if you don't like this :)


davisking:
Yeah cool. Let's try it out.

The main thing I really want to avoid is people basically writing "can you write my code for me?" posts. A particularly common version of that is "I'm incidentally using dlib and my program isn't working.  Can you debug it for me?".  99.9% of the time I get questions like that it's just that someone has a bug in their code that's unrelated to anything in dlib.  In most cases the act of turning their broken program into some small reproduction of the issue causes them to discover the bug.  Which is no surprise since one of the important steps of debugging is narrowing down the scope of the problem :)

arrufat:
Let's see if forcing people to fill the template properly improves this situation.

Maybe we'll have to tune it, since both templates now add labels to the issues, which will prevent the bot from closing them…

davisking:
Ha, yeah.  I thought about that.  The bot is just there so it's not me personally closing issues :)

arrufat:
> The main thing I really want to avoid is people basically writing "can you write my code for me?"

We already have ChatGPT for that, haha.

PR #2743, opened 2023-03-14
Add optional Python version

arrufat:
I guess a lot of people are building dlib to use it from Python, so add a field to allow them to report which version they're using.


davisking:
Yeah good idea.

PR #2747, opened 2023-03-23
Fix crash when libX11 is missing locale support

asherkin:
This fixes a crash I was running into trying the dnn_face_recognition_ex with XQuartz on a Apple Silicon Mac.

It seems that the prebuilt libX11.6.dylib bottle from Homebrew is missing full locale support, which causes the XwcTextListToTextProperty call to return -2 (XLocaleNotSupported), which caused this code to crash when calling XFree as the XTextProperty value was not initialised.

Window titles seemed generally low value here, so just avoiding the crash seemed like the easiest path forwards.


davisking:
Nice, thanks :)

PR #2748, opened 2023-03-24
test with cmake 3.8 since that's the min required cmake version for dlib

davisking:


davisking:
I've got a server in my basement that runs a bunch of nightly unit tests.  One of them was a cmake 3.8 test.  But I realized this morning that part of that stuff hadn't been running for a few months.  Moving the "does everything really work with the advertised version of cmake?" to github actions.

arrufat:
Why not just `pip3 install cmake==3.8.0`? Like in
https://github.com/davisking/dlib/blob/4d5db9940f1642ca7d1a2711aecf2605a3180189/.github/workflows/build_cpp.yml#L131

Maybe we don't want to get Python involved for this?

arrufat:
Why not just `pip3 install cmake==3.8.0`? Like in
https://github.com/davisking/dlib/blob/4d5db9940f1642ca7d1a2711aecf2605a3180189/.github/workflows/build_cpp.yml#L131

davisking:
> Why not just `pip3 install cmake==3.8.0`? Like in
> 
> https://github.com/davisking/dlib/blob/4d5db9940f1642ca7d1a2711aecf2605a3180189/.github/workflows/build_cpp.yml#L131

That work in linux?  I'm so biased against the cmake in pip on linux I didn't even check :sweat_smile: 

I've got it working now.  But I've discovered that the ffmpeg stuff doesn't work when built in cmake 3.8.0.  Trying to fix that.

arrufat:
I've just tried, and had to run:

```sh
pip install scikit-build
pip install cmake==3.8.2  ## 3.8.0 is not available on Pypi
```

But then, it tries to build CMake and fails... So, if you already got it working, then ignore this :P

davisking:
> I've just tried, and had to run:
> 
> ```shell
> pip install scikit-build
> pip install cmake==3.8.2  ## 3.8.0 is not available on Pypi
> ```
> 
> But then, it tries to build CMake and fails... So, if you already got it working, then ignore this :P

Yeah see, pip cmake on linux is just broke somehow.  It confuses me because it's been like this for years and years.

PR #2750, opened 2023-03-24
Add simple polygon class

arrufat:
I've regularly wanted to fill the insides of polygons in images.
Usually, I resorted to OpenCV's `cv::fillPoly` for that, but it bothered me to depend on OpenCV just for that.

I noticed that dlib had something similar for `dlib::canvas`, so I adapted it to work for image-like objects.

To reuse some code, I made a simple `polygon` class.
Let me know if you like this approach, or if I should rethink how this is implemented.

As an aside, it would be cool if images and canvases provided a common interface, so we could reuse all their drawing functions, whose code is very similar.

### DIFF
--- a/dlib/geometry/polygon.h
+++ b/dlib/geometry/polygon.h
@@ -1,0 +1,92 @@
+// Copyright (C) 2022  Davis E. King (davis@dlib.net), Adrià Arrufat
+// License: Boost Software License   See LICENSE.txt for the full license.
+#ifndef DLIB_POLYGOn_
+#define DLIB_POLYGOn_
+
+#include "polygon_abstract.h"
+#include "rectangle.h"
+#include "vector.h"
+
+namespace dlib
+{
+    class polygon
+    {
+    public:
+        using size_type = std::vector<point>::size_type;
+
+        polygon(const std::vector<point>& points) : points(points) {}
+
+        size_type size() const { return points.size(); }
+
+        point& operator[](const size_type idx) { return points[idx]; }
+        const point& operator[](const size_type idx) const { return points[idx]; }
+        const point& at(const size_type idx) const { return points.at(idx); }
+
+        rectangle get_rect() const
+        {
+            rectangle rect;
+            for (const auto& p : points)
+                rect += p;
+            return rect;
+        }
+
+        double area() const { return polygon_area(points); }
+
+        template <typename alloc>
+        void get_convex_shape (
+            const long top,
+            const long bottom,
+            std::vector<double, alloc>& left_boundary,
+            std::vector<double, alloc>& right_boundary
+        ) const
+        {
+            using std::min;
+            using std::max;
+
+            left_boundary.assign(bottom-top+1, std::numeric_limits<double>::infinity());
+            right_boundary.assign(bottom-top+1, -std::numeric_limits<double>::infinity());
+
+            // trace out the points along the edge of the polynomial and record them
+            for (unsigned long i = 0; i < points.size(); ++i)
+            {
+                const point p1 = points[i];
+                const point p2 = points[(i+1)%points.size()];
+
+                if (p1.y() == p2.y())
+                {
+                    if (top <= p1.y() && p1.y() <= bottom)
+                    {
+                        const long y = p1.y() - top;
+                        const double xmin = min(p1.x(), p2.x());
+                        const double xmax = min(p1.x(), p2.x());
+                        left_boundary[y]  = min(left_boundary[y], xmin);
+                        right_boundary[y] = max(right_boundary[y], xmax);
+                    }
+                }
+                else
+                {
+                    // Here we trace out the line from p1 to p2 and record where it hits.  
+
+                    // x = m*y + b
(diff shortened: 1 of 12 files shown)


dlib/geometry/polygon_abstract.h
```diff
@@ -0,0 +1,99 @@
...
+            const long top,
+            const long bottom,
+            std::vector<double, alloc>& left_boundary,
+            std::vector<double, alloc>& right_boundary
+        ) const;
+        /*!                                                                                
+            requires                                                                       
+                - 0 <= top <= bottom                                                       
+            ensures                                                                        
+                - interprets points as the coordinates defining a convex polygon.  In
```
davisking, line 82: Add a note that this isn't illegal to call if it's non-convex but it just won't yield a convex shape.  It will always yield an enclosing shape though, I think.

Can you add a few unit tests for this class? :)

dlib/geometry/polygon_abstract.h
```diff
@@ -0,0 +1,99 @@
...
+        /*!
+            ensures
+                - returns the point of the polygon at index idx.
+        !*/
+
+        /*!
+            ensures
+                - returns smallest rectangle that contains all points in the polygon.
+        !*/
+        rectangle get_rect() const;
```
davisking, line 61:
```suggestion
        rectangle get_rect() const;
        /*!
            ensures
                - returns smallest rectangle that contains all points in the polygon.
        !*/
```

dlib/geometry/polygon_abstract.h
```diff
@@ -0,0 +1,99 @@
...
+        size_type size() const;
+        /*!
+            ensures
+                - returns the number of points in the polygon.
+        !*/
+
+        point& operator[](const size_type idx);
+        /*!
+            ensures
+                - returns the point of the polygon at index idx.
```
davisking, line 42:
```suggestion
            requires
                - idx < size()
            ensures
                - returns the point of the polygon at index idx.
```

dlib/geometry/polygon_abstract.h
```diff
@@ -0,0 +1,99 @@
...
+        !*/
+
+        const point& operator[](const size_type idx) const;
+        /*!
+            ensures
+                - returns the point of the polygon at index idx.
```
davisking, line 48:
```suggestion
            requires
                - idx < size()
            ensures
                - returns the point of the polygon at index idx.
```

dlib/geometry/polygon_abstract.h
```diff
@@ -0,0 +1,99 @@
...
+        !*/
+
+        const point& at(const size_type idx) const;
+        /*!
+            ensures
+                - returns the point of the polygon at index idx.
```
davisking, line 54:
```suggestion
            ensures
                - returns the point of the polygon at index idx.
            throws
                - std::out_of_range if idx >= size()
```

dlib/geometry/polygon_abstract.h
```diff
@@ -0,0 +1,99 @@
...
+
+                Note that the origin of the coordinate system, i.e. (0,0), is located
+                at the upper left corner.  That is, points such as (1,1) or (3,5)
+                represent locations that are below and to the right of the origin.
+        !*/
+
+    public:
+        using size_type = std::vector<point>::size_type;
+
+        polygon(const std::vector<point>& points);
```
davisking, line 27:
```suggestion
        polygon(std::vector<point> points);
```

dlib/geometry/polygon.h
```diff
@@ -0,0 +1,92 @@
...
+#include "vector.h"
+
+namespace dlib
+{
+    class polygon
+    {
+    public:
+        using size_type = std::vector<point>::size_type;
+
+        polygon(const std::vector<point>& points) : points(points) {}
```
davisking, line 17:
```suggestion
        polygon(std::vector<point> points) : points(std::move(points)) {}
```
That way it can move sometimes vs always forcing a copy.

davisking:
Yeah I was just thinking about how the canvas vs image thing is annoying.  Not super sure how to do it in an efficient way though in general since a canvas is a sub-window of an image.  E.g. X11 will say "only redraw this little part of the window" and then the gui drawing stuff only computes the pixels for that little bit.  That's the wrinkle.

dlib/geometry/polygon_abstract.h
arrufat: Sure! I just put this together quickly last night before going to sleep. I submitted the PR once I got it working for my use case.

davisking: No worries :D

dlib/test/geometry.cpp
```diff
@@ -895,6 +895,39 @@ namespace
 
     }
 
+// ----------------------------------------------------------------------------------------
+
+    void test_polygon()
```
arrufat, line 900: I've added a couple of checks to test the area and make sure that the `get_convex_shape` does what we expect.

Let me know if that is enough.

davisking: Yeah, nice.

davisking (commented):
Thanks, this is great.

PR #2754, opened 2023-03-30
FFmpeg : encoding

pfeatherstone:
Just encoding this time. Trying to keep PR as small as possible.


examples/ffmpeg_video_encoding_ex.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    if (!parser.option("i"))
+    {
+        cout << "Missing -i" << endl;
+        parser.print_options();
+        return 0;
+    }
+
+    const std::string filepath = get_option(parser, "i", "");
```
arrufat, line 53: Here I would've just used this, since we already made sure there's an argument. But maybe that's just nitpicking :P.
```suggestion
    const std::string filepath = parser.option("i").argument();
```

pfeatherstone: I don't mind. I don't use dlib's command line option stuff much so I'll do whatever is best.

pfeatherstone:
@davisking let's get this done. I have a few more coming

davisking:
> @davisking let's get this done. I have a few more coming

Sweet.  Looking now :D

dlib/media/ffmpeg_abstract.h
```diff
@@ -1018,7 +1018,334 @@ namespace dlib
...
+                    - Returns true if successfully encoded, even if callback wasn't invoked.
+                    - Returns false if either EOF, i.e. flush() has been previously called,
+                      or an error occured, in which case is_open() == false.
+            !*/
+
+            void flush();
+            /*!
+                requires
+                    - is_open() == true
+
```
davisking, line 1340:
```suggestion                
```
?  The example below says calling flush() multiple times is safe and has no effect.  Is that right?  Or is there a precondition of `is_open()==true`?

davisking (commented):
Looks real nice :D

dlib/media/sink.h
```diff
@@ -0,0 +1,47 @@
...
+            };
+        }
+
+// ---------------------------------------------------------------------------------------------------
+
+        inline auto sink(std::ostream& out)
+        {
+            return [&](std::size_t ndata, const char* data) {
+                out.write(data, ndata);
+                return true;
```
davisking, line 38: Shouldn't this return `out.good()`?  I'm not sure what the bool is for if not that.

dlib/media/ffmpeg_abstract.h
```diff
@@ -1018,7 +1018,334 @@ namespace dlib
...
+            requires
+                - Byte must be a byte type, e.g. char, int8_t or uint8_t
+                - The lifetime of buf must exceed the lifetime of the encoder 
+                  instance you are passing this to.
+
+            ensures
+                - Creates a "view" type which type-erases &buf and creates
+                  a suitable callback with signature bool(std::size_t, const char*)
+                  which can be passed to encoder.
+                - When the callback is invoked, packet data is appended to buf.
```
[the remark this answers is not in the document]
pfeatherstone: Okidoc

dlib/media/ffmpeg_muxer.h
```diff
@@ -0,0 +1,534 @@
...
+                    state   = ENCODE_DONE;
+                else if (ret == AVERROR_EOF) {
+                    open_   = false;
+                    state   = ENCODE_DONE;
+                }
+                else if (ret < 0)
+                {
+                    open_   = false;
+                    state   = ENCODE_ERROR;
+                    printf("avcodec_receive_packet() failed : %i - `%s`\n", ret, get_av_error(ret).c_str());
```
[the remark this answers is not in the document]
pfeatherstone: That bit of code is technically reachable according to the libavcodec docs so we should definitely have a logging statement of some kind. This isn't the only one though. All the decoding and demuxing stuff have similar debugging statements. If we go through a phase of converting everything to `dlib::logger`, maybe that should be in another PR?

dlib/media/ffmpeg_abstract.h
pfeatherstone: You're right I need to remove that requires clause. Instead it could say something like "becomes a no-op after the first time you call flush()"

dlib/media/sink.h
pfeatherstone: Yep you're right

dlib/media/ffmpeg_muxer.h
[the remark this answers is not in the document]
pfeatherstone: Arg, now i have to use iostreams again. I hate iostreams. Oh well

dlib/media/ffmpeg_muxer.h
```diff
@@ -0,0 +1,534 @@
...
+                    state   = ENCODE_READ_PACKET_THEN_DONE;
+                } else if (ret == AVERROR(EAGAIN)) {
+                    state   = ENCODE_READ_PACKET_THEN_SEND_FRAME;
+                } else if (ret == AVERROR_EOF) {
+                    open_   = false;
+                    state   = ENCODE_DONE;
+                } else {
+                    open_   = false;
+                    state   = ENCODE_ERROR;
+                    printf("avcodec_send_frame() failed : `%s`\n", get_av_error(ret).c_str());
```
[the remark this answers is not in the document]
pfeatherstone: I've converted all print statements to `dlib::logger` statements

dlib/media/ffmpeg_muxer.h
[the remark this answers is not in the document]
pfeatherstone: Converted all print statements to `dlib::logger` statements

pfeatherstone:
I've converted all the `printf` statements to `dlib::logger`. I think it's important to have them.

dlib/media/ffmpeg_demuxer.h
```diff
@@ -449,6 +454,7 @@ namespace dlib
 // ---------------------------------------------------------------------------------------------------
 
         inline decoder::decoder(const args &a)
+        : log(std::make_shared<logger>("ffpeg::decoder"))
```
davisking, line 457:
```suggestion
        : log(std::make_shared<logger>("ffmpeg::decoder"))
```

dlib/media/ffmpeg_demuxer.h
```diff
@@ -640,6 +651,8 @@ namespace dlib
 
         inline demuxer::demuxer(const args &a)
         {
+            st.log = std::make_shared<logger>("ffpeg::demuxer");
```
davisking, line 654:
```suggestion
            st.log = std::make_shared<logger>("ffmpeg::demuxer");
```

dlib/media/ffmpeg_abstract.h
```diff
@@ -1025,33 +1025,25 @@ namespace dlib
...
-                  instance you are passing this to.
-
             ensures
-                - Creates a "view" type which type-erases &out and creates
-                  a suitable callback with signature bool(std::size_t, const char*)
-                  which can be passed to encoder.
-                - When the callback is invoked, packet data is added to the output stream.
+                - returns a function object with signature bool(std::size_t N, const char* data).  When
+                  called that function writes the first N bytes pointed to by data to out.
+                - The returned view is valid only as long as out exists.
```
davisking, line 1046:
```suggestion
                - The returned view is valid only as long as out exists.
                - Returns out.good(). I.e. returns true if the write to the stream succeeded and false otherwise.       
```

davisking (commented):
Nice, looks good.

dlib/media/ffmpeg_muxer.h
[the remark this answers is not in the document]
davisking: Ha yeah, I mean it's not so bad.  But yeah, some day std::format in dlib.  Like 2026 :sob:

pfeatherstone:
@davisking Amazing, thank you for taking time out this weekend

davisking:
> @davisking Amazing, thank you for taking time out this weekend

My pleasure :)

PR #2758, opened 2023-04-04
Add HSV support

arrufat:
I needed to use HSV instead of HSI in dlib, so I added support for it.
Code adapted from [here](https://stackoverflow.com/questions/3018313/algorithm-to-convert-rgb-to-hsv-and-hsv-to-rgb-in-range-0-255-for-both).

I've compared it to the one provided by OpenCV and the results are identical.

### DIFF
--- a/dlib/pixel.h
+++ b/dlib/pixel.h
@@ -86,6 +86,16 @@ namespace dlib
                 - min() == 0 
                 - max() == 255
                 - is_unsigned == true
+            - else if (hsv == true) then
+                - The type T will be a struct with 3 public members of type
+                  unsigned char named "h" "s" and "v".
+                - This type of pixel represents the HSV color space.
+                - num == 3
+                - has_alpha == false
+                - basic_pixel_type == unsigned char
+                - min() == 0
+                - max() == 255
+                - is_unsigned == true
              - else if (lab == true) then
                 - The type T will be a struct with 3 public members of type
                   unsigned char named "l" "a" and "b".
@@ -322,6 +332,41 @@ namespace dlib
             return !(*this == that);
         }
 
+    };
+
+// ----------------------------------------------------------------------------------------
+
+    struct hsv_pixel
+    {
+        /*!
+            WHAT THIS OBJECT REPRESENTS
+                This is a simple struct that represents an HSV colored graphical pixel.
+        !*/
+
+        hsv_pixel (
+        ) {}
+
+        hsv_pixel (
+            unsigned char h_,
+            unsigned char s_,
+            unsigned char v_
+        ) : h(h_), s(s_), v(v_) {}
+
+        unsigned char h;
+        unsigned char s;
+        unsigned char v;
+
+        bool operator == (const hsv_pixel& that) const
+        {
+            return this->h == that.h
+                && this->s == that.s
+                && this->v == that.v;
+        }
+
+        bool operator != (const hsv_pixel& that) const
+        {
+            return !(*this == that);
+        }
     };
 
     // ----------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


arrufat:
Also, I've noticed that `hsi_pixel` is actually in HSL format and not HSI.

The function that gets called to convert between `rgb_pixel` and `hsi_pixel` is even called `RGB2HSL`.
According to the Wikipedia: https://en.wikipedia.org/wiki/HSL_and_HSV
- HSI defines I as the average between R, G and B
- HSL defines L as the mid-point between max({R, G, B}) and min({R, G, B}), which is what dlib does.

Dlib changes the name from L to I when rescaling it between 0 and 255. Maybe we should leave it as is, but it's wrong :P

pfeatherstone:
Annoying how `c++ / ubuntu-18_04-gcc7` regularly gets cancelled immediately.

arrufat:
I will add some tests later :)

davisking:
> Also, I've noticed that `hsi_pixel` is actually in HSL format and not HSI.
> 
> The function that gets called to convert between `rgb_pixel` and `hsi_pixel` is even called `RGB2HSL`. According to the Wikipedia: https://en.wikipedia.org/wiki/HSL_and_HSV
> 
> * HSI defines I as the average between R, G and B
> * HSL defines L as the mid-point between max({R, G, B}) and min({R, G, B}), which is what dlib does.
> 
> Dlib changes the name from L to I when rescaling it between 0 and 255. Maybe we should leave it as is, but it's wrong :P

Ha, yeah there is disagreement with wikipedia.  But I've known a bunch of image processing people who refereed to it the way dlib does it (which is why it's like that).  It's also what tools like photoshop do (or used to do, haven't looked in a long time).

dlib/pixel.h
```diff
@@ -1024,19 +1115,142 @@ namespace dlib
...
+                out.r = p;
+                out.g = q;
+                out.b = in.x;
+                break;
+            case 4:
+                out.r = t;
+                out.g = p;
+                out.b = in.x;
+                break;
+            case 5:
```
arrufat, line 1244:
```suggestion
```
The original code had this extra case, which seems completely useless. Or am I not seeing something?

davisking: Yeah that doesn't do anything.

dlib/pixel.h
```diff
@@ -942,11 +1032,12 @@ namespace dlib
 
     // -----------------------------
 
-        struct HSL
+        // Helper struct used for HSL and HSV
+        struct HSX
```
davisking, line 1036: Everything looks good.  But why change this struct? It doesn't look like this is needed.  Like is there some generic programming I'm not seeing that benefits from this?  It seems like it would be better to have both an `HSL` and `HSV` struct type here.

arrufat: Ah, just to avoid another struct that would be exactly the same. Since this is not exposed to the user... but I can duplicate a HSV struct if you want.

arrufat:
I've noticed that the other color conversions are written in a somewhat old style.
If you want, I could modernize them a bit in a separate PR. I guess they were written before C++11 support.

dlib/pixel.h
davisking: Yeah I mean just to keep the code and function contracts super clear what is what :+1:

davisking:
> I've noticed that the other color conversions are written in a somewhat old style. If you want, I could modernize them a bit in a separate PR. I guess they were written before C++11 support.

Yeah this code is aggressively pre-C++11.  Like that `static_switch` is gnarly now. Whatever you want to do is cool :)

arrufat:
Great, I'll see what I can do, I am no expert with the type traits stuff and `enable_if`, etc… So let's see how that goes :)

PR #2761, opened 2023-04-07
Fix imglab changing the current dir too soon

arrufat:
I tried to use the `--flip` and `--flip-basic` options in imglab, but it was complaining that the XML dataset did not exist. It turns out it was changing the current directory too soon.

### DIFF
--- a/tools/imglab/src/flip_dataset.cpp
+++ b/tools/imglab/src/flip_dataset.cpp
@@ -198,14 +198,13 @@ void flip_dataset(const command_line_parser& parser)
     load_image_dataset_metadata(metadata,datasource);
     orig_metadata = metadata;
 
+    const string metadata_filename = get_parent_directory(file(datasource)).full_name() +
+        directory::get_separator() + "flipped_" + file(datasource).name();
+
     // Set the current directory to be the one that contains the
     // metadata file. We do this because the file might contain
     // file paths which are relative to this folder.
     set_current_dir(get_parent_directory(file(datasource)));
-
-    const string metadata_filename = get_parent_directory(file(datasource)).full_name() +
-        directory::get_separator() + "flipped_" + file(datasource).name();
-
 
     array2d<rgb_pixel> img, temp;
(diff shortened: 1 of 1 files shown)


tools/imglab/src/flip_dataset.cpp
```diff
@@ -198,15 +198,14 @@ void flip_dataset(const command_line_parser& parser)
...
+    const string metadata_filename = get_parent_directory(file(datasource)).full_name() +
+        directory::get_separator() + "flipped_" + file(datasource).name();
+
     // Set the current directory to be the one that contains the
     // metadata file. We do this because the file might contain
     // file paths which are relative to this folder.
     set_current_dir(get_parent_directory(file(datasource)));
 
-    const string metadata_filename = get_parent_directory(file(datasource)).full_name() +
-        directory::get_separator() + "flipped_" + file(datasource).name();
```
davisking, line 207: You sure this does something?  This line doesn't appear to depend on what the current directory is.

davisking: Oh, you want to make it so the new file is in the current working directory of the command being run instead of in the same folder as the original file?

davisking: Yeah nevermind, I see what you mean :+1: :D

davisking:
Thanks for the PR :D

tools/imglab/src/flip_dataset.cpp
arrufat: So, I just ran the program as:
```
imglab --flip dlib_faces_5points/test_cleaned.xml
```

And got the following error message because we changed the working directory just before:
```
Unable to find file dlib_faces_5points/test_cleaned.xml
```

Maybe it could have been addressed by just doing this at the beginning, so it doesn't matter when we change our current directory:

``` cpp
datasource = file(datasource)).full_name();
```

davisking: Yep, Understood. Your change was the right fix :)

PR #2764, opened 2023-04-10
FFmpeg : examples

pfeatherstone:
More audio examples

### DIFF
--- a/dlib/media.h
+++ b/dlib/media.h
@@ -5,6 +5,8 @@
 #ifndef DLIB_MEDIA 
 #define DLIB_MEDIA
 
+#include "test_for_odr_violations.h"
+
 #ifndef DLIB_USE_FFMPEG
 static_assert(false, "This version of dlib isn't built with the FFMPEG wrappers");
 #endif
--- a/dlib/media/ffmpeg_demuxer.h
+++ b/dlib/media/ffmpeg_demuxer.h
@@ -783,13 +783,15 @@ namespace dlib
                 }
                 else if (pCodecCtx->codec_type == AVMEDIA_TYPE_AUDIO)
                 {
+                    check_layout(pCodecCtx.get());
+
                     if (pCodecCtx->sample_rate == 0 ||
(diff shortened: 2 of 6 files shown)


arrufat:
You are amazing!

pfeatherstone:
Somewhere, we need to add ```#include <dlib/config.h>``` at a top level so the statement:
```
#ifndef DLIB_USE_FFMPEG
static_assert(false, "This version of dlib isn't built with the FFMPEG wrappers");
#endif
```
Doesn't catch

davisking:
> You are amazing!

Yeah these are great.

dlib/media.h
```diff
@@ -5,6 +5,8 @@
 #ifndef DLIB_MEDIA 
 #define DLIB_MEDIA
 
+#include "test_for_odr_violations.h"
```
davisking, line 8: How do you all feel about this?  One of the things that bothers me about test_for_odr_violations.h is it makes things in dlib that are header only not work without building with cmake.  Which is mostly fine.  test_for_odr_violations.h does in fact catch a lot of people screwing up and alerts them to their error.  Before it was there I got a constant stream of questions from people violating ODR in the ways noted by the comments in test_for_odr_violations.h.

Sometimes I think it's heavy handed though.  But at the same time, maybe I'm just getting soft because I get way less questions from people running afoul of ODR errors now.

pfeatherstone: I'm just trying to be consistent with the rest of the library

davisking: Yeah just curious.  You would be real surprised by the number of people who build dlib, and then take the resulting .so or .lib or whatever, copy it somewhere, and then copy down some other version of dlib from github or wherever and try to use it.  Or they don't take the config.h from the build but instead use the one from github. This is especially common on windows since there isn't any kind of default way of "installing" a built library there.  So people who don't want to use cmake get up to all kinds of mischief and this stuff gives them a clear error message rather than the awful kind of confusing ODR error they would otherwise get.

pfeatherstone: I'm all for catching errors at build time or link time, however small or unlikely.

PR #2765, opened 2023-04-11
Silence warning about unused variable

arrufat:
I was getting a warning about this.

### DIFF
--- a/dlib/media/ffmpeg_utils.h
+++ b/dlib/media/ffmpeg_utils.h
@@ -1450,11 +1450,13 @@ namespace dlib
                 f = frame(img.nr(), img.nc(), pix_traits<pixel>::fmt, {});
             }
 
+            const size_t imgsize            = img.nr()*img.nc()*sizeof(pixel);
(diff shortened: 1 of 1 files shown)


pfeatherstone:
We probably need one of the github actions workflows to build dlib with `DLIB_ASSERT` enabled

arrufat:
Oh, right, I have my text editor to highlight all references of variables, and since that macro is disabled, I saw nothing. Sorry about that. How can we silence the warning, though? Move it to `DLIB_CASSERT`?

pfeatherstone:
I would instead just add ```(void)imgsize;```

pfeatherstone:
Or add `std::ignore = imgsize`

davisking:
> We probably need one of the github actions workflows to build dlib with `DLIB_ASSERT` enabled

I've got one of the windows builds doing a debug build to at least make sure they compile.  Could have another one in linux.  Trying to not have a huge number of things running since it takes a while.

davisking:
Thanks for the PR :)

PR #2769, opened 2023-04-13
Update widgets and draw_string on images to use convert_to_utf32

arrufat:
Oops, I missed these, sorry :(

### DIFF
--- a/dlib/gui_widgets/widgets.cpp
+++ b/dlib/gui_widgets/widgets.cpp
@@ -190,7 +190,7 @@ namespace dlib
         const std::string& name
     )
     {
-        set_name(convert_mbstring_to_wstring(name));
+        set_name(convert_to_utf32(name));
     }
 
     void toggle_button::
@@ -327,7 +327,7 @@ namespace dlib
         const std::string& text
     )
     {
-        set_text(convert_mbstring_to_wstring(text));
+        set_text(convert_to_utf32(text));
     }
 
     void label::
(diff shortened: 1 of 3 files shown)


arrufat:
Do you think we should template all those methods to be:
```c++
template <typename char_type, typename traits, typename alloc>
set_text (
    const std::basic_string<char_type, traits, alloc>& text
)
{
    set_text(convert_to_utf32(text));
}
```

davisking:
No worries, thanks for the PR :D

davisking:
> Do you think we should template all those methods to be:
> 
> ```c++
> template <typename char_type, typename traits, typename alloc>
> set_text (
>     const std::basic_string<char_type, traits, alloc>& text
> )
> {
>     set_text(convert_to_utf32(text));
> }
> ```

Na.  I've never seen anyone need that for `std::basic_string`.  So I would wait for the need to arise.

arrufat:
I just noticed I didn't add `convert_to_utf32` to `draw_string` for images. So, I am adding it here, too.

I really like how I now don't need to care if my string has non ASCII characters and call `convert_utf8_to_utf32` like before. Now dlib does the right thing automatically.

Thank you for guiding me on how to achieve the allocation-free conversion :)

davisking:
> I just noticed I didn't add `convert_to_utf32` to `draw_string` for images. So, I am adding it here, too.
> 
> I really like how I now don't need to care if my string has non ASCII characters and call `convert_utf8_to_utf32` like before. Now dlib does the right thing automatically.
> 
> Thank you for guiding me on how to achieve the allocation-free conversion :)

My pleasure :D

PR #2770, opened 2023-04-13
Add ffmpeg load_frame

arrufat:
Would you be interested in such functions?

### DIFF
--- a/dlib/media/ffmpeg_demuxer.h
+++ b/dlib/media/ffmpeg_demuxer.h
@@ -1034,6 +1034,19 @@ namespace dlib
 
 // ---------------------------------------------------------------------------------------------------
 
+        template <typename image_type>
+        std::enable_if_t<is_image_type<image_type>::value, void>
+        load_frame(image_type& image, const std::string& file_name)
+        {
+            demuxer reader(file_name);
+            frame f;
+            if (!reader.is_open() || !reader.read(f) || !f.is_image())
+            {
+                throw error("ffmpeg::load_frame: error while loading " + file_name);
+            }
+            convert(f, image);
+        }
+
     }
 }
 
--- a/dlib/media/ffmpeg_demuxer.h
+++ b/dlib/media/ffmpeg_demuxer.h
@@ -1040,10 +1040,10 @@ namespace dlib
         {
             demuxer reader(file_name);
             frame f;
+
             if (!reader.is_open() || !reader.read(f) || !f.is_image())
-            {
                 throw error("ffmpeg::load_frame: error while loading " + file_name);
-            }
+
             convert(f, image);
         }
 
--- a/dlib/media/ffmpeg_muxer.h
+++ b/dlib/media/ffmpeg_muxer.h
@@ -898,6 +898,13 @@ namespace dlib
             const std::unordered_map<std::string, std::string>& codec_options = {}
         )
         {
+            const auto codecs = list_codecs();
+            const auto p = std::find_if(codecs.begin(), codecs.end(), [&](const auto& codec){ return codec_name == codec.codec_name; });
+            if (p == codecs.end())
+                throw error("ffmpeg::save_frame: codec " + codec_name + " not found");
+            if (!p->supports_encoding)
+                throw error("ffmpeg::save_frame: codec " + codec_name + " does not support encoding");
+
             muxer writer(
                 [&]
                 {
--- a/dlib/media/ffmpeg_muxer.h
+++ b/dlib/media/ffmpeg_muxer.h
@@ -900,32 +900,34 @@ namespace dlib
         {
             const auto codecs = list_codecs();
             const auto p = std::find_if(codecs.begin(), codecs.end(), [&](const auto& codec){ return codec_name == codec.codec_name; });
+
             if (p == codecs.end())
                 throw error("ffmpeg::save_frame: codec " + codec_name + " not found");
+
             if (!p->supports_encoding)
                 throw error("ffmpeg::save_frame: codec " + codec_name + " does not support encoding");
 
-            muxer writer(
-                [&]
-                {
-                    muxer::args args;
-                    args.filepath = file_name;
-                    args.enable_image = true;
-                    args.enable_audio = false;
-                    args.args_image.codec_name = codec_name;
-                    args.args_image.h = num_rows(image);
-                    args.args_image.w = num_columns(image);
-                    args.args_image.framerate = 1;
-                    args.args_image.codec_options = codec_options;
-                    args.format_options["update"] = "1";
-                    return args;
-                }());
-            if (not writer.is_open())
-                throw error("ffmpeg::save_frame: error while saving" + file_name);
+            muxer writer([&] {
+                muxer::args args;
+                args.filepath = file_name;
+                args.enable_image = true;
+                args.enable_audio = false;
+                args.args_image.codec_name = codec_name;
+                args.args_image.h = num_rows(image);
+                args.args_image.w = num_columns(image);
+                args.args_image.framerate = 1;
+                args.args_image.codec_options = codec_options;
+                args.format_options["update"] = "1";
+                return args;
+            }());
+
+            if (!writer.is_open())
+                throw error("ffmpeg::save_frame: error while saving " + file_name);
+
             frame f;
             convert(image, f);
             if (!writer.push(std::move(f)))
-                throw error("ffmpeg::save_frame: error while saving" + file_name);
+                throw error("ffmpeg::save_frame: error while saving " + file_name);
         }
     }
 
--- a/dlib/media/ffmpeg_muxer.h
+++ b/dlib/media/ffmpeg_muxer.h
@@ -890,7 +890,7 @@ namespace dlib
 // ---------------------------------------------------------------------------------------------------
 
         template <typename image_type>
-        std::enable_if_t<is_image_type<image_type>::value, void>
+        std::enable_if_t<is_rgb_image<image_type>::value, void>
         save_frame(
             const image_type& image,
             const std::string& file_name,
@@ -916,13 +916,23 @@ namespace dlib
                     args.args_image.h = num_rows(image);
                     args.args_image.w = num_columns(image);
                     args.args_image.framerate = 1;
-                    args.format_options["update"] = "1";
-                    if (pixel_traits<typename image_type::value_type>::has_alpha)
-                        args.args_image.codec_options["fmt"] = "rgba";
+                    args.args_image.codec_options = codec_options;
+                    using traits = pixel_traits<typename image_type::value_type>;
+                    if (traits::bgr_layout)
+                    {
+                        if (traits::has_alpha)
+                            args.args_image.fmt = AV_PIX_FMT_ABGR;
+                        else
+                            args.args_image.fmt = AV_PIX_FMT_BGR24;
+                    }
                     else
-                        args.args_image.codec_options["fmt"] = "rgb";
-                    for (const auto option : codec_options)
-                        args.args_image.codec_options[option.first] = option.second;
+                    {
+                        if (traits::has_alpha)
+                            args.args_image.fmt = AV_PIX_FMT_RGBA;
+                        else
+                            args.args_image.fmt = AV_PIX_FMT_RGB24;
+                    }
+                    args.format_options["update"] = "1";
                     return args;
                 }());
             if (not writer.is_open())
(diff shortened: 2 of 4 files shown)


arrufat:
I think the `ffmpeg::load_frame` would be really useful, since it would allow dlib to swallow almost any kind of image format.

I have my doubts about the `ffmpeg::save_frame`, but it could come in handy.

dlib/media/ffmpeg_muxer.h
```diff
@@ -886,7 +886,54 @@ namespace dlib
...
+                    args.args_image.h = num_rows(image);
+                    args.args_image.w = num_columns(image);
+                    args.args_image.framerate = 1;
+                    args.format_options["update"] = "1";
+                    if (pixel_traits<typename image_type::value_type>::has_alpha)
+                        args.args_image.codec_options["fmt"] = "rgba";
+                    else
+                        args.args_image.codec_options["fmt"] = "rgb";
+                    for (const auto option : codec_options)
+                        args.args_image.codec_options[option.first] = option.second;
```
pfeatherstone, line 925: can't you do ```args.args_image.codec_options = codec_options``` ?

pfeatherstone:
In ```ffmpeg::load_frame()```, there is a philosophical debate about whether or not to `throw`. Like, you might be using an installation of ffmpeg (built from source for example), that only supports h264 decoding for example. In which case, ```load_frame()``` will always fail for image types and that's expected. So should we throw or return a ```std::error_code``` for example?

I'm not sure about ```ffmpeg::save_frame()```.

dlib/media/ffmpeg_muxer.h
```diff
@@ -886,7 +886,54 @@ namespace dlib
...
+                    args.enable_audio = false;
+                    args.args_image.codec_name = codec_name;
+                    args.args_image.h = num_rows(image);
+                    args.args_image.w = num_columns(image);
+                    args.args_image.framerate = 1;
+                    args.format_options["update"] = "1";
+                    if (pixel_traits<typename image_type::value_type>::has_alpha)
+                        args.args_image.codec_options["fmt"] = "rgba";
+                    else
+                        args.args_image.codec_options["fmt"] = "rgb";
```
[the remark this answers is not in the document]
pfeatherstone: I can see people misusing ```format_options``` and ```codec_options``` but you need that level of generalisation to use the full power of ffmpeg. What you did ```args.args_image.codec_options["fmt"]``` is valid C++ but will likely not do what you want.

arrufat:
> In `ffmpeg::load_frame()`, there is a philosophical debate about whether or not to `throw`. Like, you might be using an installation of ffmpeg (built from source for example), that only supports h264 decoding for example. In which case, `load_frame()` will always fail for image types and that's expected. So should we throw or return a `std::error_code`?
> 
> I do this a lot by the way, i build ffmpeg from source with exactly what i need enabled, and nothing more. That gives me a super small binary that does exactly what I want. I don't want to carry around a 40MB binary where only 1% of the code is reachable.
> 
> I'm not sure about `ffmpeg::save_frame()`.

I don't have a strong opinion on that, I just made it behave like the `save_{png,jpeg,webp}` functions.

arrufat:
So far, I've enabled it only for RGB images (with optional alpha channel and BGR layout).
I wanted to avoid dealing with the color conversion in this function, since I would need to allocate new images, and include other headers to be able to use `array2d`.
But if you'd like, I can do it ;)

pfeatherstone:
> So far, I've enabled it only for RGB images (with optional alpha channel and BGR layout). I wanted to avoid dealing with the color conversion in this function, since I would need to allocate new images, and include other headers to be able to use `array2d`. But if you'd like, I can do it ;)

You wouldn't have to do the colour conversion. The encoder options like height, width, format are not those of the input image, but rather what ends up in the codec. Sometimes the codec forces your hand. For example, H264 usually uses YUV format. So though you're feeding the `muxer` object RGB, internally it is converting to YUV.

dlib/media/ffmpeg_muxer.h
```diff
@@ -916,13 +916,23 @@ namespace dlib
...
-                        args.args_image.codec_options["fmt"] = "rgba";
+                    args.args_image.codec_options = codec_options;
+                    using traits = pixel_traits<typename image_type::value_type>;
+                    if (traits::bgr_layout)
+                    {
+                        if (traits::has_alpha)
+                            args.args_image.fmt = AV_PIX_FMT_ABGR;
+                        else
+                            args.args_image.fmt = AV_PIX_FMT_BGR24;
+                    }
```
pfeatherstone, line 927: This isn't required.

dlib/media/ffmpeg_muxer.h
```diff
@@ -916,13 +916,23 @@ namespace dlib
...
                     else
-                        args.args_image.codec_options["fmt"] = "rgb";
-                    for (const auto option : codec_options)
-                        args.args_image.codec_options[option.first] = option.second;
+                    {
+                        if (traits::has_alpha)
+                            args.args_image.fmt = AV_PIX_FMT_RGBA;
+                        else
+                            args.args_image.fmt = AV_PIX_FMT_RGB24;
+                    }
```
pfeatherstone, line 934: This isn't required i'm pretty sure.

dlib/media/ffmpeg_muxer.h
arrufat: If I don't specify the pixel format or the framerate, ffmpeg refuses to process the image.

pfeatherstone: By default, ```encoder_image_args::fmt``` is set to ```AV_PIX_FMT_YUV420P```. It will use that by default. If the codec doesn't support that it will use something that it does. You can try set it to the same format as the input image, as you're doing, but there is no guarantee that the codec supports that and that it will be used. I guess by default, you should try use the same format as your input image.

dlib/media/ffmpeg_muxer.h
pfeatherstone: Maybe this is a good default actually.

dlib/media/ffmpeg_muxer.h
pfeatherstone: > If I don't specify the pixel format or the framerate, ffmpeg refuses to process the image.

Yeah you're right.

pfeatherstone:
I wonder, for image codecs, is the container different to the codec? I'm wondering if you need to manually specify the codec or can it be guessed from the file extension. I might play with ```ffmpeg::save_frame()``` tonight.

arrufat:
> I wonder, for image codecs, is the container different to the codec? I'm wondering if you need to manually specify the codec or can it be guessed from the file extension. I might play with `ffmpeg::save_frame()` tonight.

That would've been great, and my first try didn't have the `codec_name` hoping that it would be guessed from the extension, just like FFmpeg does, but I was getting this error:
`411 ERROR [0] ffmpeg::muxer: Codec none or  cannot be stored in this file`

dlib/media/ffmpeg_muxer.h
pfeatherstone: Maybe split ```load_frame()``` and ```save_frame()``` into separate PRs. I'm pretty happy with ```load_frame()```. I would like to do a bit of tinkering with the latter to see what's best.

pfeatherstone:
> > I wonder, for image codecs, is the container different to the codec? I'm wondering if you need to manually specify the codec or can it be guessed from the file extension. I might play with `ffmpeg::save_frame()` tonight.
> 
> That would've been great, and my first try didn't have the `codec_name` hoping that it would be guessed from the extension, just like FFmpeg does, but I was getting this error: `411 ERROR [0] ffmpeg::muxer: Codec none or cannot be stored in this file`

Yeah I can see where and why the code is doing that. That's not ffmpeg, that's the wrapper enforcing that. I'll have a look later.

pfeatherstone:
The wrappers already check if the specific codec in ```encoder_codec_args``` is supported by the container. We could change the code such that if no codec is specified then choose the best supported codec. For JPEG, PNG etc, there should only be 1 supported codec. So if we choose that, we should be good to go and everything would have been guessed from the file extension.

dlib/media/ffmpeg_muxer.h
```diff
@@ -886,7 +886,51 @@ namespace dlib
...
+        )
+        {
+            const auto codecs = list_codecs();
+            const auto p = std::find_if(codecs.begin(), codecs.end(), [&](const auto& codec){ return codec_name == codec.codec_name; });
+
+            if (p == codecs.end())
+                throw error("ffmpeg::save_frame: codec " + codec_name + " not found");
+
+            if (!p->supports_encoding)
+                throw error("ffmpeg::save_frame: codec " + codec_name + " does not support encoding");
```
arrufat, line 908: @pfeatherstone So I don't need to do this?

pfeatherstone:
Sorry, I didn't get a chance to work on this last night. Probably will this weekend though.

arrufat:
> Sorry, I didn't get a chance to work on this last night. Probably will this weekend though. 

There's no obligation  :) whenever you feel like it.

dlib/media/ffmpeg_muxer.h
```diff
@@ -886,7 +886,51 @@ namespace dlib
...
+
+            muxer writer([&] {
+                muxer::args args;
+                args.filepath = file_name;
+                args.enable_image = true;
+                args.enable_audio = false;
+                args.args_image.codec_name = codec_name;
+                args.args_image.h = num_rows(image);
+                args.args_image.w = num_columns(image);
+                args.args_image.framerate = 1;
```
pfeatherstone, line 918: You could add here if you want:
```
args.args_image.fmt = pix_traits<pixel_type_t<image_type>>::fmt;
```
This will attempt to use the same pixel format in the image file as the input image. If not supported, doesn't matter, it will automatically convert to something that the image container supports.

dlib/media/ffmpeg_demuxer.h
```diff
@@ -1034,6 +1034,19 @@ namespace dlib
 
 // ---------------------------------------------------------------------------------------------------
 
+        template <typename image_type>
+        std::enable_if_t<is_image_type<image_type>::value, void>
+        load_frame(image_type& image, const std::string& file_name)
+        {
+            demuxer reader(file_name);
```
pfeatherstone, line 1041: Here, you do in fact want to set set:
```
decoder_image_args::fmt = pix_traits<pixel_type_t<image_type>>::fmt
```
otherwise `convert()` might fail in:
```
DLIB_ASSERT(f.pixfmt() == pix_traits<pixel>::fmt, "frame doesn't have correct format");
```
By default, `demuxer` converts everything back to RGB. If `image_type` uses BGR, then it will fail. So it's important you tell `demuxer` to convert back to the same pixel format as `image`.

Maybe `demuxer` needs a templated constructor that takes in a pixel type trait for convenience so it can automatically set `decoder_image_args::fmt` to something sensible.

pfeatherstone: Alternatively, we add the function:
```
template<class image_type>
bool demuxer::read(image_type& img);
```
to ```demuxer```. This can automatically convert a `frame` object to a `image_type` and reset the underlying pixel formatter (i.e. the `details::resizer` object)  if required. But then this could potentially override  the ```demuxer::args::args_image``` parameters passed in the constructor.

arrufat: Thank you for the explanation, I think the way you have it now is good already.

arrufat:
I removed the `ffmpeg::save_frame`. I will push it to another PR.

arrufat:
@pfeatherstone the other code is here: https://github.com/arrufat/dlib/tree/add-ffmpeg-save-frame, I will grant you access to my fork :)

davisking:
Sweet

pfeatherstone:
> Here, you do in fact want to set set:
> 
> ```
> decoder_image_args::fmt = pix_traits<pixel_type_t<image_type>>::fmt
> ```
> 
> otherwise `convert()` might fail in:
> 
> ```
> DLIB_ASSERT(f.pixfmt() == pix_traits<pixel>::fmt, "frame doesn't have correct format");
> ```
> 
> By default, `demuxer` converts everything back to RGB. If `image_type` uses BGR, then it will fail. So it's important you tell `demuxer` to convert back to the same pixel format as `image`.
> 


This didn't get addressed.

pfeatherstone:
I've added #2772 to demonstrate.

arrufat:
Oh, sorry, I added it to the ffmpeg::save_frame, instead :S

pfeatherstone:
it would be good to add some unit tests for this.
Also, have you benchmarked this against the normal `load()` functions in dlib with JPEG and  PNG?
I'm sure it will be slower, but I'm interested in knowing how much slower.

arrufat:
I haven't, no. For me, the main goal is not to compete with the existing `load_image` function, but to provide an easy way to load more esoteric image files without explicitly supporting them in dlib.

PR #2771, opened 2023-04-14
Add ffmpeg save frame

arrufat:
Let's see if we can have this one, too :)

### DIFF
--- a/dlib/media/ffmpeg_abstract.h
+++ b/dlib/media/ffmpeg_abstract.h
@@ -1632,7 +1632,7 @@ namespace dlib
 // ---------------------------------------------------------------------------------------------------
 
         template <typename image_type>
-        std::enable_if_t<is_image_type<image_type>::value, void>
+        std::enable_if_t<is_rgb_image<image_type>::value, void>
         save_frame(
             const image_type& image,
             const std::string& file_name,
--- a/dlib/media/ffmpeg_abstract.h
+++ b/dlib/media/ffmpeg_abstract.h
@@ -1632,7 +1632,7 @@ namespace dlib
 // ---------------------------------------------------------------------------------------------------
 
         template <typename image_type>
-        std::enable_if_t<is_rgb_image<image_type>::value, void>
+        std::enable_if_t<is_image_type<image_type>::value, void>
         save_frame(
             const image_type& image,
             const std::string& file_name,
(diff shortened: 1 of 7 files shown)


dlib/media/ffmpeg_abstract.h
```diff
@@ -1013,6 +1013,22 @@ namespace dlib
             !*/
         };
 
+// ---------------------------------------------------------------------------------------------------
+
+        template <typename image_type>
+        std::enable_if_t<is_image_type<image_type>::value, void>
+        load_frame(
```
davisking, line 1020: This is already landed on master.  This PR needs to be merged with master I guess?  There are some things about how PRs work on github that are wacky.

arrufat: Oops... Sorry about that, fixed!

dlib/media/ffmpeg_abstract.h
```diff
@@ -1629,6 +1629,25 @@ namespace dlib
...
+        save_frame(
+            const image_type& image,
+            const std::string& file_name,
+            const std::string& codec_name,
+            const std::unordered_map<std::string, std::string>& codec_options = {}
+        );
+        /*!
+            requires
+                - image_type must be a type conforming to the generic image interface.
+                - codec_name is a valid codec from FFmpeg that suppors encoding.
```
davisking, line 1645:
```suggestion
                - codec_name is a valid codec from FFmpeg that supports encoding.
```

pfeatherstone:
I would hold the fort on this one as we shouldn't have to specify `codec_name` for image codecs. I'll work on a PR in a bit which allows us to do this.

pfeatherstone:
Or we get this in first. We can always remove `codec_name` in a future PR.

pfeatherstone:
@arrufat I forgot to say. I had a look at this a while back. So #2784 adds code to `muxer` which selects the best and default codec for that container if either not specified or not supported. I thought this would do the trick for image formats like PNG, JPEG etc. However, it turns out, ffmpeg uses the same container [image2](https://ffmpeg.org/ffmpeg-formats.html#image2-1) for all image formats, regardless of whether it's PNG, JPEG, TIFF, or whatever. The default codec for `image2` is always JPEG codec. So if you don't specify the codec, and let `muxer` choose the best codec for you, it will always choose JPEG. So doesn't work. We could add some hacky code which inspects the container `muxer` is configured with, check it's of type `image2`, if so look at the file extension, and based on that pick the codec. That should just be a little switch statement so not a big deal.

pfeatherstone:
@arrufat i've gone ahead and done it.

arrufat:
Closing as this is being done in #2800

PR #2773, opened 2023-04-15
FFmpeg : improvements to ffmpeg::list() functions 

pfeatherstone:
All `ffmpeg::list_<>()` functions return a static global instance where possible. This reduces overhead.
I've split `list_input_devices()` into:
* `list_input_device_types()` which returns a static global list of types
* `list_input_device_instances(type)` which returns a run-time look-up of all devices for that type
* ditto for output devices


davisking:
Nice :D

PR #2775, opened 2023-04-15
Add draw solid polygon

arrufat:
This is my first attempt at filling an arbitrary polygon, not just a convex one, using the scanline algorithm. With my current implementation, it doesn't seem obvious to add antialiasing for the edges, though.
However, I use this for drawing a polygon in a mask image, and then blur the mask to alpha blend whatever I want using the polygon, so I get the nice antialiasing just by blurring the mask.

Simultaneously, I fixed some other stuff:
- since I copied the `draw_solid_convex_polygon` from the widgets, where the antialiasing was not optional, when it was disabled, we were painting an extra pixel on the borders.
- there were parts that assumed a `dlib::matrix` (my fault), so I fixed that too.
- I renamed `fill_convex_polygon` to `draw_solid_convex_polygon` (to match all the other `draw_solid_` functions).

Some examples, trying to draw a star:
`antialias = true`
image

`antialias = false` (before and after this PR)
image image

The new `draw_solid_polygon` (note that I was drawing this shape in the previous examples, too)
image

### DIFF
--- a/dlib/image_transforms/draw.h
+++ b/dlib/image_transforms/draw.h
@@ -585,7 +585,6 @@ namespace dlib
 
         valid_area = shrink_rect(valid_area.intersect(bounding_box), 1);
 
-        std::vector<double> prev_intersections;
         for (long y = valid_area.top(); y <= valid_area.bottom(); ++y)
         {
             // Compute the intersections with the scanline
--- a/dlib/image_transforms/draw.h
+++ b/dlib/image_transforms/draw.h
@@ -291,7 +291,7 @@ namespace dlib
         typename pixel_type
     >
     void draw_string (
-        image_type& c,
+        image_type& image,
         const dlib::point& p,
         const std::basic_string<T,traits,alloc>& str,
         const pixel_type& color,
@@ -313,6 +313,7 @@ namespace dlib
 
         const rectangle rect(p, p);
         const font& f = *f_ptr;
+        image_view<image_type> img(image);
 
         long y_offset = rect.top() + f.ascender() - 1;
 
@@ -340,7 +341,7 @@ namespace dlib
             }
 
             // only look at letters in the intersection area
-            if (c.nr() + static_cast<long>(f.height()) < y_offset)
+            if (img.nr() + static_cast<long>(f.height()) < y_offset)
             {
                 // the string is now below our rectangle so we are done
                 return;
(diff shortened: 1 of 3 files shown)


dlib/image_transforms/draw.h
```diff
@@ -423,44 +424,53 @@ namespace dlib
...
         const unsigned char max_alpha = alpha_pixel.alpha;
 
         // we will only want to loop over the part of left_boundary that is part of the
         // valid_area.
-        long top = max(valid_area.top(),bounding_box.top());
-        long bottom = min(valid_area.bottom(),bounding_box.bottom());
+        long left = std::max(valid_area.left(), bounding_box.left());
+        long top = std::max(valid_area.top(), bounding_box.top());
+        long right = std::min(valid_area.right(), bounding_box.right());
+        long bottom = std::min(valid_area.bottom(), bounding_box.bottom());
```
davisking, line 454: This could be simplified a little.  This is computing `valid_area.intersect(bounding_box)`.  So you could replace `valid_area` with that new smaller box and then simplify some of this a bit.  Like if `antialias` was true you could do `grow_rect(valid_area.intersect(bounding_box)),1)` instead of some of the logic below, assuming I'm reading it right.

dlib/image_transforms/draw.h
```diff
@@ -561,17 +571,75 @@ namespace dlib
...
+        image_view<image_type> img(image);
+        const rectangle valid_area(get_rect(image).intersect(area));
+        const rectangle bounding_box = poly.get_rect();
+        if (bounding_box.intersect(valid_area).is_empty())
+            return;
+
+        const long left = std::max(valid_area.left(), bounding_box.left() + 1);
+        const long top = std::max(valid_area.top(), bounding_box.top() + 1);
+        const long right = std::min(valid_area.right(), bounding_box.right() - 1);
+        const long bottom = std::min(valid_area.bottom(), bounding_box.bottom() - 1);
```
davisking, line 600: Same thing here.  Totally use the rectangle methods to do this kind of stuff, it's way easier :)

Like this is `shrink_rect(valid_area.intersect(bounding_box),1)` right?

dlib/image_transforms/draw.h
```diff
@@ -534,44 +533,95 @@ namespace dlib
...
+        const rectangle bounding_box = poly.get_rect();
+        if (bounding_box.intersect(valid_area).is_empty())
+            return;
+
+        valid_area = shrink_rect(valid_area.intersect(bounding_box), 1);
+
+        for (long y = valid_area.top(); y <= valid_area.bottom(); ++y)
+        {
+            // Compute the intersections with the scanline
+            std::vector<double> intersections;
```
davisking, line 591:
```suggestion
        std::vector<double> intersections;
        for (long y = valid_area.top(); y <= valid_area.bottom(); ++y)
        {
            // Compute the intersections with the scanline.
            intersections.clear();
```
Will be way faster since it avoids repeated reallocations.

davisking:
Nice, this is great :)

PR #2777, opened 2023-04-17
FFmpeg : API change to decoder and demuxer

pfeatherstone:
The suggested API changes defer resizing and resampling to `read()` call site. This allows us to more easily implement something like:
```
template<class image_type>
bool read(image_type& img)
```
Also, there is no reason why you should specify in the constructor how you're going to resize or resample stuff.

### DIFF
--- a/dlib/gui_core/gui_core_kernel_1.cpp
+++ b/dlib/gui_core/gui_core_kernel_1.cpp
@@ -2149,7 +2149,7 @@ namespace dlib
     {
         std::wstring wstr;
         get_from_clipboard(wstr);
-        str = convert_wstring_to_utf32(wstr);
+        str = convert_to_utf32(wstr);
     }
 
     void get_from_clipboard (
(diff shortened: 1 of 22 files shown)


pfeatherstone:
Still WIP. Need to amend the tests and check i have actually made things simpler.

pfeatherstone:
Do we really want all the docs in a separate `_abstract.h` file? I'm finding it tiresome synchronizing all the files. Wouldn't it be best to inline the docs in the actual code? We've split declarations from definitions already so maybe it would be ok and the code wouldn't look so disjoint?

davisking:
> Do we really want all the docs in a separate `_abstract.h` file? I'm finding it tiresome synchronizing all the files. Wouldn't it be best to inline the docs in the actual code? We've split declarations from definitions already so maybe it would be ok and the code wouldn't look so disjoint? It's also easier to notice missing or old docs.

That's fine, so long as the headers are organized such that only public API stuff is in the header and/or only public API stuff is at the top.  So like, no `detail` namespace stuff at the top.  Either somehow move that down below the big warning comment about "implementation details below, don't look at this users!" kind of comment or put the detail namespace stuff in a separate detail header.

I.e. whatever makes it so there is some file that when someone looks at they see only information on the public API.

pfeatherstone:
Thanks. I think it's worth getting all of this right before we cut a release. Though I'm pretty sure @arrufat and I are the only users at the moment so we have a bit of freedom I think ...

PR #2779, opened 2023-04-19
circular_buffer : 10X speedup

pfeatherstone:
I don't know why, maybe because of branch prediction, or maybe because the `%` operator is very slow, but this is almost a 10X speedup improvement.

### DIFF
--- a/dlib/sliding_buffer/circular_buffer.h
+++ b/dlib/sliding_buffer/circular_buffer.h
@@ -49,7 +49,10 @@ namespace dlib
                 << "\n\t i:      " << i 
                 << "\n\t size(): " << size()
             );
-            return data[(i+offset)%data.size()]; 
+            auto index = i + offset;
+            if (index >= data.size())
+                index -= data.size();
+            return data[index]; 
         }
 
         const T& operator[] ( unsigned long i) const 
@@ -61,7 +64,10 @@ namespace dlib
                 << "\n\t i:      " << i 
                 << "\n\t size(): " << size()
             );
-            return data[(i+offset)%data.size()]; 
+            auto index = i + offset;
+            if (index >= data.size())
+                index -= data.size();
+            return data[index];
         }
 
         void resize(unsigned long size) 


davisking:
Sweet

Compaile:
Just a quick Question,

what happens when my index is more then twice the size of my buffer? 

--> should we not replace   index -= data.size(); with the % operator and still get the speedup via branch prediction?

pfeatherstone:
I'm fairly certain that `offset` will always be in range `[0, data.size())`

pfeatherstone:
https://github.com/davisking/dlib/blob/48fdace2713619fca57ebb272b238d29eab782aa/dlib/sliding_buffer/circular_buffer.h#L94 and https://github.com/davisking/dlib/blob/48fdace2713619fca57ebb272b238d29eab782aa/dlib/sliding_buffer/circular_buffer.h#L104

pfeatherstone:
It's passing all the unit tests. If there is a bug, I would be keen to know and we can add additional tests to `dtest`

davisking:
Yeah it's fine, not a bug :D

pfeatherstone:
What I like about this class in particular, aside from the fact I use it and it's very useful, is that it's a super simple implementation and of course works really well. If you look at boost's implementation of `circular_buffer`, you get a headache very quickly reading the code, and it's no faster. I don't know where the obsession with writing ridiculously complicated C++ has come from. I wonder if people actually benchmark dumb implementations vs 10000 lines of type traits, nodiscard, noexcept, constexpr, specialisations for trivial types, pod types, etc etc. I mean, boost's version has iterator support, but that's about the only bonus.

Compaile:
No it seems fine,

i was just a bit confused why we check for a out of bounds condition at all and if we check we only check it in the range 0 -> size*2 . 

did seem a bit odd to me.

thanks for the input as always

davisking:
To be super clear, there can't be any out of bounds access because of the code at https://github.com/davisking/dlib/pull/2779#issuecomment-1525766102 that @pfeatherstone pointed out.

Compaile:
Yes indeed, 
thanks again. I was just a bit confused at first when i saw there was a out of bounds check but just for a specific range

davisking:
Yeah, it looks funny at first glance :)

Issue #2595, opened 2022-06-05
[HELP] How to invert the process after alignment?


Mayorc1978:
How do I Invert the alignment after performing edits on the chipped (aligned) images?

From the code the process looks like this:
```
faces = dlib.full_object_detections()
for detection in dets:
    faces.append(sp(img, detection))
images = dlib.get_face_chips(img, faces, size=320)
```
Any API call to **get the transform**, save it in a variable, then revert the alignment? Or a direct call with the chipped image and the cropped image as parameters to get the transform, or a call to embed and render the edited chipped image back on the original uncropped image?

davisking:
There is tooling in dlib to do that.  It's not exposed to python though.  E.g. `dlib::get_mapping_to_chip()`

Mayorc1978:
> There is tooling in dlib to do that. It's not exposed to python though. E.g. `dlib::get_mapping_to_chip()`

How do I make a wrapper for that method in Python?

davisking:
The python bindings are over in the tools/python/src folder and use pybind11.

PR #2781, opened 2023-04-20
Add insert_image_chip

arrufat:
I think this function would be very convenient in the Python bindings to be able to revert the `extract_image_chip` without exposing the `transform` types in Python.

Exposing the `transform` tooling could be done at a latter time if there's a need for that, but I guess this function should cover most of the basic cases, such as #2595.

### DIFF
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -2015,17 +2015,16 @@ namespace dlib
         const interpolation_type& interp
     )
     {
-        image_view<image_type1> img(image);
-        const_image_view<image_type2> chp(chip);
-        DLIB_CASSERT(chp.nr() == location.rows && chp.nc() == location.cols,
+        image_view<image_type1> vimg(image);
+        const_image_view<image_type2> vchip(chip);
+        DLIB_CASSERT(vchip.nr() == location.rows && vchip.nc() == location.cols,
                      "The chip and the location do not have the same size.")
-        // Figure out which rectangle contains the rotated rectangle
         const auto tf = get_mapping_to_chip(location);
-        for (long r = 0; r < img.nr(); ++r)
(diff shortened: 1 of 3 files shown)


arrufat:
If you think it's needed, I could add a C++ version, but it's so easy to implement, that I don't think it's needed. The main reason why I added it to Python only, instead of exposing the `transform` type, was to avoid looping in Python.

arrufat:
Hm, that's not right, I need to perform some kind of interpolation; otherwise I don't fill the full area... I will think more about it.

dlib/image_transforms/interpolation.h
```diff
@@ -2001,6 +2001,50 @@ namespace dlib
...
+        const_image_view<image_type2> vchip(chip);
+        DLIB_CASSERT(static_cast<unsigned long>(vchip.nr()) == location.rows &&
+                     static_cast<unsigned long>(vchip.nc()) == location.cols,
+                     "The chip and the location do not have the same size.")
+        const auto tf = get_mapping_to_chip(location);
+        for (long r = 0; r < vimg.nr(); ++r)
+        {
+            for (long c = 0; c < vimg.nc(); ++c)
+            {
+                interp(vchip, tf(dlib::vector<double, 2>(c, r)), vimg[r][c]);
```
davisking, line 2028:
```suggestion
                interp(vchip, tf(dpoint(c, r)), vimg[r][c]);
```
There is a `dpoint` alias for this (i.e. double point)

davisking:
> Hm, that's not right, I need to perform some kind of interpolation; otherwise I don't fill the full area... I will think more about it.

Looks reasonable.  You happy with it or still trying stuff?

arrufat:
Ah, I'm happy with it now :)

davisking:
> Ah, I'm happy with it now :)

Sweet.  Thanks for the PR :)

PR #2782, opened 2023-04-22
Use dpoint instead of dlib::vector<double, 2>

arrufat:
I've updated the rest of the interpolation file to use `dpoint`.
Maybe that alias was introduced after all these functions?


davisking:
> I've updated the rest of the interpolation file to use `dpoint`. Maybe that alias was introduced after all these functions?

Oh yeah, forgot that came a bit later.  Yep :D 

Anyway, thanks for the PR :)

PR #2784, opened 2023-04-23
[FFmpeg] decoding and demuxing improvements

pfeatherstone:
Few things:
* resizing and resampling is done at time of read. So resizing and resampling arguments are passed to `read()` not set in constructor of `decoder` or `demuxer`
* Added templated `read()` function to `decoder` and `demuxer` so we can read directly into dlib objects.
* Split the unit tests as much as possible into decoding, demuxing, encoding and muxing. It's doing more compute overall, but i think the tests are slightly easier to follow, maintain, extend, etc. 
* I've cut the unit test example videos to make the unit tests run faster (or overall the same)
* I've added more unit tests


pfeatherstone:
Note, this is WIP, I haven't updated the examples, so expect tests to fail.

pfeatherstone:
Nearly there, will work on this tonight maybe and iron out the rest.

dlib/media/ffmpeg_demuxer.h
```diff
@@ -1457,15 +1523,12 @@ namespace dlib
         std::enable_if_t<is_image_type<image_type>::value, void>
         load_frame(image_type& image, const std::string& file_name)
         {
-            demuxer reader(file_name);
-            frame f;
-
-            if (!reader.is_open() || !reader.read(f) || !f.is_image())
+            if (!demuxer(image, video_enabled, audio_disabled).read(image))
```
pfeatherstone, line 1526: Haven't tested this yet but I bet it doesn't compile.
Need to write `demuxer({image, video_enabled, audio_disabled})` i'm pretty sure

dlib/media/ffmpeg_demuxer.h
```diff
@@ -1519,11 +1519,13 @@ namespace dlib
...
 // ---------------------------------------------------------------------------------------------------
 
-        template <typename image_type>
-        std::enable_if_t<is_image_type<image_type>::value, void>
-        load_frame(image_type& image, const std::string& file_name)
+        template <
+          class image_type,
+          is_image_check<image_type>
+        >
+        inline void load_frame(image_type& image, const std::string& file_name)
```
pfeatherstone, line 1526: Thank you @arrufat for spotting the declaration bug.

pfeatherstone:
Sorry i didn't get a chance to work on this last night. Will do tonight.

davisking:
> Sorry i didn't get a chance to work on this last night. Will do tonight.

No worries.  No deadlines here :D

pfeatherstone:
There is still one thing I want to do which will substantially reduce memory allocations. Change `dlib::ffmpeg::details::resizer` and `dlib::ffmpeg::details::resampler` such that they can scale or resample directly into dlib objects. At the moment, we have to resize or resample into `dlib::ffmpeg::frame` objects, then basically memcpy into dlib objects. So we are doing 2X too many allocations and frees. I don't like this and it's definitely fixable. I should find time to work on this this weekend. This has zero impact on public API.

pfeatherstone:
Also, I would be keen to get someone's opinion on the following. Currently, `encoder`'s API is:

* push frames -> callback(size_t nencoded_bytes, const char* encoded_bytes) gets called
 
I was wondering if we wanted something similar for `decoder`:

* push raw data -> callback(frame& f) gets called

At the moment, the API for `decoder` is to call `read()` until either the object is closed or you get typed EAGAIN return code, which is a bit closer to how `demuxer` works. 

So we either strive for `decoder`'s API to be similar to `demuxer` or to `encoder`. I'm actually leaning more towards `encoder`. Note, we would be leaving `demuxer`'s and `muxer`'s API alone.

pfeatherstone:
Hmm, using `sws_scale()` directly with dlib objects can sometimes crash because ffmpeg assumes all memory is over-aligned. https://trac.ffmpeg.org/ticket/8747 confirms that this can happen. Need to see if there is a workaround.

davisking:
> Also, I would be keen to get someone's opinion on the following. Currently, `encoder`'s API is:
> 
> * push frames -> callback(size_t nencoded_bytes, const char* encoded_bytes) gets called
> 
> I was wondering if we wanted something similar for `decoder`:
> 
> * push raw data -> callback(frame& f) gets called
> 
> At the moment, the API for `decoder` is to call `read()` until either the object is closed or you get typed EAGAIN return code, which is a bit closer to how `demuxer` works.
> 
> So we either strive for `decoder`'s API to be similar to `demuxer` or to `encoder`. I'm actually leaning more towards `encoder`. Note, we would be leaving `demuxer`'s and `muxer`'s API alone.
> 
> Just a thought, we would want the callback for `decoder` to be generic so it could take signature `void(frame&)` or `void(image_type& img)` where `image_type` is a dlib object. That should be doable with some type erasure.

Yeah I would go more for `decoder` being like `encoder` too.  I.e. someone just calls `decoder::push_encoded()` in a loop and the get called back whenever there is a new frames worth of data.

pfeatherstone:
> > Also, I would be keen to get someone's opinion on the following. Currently, `encoder`'s API is:
> > 
> > * push frames -> callback(size_t nencoded_bytes, const char* encoded_bytes) gets called
> > 
> > I was wondering if we wanted something similar for `decoder`:
> > 
> > * push raw data -> callback(frame& f) gets called
> > 
> > At the moment, the API for `decoder` is to call `read()` until either the object is closed or you get typed EAGAIN return code, which is a bit closer to how `demuxer` works.
> > So we either strive for `decoder`'s API to be similar to `demuxer` or to `encoder`. I'm actually leaning more towards `encoder`. Note, we would be leaving `demuxer`'s and `muxer`'s API alone.
> > Just a thought, we would want the callback for `decoder` to be generic so it could take signature `void(frame&)` or `void(image_type& img)` where `image_type` is a dlib object. That should be doable with some type erasure.
> 
> Yeah I would go more for `decoder` being like `encoder` too. I.e. someone just calls `decoder::push_encoded()` in a loop and the get called back whenever there is a new frames worth of data.

Also, do we like the callback being set in the constructor, or shall we pass it by reference in `push()` method? We could do the same for both `encoder` and `decoder`. There is no particular reason why it needs to be set in constructor. It also means we don't have to store it if we pass it to `push()`

davisking:
> > > Also, I would be keen to get someone's opinion on the following. Currently, `encoder`'s API is:
> > > 
> > > * push frames -> callback(size_t nencoded_bytes, const char* encoded_bytes) gets called
> > > 
> > > I was wondering if we wanted something similar for `decoder`:
> > > 
> > > * push raw data -> callback(frame& f) gets called
> > > 
> > > At the moment, the API for `decoder` is to call `read()` until either the object is closed or you get typed EAGAIN return code, which is a bit closer to how `demuxer` works.
> > > So we either strive for `decoder`'s API to be similar to `demuxer` or to `encoder`. I'm actually leaning more towards `encoder`. Note, we would be leaving `demuxer`'s and `muxer`'s API alone.
> > > Just a thought, we would want the callback for `decoder` to be generic so it could take signature `void(frame&)` or `void(image_type& img)` where `image_type` is a dlib object. That should be doable with some type erasure.
> > 
> > 
> > Yeah I would go more for `decoder` being like `encoder` too. I.e. someone just calls `decoder::push_encoded()` in a loop and the get called back whenever there is a new frames worth of data.
> 
> Also, do we like the callback being set in the constructor, or shall we pass it by reference in `push()` method? We could do the same for both `encoder` and `decoder`. There is no particular reason why it needs to be set in constructor. It also means we don't have to store it if we pass it to `push()`.
> 
> In general, I like making things as stateless as possible.

Yeah it would be cool if the callback wasn't given to the constructor.

pfeatherstone:
Sorry about delays again. I'm only getting like 20mins per day to work on this...

arrufat:
> Sorry about delays again. I'm only getting like 20mins per day to work on this...

What!? This is unacceptable!

Seriously, don't apologize, we're all doing this for fun in our spare time :)

davisking:
Indeed :D

pfeatherstone:
> Yeah I would go more for decoder being like encoder too. I.e. someone just calls decoder::push_encoded() in a loop and the get called back whenever there is a new frames worth of data.

While doing this for `decoder` I found myself needing some type traits for callable objects so that I could wrap different callbacks into a uniform way, a bit like what the `sink()` functions do, except it's a bit more difficult here since you want to support callbacks with signature `void(frame& f)` and others with signature `void(image_type& img)` where `image_type` satisfies image interface. Depending on the signature, you want the wrapper callaback to do different things. Anyway, I wrote those callable type traits and waiting for #2790 to merge so I can do this properly.

pfeatherstone:
Yeah this PR is too large now. I'm gonna have to split it I think.

pfeatherstone:
I might need a couple more SFINAE tricks for earlier compilers (I've been testing with gcc 12). Not an issue, I can see what's required.

pfeatherstone:
#2794 has what i need to fix the remaining issues. Also, maybe we can try merge this PR rather than split. It's gonna create me more work and like it's all there.

davisking:
> #2794 has what i need to fix the remaining issues. Also, maybe we can try merge this PR rather than split. It's gonna create me more work and like it's all there.

Yeah, that's fine.  Let me know when it's ready :+1:

pfeatherstone:
The docs for `encoder`  should say that you mustn't call `encoder::push` in callback passed to `encoder::push`, i.e. you can't create a recursive loop inside `push()`. Same goes for `decoder::push`. This can be checked by adding an extra `bool` member variable, which gets turned on at start of `push` and turned off at the end. Then add a `DLIB_ASSERT` at the beginning of `push()` to check the bool is false before you do anything else. This should check you're not in a recursive loop.

davisking:
> I'll go through everything tonight or tomorrow, check I'm happy. The only thing is that now that `decoder` uses callback based API, there are so many ways you can use it and the examples only show 2 of them. I'm wondering if examples need to be more exhaustive. In any case that can be a future PR.

Na examples just need to get people oriented and highlight whatever it is you want to highlight, not be any kind of exhaustive show of all that is possible.  The API docs are the thing that exhaustively shows everything.  Examples are good to show example use cases.  Like "hey you can put these 3 things together like this and it's super useful!" kind of deal.  So the examples that are there now are plenty.

pfeatherstone:
@davisking Hi, could you have a first review?

davisking:
> @davisking Hi, could you have a first review?

Yeah, I'll look later tonight probably :+1:

examples/ffmpeg_rtsp_ex.cpp
```diff
@@ -46,13 +48,6 @@ try
         return 0;
     }
 
-    if (!parser.option("i"))
-    {
-        cout << "Missing -i" << endl;
-        parser.print_options();
-        return 0;
-    }
-
```
davisking, line 55: Keep this, since `i` is a required option.  The error message you get otherwise is less clear.

pfeatherstone: Yeah I was wondering if the command line parser could be improved at some later stage. otherwise the error message is a stack trace.

davisking: Yeah.  There are a bunch of check functions you can call which throw errors that look nice on the terminal.  Use those to validate the CLI.  And we can add more of them for whatever anyone wants.

davisking:
#> @davisking Hi, could you have a first review?

Just looked this over for a while.  Looks great.  Let me know when you want to merge it :D

pfeatherstone:
> #> @davisking Hi, could you have a first review?
> 
> Just looked this over for a while. Looks great. Let me know when you want to merge it :D

Let's go for it.

PR #2785, opened 2023-04-24
Fix compilation error in ffmpeg::load_frame due to doc inlining

arrufat:


dlib/media/ffmpeg_demuxer.h
```diff
@@ -1453,9 +1453,11 @@ namespace dlib
...
 // ---------------------------------------------------------------------------------------------------
 
-        template <typename image_type>
-        std::enable_if_t<is_image_type<image_type>::value, void>
-        load_frame(image_type& image, const std::string& file_name)
+        template <
+          class image_type,
+          is_image_check<image_type>
+        >
+        void load_frame(image_type& image, const std::string& file_name)
```
pfeatherstone, line 1460: Don't you need to add `inline` here too?

arrufat:
I've just seen that @pfeatherstone just added this change in #2784. Closing.

PR #2790, opened 2023-05-04
[Meta] added callable traits

pfeatherstone:
This is going to come in handy in a second.

### DIFF
--- a/dlib/type_traits.h
+++ b/dlib/type_traits.h
@@ -95,7 +95,7 @@ namespace dlib
     template <typename T> 
     using is_built_in_scalar_type = std::is_arithmetic<T>;
 
- // ----------------------------------------------------------------------------------------
+// ----------------------------------------------------------------------------------------
 
     template<class Byte>
     using is_byte = std::integral_constant<bool, std::is_same<Byte,char>::value
@@ -114,7 +114,7 @@ namespace dlib
     template <typename T>
     struct basic_type { using type = remove_cvref_t<T>; };
 
- // ----------------------------------------------------------------------------------------
+// ----------------------------------------------------------------------------------------
 
     template<class...>
     struct conjunction : std::true_type {};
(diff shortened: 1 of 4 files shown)


pfeatherstone:
The only thing is that this doesn't work with `std::bind_front`, `dlib::bind_front` (back variants too). Then again, nor does ` boost::callable_traits::args_t`. But those are niche use cases.

pfeatherstone:
By the way, this should be ready to go

dlib/type_traits.h
```diff
@@ -210,8 +210,102 @@ namespace dlib
...
+#else
+        template<std::size_t I, class... Ts>
+        struct nth_type;
+
+        template<std::size_t I, class T0, class... Ts>
+        struct nth_type<I, T0, Ts...> { using type = typename nth_type<I-1, Ts...>::type; };
+
+        template<class T0, class... Ts>
+        struct nth_type<0, T0, Ts...> { using type = T0; };
+#endif
```
davisking, line 245:
```suggestion
        template<std::size_t I, class... Ts>
        struct nth_type;

        template<std::size_t I, class T0, class... Ts>
        struct nth_type<I, T0, Ts...> { using type = typename nth_type<I-1, Ts...>::type; };

        template<class T0, class... Ts>
        struct nth_type<0, T0, Ts...> { using type = T0; };
```

Why not this?  If `__type_pack_element` was a "when we get to C++-some-new-version" use this and delete the `#else` block then that's one thing, but `__type_pack_element` is a non-standard language extension isn't it?

davisking (commented):
Looks cool.  Add docs for the things though :)

davisking (commented):
Looks cool.  Add docs for the things though :)

dlib/type_traits.h
pfeatherstone: Yes but it's supported by pretty much every version of clang and later versions of GCC. It's an O(1) lookup so it's the fastest possible implementation. It this were used heavily in template meta-programming then it could make a huge difference. It's also appropriately guarded by the `__has_builtin` macro so it's harmless.

pfeatherstone: We should strive to avoid template recursion as it absolutely kills compilation times.

pfeatherstone: If you disagree with using `__type_pack_element` then we can go without. The implementation is a bit naive but at the moment it's ok because it's only used in `type_safe_union`, which is probably only used for like up to 5 types, and now `callable_traits` which again is only for a handful of types. If it were used by the super heavily templated dnn stuff, then it could be a game changer if we used a proper implementation, or `__type_pack_element` which is the most optimal solution.

davisking: Na, you are right. Making the build faster is worth it 😁

dlib/type_traits.h
```diff
@@ -210,8 +210,135 @@ namespace dlib
...
+    }
+
+    template<std::size_t I, class... Ts>
+    struct nth_type;
+    /*!
+        WHAT THIS OBJECT REPRESENTS
+            This is a type trait for getting the n'th argument of a parameter pack.
```
davisking, line 252:
```suggestion
            This is a type trait for getting the n'th argument of a parameter pack.
            In particular, nth_type<n, some_types...>::type is the nth type in some_types.
```

dlib/type_traits.h
```diff
@@ -210,8 +210,136 @@ namespace dlib
...
+                args        : a parameter pack packaged in a types_<> meta container containing
+                              all the function argument types
+            It also provides the following static members:
+                nargs       : the number of function arguments
+            
+            For example, a function type F with signature R(T1, T2, T3) has the following traits:
+                callable_traits<F>::return_type == R
+                callable_traits<F>::args        == types_<T1,T2,T3>
+                callable_traits<F>::nargs       == 3
+    !*/
```
davisking, line 283: This is nice

dlib/type_traits.h
```diff
@@ -210,8 +210,136 @@ namespace dlib
...
+            This returns a compile-time constant for callable_traits::nargs 
+            packaged in a std::integral_constant.
+    !*/
+
+    template<class Callable>
+    using callable_return = typename callable_traits<Callable>::return_type;
+    /*!
+        WHAT THIS OBJECT REPRESENTS
+            This is a type alias for callable_traits::return_type
+    !*/
```
davisking, line 339:
```suggestion
    template<class Callable>
    using callable_args = typename callable_traits<Callable>::args;

    template<std::size_t I, class Callable>
    using callable_arg = nth_type_t<I, callable_args<Callable>>;
    
    template<class Callable>
    using callable_nargs = std::integral_constant<std::size_t, callable_traits<Callable>::nargs>;

    template<class Callable>
    using callable_return = typename callable_traits<Callable>::return_type;
```
No docs when the only thing to say is to restate what the code trivially says :)

PR #2793, opened 2023-05-06
[FFT] split FFT API into different layers

pfeatherstone:
* Splits the FFT API into different files for: span-style, STL and matrix overloads
* Span-style and STL overloads, which are building blocks for other overloads, are put in `dlib/fft/` folder
* Backends are compiled rather than header only.

This allows users to use dlib's FFT functions without having to include the entire matrix tools. It also modularises the API into layers and facilitates future APIs.


davisking:
Yeah, nice.

PR #2794, opened 2023-05-08
[Meta] fixed callable_traits

pfeatherstone:
`callable_traits` works on non-callable types, in which case it simply provides `callable_traits::is_callable == false`


pfeatherstone:
Oh, and added `is_callable`, not to be confused with `is_invocable`

davisking:
Thanks, looks good.

PR #2796, opened 2023-05-16
Fix CUDA memory alignment issue in `loss_multiclass_log_per_pixel_weighted`

reunanen:
Problem: CUDA error having `code: 716, reason: misaligned address`

Why does this happen: `float` data [should be aligned to 4 bytes](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses), but this is not guaranteed when the truth data is `uint16_t` and is in the buffer before the weights data

Solution: re-arrange the buffer so that the weights data (type `float`) comes before the truth data (type `uint16_t`), essentially making it so that the `float` data is correctly aligned to 4 bytes

### DIFF
--- a/dlib/cuda/cuda_dlib.h
+++ b/dlib/cuda/cuda_dlib.h
@@ -707,7 +707,7 @@ namespace dlib
 
                 cuda_data_ptr<float> loss_buf = static_pointer_cast<float>(buf, 1);
                 buf = buf+sizeof(float);
-                const auto weights_offset = subnetwork_output.num_samples() * bytes_per_plane;
+                const auto truth_offset = subnetwork_output.num_samples() * weight_bytes_per_plane;
                 // copy the truth data into a cuda buffer.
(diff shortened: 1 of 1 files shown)


reunanen:
Some background: I was running on an older version of dlib without any trouble. I updated to essentially the latest `master` version (in order to try out the Barlow Twins loss actually), and my test automation suite started to fail because of this problem. Not really sure what exact change in `master` caused this problem to surface; I didn't really touch (the versions of) CUDA or cuDNN or my test automation suite or anything.

reunanen:
Ah, I was previously on an (old) version that only had a CPU implementation for this specific loss layer. So that's why.

davisking:
Yeah, nice catch.  Thanks for the PR :)

arrufat:
@reunanen thank you for cleaning up after me :sweat_smile:

Let me know how the Barlow Twins loss works for you. I've never used for real-world stuff, it would be nice to see it solving non-academic problems.

PR #2800, opened 2023-05-22
[FFmpeg] added save_frame() and improved picking of codecs when not specified

pfeatherstone:
- Added `save_frame()`
- Added more code inside `muxer` for picking appropriate codecs when either not specified or not suitable.

### DIFF
--- a/dlib/media/ffmpeg_details.h
+++ b/dlib/media/ffmpeg_details.h
@@ -508,6 +508,14 @@ namespace dlib { namespace ffmpeg { namespace details
                 return AV_CODEC_ID_AAC;
             else if (ext == "ac3")
                 return AV_CODEC_ID_AC3;
+            else if (ext == "jls")
+                return AV_CODEC_ID_JPEGLS;
+            else if (ext == "jp2")
+                return AV_CODEC_ID_JPEG2000;
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(59, 37, 100)
+            else if (ext == "jxl")
+                return AV_CODEC_ID_JPEGXL;
+#endif
         }
 
         return AV_CODEC_ID_NONE;
--- a/dlib/media/ffmpeg_details.h
+++ b/dlib/media/ffmpeg_details.h
@@ -482,6 +482,39 @@ namespace dlib { namespace ffmpeg { namespace details
 
 // ---------------------------------------------------------------------------------------------------
 
+    inline AVCodecID pick_codec_from_filename(const std::string& filename)
+    {
+        const auto ext_pos = filename.find_last_of(".");
+
+        if (ext_pos != std::string::npos)
+        {
+            const std::string ext = filename.substr(ext_pos + 1);
+
+            if (ext == "png" || ext == "PNG")
+                return AV_CODEC_ID_PNG;
+            else if (ext == "jpeg" || ext == "jpg" || ext == "JPEG")
+                return AV_CODEC_ID_MJPEG;
+            else if (ext == "tiff")
+                return AV_CODEC_ID_TIFF;
+            else if (ext == "webp")
+                return AV_CODEC_ID_WEBP;
+            else if (ext == "bmp")
+                return AV_CODEC_ID_BMP;
+            else if (ext == "h264")
+                return AV_CODEC_ID_H264;
+            else if (ext == "h265" || ext == "hevc")
+                return AV_CODEC_ID_H265;
+            else if (ext == "aac")
+                return AV_CODEC_ID_AAC;
+            else if (ext == "ac3")
+                return AV_CODEC_ID_AC3;
+        }
+
+        return AV_CODEC_ID_NONE;
+    }
+
+// ---------------------------------------------------------------------------------------------------
+
 }}}
 
 #endif //DLIB_FFMPEG_DETAILS
--- a/dlib/media/ffmpeg_muxer.h
+++ b/dlib/media/ffmpeg_muxer.h
@@ -822,9 +822,67 @@ namespace dlib
                 }
 #endif
             }
-        }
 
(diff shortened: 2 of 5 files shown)


dlib/media/ffmpeg_details.h
```diff
@@ -480,6 +480,39 @@ namespace dlib { namespace ffmpeg { namespace details
...
+                return AV_CODEC_ID_BMP;
+            else if (ext == "h264")
+                return AV_CODEC_ID_H264;
+            else if (ext == "h265" || ext == "hevc")
+                return AV_CODEC_ID_H265;
+            else if (ext == "aac")
+                return AV_CODEC_ID_AAC;
+            else if (ext == "ac3")
+                return AV_CODEC_ID_AC3;
+        }
```
arrufat, line 511: How about adding JPEG XL support?
```suggestion
            else if (ext == "jxl")
                return AV_CODEC_ID_JPEGXL;
        }
```

dlib/test/ffmpeg.cpp
```diff
@@ -146,6 +146,21 @@ namespace
         DLIB_TEST_MSG(similarity > 25.0, "psnr " << similarity);
     }
 
+    template<class pixel_type>
+    void test_load_save_frame(const std::string& filename)
+    {
+        matrix<pixel_type> img1;
```
arrufat, line 152:
```suggestion
        matrix<pixel_type> img1, img2;
```

dlib/media/ffmpeg_details.h
arrufat: At some point, I would like to add proper JPEG XL support
- `rgb_pixel` / `rgba_pixel` images
- animations
- maybe other pixel formats (more pixel formats: bit depth and channels)

When I was working on WebP support, I already saw it was a nice improvement over JPEG, PNG, and GIF (I have a branch with partial support for WebP animations in dlib)

However, in practice, I often hit the limitations in size (it is a VP8 intra frame, so the width and height is stored in 14 bits, which means 16383 x 16383 pixels, at most).

Then I discovered JPEG XL and it is an outstanding image format. I mean, it's a proper image format, not a video frame repurposed as an image format, like WebP or AVIF.

I wish [Google added support back to Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1178058) (not because I use Chrome, but because it will kill the format).

Sorry for the small rant...

pfeatherstone: Only thing is that JPEGXL support didn't land until FFmpeg 5.1

pfeatherstone:
Probs need to make tests use lossless compression for image formats so that PSNR test is reproducible.

dlib/media/ffmpeg_details.h
arrufat: Thank you for adding that :)

dlib/media/ffmpeg_details.h
```diff
@@ -508,6 +508,14 @@ namespace dlib { namespace ffmpeg { namespace details
...
             else if (ext == "ac3")
                 return AV_CODEC_ID_AC3;
+            else if (ext == "jls")
+                return AV_CODEC_ID_JPEGLS;
+            else if (ext == "jp2")
+                return AV_CODEC_ID_JPEG2000;
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(59, 37, 100)
+            else if (ext == "jxl")
+                return AV_CODEC_ID_JPEGXL;
+#endif
```
arrufat, line 518: Thank you!

pfeatherstone:
@arrufat do you mind adding bits to the tests so save_frame() is lossless? Otherwise PSNR check could fail.

arrufat:
> @arrufat do you mind adding bits to the tests so save_frame() is lossless? Otherwise PSNR check could fail.

I will try tonight :)

arrufat:
> @arrufat do you mind adding bits to the tests so save_frame() is lossless? Otherwise PSNR check could fail.

I will try tonight. However, we're saving JPEG images, it can never be lossless, right?

pfeatherstone:
You can set the quality to maximum then I'm pretty sure it is

pfeatherstone:
> > @arrufat do you mind adding bits to the tests so save_frame() is lossless? Otherwise PSNR check could fail.
> 
> I will try tonight. However, we're saving JPEG images, it can never be lossless, right?

Done.

dlib/media/ffmpeg_muxer.h
```diff
@@ -1381,6 +1424,39 @@ namespace dlib
...
+                args.args_image.codec_options = codec_options;
+                args.format_options["update"] = "1";
+                return args;
+            }());
+
+            if (!writer.is_open())
+                throw error("ffmpeg::save_frame: error while saving " + file_name);
+
+            if (!writer.push(image))
+                throw error("ffmpeg::save_frame: error while saving " + file_name);
```
arrufat, line 1457: Do you think we should do this? And similarly add `EIMAGE_LOAD` in the `ffmpeg::load_frame` function?
```suggestion
                throw error(EIMAGE_SAVE, "ffmpeg::save_frame: error while saving " + file_name);
```

pfeatherstone: Yeah, why not. I'm not familiar with that, but if it's in keeping with the rest of dlib, then go for it.

arrufat: Let's see what @davisking thinks, I'd go for it, but I'm also not that familiar with that. It seems better than just having a generic error.

pfeatherstone: yeah ok, feel free to commit

pfeatherstone: sod it, done it.

pfeatherstone:
@davisking This should be good to go

PR #2804, opened 2023-05-30
Add minimal pyproject.toml

### LINKED ISSUE #2803: [Bug]: PIP deprecation warning for legacy install method

Linux (x86-64)

### dlib version

19.24.1

### Python version

3.11

### Compiler

pip 23.0.1

### Expected Behavior

`pip install dlib`

compile 

installed dlib

### Current Behavior

```
DEPRECATION: dlib is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
  Running setup.py install for dlib ... done
```

### Steps to Reproduce

Install dlib with latest python and pip versions

### Anything else?

Should not be a very impactful change, but according to the warning, the next pip version will break the install.

rvandernoort:
Fixes #2803  by creating a minimal pyproject.toml that still uses the existing setup.py details to remove the deprecation warning. Updates the way Python package is compiled and adds the requirement `build`. I am not an expert but it seems to work!

Old method gave the following warnings:
```
/home/rover/Documents/tmp/dlib/venv-old/lib/python3.11/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/home/rover/Documents/tmp/dlib/venv-old/lib/python3.11/site-packages/setuptools/command/easy_install.py:146: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
```

New methods removes these warnings and compiles to wheel package. 
`Successfully built dlib-19.24.99-cp311-cp311-linux_x86_64.whl`

I am not able to test the Pypi package and the original deprecation warning.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ docs/docs/python/classes.txt
 docs/docs/python/functions.txt
 docs/docs/python/constants.txt
 **/.vscode
+**/venv
--- a/README.md
+++ b/README.md
@@ -37,12 +37,22 @@ vcpkg install dlib
 
 ## Compiling dlib Python API
 
-Before you can run the Python example programs you must compile dlib. Type:
-
+Before you can run the Python example programs you must install the build requirement.
 ```bash
-python setup.py install
+python -m venv venv
+pip install build
 ```
 
+Then you must compile dlib and install it in your environment. Type:
+```bash
+python -m build --wheel
+pip install dist/dlib-<version>.whl
+```
+
+Or download dlib using PyPi:
+```bash
+pip install dlib
+```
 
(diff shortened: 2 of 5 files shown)


davisking:
Thanks.  I'll try this out in a bit.

davisking:
Yeah looks good. Thanks for the PR :)

PR #2806, opened 2023-06-01
Fix pixel saturation in interpolate_quadratic

arrufat:
When using `interpolate_quadratic` some really bright/dark areas would present artifacts.
This PR fixes that by clamping the pixel values to their appropriate range before converting them.

(image resized from 512x512 to 5800+ using `interpolate_quadratic` and a pyramid, and then resized down to 512 using `interpolate_nearest_neighbor` for display purposes here)
lenna-before

### DIFF
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -310,7 +310,9 @@ namespace dlib
                     pixel_to_vector<double>(img[r+1][c  ])(i),
                     pixel_to_vector<double>(img[r+1][c+1])(i));
             typename image_view_type::pixel_type temp;
-            vector_to_pixel(temp, pvout);
+            const auto min_val = pixel_traits<typename image_view_type::pixel_type>::min();
+            const auto max_val = pixel_traits<typename image_view_type::pixel_type>::max();
+            vector_to_pixel(temp, clamp(pvout, min_val, max_val));
             assign_pixel(result, temp);
             return true;
         }
--- a/dlib/image_transforms/interpolation.h
+++ b/dlib/image_transforms/interpolation.h
@@ -310,8 +310,8 @@ namespace dlib
                     pixel_to_vector<double>(img[r+1][c  ])(i),
                     pixel_to_vector<double>(img[r+1][c+1])(i));
             typename image_view_type::pixel_type temp;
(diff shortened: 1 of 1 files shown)


dlib/image_transforms/interpolation.h
```diff
@@ -310,7 +310,9 @@ namespace dlib
                     pixel_to_vector<double>(img[r+1][c  ])(i),
                     pixel_to_vector<double>(img[r+1][c+1])(i));
             typename image_view_type::pixel_type temp;
-            vector_to_pixel(temp, pvout);
+            const auto min_val = pixel_traits<typename image_view_type::pixel_type>::min();
+            const auto max_val = pixel_traits<typename image_view_type::pixel_type>::max();
+            vector_to_pixel(temp, clamp(pvout, min_val, max_val));
             assign_pixel(result, temp);
```
pfeatherstone, line 316: You can use the alias `pixel_type_t<image_view_type>` if you want.
So:
* `const auto min_val = pixel_type_t<image_view_type>::min();`
* `const auto max_val = pixel_type_t<image_view_type>::max();`
It's personal preference though...

arrufat: Sure, why not?

davisking:
Thanks, looks good :D

dlib/image_transforms/interpolation.h
```diff
@@ -310,7 +310,9 @@ namespace dlib
                     pixel_to_vector<double>(img[r+1][c  ])(i),
                     pixel_to_vector<double>(img[r+1][c+1])(i));
             typename image_view_type::pixel_type temp;
-            vector_to_pixel(temp, pvout);
+            const auto min_val = pixel_traits<pixel_type_t<image_view_type>>::min();
+            const auto max_val = pixel_traits<pixel_type_t<image_view_type>>::max();
```
pfeatherstone, line 314: I highly doubt this would have made any difference at all since the compiler would have probs inlined them, but `min_val` and `max_val` could have been `constexpr`.

davisking: Yeah I wouldn't worry about it.  `constexpr` is useful for compile time stuff (e.g. doing template magic).  It shouldn't make a difference to things like this.

PR #2815, opened 2023-06-25
dlib::optional

pfeatherstone:
This is work in progress.
This is inspired by Sy Brand's tl::optional repo and https://github.com/bitwizeshift/BackportCpp.git
I've tried a few different ways of doing things. For example, i didn't like using `union` but it turns out you can't use `aligned_storage` if you want things to be constexpr and direct initialization to work.
I have to say, this is an education in C++ initialization. It's laughable how complicated an object that is either active or None can be. C++ is so messed up. And yet, I love it

### DIFF
--- a/dlib/optional.h
+++ b/dlib/optional.h
@@ -5,8 +5,7 @@
 
 #include <exception>
 #include <initializer_list>
-#include "type_traits.h"
-#include "utility.h"
+#include "functional.h"
 
 namespace dlib
 {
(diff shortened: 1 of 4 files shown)


pfeatherstone:
It's ready. I think it's a good addition

davisking:
Sweet, I'll look in a bit :D

davisking:
Sorry, not yet. Haven't forgotten though 😅

pfeatherstone:
I quite like the `.then()` pattern, i.e. continuations. I wonder how hard it would be to enhance `dlib::thread_pool` so support continuations on tasks and combine them using free functions like `when_all()`, `when_one()`, `when_seq()`, or alternatively using `&&`, `||`, `>>`. Then use another function `wait()` to wait for the task graph to complete.

davisking:
> I quite like the `.then()` pattern, i.e. continuations. I wonder how hard it would be to enhance `dlib::thread_pool` to support continuations on tasks and combine them using free functions like `when_all()`, `when_one()`, `when_seq()`, or alternatively using `&&`, `||`, `>>`. Then use another function `wait()` to wait for the task graph to complete.

Yeah, depends on how heavy handed you want the implementation to be.  Like if every call to any of these functions spawned a new thread that just waited on the future(s) to be ready that's easy.  But something less heavyweight gets tricky fast.

PR #2822, opened 2023-07-10
Optimize pointer to matrix assignment

facug91:
As discussed [here](https://github.com/davisking/dlib/discussions/2814), this assignment could be optimized, and that't what this PR does.  
Here, I also check for data size to be different. If it's not, it doesn't call set_size, which make the assignment even faster that the benchmarks I've said in the discussion.  
I've implemented it with op_pointer_to_mat as Davis suggested, which is great because now everyone who already uses this method will have an efficiency improvement just by upgrading the library.

PR #2826, opened 2023-07-13
Make assignment from a mat(ptr,rows,cols) faster.

davisking:


davisking:
Merged in https://github.com/davisking/dlib/pull/2822

Compaile:
Quick question, why do we check for size!= 0 and that the stride fits the nc.
If we want to check the whole thing (which the second check suggest) should we not compare the whole size? instead of null check it.

And if thats for the user to check (which i dont think is dlibs philosophy) why do we check the second part at all?

davisking:
> Quick question, why do we check for size!= 0 and that the stride fits the nc. If we want to check the whole thing (which the second check suggest) should we not compare the whole size? instead of null check it.
> 
> And if thats for the user to check (which i dont think is dlibs philosophy) why do we check the second part at all?

It's because `matrix_assign_default()` has these preconditions (from dlib/matrix/matrix_assign_fwd.h):
```
            - dest.nr() == src.nr()
            - dest.nc() == src.nc()
```
So we know the sizes match already.  We just have to determine when we can use what kind of memcpy() from there.

Compaile:
ah perfect, thanks!

davisking:
No problem :)

PR #2827, opened 2023-07-15
dlib::expected

pfeatherstone:
WIP.
This needs a crazy amount of unit tests.
This should be a language feature given how complicated it is to implement correctly.


pfeatherstone:
@davisking Writing our own unit tests for this might be a ball ache. Do you think we can copy MSVC's unit tests and refactor to make it work with dlib's unit testing framework?:
https://github.com/microsoft/STL/blob/main/tests/std/tests/P0323R12_expected/test.cpp
https://github.com/microsoft/STL/blob/main/tests/std/tests/P2505R5_monadic_functions_for_std_expected/test.cpp

MSVC STL's license 
```
The Microsoft C++ Standard Library is under the Apache License v2.0 with LLVM Exception:

                                 Apache License
                           Version 2.0, January 2004
                        http://www.apache.org/licenses/
```

arrufat:
Maybe you can ask ChatGPT what tests should be used for a self written std::expected in C++. ChatGPT is a license laundry system :P. Just kidding.

davisking:
> Maybe you can ask ChatGPT what tests should be used for a self written std::expected in C++. ChatGPT is a license laundry system :P. Just kidding.

Yeah and I don't trust it to be able to make sensible tests anyway :)

Although ChatGPT is super amazing.   Still though, making tests is nearly the same thing as designing and understanding an interface contract.  Which is far and away the hardest part of making software.  It's also part of the "so what do you want the computer to do?" question.  Which somehow has to be communicated to the computer (or even other human).  It's unknowable by a third party what the tests should even be without an interface contract.  

What happens often though is people tacitly some contract (just silently in their heads) while editing the code.  I've seen plenty of that in real life (I'm sure everyone has). You get a bunch of people looking at a function's name and bringing their assumptions along for the ride to tacitly assume some contract.  Invariably different people assume different contracts and then there are bugs when those contract definitions don't align.  E.g. person A writes code calling our undocumented function assuming the contract is one way and person B writes code assuming the contract is some different way.  And maybe it's working fine at the moment.  But then person C comes along, assumes some 3rd contract, and now there are 3 different bits of software depending on the undocumented thing.  Invariably someone made a mistake at some point and now things are not working well (maybe it's seg faulting).  So one of the three people (or even a 4th) comes along and "fixes" the undocumented function.  Their fix assumes some contract, which could be any of the three noted or some 4th different one.  And maybe that new contract they implemented the fix with respect to is now incompatible with one of these other silently assumed contracts so now there is some new bug in the code using this stuff.  Which will surface later in some confusing way and it won't even be clear which code is "wrong".  Since depending on the contracts you want to say things uphold it could be the client code that's wrong or it could be the library code that's wrong.  And it just spirals out of control from there until no one can understand what changes will or won't break the system because each bit of code makes incoherent assumptions about what the contracts of the other parts of the system are.

Anyway, contracts are life :D

davisking:
ChatGPT is amazing though.  I love it.

You all try to play tic tac toe with it though?  It can't do it.  You get it onto something that isn't well represented in the training data and it just can't reason about it if it involves any non-trivial number of steps in the reasoning process to do it.

arrufat:
Yes, I also like ChatGPT. Some of my [latest PR](https://github.com/davisking/dlib/pull/2775) may or may have not been drafted with it.

It wasn't a one-prompt thing experience, though… It knows many things about dlib, but it makes up a lot of stuff. I couldn't get it to produce fully working code for that PR, so I had to step in. But it was a fun and amazing experience.

davisking:
Yeah.  It's real nice for API reference questions.  Like "what's the function for doing X again?".  I've also used to find out random stuff like, what episode of star trek did they decide you couldn't go faster than warp whatever?  It's got a bead on pretty much everything.  Not always right but more right than a vanilla google search most of the time.

pfeatherstone:
I don't have much interest in large LLMs and training architectures I can't train myself on my own datasets to solve real world problems. I could be wrong, but you need a huge cluster, lots of money, a dataset the size of the internet and possibly 100s of humans to train the damn thing. It's cool but only a handful of organizations can work on this. I'm looking forward to TinyCorp releasing commodity hardware.

pfeatherstone:
I find it frustrating that even a 100 million parameter transformer requires at least 2 large GPUs like the A6000.

davisking:
Oh yeah, it costs a ton of money to train something like ChatGPT.  It is well out of reach of one person doing it on their computer on the side.  Just the costs in electricity are enormous.

You should play with it though. It's still really cool.  Like I've got this nice large flat screen TV.  I could never make that on my own but I still love it :D

pfeatherstone:
I want to try LLaMa

pfeatherstone:
I also think MuZero and the other models from DeepMind are more impressive. I need to get into reinforcement learning more...

pfeatherstone:
TODO: There are some strong exception guarantees that need to be satisfied.

davisking (commented):
>https://github.com/davisking/dlib/commit/b8806757fbc468ea5638c35a6c02194537a7f44b an extra assignment operator. jesus, this is a headache. I doubt Davis will be bothered to review this. This isn't a 2 hour job.

Na I always review 😭

pfeatherstone:
Making use of dlib's `switch_()` function which approximates `if constexpr`. Beautiful

pfeatherstone:
I would like to benchmark compilation speed vs `std::expected`. Vocabulary types should have almost instant compile times otherwise pointless.

pfeatherstone:
@davisking This is ready for a review. Cheers.

davisking:
Cool. Will look in a bit :)

pfeatherstone:
I think there is a way to use c++20 coroutines to inline the monads. So rather than do:
```
dlib::expected<int, error_code> foo()
{
    return dlib::expected<void, error_code>{}
        .transform(return fn1())
        .transform(return fn2())
        .transform(return fn3());
}
```
You could have 
```
dlib::expected<int, error_code> foo()
{
    co_await fn1();
    co_await fn2();
    co_return fn3();
}
```
Something like that. If any of `fn1`, `fn2` or `fn3` fails, the coroutine returns immediately with `dlib::expected<int, error_code>` in an error state. It would almost be like exceptions, but without exceptions, and the error would propagate to the caller only, not all the way up an unwinded stack. Pretty cool. But that could be abusing the language a bit. In any case, that would be for another PR if I thought it useful and lent itself to cleaner better code.

pfeatherstone:
The nice thing about `dlib::expected` (or `std::expected`) is that you can make them behave like error codes or exceptions. Furthermore, the caller can choose to handle the error case however they want. They are not forced to put something in a try-catch statement or think about how one tiny function call could abort their whole program in an exceptional case. Big fan. Though it would be better implemented as a language feature.

pfeatherstone:
@davisking Have you had a chance to look at this? Do you like it? Or not a fan of this type of error handling?

davisking:
> @davisking Have you had a chance to look at this? Do you like it? Or not a fan of this type of error handling?

Sorry, just been busy.  Na it's fine.  My opinion on this kind of stuff is still in line with what http://dlib.net/howto_contribute.html#8 says :)    Where std::excepted is just nicer error codes.

Anyway, looks ok on the whole.  I'll look more later, probably this weekend.  Add some docs to the top of the main header though that notes what classes are publically exposed.  I.e. what are at the dlib:: namespace scope and intended for people to use with some comments about what they are.   Then put a big comment below them that makes it clear everything below that is implementation details users are not supposed to read.

pfeatherstone:
> My opinion on this kind of stuff is still in line with what http://dlib.net/howto_contribute.html#8 says :) Where std::excepted is just nicer error codes.

The only thing I will say is writing `dlib::optional` and `dlib::expected` has taught me to think more carefully about exceptions. Before, I used to think the strong exception guarantee was a nice thing to have. Now, I really get why it's important. For example `type_safe_union` does not have such guarantees. What is the state of TSU when an exception is thrown mid-way through a placement-new call? It's undocumented, and likely undefined behaviour. Exceptions can put your program in a weird state that you very likely didn't test against. What happens when you throw in a destructor? The language says destructors should be nothrow but doesn't enforce it. Loads of things like that. Maybe I'm overthinking it and should remain naive and accept stuff "just works". However, it feels like something like `std::expected`, or what [other languages are doing](https://ziglang.org/documentation/master/#try), is more predictable IMO. But dunno. I'm not a c++ expert, or language expert for that matter. Don't get me wrong, i don't want to litter my code with checking return codes like in C, but now exceptions worry me. If i were to write my own library, I think I would consider using `std::expected` for all error handling.

pfeatherstone:
https://learn.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170 
This article is an interested read. It roughly echoes what @davisking recommends but doesn't discuss `std::expected`. 
The interesting additional point it makes is the following:
> For every function that might throw or propagate an exception, provide one of the three exception guarantees: the strong guarantee, the basic guarantee, or the nothrow (noexcept) guarantee. For more information, see [How to: Design for exception safety](https://learn.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-170).

davisking:
> > My opinion on this kind of stuff is still in line with what http://dlib.net/howto_contribute.html#8 says :) Where std::excepted is just nicer error codes.
> 
> The only thing I will say is writing `dlib::optional` and `dlib::expected` has taught me to think more carefully about exceptions. Before, I used to think the strong exception guarantee was a nice thing to have. Now, I really get why it's important. For example `type_safe_union` does not have such guarantees. What is the state of TSU when an exception is thrown mid-way through a placement-new call? It's undocumented, and likely undefined behaviour. 

Yeah it is fundamentally important that all C++ code be exception safe.  That's a hill I'm going to die on :)   I need to look at `type_safe_union` and double check that this is so, because it needs to be.  Do you think you see a bug (i.e. that an exception would cause a bad thing to happen)?

> Exceptions can put your program in a weird state that you very likely didn't test against.

I've heard this argument, but I don't get how `std::expected` or the like makes it any different.  The fundamental issue is there are rare events (the unexpected or exceptional things) and people sometimes ignore them.  All std::expected is doing is making is to the caller has to do manual stack unwinding and exception propagation.  Which for cases where the event in question is expected to be handled near to the place its generated is good.  E.g. if you write `foo()` and there is this non-nominal event that when it happens the immediate caller of `foo()` will want to handle with special code in virtual all use cases for `foo()` then totally use something like `std::expected`.  But otherwise totally use an exception.  In either case calling code has to be able to handle whatever that non-nominal event is.

The deal with exceptions is that there are lots of events that just fundamentally can't be handled by the immediate caller.  So there has to be a mechanism to propagate that event up higher and higher until we get to the place where it can be handled.  Sometimes you see code that wraps every single function call in something like 
```
if (DID_NON_NOMINAL_THING_HAPPEN?(foo())) 
    return some_non_expected_thing;
```
I.e. people are literally writing the exact kind of code exception unwinding in gcc would write for them. Like imagine a code base where literally every function returns something like `std::expected<foos_normal_output, SomeVagueGeneralizedErrorNotSpecificToAnyFunctionThatHasNoContractSoIsntHandlableByAnyImmedateCaller>`.  I've seen lots of that.  What you end up with is needing to wrap some "return if not expected" boilerplate around every single function call.  It makes the code harder to read because everything is covered in boilerplate and even worse people invariably forget to do it sometimes and silently drop error events.

That's to say nothing on how such code deals with stuff like iterators, constructors, or the like.  That is, there are all kinds of contexts where you just really can't signal an error in a reasonable way with anything other than an exception.

To say that all another way, some events are things that intermediate code shouldn't know about and has no way of dealing with.  Those things are great for exceptions, since you know the event will be delivered to the right place because the compiler does it all for you automatically.

>  What happens when you throw in a destructor? 

Yeah don't do that :)
It's not undefined behavior though. 

You have the same problem though with std::expected.  I.e. what do you do when some function you call in a destructor returns a non-expected value you don't know how to deal with there?  You are just as screwed as getting an exception in that context.  So it's important to not do things that could fail in a destructor.

> However, it feels like something like `std::expected`, or what [other languages are doing](https://ziglang.org/documentation/master/#try), is more predictable IMO. But dunno. I'm not a c++ expert, or language expert for that matter. Don't get me wrong, i don't want to litter my code with checking return codes like in C, but now exceptions worry me. If i were to write my own library, I think I would consider using `std::expected` for all error handling.

I mean, don't get me wrong.  I'm a fan of `std::expected`.  But there is also a bunch of "because google" cargo cult programming in some languages and communities now around "don't do exceptions" because google doesn't use exceptions.  Except the reason google doesn't use exceptions is because they started with a big C codebase (i.e. they started with code that didn't use RAII so wasn't exception safe, because it's C), not because exceptions are bad.

The TLDR of what I'm saying is that if people have to wrap every function call in something that is just manually making them write exception unwinding then that is wack :)  Let the compiler do that for you.

pfeatherstone:
I think with `std::expected` there is a whole bunch of SFINAE and `static_assert()` statements that carefully monitor your types so that a `std::expected` object is (or at least should) always in a defined state, even when something throws. 

The cool thing with `std::expected` is that you can use `.value()` everywhere and if the object is an error state, it will behave like exceptions and propagate upward. So my understanding is:
* use `obj.value()` on all your `std::expected` objects and  you get exceptions basically.
* use `if (obj) *obj` and you get error codes.

davisking:
> I think with `std::expected` there is a whole bunch of SFINAE and `static_assert()` statements that carefully monitor your types so that a `std::expected` object is (or at least should) always in a defined state, even when something throws.

Yeah `std::expected` is totally exception safe so no worries there.

 
> The cool thing with `std::expected` is that you can use `.value()` everywhere and if the object is an error state, it will behave like exceptions and propagate upward. So my understanding is:
> 
> * use `obj.value()` on all your `std::expected` objects and  you get exceptions basically.
> * use `if (obj) *obj` and you get error codes.

Right, that's the pattern.  But you aren't really supposed to call `.value()` if there isn't a value.  The exception it throws is of the "you can't handle this anywhere near here so basically let the whole program/module unravel and terminate" variety of exception.  Which is, generally speaking, what I'm saying are good exceptions.  I.e. ones where the immediate caller will not have any business catching or even knowing about them.

> 
> So you're not forced to handle `std::expected` objects locally if you don't want to.

Well, you really are.  APIs that make good use of `std::expected` are ones where the non-expected value is something that the immediate caller has reason to know about.  So they shouldn't be ignoring it.  E.g. you could have some function that reads something from a file and it's returning `std::expected<TheThingYouMeantToRead, EndOfFile>`.  So you totally want to be doing something in the caller or near to the caller about that end of file event.  Having "we hit the end of the file" turn into `std::bad_expected_access` would be real bad.  Callers higher up getting `std::bad_expected_access` will have no concept of what `std::bad_expected_access` means since it's so general it could be anything.  I don't know what could meaningfully be done except abort the program upon `std::bad_expected_access` and say go fix your buggy program.

But on the other hand, a good use of exceptions, for example, is the command line parser in dlib.  That thing throws an exception on a command line validation error which let's the huge number of places such user input errors are identified get immediately jumped into the one catch block that can print "hey user, you messed up!" and then terminate the program.  If all the CLI functions returned `std::excepted<TheirNominalOutput, UserGaveBadArgumentsError>` the API would be awful to use and make the user code much less readable since there would be "did the user give a bad input? ok tell them about that and exit" duplicated everywhere.
 
> EDIT: I dunno, I submit to the wisdom of a higher being.

I guess I'm saying there just isn't any right way :)  It all depends on the use cases and the API being designed.

Although maybe you are worried about undefined behavior with exceptions?  That's not an issue.  Like it's fine to throw in a destructor as far as the language goes.  The issue is just that you then will get into a situation where there is already some error event that you will want to report to the user to do something with, but then some other error happens and now you aren't sure what to do about it.  It just makes APIs awful so the general advice is to ensure destructors can't fail.  Which is pretty good advice.

davisking:
> https://github.com/davisking/dlib/blob/c6c865ba71fb41ff71f653358f4077c12ca5fe28/dlib/type_safe_union/type_safe_union_kernel.h#L192
> 
> If this line throws, TSU could be in destroyed state, when it shouldn't really be. There should be some strong exception guarantee there for example. Or we document what happens.

You sure?  It looks correct to me.  What's an example of it being bad?  Like just calling `destruct()` is the same as just calling `clear()`, which puts it into a valid state (the empty state).

Or do you just mean the docs are unclear about what should happen.  E.g. If T's constructor throws does it leave the TSU unmodified or empty?  Yeah I guess that could be more clear.

pfeatherstone:
This is a great discussion! I think I'm generally not happy with error handling in C++. You have to rely on best practices and the language doesn't really help you to do it correctly all the time. It requires discipline to write good APIs. I think std::expected is a good way to formalise certain things. But you're right it's not the only thing.

davisking:
> This is a great discussion! I think I'm generally not happy with error handling in C++. You have to rely on best practices and the language doesn't really help you to do it correctly all the time. It requires discipline to write good APIs. 

You could substitute error handling in that sentence with any other part of c++ and all that still holds 😭

davisking:
Yeah. Honestly I think the combination of exceptions and std::expected allows for making really nice clean APIs.

pfeatherstone:
> Yeah. Honestly I think the combination of exceptions and std::expected allows for making really nice clean APIs.

Cool, well hopefully this is a good PR

davisking (commented):
I've been looking over this for a while and it looks good (as expected :) ).  Anything else you want to do to it before I merge it?

pfeatherstone:
You made some comments regarding docs. I need to do that.

pfeatherstone:
Also, ideally there would be more tests. I wish there was an automated way to test good code gen in release mode.

pfeatherstone:
So I tried moving all the implementation details at the bottom. In order to do that I have to do a lot of forward declaration and splitting declarations from definitions. I could do that but given that the user never really has to read this file, only cppreference, I'm reluctant to do a massive refactor. I've added the block comment at the top of the file that explains what objects are implemented and points to cppreference for docs.

Otherwise, I've pretty much copy pasted unit tests from MSVC's unit tests. You can never have enough unit tests. The green button is life.

davisking:
> So I tried moving all the implementation details at the bottom. In order to do that I have to do a lot of forward declaration and splitting declarations from definitions. I could do that but given that the user never really has to read this file, only cppreference, I'm reluctant to do a massive refactor. I've added the block comment at the top of the file that explains what objects are implemented and points to cppreference for docs.
> 
> 
> 
> Otherwise, I've pretty much copy pasted unit tests from MSVC's unit tests. You can never have enough unit tests. The green button is life.

Cool, yeah don't do any refactors. Just some comments at the top is good.

pfeatherstone:
I've added some tests from MSVC. Lots of stuff doesn't work. And i've commented out a whole bunch of static_asserts. Pfff, starting to think life is too short. Just when you think you know a language, you find out you can't even write a fancy std::pair. WTF

davisking:
Yeah it's got a hefty amount of corner cases.

pfeatherstone:
So I've added all the MSVC tests. I've had to disable a lot of static_assert statements. It could be that a couple noexcept specifiers are slightly off. Otherwise very close. This is hard work.

pfeatherstone:
the error windows-latest is complaining about I don't get. I wonder if that is a legit bug with MSVC and/or their standard library.

pfeatherstone:
We can also use:
https://github.com/viboes/std-make/blob/master/test/expected/expected_pass.cpp
https://github.com/viboes/std-make/blob/master/test/expected/unexpected_pass.cpp
https://github.com/viboes/std-make/blob/master/test/expected/expected_no_move_constructible_fail.cpp
https://github.com/viboes/std-make/blob/master/test/expected2/trivial_expected_pass.cpp
https://github.com/viboes/std-make/blob/master/test/expected2/trivial_unexpected_pass.cpp

pfeatherstone:
I'll get round to this at some point. Not sure if many people will use it though. so maybe not worth the effort.

davisking:
> I'll get round to this at some point. Not sure if many people will use it though. so maybe not worth the effort.

Yeah I wouldn't bother with it unless you are personally jazzed about doing it :)

pfeatherstone:
Yeah let's just close it. No one cares. It was educational at least.

PR #2830, opened 2023-07-17
build (pyproject.toml): Add [project] info ...

goyalyashpal:
* Via `pdm import setup.py`
* pdm: npm like pac-manager for python

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,33 @@
 [build-system]
 requires = ["setuptools", "wheel", "cmake"]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "dlib"
+version = "19.24.99"
+description = "A toolkit for making real world machine learning and data analysis applications"
+keywords = ["dlib", "Computer Vision", "Machine Learning"]
+readme = "README.md"
+authors = [
+    {name = "Davis King", email = "davis@dlib.net"},
+]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: Microsoft",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: POSIX",
+    "Operating System :: POSIX :: Linux",
+    "Programming Language :: C++",
+    "Programming Language :: Python",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence",
+    "Topic :: Scientific/Engineering :: Image Recognition",
+    "Topic :: Software Development",
+]
+license = {text = "Boost Software License"}
+
+[project.urls]
+Homepage = "https://github.com/davisking/dlib"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "dlib"
-version = "19.24.99"
+# version = "19.24.99"
 description = "A toolkit for making real world machine learning and data analysis applications"
 keywords = ["dlib", "Computer Vision", "Machine Learning"]
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ requires = ["setuptools", "wheel", "cmake"]
 build-backend = "setuptools.build_meta"
 
 [project]
+# https://packaging.python.org/en/latest/specifications/declaring-project-metadata/
 name = "dlib"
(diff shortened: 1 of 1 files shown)


goyalyashpal:
```console
$ pdm import setup.py
Inside an active virtualenv D:\15_devt\GitHub\dlib-master\.venv, reusing it.
Set env var PDM_IGNORE_ACTIVE_VENV to ignore it.
The project's requires-python has been set to >=3.11. You can change it later if necessary.
Changes are written to pyproject.toml.
```

```console
$ pdm import setup.py
Virtualenv D:\15_devt\GitHub\dlib-master\.venv is reused.
The project's requires-python has been set to >=3.11. You can change it later if necessary.
Changes are written to pyproject.toml.
```

goyalyashpal:
Please review :smiley:

The reason I marked it as draft earlier was as I wanted to explore if [py-build-cmake] may help deal with that MS VS problem. 

but I couldn't get enough time & felt that its aim might be smth different as well. so, dropped that plan. maybe will do it someday later.

[py-build-cmake]: https://github.com/tttapa/py-build-cmake

goyalyashpal:
> what's the consequence of this change?

i don't know 😅. i don't have any experience whatsoever with packaging. these changes were autogenerated with that tool called `pdm`

it's just i read at 2-3 places that avoid setup.py etc, put things in pyproject.toml etc... so, i thought maybe they are right....

i have added the relevant links to my previous comment - maybe they can contain more concrete wordings to answer you.

> Asking because I would kinda prefer this file to have a little stuff in it as possible.

yep, that's perfectly fine with me 😃 - feel free to close this if you so wish.

davisking:
Yeah if it's auto-generated then it's not needed and just duplication :)

The setup.py is also totally needed to build the C++ code.  So :shrug:   Python packaging is a mess.

goyalyashpal:
> The setup.py is also totally needed to build the C++ code

yeah, that was what my motive was when trying to investigate for that py-build-cmake thingy... but i got pretty exhausted already in micromamba thing that i couldn't persue this further

goyalyashpal:
> The setup.py is also totally needed to build the C++ code

yeah, that was what my motive was when trying to investigate for that py-build-cmake thingy... but i got pretty exhausted already in micromamba thing that i couldn't persue this further

goyalyashpal:
One example of migration I found online (though that is only python based build)

* [slott56/py-web-tool#15] for publishing on pypi
 
> Fiddle with pyproject.toml and tox.ini to eliminate setup.py.


[slott56/py-web-tool#15]: https://github.com/slott56/py-web-tool/pull/15/files?file-filters%5B%5D=.ini&file-filters%5B%5D=.toml&show-deleted-files=true&show-viewed-files=true "File filter: .toml .ini on Publish to Pypi"

goyalyashpal:
> Asking because I would kinda prefer this file to have a little stuff in it as possible.

hey @ davisking feel free to reopen this mr when u change ur mind in this regard 😃

goyalyashpal:
found one related comment regarding this in wild:

> _"Cool to see [pyproject.toml](https://github.com/aclark4life/oiio/blob/python-wheels/pyproject.toml) creating a wheel with no setuptools in sight."_
>
> \- @ aclark4life at https://github.com/python-pillow/Pillow/issues/7150#issuecomment-1739784299

aclark4life:
@goyalyashpal @davisking Right, `setup.py` no longer required to build C++, in this case, due to the inclusion of [scikit-build-core](https://github.com/scikit-build/scikit-build-core) which in turn includes other modern Python packaging standards.

davisking:
I don't follow.  This PR doesn't appear to do anything.  It's just duplicating stuff from setup.py isn't it?

And the setup.py is still required to build the C++ code.  There is this stuff at the end of this PR's pyproject.toml but it's all commented out:
```

# # ext_modules=[CMakeExtension('_dlib_pybind11','tools/python')],
# # https://stackoverflow.com/questions/74409966/#74421415
# # cmdclass=dict(build_ext=CMakeBuild, test=PyTest),
# # # We need an older more-itertools version because v6 broke pytest (for everyone, not just dlib)
# # tests_require=['pytest==3.8', 'more-itertools<6.0.0'],
# # #install_requires=['cmake'], # removed because the pip cmake package is busted, maybe someday it will be usable.
```

I mean, I would be well happy to be rid of setup.py and all it's grit too.  But something has to invoke cmake and do everything but I'm not seeing it here.

davisking:
I mean, let me put it this way, send me a PR that deletes setup.py where dlib is still installable and works and all that and I'll be happy to merge that :)

aclark4life:
@davisking I may just do that, stay tuned! At least I'll give it a whirl and follow up with the results.

PR #2836, opened 2023-07-21
Only trigger CI when relevant files are changed

arrufat:


davisking:
Yeah I was just thinking about this.  It's pretty much still a "run everything that takes any time to run" situation though :shrug:

davisking:
Thanks :)

arrufat:
Yeah, maybe we could go crazy with the paths and make a really complicated configuration that only runs the tests associated with the paths that changed…

At least, with this PR, editing documentation or other non-code stuff won't trigger GitHub Actions.

davisking:
Yep. And editing the ci config files only reruns that one config. Which is nice for trying to puzzle out what the config script is supposed to be for somethings.

arrufat:
Maybe it would be easier to have the CI scripts just run a Python or Shell script, that performs the git diff between `HEAD` and `HEAD~1`, checks which files changed and has all the logic to decide which tests to build and run.

davisking:
Eh don't worry about it. Current setup is plenty good :)

PR #2839, opened 2023-07-22
Perform convolution with bias addition combined for a specific algorithm 

facug91:
In PR #2656 we talked about separate convolution forward pass and bias addition, because for some reason, it was slower.

I was reading cuDNN documentation (https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnActivationMode_t), and it says that function `cudnnConvolutionBiasActivationForward` with `CUDNN_ACTIVATION_IDENTITY` (as used in dlib at the moment), should only be called with algorithm `CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM`. Maybe that's why it was running slower.

In this PR I combined both operations again, but only when forward algorithm is `CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM`. I tested it with https://github.com/dlibml/dnn, using alexnet and darknet, and these were the results:

alexnet avg 23451 us -> 23267.5 us (the same)
darknet19 avg 24320 us -> 22226 us (8.6% faster)
darknet53 avg 61156.5 us -> 55745.5 us (8.8% faster)
darknet53csp avg 55172 us -> 47809.5 us (13.34% faster)

Although with AlexNet there was no improvement, it didn't run slower. And with darknet, it was quite faster. Maybe more testing in other gpus would help. Be sure to manually set `forward_best_algo` to `CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM` to test it. Also, remember to use `fuse_layers(net)`, otherwise it won't use the merged layers.

### DIFF
--- a/dlib/cuda/cudnn_dlibapi.cpp
+++ b/dlib/cuda/cudnn_dlibapi.cpp
@@ -928,7 +928,7 @@ namespace dlib
                 backward_filters_best_algo = CUDNN_CONVOLUTION_BWD_FILTER_ALGO_0;
             }
 #endif
-	    backward_filters_algo = backward_filters_best_algo;
+            backward_filters_algo = backward_filters_best_algo;
 
             // Save this algorithm selection in the cache
             config_to_algo_cache[cache_key] = std::make_tuple(forward_algo, backward_data_algo, backward_filters_algo);
@@ -1177,6 +1177,19 @@ namespace dlib
             const tensor& biases
         )
         {
+
+            // Function cudnnConvolutionBiasActivationForward should only be called with CUDNN_ACTIVATION_IDENTITY when
+            // the chosen forward algorithm is CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM, as cuDNN documentation explicitly says.
+            // In case the algorithm is different, perform the forward pass and bias addition separately.
+            if (forward_algo != CUDNN_CONVOLUTION_FWD_ALGO_IMPLICIT_PRECOMP_GEMM)
+            {
+                (*this)(add_to_output, output, data, filters);
+
+                tt::add(1, output, 1, biases);
+
+                return;
+            }
+
             DLIB_CASSERT(is_same_object(output,data) == false);
             DLIB_CASSERT(is_same_object(output,filters) == false);
             DLIB_CASSERT(filters.k() == data.k());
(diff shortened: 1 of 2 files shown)


arrufat:
Great news, I will try tonight if I have time / energy

arrufat:
Ah, AlexNet doesn't have Batch Normalization layers, so the convolutions there always have biases. As a result, it's unaffected by `fuse_layers`.

facug91:
> Ah, AlexNet doesn't have Batch Normalization layers, so the convolutions there always have biases. As a result, it's unaffected by `fuse_layers`.

Good point. Anyway, the execution of the combined layer should also change.
I ran some more tests with other models and again, except for AlexNet, they all got better. The second column represents how much faster it is with this modification:

```
alexnet		-1,40%
sqznet1.0	10,79%
sqznet1.1	16,96%
vggnet11	 2,63%
vggnet13	 6,16%
vggnet16	 6,61%
vggnet19	 5,14%
googlenet	 5,57%
resnet18	 7,37%
resnet34	 5,78%
resnet50	11,42%
resnet101	 8,94%
resnet152	 8,51%
darknet19	 7,18%
darknet53	 5,11%
darknet53csp	11,42%
```

arrufat:
I can confirm similar results on my machine.

This is the pattern that the `fuse_layers` visitor is looking for:
https://github.com/davisking/dlib/blob/a50bd167c4d5b40374706dcdc4369552d57172b8/dlib/dnn/visitors.h#L406

Calling that visitor on a network that doesn't have that pattern (convolution layer followed by affine layer, which has the scaling and bias of the batch norm) won't do anything at all.

I should've paid more attention to the CuDNN documentation when doing all this stuff. Thank you for pointing it out. TBH, I am still surprised that I made it work at all.

facug91:
I was looking at that PR yesterday, where you added all of this, and it was huge. This is just a very small and specific use case hahah. I'll probably need your help for another PR later, where I plan to modify more things about this, but I'm not entirely sure how to do it.

Thank you for testing and confirming results :)

arrufat:
Don't hesitate to open a discussion and we can talk about it :)

facug91:
I think I was able to solve it, but thanks arrufat. When this PR is accepted (or not hahah), I'll create the other one. I depend on this code, but I managed to fuse convolution with bias and activation function, and the improvement is really nice :)

davisking (commented):
Nice, this is great. Thanks for the PR :)

PR #2843, opened 2023-07-29
updated cudnn path

prashanthar2000:
I faced an issue as neither CUDNN_INCLUDE_DIR nor CUDNN_HOME environment variables were set. This resulted in a "Found cuDNN" error. Therefore, I updated the path with /cuda, as it was most suggested installation path for installing cudnn.

### DIFF
--- a/dlib/cmake_utils/test_for_cudnn/find_cudnn.txt
+++ b/dlib/cmake_utils/test_for_cudnn/find_cudnn.txt
@@ -4,7 +4,7 @@ message(STATUS "Looking for cuDNN install...")
 # libraries and also a few other places as well.
 find_path(cudnn_include cudnn.h
     HINTS ${CUDA_INCLUDE_DIRS} ENV CUDNN_INCLUDE_DIR  ENV CUDNN_HOME
-    PATHS /usr/local ENV CPATH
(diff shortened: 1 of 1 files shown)


dlib/cmake_utils/test_for_cudnn/find_cudnn.txt
```diff
@@ -4,7 +4,7 @@ message(STATUS "Looking for cuDNN install...")
 # libraries and also a few other places as well.
 find_path(cudnn_include cudnn.h
     HINTS ${CUDA_INCLUDE_DIRS} ENV CUDNN_INCLUDE_DIR  ENV CUDNN_HOME
-    PATHS /usr/local ENV CPATH
+    PATHS /usr/local/cuda ENV CPATH
```
davisking, line 7: On most systems it is in /usr/local, not /usr/local/cuda.  So this will break things for most people.  Does listing both paths work for you?  E.g. as is done immediately below.

prashanthar2000: Yes it did work  and i have updated the pull request . please have a look

davisking:
Thanks :)

PR #2862, opened 2023-09-18
Use snprintf instead of sprintf

pfeatherstone:
Apple Clang complains about sprintf and suggests using snprintf due to CVE bugs.


davisking:
Thanks.

PR #2863, opened 2023-09-19
[FFMPEG] bug fix in demuxer when more than 2 streams are present

pfeatherstone:
This is an urgent bug fix. Also, I can smell a dlib release coming. This needs to go in before that happens.


davisking:
Sweet, thanks. Yeah was preparing the next dlib release :)

Issue #2865, opened 2023-09-20
[Bug]: Webcam example does not work properly with FFmpeg 6


arrufat:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### dlib version

master

### Compiler

GCC 13.2.1

### Expected Behavior

Run normally.

### Current Behavior

![Screenshot from 2023-09-20 10-42-05](https://github.com/davisking/dlib/assets/1671644/68085c47-e936-41e0-a08b-1d78d6ddeea1)

There's seems to be some kind of size or stride problem somewhere.
@pfeatherstone can you check if it still works on previous versions of FFmpeg? I am not sure it's a 6.0 specific thing.

### Steps to Reproduce

```bash
git clone https://github.com/davisking/dlib
cd dlib/examples
wget http://dlib.net/files/shape_predictor_68_face_landmarks.dat.bz2
bzip2 -d shape_predictor_68_face_landmarks.dat.bz2
cmake -B build -G Ninja
cmake --build build --target ffmpeg_webcam_face_pose_ex
./build/ffmpeg_webcam_face_pose_ex --height 480 --width 640
```

Relevant CMake output:
```
-- Searching for FFMPEG/LIBAV
-- Checking for modules 'libavdevice;libavfilter;libavformat;libavcodec;libswresample;libswscale;libavutil'
--   Found libavdevice, version 60.1.100
--   Found libavfilter, version 9.3.100
--   Found libavformat, version 60.3.100
--   Found libavcodec, version 60.3.100
--   Found libswresample, version 4.10.100
--   Found libswscale, version 7.1.100
--   Found libavutil, version 58.2.100
-- Found FFMPEG/LIBAV via pkg-config in `/usr/lib`
```

### Anything else?

Just ran this example to make sure it worked before a new dlib version is released.

pfeatherstone:
Yep will check when I get to work

pfeatherstone:
Yep. It's not a bug with the implementation details or ffmpeg version, it's the example. When reading "ffmpeg::frame" instances, we're not asking it to convert pixel format to RGB. They are coming out as YUV. The easiest thing is to use the simpler API and read directly into `array2d<rgb_pixel>`. I'll submit PR.

pfeatherstone:
I've a PR. Let me know if that fixes it on your end. Cheers

arrufat:
Yes, it works, thank you!

davisking:
[...]

PR #2866, opened 2023-09-20
[FFMPEG] bug fix in webcam example

pfeatherstone:
Fixes https://github.com/davisking/dlib/issues/2865


davisking:
Nice, thanks.

PR #2870, opened 2023-09-24
[Functional] fix bind_front() and bind_back() with gcc 5 

pfeatherstone:


davisking:
Can you add a unit test that works with this change but not without it?

pfeatherstone:
It would have to be a GitHub Actions worker/runner/(whatever it's called) that uses say gcc 5.1. Anything after gcc 6 works fine either the old impl and all c++14 complaint clang seemed to work to.

pfeatherstone:
I think the issue was due to some subtle rule in direct initialisation that wasn't obeyed correctly in gcc 5. So the test would be to test with that compiler. I mean this could all be irrelevant if dlib will soon adopt c++17.

davisking:
> I think the issue was due to some subtle rule in direct initialisation that wasn't obeyed correctly in gcc 5. So the test would be to test with that compiler. I mean this could all be irrelevant if dlib will soon adopt c++17.

Yeah.  I'm not going to worry about it.  And yeah, C++17 soon maybe.  I think I'll finish making the next dlib release and then the release after that can be C++17.

PR #2875, opened 2023-10-03
[scope] Adding scope_exit

pfeatherstone:
Adding https://github.com/davisking/dlib/discussions/2874


pfeatherstone:
Weird the macos-latest runner keeps stalling...

davisking:
Nice, that's great.

PR #2878, opened 2023-10-11
Remove cmake from pyproject.toml

avkarenow:
It should be used system cmake instead of compiling own package. Compiling cmake is problematic because it takes a long time and needs a lot of additional packages that many systems doesn't provide.


davisking:
Thanks

PR #2879, opened 2023-10-11
Make use of CMAKE_BUILD_PARALLEL_LEVEL

avkarenow:
`CMAKE_BUILD_PARALLEL_LEVEL` is a variable that can limit resources used during build: https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html

Unfortunately, when `-j` flag is set the variable is ignored. The patch allows to use of `CMAKE_BUILD_PARALLEL_LEVEL` and passes the value to `-j` flag.


davisking:
Thanks

PR #2889, opened 2023-11-10
Add find_convex_hull using Graham scan

arrufat:
So, I know I've not been very active on dlib really...
The thing is, at my current job I am using Zig quite a bit, and therefore, not using dlib as much as I'd like.

However, today I had to compute the convex hull of a set of points at work and checked if an implementation on dlib existed to port it to Zig, but I couldn't find it.

So I coded it in Zig, and now I am back porting it to dlib, as I think it may be useful.

Let me know if you think the stuff in the `impl` namespace should also be exposed to the users.

### DIFF
--- a/dlib/geometry/vector.h
+++ b/dlib/geometry/vector.h
@@ -1310,6 +1310,73 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    namespace impl
+    {
+        enum class points_orientation
+        {
+            collinear,
+            clockwise,
+            couter_clockwise,
+        };
+
+        template <typename T>
+        std::enable_if_t<std::is_same<T, point>::value || std::is_same<T, dpoint>::value, points_orientation>
+        find_points_orientation(const T& a, const T& b, const T& c)
+        {
+            const auto v = a.x() * (b.y() - c.y()) + b.x() * (c.y() - a.y()) + c.x() * (a.y() - b.y());
+            if (v < 0) return points_orientation::clockwise;
+            if (v > 0) return points_orientation::couter_clockwise;
+            return points_orientation::collinear;
+        }
+    }
+
+    template <typename T>
+    std::enable_if_t<std::is_same<T, point>::value || std::is_same<T, dpoint>::value, std::vector<T>>
+    find_convex_hull(std::vector<T>& points)
+    {
+        if (points.size() < 3)
+            return {};
+
+        // find the point with the lowest y coordinate, and the left-most in case of ties.
+        const auto p0 = *std::min_element(points.begin(), points.end(), [](const auto& a, const auto& b){
+            return std::make_pair(a.y(), a.x()) < std::make_pair(b.y(), b.x());
+        });
+
+        // sort the points by polar angle in clockwise order
+        std::sort(points.begin(), points.end(), [&p0](const auto& a, const auto& b){
+            switch (impl::find_points_orientation(p0, a, b))
+            {
+                case impl::points_orientation::clockwise:
+                    return true;
+                case impl::points_orientation::couter_clockwise:
+                    return false;
+                case impl::points_orientation::collinear:
+                    return (p0.x() - a.x()) * (p0.x() - a.x()) + (p0.y() - a.y()) * (p0.y() - a.y()) <
+                           (p0.x() - b.x()) * (p0.x() - b.x()) + (p0.y() - b.y()) * (p0.y() - b.y());
+                default:
+                    throw std::logic_error("unreachable: invalid points_orientation");
+            }
+        });
+
+        std::vector<T> hull;
+        for (const auto& p : points)
+        {
+            while (hull.size() > 1 &&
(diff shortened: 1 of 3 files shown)


pfeatherstone:
Interesting companies are already using Zig even though it hasn't reached 1.0 yet. Do you find there are enough libraries in the Zig ecosystem to be productive ? I'm genuinely interested. I quite like the look of it, but don't fancy rewriting the world.

arrufat:
@pfeatherstone, well, I did have to rewrite the world, basically. We built a virtual try on for makeup that runs completely on the browser using WebAssembly.
First, I made a version in C++ and Emscripten.
I was using dlib, of course, but there are some functions that implicitly allocate, and couldn't be used for WebAssembly directly. At least I didn't know how to, without modifying them.

Eventually, we got interested in Zig because it has native support for Wasm, and all functions that allocate take an allocator as a parameter. So it's easy to just allocate a bunch of memory from JavaScript, pass it to Zig and create a fixed buffer allocator, and use that as a memory pool. But I had to write everything (some PRs made it into dlib from the C++ version):
- color space conversions (#2758)
- blurring and cropping images
- functions for drawing lines and filling polygons (#2775)
- projective transforms (including an SVD... That was fun)
- a matrix class that doesn't allocate memory
- and of course, convex hull (#2889)

It's a nice little language, though.

pfeatherstone:
Great work! I still think I'll only start paying attention to Zig when there is something like onnxruntime and a good tasking system. Though it looks like Zig supports coroutines so on the latter point, it seems like that's on its way. In general, these days I hate not being productive. So I'm using Python a lot more and if you use it correctly, the performance gap between Python and C can be quite small. And there is a library for anything possibly imaginable in Python. I basically only use C++ for systems programming now.

dlib/geometry/vector.h
```diff
@@ -1308,6 +1308,73 @@ namespace dlib
...
+            if (v < 0) return points_orientation::clockwise;
+            if (v > 0) return points_orientation::couter_clockwise;
+            return points_orientation::collinear;
+        }
+    }
+
+    template <typename T>
+    std::enable_if_t<std::is_same<T, point>::value || std::is_same<T, dpoint>::value, std::vector<T>>
+    find_convex_hull(std::vector<T>& points)
+    {
```
davisking, line 1336:
```suggestion
    std::vector<T> find_convex_hull(std::vector<T>& points)
    {
        static_assert(std::is_same<T, point>::value || std::is_same<T, dpoint>::value, "find_convex_hull() only works for dlib::vector types.");
```
Why not this?  I would expect this to give more readable build errors.

dlib/geometry/vector.h
```diff
@@ -1308,6 +1308,73 @@ namespace dlib
...
+        std::sort(points.begin(), points.end(), [&p0](const auto& a, const auto& b){
+            switch (impl::find_points_orientation(p0, a, b))
+            {
+                case impl::points_orientation::clockwise:
+                    return true;
+                case impl::points_orientation::couter_clockwise:
+                    return false;
+                case impl::points_orientation::collinear:
+                    return (p0.x() - a.x()) * (p0.x() - a.x()) + (p0.y() - a.y()) * (p0.y() - a.y()) <
+                           (p0.x() - b.x()) * (p0.x() - b.x()) + (p0.y() - b.y()) * (p0.y() - b.y());
```
davisking, line 1355:
```suggestion
                    return length_squared(p0-a) < length_squared(p0-b);
```
Should be just as fast.

dlib/geometry/vector.h
```diff
@@ -1308,6 +1308,73 @@ namespace dlib
...
+                default:
+                    throw std::logic_error("unreachable: invalid points_orientation");
```
davisking, line 1357:
```suggestion
                    DLIB_CASSERT(false, "this should be impossible");
```
That way if it were to happen there would be a bit more information, like the file and line number.  I know you can get that with gdb but it's a hassle in many cases.  Like if it's running in some server container you often just get a terminal log.

dlib/geometry/vector_abstract.h
```diff
@@ -456,6 +456,21 @@ namespace dlib
...
     !*/
 
+// ----------------------------------------------------------------------------------------
+
+    template <typename T>
+    std::enable_if_t<std::is_same<T, point>::value || std::is_same<T, dpoint>::value, std::vector<T>>
+    find_convex_hull(
+        std::vector<T>& points
+    );
+    /*!
```
davisking, line 466:
```suggestion
    std::vector<T> find_convex_hull(
        std::vector<T>& points
    );
    /*!
        requires
            - T == dlib::point or dlib::dpoint
```
A little more readable :)

davisking:
Yeah, this is cool :)

davisking:
And the `impl` namespace stuff is fine as is :)

arrufat (commented):
Add Davis' suggestions

davisking:
Thanks, this is great :)

PR #2899, opened 2023-12-06
extend letterbox behavior

arrufat:
Before it required to specify a size and would perform a square letterbox.

I think it's better if we make the input image fit into the shape of the output image.
Moreover, that makes the function never allocate the output image.


arrufat:
I know this is a breaking change... Let's see if the tests pass and if we need to provide the old interface, too.

davisking:
Cool, yeah this is better :)

PR #2900, opened 2023-12-13
Handled equality case

hafiz031:
Without equality, say if, `percent_covered_thresh == 1` and `iou_thresh == 1`, then the operator() would return `False` when there is indeed 100% overlap.

### DIFF
--- a/dlib/image_processing/box_overlap_testing.h
+++ b/dlib/image_processing/box_overlap_testing.h
@@ -93,9 +93,9 @@ namespace dlib
                 return false;
 
             const double outer = (a+b).area();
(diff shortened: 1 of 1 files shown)


davisking:
Thanks for the PR, but it's this way on purpose so that setting the threshold to 0 doesn't turn it into something that says that all boxes overlap no matter what.  It also means that a setting of 1 disables one of the tests, which is also intentional.  For instance, the default setting disables the "percent covered" rule.

Agrahariaayush:
Hey can I get the precompiled binary file of dlib 19.24.2

PR #2906, opened 2024-01-08
loss_barlow_twins: add get_eccm member function

arrufat:
This allows us to greatly simplify the self supervised learning example:
- the computation in user code was a bit too distracting
- avoids duplicated computation/allocation of this matrix
- avoids edge case where net outputs are zero due to trainer synchronization

### DIFF
--- a/dlib/dnn/loss_abstract.h
+++ b/dlib/dnn/loss_abstract.h
@@ -2144,6 +2144,20 @@ namespace dlib
                   in WHAT THIS OBJECT REPRESENTS for details.
         !*/
 
+        tensor& get_eccm();
+        /*!
+            ensures
+                - returns the empirical cross-correlation matrix computed by the loss.
+                - this is only meant to be used for visualization/debugging purposes.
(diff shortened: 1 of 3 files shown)


dlib/dnn/loss_abstract.h
```diff
@@ -2144,6 +2144,20 @@ namespace dlib
                   in WHAT THIS OBJECT REPRESENTS for details.
         !*/
 
+        tensor& get_eccm();
+        /*!
+            ensures
+                - returns the empirical cross-correlation matrix computed by the loss.
+                - this is only meant to be used for visualization/debugging purposes.
+        !*/
```
arrufat, line 2152: I am not sure if this will ever be used. But even if the user modifies the `eccm`, it doesn't really affect the loss, since it's re-computed at every step.

So, in case someone has to modify the `eccm` between training steps, they don't need to copy it.

davisking:
Yeah, nice :D

PR #2909, opened 2024-01-28
Return shortcut in disturb_colors

arrufat:
No need to do anything if both gamma and color magnitudes are set to 0.


davisking:
Thanks :D

PR #2911, opened 2024-02-04
fix letterbox_image in yolo example

### LINKED ISSUE #2910: [Bug]: Zero Sized Image (letterbox_image function: img_out size must be bigger than zero, but was 0x0)

Linux (x86-64)

### dlib version

https://github.com/davisking/dlib/commit/19a952c3a4d983b456d497aa888942694d98a809

### Python version

_No response_

### Compiler

gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

### Expected Behavior

In the current implementation, the variable temp.first.size is expected to be equal to (image_size, image_size). However, during execution, it appears that the size is not set correctly, leading to unexpected behavior.

### Current Behavior

Line 444 `train_data.enqueue(temp);` Set the size of the temp.first to zero.

### Steps to Reproduce

Just run.

### Anything else?

As a suggestion: Move the Line 398 : `temp.first.set_size(image_size, image_size);`in a loop.

arrufat:
Fixes #2910

### DIFF
--- a/examples/dnn_yolo_train_ex.cpp
+++ b/examples/dnn_yolo_train_ex.cpp
@@ -395,7 +395,6 @@ try
         dlib::rand rnd(time(nullptr) + seed);
         matrix<rgb_pixel> image, rotated;
         std::pair<matrix<rgb_pixel>, std::vector<yolo_rect>> temp;
-        temp.first.set_size(image_size, image_size);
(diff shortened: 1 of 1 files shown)


davisking:
Thanks :)

PR #2914, opened 2024-02-19
Add color transform and inv color transform

arrufat:
I've had the need to revert the output of dlib's `random_color_transform`.

As a result, I've refactored the code so that we now have
- `color_transform` object, which deterministically returns a transform for some given parameters
- `inv_color_transform` object, which returns the inverse color transform of the given color transform
- `random_color_transform` is now a function that returns the `color_transform` object generated with random parameters.

I understand this might break people's code for seemingly no reason, so, if you would rather not merge this, I'll figure a way around it :)

Let's see if all tests pass, first.

### DIFF
--- a/dlib/image_transforms/random_color_transform.h
+++ b/dlib/image_transforms/random_color_transform.h
@@ -82,6 +82,9 @@ namespace dlib
             const auto red_scale = tform.get_red_scale();
             const auto green_scale = tform.get_green_scale();
             const auto blue_scale = tform.get_blue_scale();
+
+            // Now compute a lookup table for all the color channels.  The table tells us
+            // what the transform does.
             table.resize(256 * 3);
             unsigned long i = 0;
             for (int k = 0; k < 256; ++k)
--- a/dlib/image_transforms/random_color_transform.h
+++ b/dlib/image_transforms/random_color_transform.h
@@ -7,36 +7,31 @@
 #include "../image_processing/generic_image.h"
 #include "../pixel.h"
 #include "../rand.h"
+#include "../matrix.h"
 
 namespace dlib
 {
 
 // ----------------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


dlib/image_transforms/random_color_transform.h
```diff
@@ -7,36 +7,31 @@
...
+            DLIB_CASSERT(0 <= green_scale_ && green_scale_ <= 1)
+            DLIB_CASSERT(0 <= blue_scale_ && blue_scale_ <= 1)
+            const double m = 255 * std::max({red_scale_, green_scale_, blue_scale_});
             red_scale /= m;
             green_scale /= m;
             blue_scale /= m;
-
-            // Now compute a lookup table for all the color channels.  The table tells us
-            // what the transform does.
             table.resize(256 * 3);
```
davisking, line 35:
```suggestion
            // Now compute a lookup table for all the color channels.  The table tells us
            // what the transform does.
            table.resize(256 * 3);
```

dlib/image_transforms/random_color_transform.h
```diff
@@ -61,25 +58,90 @@ namespace dlib
...
+    public:
+        inv_color_transform(
+            const color_transform& tform
+        )
+        {
+            const auto gamma = tform.get_gamma();
+            const auto red_scale = tform.get_red_scale();
+            const auto green_scale = tform.get_green_scale();
+            const auto blue_scale = tform.get_blue_scale();
+            table.resize(256 * 3);
```
davisking, line 85:
```suggestion

            // Now compute a lookup table for all the color channels.  The table tells us
            // what the transform does.
            table.resize(256 * 3);
```

davisking:
Na, it's a good change and one that I don't think would be very disruptive to users.  They will just get a compiler error that is very easily fixed.

PR #2917, opened 2024-02-25
Add JXL support

arrufat:
This PR adds basic support for the JPEG XL format: loading and saving JXL images from and to all supported dlib pixel formats with appropriate conversions.

The utility of this PR is limited, since it can already be done (at least the loading part) with `dlib::ffmpeg::load_frame`.

However, I intend to use JXL for all my datasets from now on, since I will save about 20-40% of disk space with reversible pixel perfect conversion from JPEG and PNG. All that while decoding images faster than libjpeg-turbo.

So, I might add more features specific to JPEG XL in the future (animations, grayscale + alpha, lossless JPEG transcoding).


arrufat:
I think it's done. However, maybe it comes at a wrong time, with all the CMake stuff being refreshed…

arrufat:
OK, now it's done for real: added Python bindings and updated imglab, too.

dlib/image_loader/jxl_loader.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+
+    static std::vector<unsigned char> load_contents(const std::string& filename)
+    {
+        std::ifstream stream(filename, std::ios::binary);
+        stream.exceptions(std::ifstream::failbit | std::ifstream::badbit | std::ifstream::eofbit);
+        stream.seekg(0, std::ios_base::end);
+        std::vector<unsigned char> buffer(stream.tellg());
+        stream.seekg(0);
+        stream.read(reinterpret_cast<char*>(buffer.data()), buffer.size());
```
davisking, line 23:
```suggestion
        std::vector<unsigned char> buffer;
        vectorstream temp(buffer);
        temp << stream.rdbuf();
```
This might be faster.  Used to be the fastest way to do this kind of thing with streams.  It's more readable in any case :)

dlib/image_loader/jxl_loader.h
```diff
@@ -0,0 +1,127 @@
...
+
+// ----------------------------------------------------------------------------------------
+
+    template <
+        typename image_type
+        >
+    void load_jxl (
+        image_type& image,
+        const std::string& file_name
+   )
```
davisking, line 90:
```suggestion
    )
```

dlib/image_loader/jxl_loader.h
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    template <
+        typename image_type
+        >
+    void load_jxl (
+        image_type& image,
+        const unsigned char* imgbuff,
+        size_t imgbuffsize
+   )
```
davisking, line 102:
```suggestion
    )
```

dlib/image_loader/jxl_loader.h
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    template <
+        typename image_type
+        >
+    void load_jxl (
+        image_type& image,
+        const char* imgbuff,
+        size_t imgbuffsize
+   )
```
davisking, line 114:
```suggestion
    )
```

dlib/image_loader/jxl_loader_abstract.h
```diff
@@ -0,0 +1,155 @@
...
+#include "image_loader_abstract.h"
+#include "../algs.h"
+#include "../pixel.h"
+#include "../dir_nav.h"
+#include "../image_processing/generic_image.h"
+
+namespace dlib
+{
+
+    class jxl_loader : noncopyable
```
davisking, line 15: How about only having this class in the .cpp file?  The png and jpeg ones have other methods that let you quickly query stuff about the image.  But if there is only the get_image() method then there isn't any reason for the class to be public to users, and they are better off just always using the global methods.

arrufat: OK, I'll try that later.

dlib/image_saver/save_jxl_abstract.h
```diff
@@ -0,0 +1,55 @@
...
+              dlib/image_processing/generic_image.h or a matrix expression
+            - image.size() != 0
+            - quality >= 0
+        ensures
+            - writes the image to the file indicated by filename in the JPEG XL format.
+            - image[0][0] will be in the upper left corner of the image.
+            - image[image.nr()-1][image.nc()-1] will be in the lower right corner of the
+              image.
+            - This routine can save images containing any type of pixel.  However,
+              save_jxl() can only natively store rgb_pixel, and rgb_alpha_pixel pixel types.
```
davisking, line 35: Doesn't it also save grayscale too?

arrufat: Right! You can see I reused the WebP saver :sweat_smile:

dlib/image_saver/save_jxl.h
```diff
@@ -0,0 +1,118 @@
...
+            );
+
+        auto data = reinterpret_cast<const uint8_t*>(image_data(img));
+        const int width = img.nc();
+        const int height = img.nr();
+        const int depth = pixel_traits<pixel_type>::num;
+        // Fast path: rgb, rgb_alpha, grayscale
+        if (pixel_traits<pixel_type>::rgb ||
+            pixel_traits<pixel_type>::rgb_alpha ||
+            pixel_traits<pixel_type>::grayscale)
```
davisking, line 74: Need to check the size of the grayscale pixel.  It could be `unsigned char` but also `float` or `int` or whatever.  Presumably this path only works for `unsigned char` grayscale images?

dlib/image_saver/save_jxl_abstract.h
davisking: Yeah, that one might as well be private in the .cpp file too.  But :shrug:  :D

dlib/image_saver/save_jxl.h
arrufat: Done!

dlib/image_loader/jxl_loader_abstract.h
arrufat: I've ended up adding those methods to the `jxl_loader`.
I didn't know dlib had those.

Furthermore, I took the liberty to add `nr()` and `nc()` to query the height and width, respectively. With these, I could even discard images with undesirable sizes before fully loading them.

Maybe `width()` and `height()` would be better for method names? I just used `nr` and `nc` to be consistent with how the ones in the `matrix` class are named.

dlib/image_saver/save_jxl_abstract.h
arrufat: I will add those methods in another PR, maybe I can add `nr` and `nc` too to jpeg and png loaders, too.

arrufat:
Done. If I (or someone) have the need for more JPEG XL features at some point, I will add them:
- grayscale_alpha_pixel: it would require a new pixel type
- storing the image in floating point: it would require `rgb_float`, `rgb_alpha_float`
- saving and loading animations: overload the load and save functions with `std::vector`
- transcoding from and to JPEG: this might be useful to convert imglab datasets.

dlib/image_loader/jxl_loader_abstract.h
davisking: Yeah `nr()` and `nc()` is good.

dlib/image_loader/jxl_loader.h
```diff
@@ -23,6 +23,14 @@ namespace dlib
         jxl_loader(const dlib::file& f);
         jxl_loader(const unsigned char* imgbuffer, size_t buffersize);
 
+        bool is_gray() const;
+        bool is_graya() const;
+        bool is_rgb() const;
+        bool is_rgba() const;
+        unsigned int bit_depth() const;
+        unsigned int nr() const;
+        unsigned int nc() const;
```
davisking, line 32:
```suggestion
        long nr() const;
        long nc() const;
```
Any objection to this?  That way it's consistent with all the other `nr()` and `nc()` functions.  Mostly doesn't matter, but someone will eventually do `nr()-something` and get surprised it's unsigned at some point.  Or they will get compiler warnings about signed vs unsigned comparisons and be annoyed by it.

arrufat: Actually, I implemented it like this at the beginning, but somehow I saw all the other methods returned unsigned int, so I changed it. But yeah, long is better.

dlib/image_loader/jxl_loader.h
```diff
@@ -0,0 +1,136 @@
...
+#endif
+            image_view<image_type> vimg(image);
+            vimg.set_size(height, width);
+            using pixel_type = typename image_traits<image_type>::pixel_type;
+
+            // Fast path: rgb, rgb_alpha, grayscale with matching input depth
+            if (pixel_traits<pixel_type>::num == depth && (
+                pixel_traits<pixel_type>::rgb ||
+                pixel_traits<pixel_type>::rgb_alpha ||
+                pixel_traits<pixel_type>::grayscale))
```
davisking, line 54: This needs to check if it's grayscale and of base pixel type unsigned char too doesn't it?

arrufat: Right, I was thinking that even if it's not an unsigned char, it would work, since JPEG XL supports ints and floats. But I should also tell the encoder the type. So let's just restrain ourselves to unsigned char in this PR.

davisking:
Nice, this is sweet.

hscodesolution:
there's a problem with test_for_libjxl and it failed all the time, i tried to run libjxl_test.cpp manually in visual studio and it worked but with cmake its not working, getting error message, System copy of libjxl is either too old or broken.  Will disable JPEG XL support.

I found out that in CMakeLists under test_for_libjxl directory, there's typing mistake, it should be JXL_INCLUDE_DIRS not JXL_INCLUDE_DIR  &&   JXL_LIBRARIES not JXL_LIBRARY

I changed them but CMAKE still failed to compile test_for_libjxl 

What could be the problem ?

PR #2922, opened 2024-03-02
Use STD in testing suite

pfeatherstone:
I stumbled across this while doing the CMake stuff.
I don't see any good reason why it's using dlib's own map container when there is a perfectly good one in STL. Also, use std::mutex and std::atomic instead of dlib pthread wrappers.


davisking:
Cool.

PR #2923, opened 2024-03-04
CMake: Set LAPACK libs as public

jschueller:
lapack/cblas libs are needed through the dlib/matrix/lapack headers

it can be verified on the optimization example:
https://github.com/jschueller/dlib-cmake-example

### DIFF
--- a/dlib/CMakeLists.txt
+++ b/dlib/CMakeLists.txt
@@ -622,7 +622,7 @@ if (NOT TARGET dlib)
 
          if (DLIB_USE_LAPACK)
             if (lapack_found)
-               list (APPEND dlib_needed_private_libraries ${lapack_libraries})
+               list (APPEND dlib_needed_public_libraries ${lapack_libraries})
                if (lapack_with_underscore)
                   set(LAPACK_FORCE_UNDERSCORE 1)
                   enable_preprocessor_switch(LAPACK_FORCE_UNDERSCORE)


davisking:
What's the issue exactly?  I don't totally follow.  All the stuff I know about compiles fine how it is on master.  The example you linked to just looks like the examples in dlib's repository. But something is different?

jschueller:
the problem the executable fails to link if I dont artificially link to LAPACK (comment the target_link_libraries):
```
[ 25%] Building CXX object CMakeFiles/optim_ex.dir/optimization_ex.cpp.o
[ 50%] Linking CXX executable optim_ex
/usr/bin/ld: CMakeFiles/optim_ex.dir/optimization_ex.cpp.o: undefined reference to symbol 'cblas_dger'
/usr/bin/ld: /usr/lib/libcblas.so.3: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

davisking:
Are you sure you are using the current version of dlib for this?

jschueller:
yes, master, but I should add that this only happens for shared libs (dlib defaults to static isnt it ?)

davisking:
Sure.  Can you post exactly the commands to reproduce this?  I tried building and install dlib as a shared library on ubuntu and building https://github.com/jschueller/dlib-cmake-example.  It all built and ran fine.  I git cloned the current master dlib though.

davisking:
Sorry, I meant to say I also deleted that line when I tested it.  So what’s the exact set of commands to reproduce this?On Mar 9, 2024, at 3:39 AM, Julien Schueller ***@***.***> wrote:﻿
yes, the example works because I manually link LAPACK; you can comment lines 14-15 to see the error:
https://github.com/jschueller/dlib-cmake-example/blob/main/CMakeLists.txt#L14

jschueller:
maybe you are using static lapack libs in dlib, in that case it wont show either, as ubuntu provides both, I'm using the shared ones

jschueller:
anyway, here is how to reproduce from a clean ubuntu install:
```
FROM ubuntu:22.04
ENV DEBIAN_FRONTEND=noninteractive MAKEFLAGS="-j8"

RUN apt-get -y update && apt-get install -y git g++ cmake liblapack-dev libjpeg-turbo8-dev libpng-dev zlib1g-dev

WORKDIR /tmp
RUN git clone --depth 1 https://github.com/davisking/dlib.git && cd dlib && cmake -B build -DBUILD_SHARED_LIBS=ON . && make install -C build -j12

RUN git clone https://github.com/jschueller/dlib-cmake-example.git && cd dlib-cmake-example && sed -i "s|LAPACK_LIBRARIES|skip|g" CMakeLists.txt && cmake . && make
```

davisking:
Huh yeah, sorry for the late reply.   And thanks for the PR :)

PR #2924, opened 2024-03-05
Add missing magic number for JPEG XL

arrufat:
For some reason, JPEG XL can have two magic numbers. Adding the one that was missing here.

### DIFF
--- a/dlib/image_loader/load_image.h
+++ b/dlib/image_loader/load_image.h
@@ -57,11 +57,12 @@ namespace dlib
                 return DNG;
             else if (buffer[0]=='G' && buffer[1]=='I' && buffer[2] == 'F') 
                 return GIF;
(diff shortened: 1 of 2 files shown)


arrufat:
This is the reason:

[JXL_SIG_CODESTREAM](https://libjxl.readthedocs.io/en/latest/api_decoder.html#_CPPv4N12JxlSignature18JXL_SIG_CODESTREAME)

    A valid JPEG XL codestream signature was found, that is a JPEG XL image without container.

[JXL_SIG_CONTAINER](https://libjxl.readthedocs.io/en/latest/api_decoder.html#_CPPv4N12JxlSignature17JXL_SIG_CONTAINERE)

    A valid container signature was found, that is a JPEG XL image embedded in a box format container.

davisking:
Sweet, thanks.

PR #2929, opened 2024-03-10
Fix `std::unique_lock` usage in the Windows implementation of `dlib::signaler`

### LINKED ISSUE #2926: [Bug]: Pipe tests fail on latest Visual Studio

Windows

### dlib version

d5909ed9773f89f78203ba4d8324c066b8159925

### Python version

N/A

### Compiler

Visual Studio 2022, version 17.9.2

### Expected Behavior

Tests pass on latest Visual Studio 2022.

### Current Behavior

On the latest Visual Studio 2022 (version 17.9.2), [the pipe test](https://github.com/davisking/dlib/blob/master/dlib/test/pipe.cpp) fails for Debug builds.

The reason appears to be that Microsoft seems to have added [this check](https://github.com/microsoft/STL/commit/acb342c03a2b7978b5d11481af5618ea50b3dbfe#diff-e24ffc00f5c103372fd883f6c4d31c61c95262684010e4957c2faa86cc0f48b4R596-R597) to their STL implementation:
```
"wait_until's caller must own the lock argument (N4958 [thread.condition.condvar]/17)"
```

### Steps to Reproduce

Get latest Visual Studio 2022 (version 17.9.2), build Debug tests, and execute [the pipe test](https://github.com/davisking/dlib/blob/master/dlib/test/pipe.cpp).

### Anything else?

It seems that the `std::unique_lock` object constructed with the `std::defer_lock` option in `dlib::signaler::wait_or_timeout` does not end up "owning" the mutex when calling the condition variable's `wait_until`, which appears to be something Microsoft's updated STL does not really like.

I am not sure if this is a bug in the Microsoft STL implementation (it might very well be), or something to change in dlib, but at bare minimum I need to find some workaround. If I should just `#define _CONTAINER_DEBUG_LEVEL 0` for now, then so be it – but in any case wanted to at least post this for awareness, because expecting that other dlib users may stumble upon this as well, when they update their Visual Studio to the latest versions.

Please note that the underlying code has worked just fine on previous Visual Studio versions. In addition, Release builds (which do not have this check) still work great.

For the record: It seems that Microsoft [have updated](https://github.com/microsoft/STL/issues/3969) to [N4958](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4958.pdf), which does say this for `condition_variable::wait_until` (on page 1893):
> Preconditions: `lock.owns_lock()` is `true`

So I'm guessing that's where this precondition check basically comes from.

reunanen:
Resolves #2926

### DIFF
--- a/dlib/threads/threads_kernel_1.h
+++ b/dlib/threads/threads_kernel_1.h
@@ -92,16 +92,22 @@ namespace dlib
         void wait (
         ) const
         { 
-            std::unique_lock<std::mutex> cs(m.cs, std::defer_lock);
+            std::unique_lock<std::mutex> cs(m.cs, std::adopt_lock);
             cv.wait(cs);
+            // Make sure we don't actually modify the mutex. Since the calling code will have locked
+            // it and it should remain locked.
+            cs.release();			
         }
 
         bool wait_or_timeout (
             unsigned long milliseconds
         ) const
         { 
-            std::unique_lock<std::mutex> cs(m.cs, std::defer_lock);
+            std::unique_lock<std::mutex> cs(m.cs, std::adopt_lock);
             auto status = cv.wait_until(cs, std::chrono::system_clock::now() + std::chrono::milliseconds(milliseconds));
+            // Make sure we don't actually modify the mutex. Since the calling code will have locked
+            // it and it should remain locked.
+            cs.release();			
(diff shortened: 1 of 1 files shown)


davisking:
Thanks :D

PR #2938, opened 2024-03-30
Reduce FOREVER from 290 to 200 years

Chainfire:
Prevents wrap observed in emcc, causing find_max_global to exit immediately unless a custom (and non-wrapping) max_runtime is provided.

### DIFF
--- a/dlib/global_optimization/find_max_global.h
+++ b/dlib/global_optimization/find_max_global.h
@@ -114,7 +114,7 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
 
 // ----------------------------------------------------------------------------------------
 
-    const auto FOREVER = std::chrono::hours(24*365*290); // 290 years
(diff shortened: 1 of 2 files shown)


Chainfire:
Problem observed when using dlib compiled to 32-bit WebAssembly using Emscripten (emcc) and run in the browser.

I'm not sure what the real reason for it is, but trying out the code listed below (not mine, found on StackOverflow) in multiple different browsers all output the max addition was ~ 238 years. 

Perhaps a better solution would eventually be using `steady_clock::time_point::max()` somehow, but using 200 instead of 290 years might be acceptable for now.

```
    using namespace std::chrono;
    using namespace std;
    using days = duration<int, ratio<86400>>;
    using years = duration<double, ratio_multiply<ratio<146097, 400>, days::period>>;
    cout << years{steady_clock::time_point::max() - steady_clock::now()}.count() << " years to go\n";
```

davisking:
Cool, yeah thanks.  This is a good solution.

PR #2939, opened 2024-03-30
Remove MMX header

Chainfire:
MMX is not actually used by dlib and including the header breaks SIMD support detection and usage on platforms that support SSE but not MMX, such as emcc

### DIFF
--- a/dlib/simd/simd_check.h
+++ b/dlib/simd/simd_check.h
@@ -91,7 +91,6 @@
 #ifdef DLIB_HAVE_SSE2
     #include <xmmintrin.h>
     #include <emmintrin.h>
-    #include <mmintrin.h>
 #endif
 #ifdef DLIB_HAVE_SSE3
     #include <pmmintrin.h> // SSE3


Chainfire:
When using emcc (Emscripten) to compile to 32-bit WebAssembly with SIMD support (`-msse42 -msimd128`), support is available for SSE through SSE4.2. These calls are translated (or emulated if not compatible) under the hood to WebAssembly SIMD instructions.

dlib pulls in the MMX intrinsics header as well, but MMX instructions are not supported by emcc, and compilation fails.

I went through the MMX intrinsics header file and searched for all the symbols in the dlib codebase, but could not find any references to any of them. I believe it is not (or perhaps no longer) used. Thus removal shouldn't be a problem, and it fixes SIMD compatibility with emcc.

I have additionally compiled my own code with clang in SSE2 mode and forced SSE2 usage in dlib, and it compiled and worked without issue. However, I did not run the entire dlib test suite.

davisking:
Cool, yeah shouldn't be needed.  Thanks for the PR :D

PR #2941, opened 2024-04-01
Thread pools and WebAssembly

Chainfire:
Two commits:

**Add num_threads_started() to thread_pool**

When running in the browser, creating threads requires returning control to the browser. Calling `std::thread::create()` and waiting (forever) does nothing. So the required procedure is this:

- Call from JavaScript into native code (where dlib runs compiled into WebAssembly)
- Native code calls `std::thread::create()`
- Return to JavaScript
- Set a timer with a callback in JavaScript (just busy-waiting in JavaScript doesn't do anything either)
- In the callback, call into native code and check if the thread has started
- Do the work

Emscripten (emcc) can optionally transparently create a thread pool for you before the native code is run (where `std::thread::create()` will pull from), but the size of the pool must be determined at compile time, and it litters the user's browser with unused threads. Also, the browser limits the number of threads that can be created (typically `(n_cpu / 2) + 4`) and you don't know the number of cpu cores the user has before hand, so that is tricky.

Luckily `worker_thread_ids.size()` gives us exactly the info we need to be able to handle this ourselves, so we add `num_threads_started()` that exposes this information. Now in the timer-based callback into native code we can check if `num_threads_started() == num_threads_in_pool()` and we are good to go.

**Move thread creation and joining from thread_pool internals to an external thread_pool_thread_manager**

As you probably understand from the above, having multiple libraries inside your WebAssembly code managing their own thread pools can be problematic. So when the previous commit is not sufficient, this commit allows us to manage our own thread pools entirely, while still being able to use multithreading with calls like `find_max_global`.

We do this by moving `std::thread::create()` and `std::thread::join()` outside of the thread_pool internals. By subclassing the `thread_pool_thread_manager` class essentially it allows us to run `thread_pool::thread()` as a payload in threads from our own external thread pools, where semantically creating a thread becomes acquiring it, and joining it becomes releasing it. It moves all responsibility for these things away from dlib.

I've tried implementing this in a couple of different ways but ultimately settled on using a pointer to this class (default nullptr, in which case thread_pool creates and releases its own). IMHO it's less pretty than using callbacks, object references, or constructed objects as default parameters, but holding and managing std::thread references is tricky in callbacks, using a global default object doesn't seem ideal (and it would have to be implemented better than this default implementation), and constructing an object inside the default parameter leads to several unnecessary allocations.

### DIFF
--- a/dlib/threads/thread_pool_extension_abstract.h
+++ b/dlib/threads/thread_pool_extension_abstract.h
@@ -276,6 +276,15 @@ namespace dlib
                   the maximum number of tasks that this object will process concurrently.
         !*/
 
+        unsigned long num_threads_started (
+        ) const;
+        /*!
+            ensures
+                - returns the number of threads that have actually been started.  This should
+                  ultimately equal num_threads_in_pool, but may less during pool startup.  This
+                  number does not decrease when shutting down the pool.
+        !*/
+
         template <typename F>
         uint64 add_task_by_value (
             const F& function_object
--- a/dlib/threads/thread_pool_extension_abstract.h
+++ b/dlib/threads/thread_pool_extension_abstract.h
@@ -204,6 +204,70 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    class thread_pool_thread_manager
+    {
+        /*!
+            WHAT THIS OBJECT REPRESENTS
+                This object represents the methods used to create (acquire) and join (release)
+                threads for a thread_pool. The default implementation uses std::thread() to
+                acquire and std::thread::join() to release threads.
+
+                This class can be subclassed so threads from an external thread pool can be
+                reused rather than managing our own. This can be useful in situations where
+                threads are a scarce resource or their creation is slow or otherwise limited.
+
+                When using a subclass, thread_pools should never be created with a higher
+                num_threads than are available to be returned from create_thread() immediately.
+
+                The user of the subclass is responsible for ensuring its lifecycle extends
+                at least to that of the thread_pools it is passed to.
+
+                The default implementation is meant to be used with a single thread_pool, and
+                thread_pool will create and manage an instance internally if the manager
+                parameter is omitted or nullptr.
+        !*/
+    public:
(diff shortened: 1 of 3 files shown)


dlib/threads/thread_pool_extension_abstract.h
```diff
@@ -234,11 +298,17 @@ namespace dlib
 
     public:
         explicit thread_pool (
-            unsigned long num_threads
+            unsigned long num_threads,
+            thread_pool_thread_manager* manager = nullptr
```
davisking, line 302: This all looks sensible.  But why not use a std::shared_ptr or a std::unique_ptr since it sounds like a manager shouldn't be shared between pools.

Chainfire: It's fine to share managers between pools if the manager is built to handle that. In my code running in the browser, I indeed have exactly one manager that is passed to every thread pool. But that is a controlled situation.

The default implementation for when no custom manager is supplied is however naive. It just forever adds to a `std::vector<std::thread>` and never cleans up until destruction. When using one instantiation per pool that is not a problem (exactly num_threads are created once at thread pool creation and joined once at thread pool shutdown, after which the object is destroyed, the vector is never larger than it needs to be), but when sharing it'll grow to the size of num_threads_created_ever. Of course that could be corrected, but it's not necessary in the current implemenation, and I wanted to keep it as simple as possible as C++ is a trip outside my regular programming.

Furthermore I wasn't sure if a thread in a pool can in turn create its own pool, in which case a reused manager would have to be thread safe (with the associated cost), which it doesn't have to be if every pool creates its own instance, as the method are then always called from the same thread.

If you still think shared_ptr or unique_ptr is the way to go, I guess I'll have to read up on those :)

davisking: Yeah that's fine.  I just mean, use a `std::shared_ptr` so that ownership of the pointer is clear.  E.g. then you don't have to call `delete`.  You should never have to manually call delete in C++ :D

PR #2942, opened 2024-04-02
Actually load rgb_alpha images.

arrufat:
This should've been added in #2925.

The only doubt I have around the function name:

- `load_rgb_alpha_image`: following the `rgb_alpha_pixel` in dlib
- `load_rgba_image`: shorter and maybe nicer?


davisking:
Thanks.  Current name is good I think :D

PR #2970, opened 2024-06-19
Update Pybind11 to v2.12.0 for numpy v2 support

yuzawa-san:
- I upgraded to numpy 2.0.0 recently and found the numpy image logic was broken
- I found that numpy had some [breaking changes](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#c-api-changes) which caused the itemsize() call to return 0. see https://github.com/pybind/pybind11/issues/5009
- this was fixed https://github.com/pybind/pybind11/pull/5050 and released in https://github.com/pybind/pybind11/releases/tag/v2.12.0
- i used the same methodology as https://github.com/davisking/dlib/pull/2660/ (copying the files)
- i tested locally with the numpy 1.26.4 and 2.0.0
- open question: should i make this a git submodule?


davisking:
Nice, thanks for doing this :)

PR #2971, opened 2024-06-22
Update Pybind11 to v2.12.0 for numpy v2 support (with submodule)

yuzawa-san:
same as https://github.com/davisking/dlib/pull/2970 but with git submodule


davisking:
Thanks for doing this too.  I merged your other PR instead since I know a decent number of users get thrown off by git submodules. Although personally I use them and think they are fine. :shrug:   But it's one less hurdle for users to jump though.

Issue #2977, opened 2024-07-08
[Bug]: runtime error importing dlib


norflin321:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### dlib version

0947215ddb268fef192e45c140c9430f77648166

### Python version

3.10.12

### Compiler

clang 15.0.7

### Expected Behavior

`import dlib` without errors

### Current Behavior

```
Python 3.10.12 | packaged by conda-forge | (main, Jun 23 2023, 22:41:52) [Clang 15.0.7 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import dlib
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/norflin/code/miniforge3/lib/python3.10/site-packages/dlib-19.24.99-py3.10-macosx-11.0-arm64.egg/dlib/__init__.py", line 19, in <module>
    from _dlib_pybind11 import *
ImportError: dlopen(/Users/norflin/code/miniforge3/lib/python3.10/site-packages/dlib-19.24.99-py3.10-macosx-11.0-arm64.egg/_dlib_pybind11.cpython-310-darwin.so, 0x0002): Library not loaded: /opt/homebrew/opt/openvino/lib/libopenvino_onnx_frontend.2410.dylib
  Referenced from: <4E8ECBBB-080F-3260-882E-4014BB1E62AE> /opt/homebrew/Cellar/ffmpeg/7.0_1/lib/libavdevice.61.1.100.dylib
[...]

PR #2978, opened 2024-07-09
Disable FFmpeg support when building the Python wheel

arrufat:
May solve stuff like https://github.com/davisking/dlib/issues/2977


davisking:
Yeah, seen a few cases like that.  So this is a good idea.

PR #2999, opened 2024-08-25
Add RMS Normalization Layer

Cydral:
This PR introduces a new RMS (Root Mean Square) Normalization layer to Dlib. RMS Normalization is a variant of Layer Normalization that has shown promising results in various deep learning tasks, particularly in Natural Language Processing.

Key changes:
- Add rms_norm_ class implementing the RMS Normalization layer
- Implement rms_normalize and rms_normalize_gradient utility functions
- Add CPU and CUDA implementations for RMS Normalization
- Include unit tests for the new layer

This new layer provides an alternative to the existing layer_norm_, offering potential performance benefits and improved training stability in certain scenarios.

Usage Example:
For a comprehensive example of how to use this new RMS Normalization layer in a Transformer-based architecture, please refer to the ERNIE project: https://github.com/Cydral/ERNIE

### DIFF
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -2345,7 +2345,8 @@ namespace dlib
             const tensor& src,
             const tensor& gamma,
             tensor& src_grad,
-            tensor& gamma_grad
+            tensor& gamma_grad,
+            tensor& dscale
         )
         {
             const long num = src.k() * src.nr() * src.nc();
@@ -2359,8 +2360,6 @@ namespace dlib
             DLIB_CASSERT(eps > 0);
 
             gamma_grad = 0;
-            resizable_tensor dscale;
-            dscale.copy_size(scale);
             dscale = 0;
             launch_kernel(_cuda_rms_normalize_gradient, max_jobs(num, src.num_samples()),
                 src_grad.device(), gamma_grad.device(), src.device(),
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -2242,6 +2242,132 @@ namespace dlib
                           dmeans.device(), dvars.device(), eps, src.num_samples(), num);
         }
 
+   // ----------------------------------------------------------------------------------------
+
+        __global__ void _cuda_rms_normalize(float* out, const float* s, float* scale, const float* g, float eps, size_t ns, size_t num)
+        {
+            // Compute sum of squares
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                auto p = s + n * num;
+                float sum_squares = 0;
+                for (auto i : grid_stride_range(0, num))
+                {
+                    sum_squares += p[i] * p[i];
+                }
+                warp_reduce_atomic_add(scale[n], sum_squares / num);
+            }
+            __syncthreads();
+
+            // Compute RMS inverse
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                for (auto i : grid_stride_range(0, 1))
+                {
+                    scale[n] = 1.0f / std::sqrt(scale[n] + eps);
+                }
+            }
+            __syncthreads();
+
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                for (auto i : grid_stride_range(0, num))
+                {
+                    const float val = s[n * num + i] * scale[n];
+                    out[n * num + i] = val * g[i];
+                }
+            }
+        }
+
+        __global__ void _cuda_rms_normalize_gradient(float* out, float* gg, const float* s, const float* gi, const float* scale, const float* g, float* dscale, float eps, size_t ns, size_t num)
+        {
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                float temp_dscale = 0;
+                for (auto i : grid_stride_range(0, num))
+                {
+                    auto idx = n * num + i;
+                    const float x_hat = s[idx] * scale[n];
+                    gg[i] += gi[idx] * x_hat;
+
+                    const float dx = gi[idx] * g[i];
+                    temp_dscale += dx * s[idx] * -0.5 * scale[n] * scale[n] * scale[n];
+                }
+                warp_reduce_atomic_add(dscale[n], temp_dscale);
+            }
+            __syncthreads();
+
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                for (auto i : grid_stride_range(0, num))
+                {
+                    auto idx = n * num + i;
+                    const float dx = gi[idx] * g[i];
+                    out[idx] += dx * scale[n] + dscale[n] * 2 * s[idx] / num;
+                }
+            }
+        }
+
+        void rms_normalize(
+            const double eps,
+            resizable_tensor& dest,
+            resizable_tensor& scale,
+            const tensor& src,
+            const tensor& gamma
+        )
+        {
+            const long num = src.k() * src.nr() * src.nc();
+            DLIB_CASSERT(
+                src.k() == gamma.k() &&
+                src.nr() == gamma.nr() &&
+                src.nc() == gamma.nc() &&
+                eps > 0,
+                "\ngamma.k():  " << gamma.k() <<
+                "\ngamma.nr(): " << gamma.nr() <<
+                "\ngamma.nc(): " << gamma.nc() <<
+                "\nsrc.k():    " << src.k() <<
+                "\nsrc.nr():   " << src.nr() <<
+                "\nsrc.nc():   " << src.nc() <<
+                "\neps:  " << eps
+            );
+
+            dest.copy_size(src);
+            scale.set_size(src.num_samples());
+            scale = 0;
+            launch_kernel(_cuda_rms_normalize, max_jobs(num, src.num_samples()), dest.device(), src.device(),
+                scale.device(), gamma.device(), eps, src.num_samples(), num);
+        }
+
+        void rms_normalize_gradient(
+            const double eps,
+            const tensor& gradient_input,
+            const tensor& scale,
+            const tensor& src,
+            const tensor& gamma,
+            tensor& src_grad,
+            tensor& gamma_grad
+        )
+        {
+            const long num = src.k() * src.nr() * src.nc();
+            DLIB_CASSERT(src.num_samples() == scale.size());
+            DLIB_CASSERT(src.k() == gamma.k());
+            DLIB_CASSERT(src.nr() == gamma.nr());
+            DLIB_CASSERT(src.nc() == gamma.nc());
+            DLIB_CASSERT(have_same_dimensions(gradient_input, src));
+            DLIB_CASSERT(have_same_dimensions(gradient_input, src_grad));
+            DLIB_CASSERT(have_same_dimensions(gamma_grad, gamma));
+            DLIB_CASSERT(eps > 0);
+
+            gamma_grad = 0;
+            resizable_tensor dscale;
+            dscale.copy_size(scale);
+            dscale = 0;
+            launch_kernel(_cuda_rms_normalize_gradient, max_jobs(num, src.num_samples()),
+                src_grad.device(), gamma_grad.device(), src.device(),
+                gradient_input.device(), scale.device(), gamma.device(),
+                dscale.device(), eps, src.num_samples(), num);
+        }
+
     // ----------------------------------------------------------------------------------------
 
         __global__ void _cuda_copy_tensor_add_to (float* dest, size_t size,  const float* src,  size_t dest_stride, size_t src_stride, size_t block_size)
--- a/dlib/cuda/cuda_dlib.cu
+++ b/dlib/cuda/cuda_dlib.cu
@@ -2282,22 +2282,29 @@ namespace dlib
 
    // ----------------------------------------------------------------------------------------
 
-        __global__ void _cuda_rms_normalize(float* out, const float* s, float* scale, const float* g, float eps, size_t ns, size_t num)
+        __global__ void _cuda_rms_normalize(
+            float* dest,
+            float* scale,
+            const float* src,
+            const float* gamma,
+            float eps,
+            size_t ns,
+            size_t ks,
+            size_t num
+        )
         {
-            // Compute sum of squares
             for (auto n : grid_stride_range_y(0, ns))
             {
-                auto p = s + n * num;
-                float sum_squares = 0;
-                for (auto i : grid_stride_range(0, num))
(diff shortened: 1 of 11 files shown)


dlib/test/dnn.cpp
```diff
@@ -652,6 +652,76 @@ namespace
 #endif
     }
 
+// ----------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
Cydral: I've just made the change. I saw the discussion about the problem (bug?) noted and related to the behaviour of the other layer_norm_ normalisation function, but I don't think I have this ‘type’ of issue with rms_norm_.

Cydral (commented):
Adding a complementary test to comply with the other assessments.

dlib/test/dnn.cpp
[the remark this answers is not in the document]
davisking: Yeah, your stuff should be fine.

dlib/cuda/cuda_dlib.cu
```diff
@@ -2242,6 +2242,132 @@ namespace dlib
...
+            DLIB_CASSERT(src.k() == gamma.k());
+            DLIB_CASSERT(src.nr() == gamma.nr());
+            DLIB_CASSERT(src.nc() == gamma.nc());
+            DLIB_CASSERT(have_same_dimensions(gradient_input, src));
+            DLIB_CASSERT(have_same_dimensions(gradient_input, src_grad));
+            DLIB_CASSERT(have_same_dimensions(gamma_grad, gamma));
+            DLIB_CASSERT(eps > 0);
+
+            gamma_grad = 0;
+            resizable_tensor dscale;
```
davisking, line 2362: The tensors have to be allocated once up front.  Having this here causes the cuda kernel to run on a dangling pointer.  Which is probably why the unit test fails when run on a GPU.  See my comment here https://github.com/davisking/dlib/pull/3001#issuecomment-2316610282 about it.

Cydral: Indeed, change done

Cydral:
Fix dangling pointer issue in CUDA implementation of rms_normalize_gradient:

Key changes include:
1. Moving the allocation of the dscale tensor to the rms_norm_ class as a private member.
2. Allocating memory for dscale in the setup method of rms_norm_.
3. Using the member variable dscale in the rms_normalize_gradient function instead of a local declaration.
4. Consistently passing dscale through all relevant functions to ensure an identical interface between the CPU and CUDA implementations.

davisking:
Looks good I think.  Try building and running the tests locally though with cuda.  When I do that I get this build error
```
/home/davis/source/dlib/dlib/test/dnn.cpp: In function ‘void {anonymous}::test_rms_normalize()’:
/home/davis/source/dlib/dlib/test/dnn.cpp:718:107: error: too few arguments to function ‘void dlib::cpu::rms_normalize_gradient(double, const dlib::tensor&, const dlib::tensor&, const dlib::tensor&, const dlib::tensor&, dlib::tensor&, dlib::tensor&, dlib::tensor&)’
         cpu::rms_normalize_gradient(eps, gradient_input, scale_cpu, x, gamma, src_grad_cpu, gamma_grad_cpu);
```

I.e. just do
```
cd dlib/test
mkdir build
cmake ..
make -j6 dtest && ./dtest --test_dnn -d
```
and make sure you see it say it's going to use cuda. Presumably it will since you are using cuda with dlib already though.

Cydral:
Damn! There's also an update to be made in the test for this new layer. I'll try and have a look tomorrow.

Cydral:
The implementation now seems to be correct and the compilation tests have been passed for the various platforms.

davisking:
The tests that run on github don't compile any GPU code since github doesn't have (or didn't anyway, maybe we can get some now, not sure) GPU machines to test on.  So you have to test it locally yourself.  Like 
```
diff --git a/dlib/test/dnn.cpp b/dlib/test/dnn.cpp
index f8c18e57..2f24084a 100644
--- a/dlib/test/dnn.cpp
+++ b/dlib/test/dnn.cpp
@@ -2079,12 +2079,12 @@ namespace
             auto res = test_layer(l);
             DLIB_TEST_MSG(res, res);
         }
-        {
-            print_spinner();
-            layer_norm_ l;
-            auto res = test_layer(l);
-            DLIB_TEST_MSG(res, res);
-        }
+        // {
+        //     print_spinner();
+        //     layer_norm_ l;
+        //     auto res = test_layer(l);
+        //     DLIB_TEST_MSG(res, res);
+        // }
         {  
             print_spinner();  
             rms_norm_ l;  
```
And then `make -j6 dtest && ./dtest --test_dnn -d`.  For me I still see 
```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!! TEST FAILED: test_dnn !!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Failure message from test: 

Error occurred at line 2092.
Error occurred in file /home/davis/source/dlib/dlib/test/dnn.cpp.
Failing expression was res.
Gradient error in parameter #0.  Relative error: 0.208587
expected derivative: -2.85984
output derivative:   -2.26331
iteration:           0
```

Probably just need the other fixes @arrufat has over at https://github.com/davisking/dlib/pull/3001.

Cydral:
Thank you for your feedback, Davis. I had indeed tested the implementation using an external program that allowed me to compare the CPU and CUDA executions against each other and against the theoretical expected results. The CPU version was tested against the expected theoretical output (as given in the dnn.cpp test), and I hadn't initially noticed any issues with the expected results.

```
resizable_tensor y_cuda(x);
resizable_tensor scale_cuda(x.num_samples());
cuda::rms_normalize(eps, y_cuda, scale_cuda, x, gamma);
DLIB_TEST_MSG(max(abs(mat(y_cpu) - mat(y_cuda))) < 1e-5, "rms_norm layer, max(abs(mat(y_cpu) - mat(y_cuda))) < 1e-5");
DLIB_TEST_MSG(max(abs(mat(scale_cpu) - mat(scale_cuda))) < 1e-5, "rms_norm layer, max(abs(mat(scale_cpu) - mat(scale_cuda))) < 1e-5");

resizable_tensor gradient_input(x);
resizable_tensor src_grad_cpu(x), gamma_grad_cpu(1, x.k(), x.nr(), x.nc());
resizable_tensor src_grad_cuda(x), gamma_grad_cuda(1, x.k(), x.nr(), x.nc());
resizable_tensor dscale_cpu(x.num_samples()), dscale_cuda(x.num_samples());
rnd.fill_gaussian(gradient_input);
src_grad_cpu = 0;
src_grad_cuda = 0;
cpu::rms_normalize_gradient(eps, gradient_input, scale_cpu, x, gamma, src_grad_cpu, gamma_grad_cpu, dscale_cpu);
cuda::rms_normalize_gradient(eps, gradient_input, scale_cuda, x, gamma, src_grad_cuda, gamma_grad_cuda, dscale_cuda);
DLIB_TEST_MSG(max(abs(mat(src_grad_cpu) - mat(src_grad_cuda))) < 1e-5, "rms_norm layer, max(abs(mat(src_grad_cpu) - mat(src_grad_cuda))) < 1e-5");
DLIB_TEST_MSG(max(abs(mat(gamma_grad_cpu) - mat(gamma_grad_cuda))) < 1e-5, "rms_norm layer, max(abs(mat(gamma_grad_cpu) - mat(gamma_grad_cuda))) < 1e-5");
```

These tests passed in my external program. However, I've now been able to reproduce the gradient error you mentioned when running:
```
layer_norm_ l;
auto res = test_layer(l);
DLIB_TEST_MSG(res, res);
```

```
test failed: Gradient error in parameter #0.  Relative error: 0.203608
expected derivative: -3.19285
output derivative:   -2.54276
iteration:           0
```

Given that I can now reproduce this issue, I will continue to investigate the discrepancy between my initial tests and the test_layer function results. I appreciate your pointing this out, and I'll work on resolving this inconsistency.

davisking:
Sweet, no worries :D   Definitely look at the change @arrufat just made.  Since it looks like this PR is based on the layer norm code and probably ends up with the same kind of bug the layer norm code had, which @arrufat just fixed :D

dlib/cuda/cuda_dlib.cu
```diff
@@ -2242,6 +2242,131 @@ namespace dlib
...
+        __global__ void _cuda_rms_normalize_gradient(float* out, float* gg, const float* s, const float* gi, const float* scale, const float* g, float* dscale, float eps, size_t ns, size_t num)
+        {
+            for (auto n : grid_stride_range_y(0, ns))
+            {
+                float temp_dscale = 0;
+                for (auto i : grid_stride_range(0, num))
+                {
+                    auto idx = n * num + i;
+                    const float x_hat = s[idx] * scale[n];
+                    gg[i] += gi[idx] * x_hat;
```
davisking, line 2291: This needs to use a `warp_reduce_atomic_add()`.  Can't have multiple threads adding to the same output.

Cydral:
I've reviewed the discussion regarding the layer_norm issue, and I acknowledge that the problem might indeed stem from poorly managed concurrency. However, at this point, I'm still uncertain about the exact source of the error...

In light of this, I'm in the process of completely rewriting the layer. While the current implementation correctly handles batches, it does so via channels rather than treating each matrix in the tensor individually. I'll be posting updated versions addressing this aspect soon.

Simultaneously, to investigate the CUDA-related issue, I've initially adopted a simpler approach to writing the kernels, essentially mirroring the CPU code. This should help in isolating any problems specifically related to gradient calculation.

Additionally, I'll be updating the test program in dnn.cpp to reflect these changes and provide more comprehensive testing.

dlib/cuda/cuda_dlib.cu
Cydral: Regarding the concurrent write issue, my initial code already included a mechanism to protect this operation:

`atomicAdd(&gamma_grad[c], sum_grad_x_hat_for_gamma);`

Cydral: The updated code here after:

```
        __global__ void rms_normalize_gradient_kernel(
            const float* gradient_input,
            const float* src,
            const float* gamma,
            const float* scale,
            float* src_grad,
            float* gamma_grad,
            float* dscale,
            const long num_samples,
            const long num_channels,
            const long spatial_dim
        )
        {
            const long n = blockIdx.x;
            const long c = threadIdx.x;

            if (n < num_samples && c < num_channels)
            {
                float sum_grad_x_hat = 0;
                float sum_grad_x_hat_x = 0;
                float sum_grad_x_hat_for_gamma = 0;
                float rms_inv = scale[n * num_channels + c];

                for (long s = 0; s < spatial_dim; ++s)
                {
                    long idx = (n * num_channels + c) * spatial_dim + s;
                    float x = src[idx];
                    float grad = gradient_input[idx];
                    float x_hat = x * rms_inv;

                    sum_grad_x_hat += grad * gamma[c];
                    sum_grad_x_hat_x += grad * gamma[c] * x_hat * x;
                    sum_grad_x_hat_for_gamma += grad * x_hat;
                }

                float dscale_val = -0.5f * rms_inv * rms_inv * rms_inv * sum_grad_x_hat_x;
                dscale[n * num_channels + c] = dscale_val;

                atomicAdd(&gamma_grad[c], sum_grad_x_hat_for_gamma);

                for (long s = 0; s < spatial_dim; ++s)
                {
                    long idx = (n * num_channels + c) * spatial_dim + s;
                    float x = src[idx];
                    float grad = gradient_input[idx];
                    float x_hat = x * rms_inv;

                    src_grad[idx] = gamma[c] * rms_inv * (grad - (sum_grad_x_hat * x_hat + dscale_val * x * 2) / spatial_dim);
                }
            }
        }
```
Looking at it this way, I don't see much of a concurrency problem...

Cydral:
It's getting late, I'll continue searching tomorrow.

pfeatherstone:
Is the plan to slowly merge all the transformer stuff ?

Cydral:
That's exactly the point... I already have a technically functional implementation, at least on simple examples, and I'm gradually transforming the new layers to try and make them as efficient as possible and in line with Dlib practices. It's a time-consuming job... but I'm making progress.

pfeatherstone:
Awesome! Are you going to implement flash attention somehow ?

Cydral:
I'm currently working on implementing the basic mechanism, specifically the multihead attention.

dlib/cuda/cuda_dlib.cu
davisking: Yeah that doesn't look to have any data races in it.   Totally use a warp reduction with a grid stride loop though.  Those are super fast.  Since the warp reduction stuff is extremely fast and using a nested grid stride loop is way faster than a regular `for` loop since that uses cuda threads.

I.e. this pattern:
```
                        // loop over some 2D thing and sum and store things into out.
                        for (auto i : grid_stride_range_y(0, nr))
                        {
                            float temp = 0;
                            for (auto j : grid_stride_range(0, nc))
                                temp += whatever[i*nc+j];

                            // store the sum into out[i]
                            warp_reduce_atomic_add(out[i], temp);
                        }
```
It's wildly fast.  More about it at https://developer.nvidia.com/blog/faster-parallel-reductions-kepler/ which is what that warp_reduce_atomic_add() is doing.

Cydral: Yes, I've completely reworked the code to fix the competition problems and use the reduction mechanisms.

Cydral:
I'm currently reviewing the automated checks and recompilations under GitHub to ensure everything is working correctly. I've performed a general update of the various implementations related to the rms_norm_ layer. After testing in my own programs, everything now seems to be working correctly, both under CPU and GPU architectures.
There were some concurrency issues, particularly in the GPU part, related to the gradient update. I took the opportunity to simplify the calculations a bit. Also, the normalization values are no longer saved in the layer, as they are actually intermediate variables that are recalculated in the forward pass (the same behavior applies for the derivative in the backward pass). Only the learned gamma weights remain in the layer's parameter storage. For now, there is one gamma value per channel, but it would be interesting to study the impact of applying a gamma matrix per channel to allow for better and more independent local adaptation.

Cydral:
I believe the pull request is now ready for @davis to review and merge, pending his final approval.

davisking:
Nice, this is great. Thanks for the PR :)

PR #3000, opened 2024-08-26
Add customizable dropout layer with compile-time rate specification

Cydral:
This PR introduces a new customizable dropout layer, `dropout_custom_`, which allows specifying the dropout rate at compile-time. This enhancement is particularly beneficial for deep neural networks with numerous layers, where manually setting different dropout rates for each layer can be cumbersome.

Key features and benefits:

1. Compile-time dropout rate specification: Allows for clearer and more concise network definitions.
2. Inherits from the existing `dropout_` class: Maintains all functionality of the original dropout layer.
3. Template-based implementation: Provides type-safety and potential performance benefits.
4. Includes a pre-defined `dropout_10` alias: Offers a convenient 10% dropout option for common use cases.

### DIFF
--- a/dlib/dnn/layers_abstract.h
+++ b/dlib/dnn/layers_abstract.h
@@ -1466,7 +1466,8 @@ namespace dlib
 
     template <int DROP_RATE, typename SUBNET>
     using dropout_rate = add_layer<dropout_rate_<DROP_RATE>, SUBNET>;
-
+    template <typename SUBNET>
+    using dropout_10 = add_layer<dropout_rate_<10>, SUBNET>;
 // ----------------------------------------------------------------------------------------
 
     class multiply_
--- a/dlib/dnn/layers_abstract.h
+++ b/dlib/dnn/layers_abstract.h
@@ -1436,7 +1436,7 @@ namespace dlib
 // ----------------------------------------------------------------------------------------
 
     template <float DROP_RATE>
-    class dropout_custom_ : public dropout_
+    class dropout_rate_ : public dropout_
     {
         /*!
             WHAT THIS OBJECT REPRESENTS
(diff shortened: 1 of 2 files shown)


arrufat:
This is not good enough?

```c++
visit_computational_layers(net, [](dropout_& l){ l = dropout_(0.1); });
```

Cydral:
No, because it doesn't allow you to precisely identify the layer to be modified. In an LLM-type network (but this would also be true for a convolution-based network for specific image processing), it may be necessary to add layers whose outputs are filtered by a dropout of a different rate.

Cydral:
However, if we don't want to break the interface and to make the layer more flexible, we can also add a template parameter to the dropout_ layer, keep the dropout instantiation with a default parameter of 0.5 and add a dropout_c (c for custom) to specify the rate when defining the layer.

davisking:
Yeah this is cool the way it is.  How about a different name for it though, maybe `dropout_at_rate` or `dropout_rate` maybe?  IDK, what do you guys think?

Cydral:
"dropout_rate" seems pretty good. I'll rename the new class accordingly if that's OK with you.

dlib/dnn/layers_abstract.h
```diff
@@ -1433,6 +1433,40 @@ namespace dlib
...
+        /*!
+            ensures
+                - Constructs a dropout layer with a dropout rate of DROP_RATE.
+                - Calls the base class constructor dropout_(DROP_RATE).
+        !*/
+    };
+
+    template <int DROP_RATE, typename SUBNET>
+    using dropout_rate = add_layer<dropout_rate_<DROP_RATE>, SUBNET>;
+
```
davisking, line 1469:
```suggestion
    template <typename SUBNET>
    using dropout_10 = add_layer<dropout_rate_<10>, SUBNET>;
```

PR #3016, opened 2024-09-19
Cleanup: remove using namespace std;

### LINKED ISSUE #3015: [Bug]: Conflicts with other codes due to "using namespace std;" in header files

Linux (x86-64)

### dlib version

19.24

### Python version

3.12

### Compiler

GCC 13.2

### Expected Behavior

_No response_

### Current Behavior

Several headers of dlib declare "using namespace std;". It creates conflicts with other codes having functions/classes in their own namespace having the same name than the std lib and having the same practice.

It's usually considered a very bad practice in headers: stackoverflow.com/questions/1452721/whats-the-problem-with-using-namespace-std

### Steps to Reproduce

/* This code reproduces the problem */
namespace my_namespace { template <typename T> struct numeric_limits {}; }
using namespace my_namespace;
# include <dlib/optimization.h>
int main(){ return 0; }

### Anything else?

_No response_

arrufat:
Fixes #3015

### DIFF
--- a/dlib/global_optimization/find_max_global.h
+++ b/dlib/global_optimization/find_max_global.h
@@ -178,18 +178,21 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
             running_stats_decayed<double> objective_funct_eval_time(functions.size()*5);
             std::mutex eval_time_mutex;
 
-            const auto time_to_stop = std::chrono::steady_clock::now() + max_runtime;
+            using std::chrono::steady_clock;
+            using std::chrono::duration_cast;
+            using std::chrono::nanoseconds;
+            const auto time_to_stop = steady_clock::now() + max_runtime;
             //atomic<bool> doesn't support .fetch_or, use std::atomic<int> instead
             std::atomic<int> this_should_stop{false};
 
             double max_solver_overhead_time = 0;
 
             // Now run the main solver loop.
-            for (size_t i = 0; i < num.max_calls && std::chrono::steady_clock::now() < time_to_stop && !this_should_stop.load(); ++i)
+            for (size_t i = 0; i < num.max_calls && steady_clock::now() < time_to_stop && !this_should_stop.load(); ++i)
             {
-                const auto get_next_x_start_time = std::chrono::steady_clock::now();
+                const auto get_next_x_start_time = steady_clock::now();
                 auto next = std::make_shared<function_evaluation_request>(opt.get_next_x());
-                const auto get_next_x_runtime = std::chrono::steady_clock::now() - get_next_x_start_time;
+                const auto get_next_x_runtime = steady_clock::now() - get_next_x_start_time;
 
                 auto execute_call = [&functions,&ymult,&log_scale,&eval_time_mutex,&objective_funct_eval_time,next,&should_stop,&this_should_stop]() {
                     matrix<double,0,1> x = next->x();
@@ -200,9 +203,9 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
                         if (log_scale[next->function_idx()][j])
                             x(j) = std::exp(x(j));
                     }
-                    const auto funct_eval_start = std::chrono::steady_clock::now();
+                    const auto funct_eval_start = steady_clock::now();
                     double y = ymult*call_function_and_expand_args(functions[next->function_idx()], x);
-                    const double funct_eval_runtime = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now() - funct_eval_start).count();
+                    const double funct_eval_runtime = duration_cast<nanoseconds>(steady_clock::now() - funct_eval_start).count();
                     this_should_stop.fetch_or(should_stop(y*ymult));
                     next->set(y);
                     
@@ -214,7 +217,7 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
 
                 std::lock_guard<std::mutex> lock(eval_time_mutex);
                 const double obj_funct_time = objective_funct_eval_time.mean()/std::max(1ul,tp.num_threads_in_pool());
-                const double solver_overhead_time = std::chrono::duration_cast<std::chrono::nanoseconds>(get_next_x_runtime).count();
+                const double solver_overhead_time = duration_cast<nanoseconds>(get_next_x_runtime).count();
                 max_solver_overhead_time = std::max(max_solver_overhead_time, solver_overhead_time);
                 // Don't start thinking about the logic below until we have at least 5 objective
                 // function samples for each objective function.  This way we have a decent idea how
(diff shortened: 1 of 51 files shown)


ChristopheDumeunier:
Thanks for the quick reaction.  This fixes my problem.
Note that you may encounter the same problem with some 'using namespace std::chrono' in some headers too. ;-)

arrufat:
Yes, I saw a couple of those, too. Might fix it later.

arrufat:
Not too sure about this one https://github.com/davisking/dlib/pull/3016/commits/ab82c4f92f2b1245c28bfde85b30d018ef7534ae

I am ready to revert it.

davisking:
Eh, what is the specific error we are talking about @ChristopheDumeunier ?  Because doing `using namespace` inside a function scope is totally fine.  It doesn't matter if that function is in a header or not.  What's bad is doing it at a global scope in a header, as that brings symbols into client code.

Was there any of that in dlib?  There shouldn't be.

I don't really want to go mass changing code that isn't breaking.  That's invariably how we find out later on some platform something is now broken.

ChristopheDumeunier:
@davisking Doing `using namespace` inside a function is not fine if the inline/template function is defined in a header because you don't know the compilation context of your code.

Here is a code that reproduce the problem: https://godbolt.org/z/r7T5nqMr5

arrufat:
So, in that example, you declare your namespace in a CPP file and include dlib **after** that?

If you just move the code that simulates including dlib to the top of the C++ (where the `#include` directives usually are), the error is gone.

I am just curious, why would you do that?

ChristopheDumeunier:
It's a reduced example. In our application, some parts of the unit tests of the code are generated by scripts and macro, the the order of includes can be strange and vary depending on (de)activated dependencies. We use dlib from a long time, without this error. The error appeared when we added a struct named "numeric_limits" somewhere else in our namespace.

arrufat:
The `using namespace` thing in dlib isn't the issue, if you name anything like something that's already inside dlib and then use `using namespace <your_namespace>` before including dlib, you'll have the same problem.

That's why they say you shouldn't use them globally in header files: so that you don't clash with other people's code.

But what you're doing amounts to that: `using namespace` before importing a header is akin to using it globally in a header file.

I am closing this PR.

davisking:
Well, it's not exactly the same @arrufat 

Since name lookup is preferred for names in your own namespace.   Like if you did `class matrix{};` before including any dlib headers it's still fine and nothing bad would happen, since dlib is all in the one monolithic dlib namespace any code in it will still prefer the `dlib::matrix`.  But for symbols outside the dlib namespace, if someone goes and defines them at global scope like in the example yeah it's going to become ambiguous.  

Which is generally a thing you shouldn't be doing.  But that's fine, dlib is all about being convenient for users, so making these changes is fine with me.  I get that people get pushed into doing stuff that is sub-optimal in software and we should make those cases not a pain when we can :)

arrufat:
I see. Actually, that's why I hurried to make this PR so that I could solve @ChristopheDumeunier's issue.

However, you scared me that I might have just broken some quirky compilers/setups by doing that stuff (I saw workarounds for GCC 4.9 while doing this PR).

And I wouldn't want to take more of your time with more people opening issues because of me...

dlib/global_optimization/find_max_global.h
```diff
@@ -177,20 +177,19 @@ template <typename T> static auto go(T&& f, const matrix<double, 0, 1>& a) -> de
 
             running_stats_decayed<double> objective_funct_eval_time(functions.size()*5);
             std::mutex eval_time_mutex;
-            using namespace std::chrono;
```
davisking, line 180: How about this?
```
using std::chrono::steady_clock;
using std::chrono::duration_cast;
using std::chrono::nanoseconds;
```
and then the code below reads more clearly.

dlib/image_loader/image_loader.h
```diff
@@ -14,7 +14,6 @@
 #include "../uintn.h"
 #include "../image_transforms/assign_image.h"
 #include <algorithm>
-#include "../vectorstream.h"
```
davisking, line 17: `load_dng()` down below uses this though.

davisking: This an accidental change?

dlib/matrix/matrix_blas_bindings.h
```diff
@@ -13,13 +13,8 @@
 #include "cblas_constants.h"
 #include <stdint.h>
 
-//#include <iostream>
-//using namespace std;
-
 namespace dlib
 {
-
-
```
davisking, line 22: Don't forget to revert these.

dlib/media/ffmpeg_demuxer.h
```diff
@@ -995,7 +996,6 @@ namespace dlib
             Callback&& clb
         )
         {
-            using namespace std::chrono;
```
davisking, line 998: Same thing here, just do `using std::chrono::whatever` to bring in the symbols needed.  Name lookup is such that that doesn't become ambiguous when people do funny things ahead of the code.

dlib/image_loader/image_loader.h
davisking: huh, it's totally used down below though in that function :shrug:

davisking:
> I see. Actually, that's why I hurried to make this PR so that I could solve @ChristopheDumeunier's issue.
> 
> However, you scared me that I might have just broken some quirky compilers/setups by doing that stuff (I saw workarounds for GCC 4.9 while doing this PR).
> 
Yeah, maybe just leave that one. IDK.  It's an old compiler so I doubt anyone cares.  

And in general, the easy thing to do is often to replace any `using namespace std;` with a `using std::whatever_the_thing_is_its_using;` and then leave the rest of the code alone.  Kinda depends on how many lines need to change and how readable/unreadable it makes them.  Like the chrono stuff gets kinda unreadable if it's all fully scoped on each use.

dlib/test/binary_search_tree.h
```diff
@@ -18,7 +18,6 @@ namespace
 {
 
     using namespace test;
-    using namespace std;
```
davisking, line 21: Don't need to edit any of the stuff in dlib/test.  That is all fine since nothing should be including it.

davisking: Although I guess it doesn't hurt anything to do it too.

arrufat: I thought, since, I was at it, we might as well do it in the same PR... I can revert those if you want.

arrufat:
> > I see. Actually, that's why I hurried to make this PR so that I could solve @ChristopheDumeunier's issue.
> > However, you scared me that I might have just broken some quirky compilers/setups by doing that stuff (I saw workarounds for GCC 4.9 while doing this PR).
> 
> Yeah, maybe just leave that one. IDK. It's an old compiler so I doubt anyone cares.

I think it's safe to remove, as dlib now uses C++14, it doesn't look like GCC 4.9 fully supports C++14: https://en.cppreference.com/w/cpp/compiler_support/14
GCC 5 was the first version to fully support C++14.

dlib/test/binary_search_tree.h
davisking: Na either way is fine.

davisking:
Thanks, looks good :)

PR #3020, opened 2024-09-26
Add multm_prev_ layer and enhance gemm() function for PLANE_WISE operations

Cydral:
This pull request introduces a new layer, `multm_prev_`, and enhances the `gemm()` function to support PLANE_WISE operations. These changes aim to improve the flexibility and performance of matrix multiplications in deep learning models, particularly for attention mechanisms.

### New layer: multm_prev_

The `multm_prev_` layer performs element-wise matrix multiplication between the current layer's input and the previous layer's output. This new layer is particularly useful for implementing attention mechanisms and other operations that require element-wise interactions between tensors.

Key features of multm_prev_:
- Supports PLANE_WISE matrix multiplication
- Preserves sample and channel dimensions
- Efficiently handles 4D tensors

### Enhancement to gemm() function:

The `gemm()` function has been updated to support two modes of operation: CHANNEL_WISE (default) and PLANE_WISE. This modification allows for more efficient and flexible matrix multiplications, especially when dealing with 4D tensors.

Key changes to gemm():
1. Added a new parameter `g_mode` to specify the operation mode (0 for CHANNEL_WISE, 1 for PLANE_WISE)
2. Implemented PLANE_WISE mode, which performs matrix multiplication for each corresponding 2D plane across all samples and channels
3. Updated documentation to reflect the new functionality and requirements for both modes

These changes provide greater flexibility in implementing complex neural network architectures, particularly those involving attention mechanisms or other operations requiring element-wise interactions between tensors.

A new test function, `test_multm_prev()`, has been added to verify the correct functionality of the `multm_prev` layer and the enhanced `gemm()` function.


arrufat:
Nice, I was just wondering if `matmul_prev` wouldn't be a better name.

Cydral:
We can change the name without any problem. I am already dealing with compilation issues likely due to static uses of mult_prev_ (in the template part), and we will decide on the name to keep afterward.

Cydral:
On the other hand, I was thinking of using the same convention for the transformation to be applied to softmax and thus have a special layer named softmaxm. So we would take mat_softmax or perhaps better msoftmax?

arrufat:
Would it be too difficult to have just an `attention_` layer? I know that would mean doing the backpropagation by hand inside that layer, just like the `loss_barlowtwins` is doing it (but that one is just a `bn_con`).

Cydral:
It would be simpler to use for some people, but we would lose the flexibility to build attention in a potentially specific way (even though it currently follows fairly standard and structured steps). For instance, we can decide whether or not to mask, apply an additional filter for whether or not to remove a pad token before applying softmax, and so on. I was thinking more of providing, as you did for ResNet, an external definition file that gives a certain definition of the network...

arrufat:
Yes, we would lose flexibility, or maybe that layer could be initialized with a struct of options that control the behavior/features of the attention layer. But yes, it would still be less flexible.

pfeatherstone:
It would be harder to implement something like flash attention without an explicit `attention_` layer

Cydral:
Indeed, I can add a high-level declaration in the layer definition file, similar to what was done for the inception layer, like :
```
template <int embedding_dim, int nb_heads, typename SUBNET>
using attention_ = (...)
```

davisking:
Sorry, I'm just catching up on these threads.  Seems like this PR is still being worked on?  There are conflicts with master in any case.   Let me know when I should look it over :)

Cydral:
@davis, no you can do the merging. I think the conflicts with the master come from the fact that I created several branches from my own Dlib fork to be able to work on several layers in parallel. Any new layers currently being created are finished and can be integrated please.
Technically, I still have a new layer to release but I'm going to wait until all the changes have been merged into the master branch to avoid any further conflicts... let me know if that's OK with you.

Translated with DeepL.com (free version)

davisking:
> @davis, no you can do the merging.

You please merge them :)

I'll review them once there aren't merge conflicts.

>  Technically, I still have a new layer to release but I'm going to wait until all the changes have been merged into the master branch to avoid any further conflicts... let me know if that's OK with you.

Yeah that's fine :D

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
davisking, line 1: Do you intend to add this?  Like does this make life easier for vscode users in some way?  I'm not necessarily adverse to it but it seems like the sort of thing that would normally be purely local to individual users.

Cydral: No, of course. I don't know where that came from... obviously it was added by GitHub tooling when I made fixes

Cydral:
@Davis, could you please review this PR?

dlib/cuda/cublas_dlibapi.cpp
```diff
@@ -101,55 +101,118 @@ namespace dlib
             const tensor& lhs,
             bool trans_lhs,
             const tensor& rhs,
-            bool trans_rhs
+            bool trans_rhs,
+            size_t g_mode
```
davisking, line 105:
```suggestion
            gemm_mode g_mode
```
Why not use the enum here?

dlib/dnn/layers_abstract.h
```diff
@@ -4087,6 +4166,152 @@ namespace dlib
     template <long diag, long num, long den, typename SUBNET>
     using tril_diag = add_layer<tril_<diag, void, num, den>, SUBNET>;
 
+// ----------------------------------------------------------------------------------------
+
+    class positional_encodings_
```
davisking, line 4171: This class is already in this file on master (so don't add another copy of it) :D

dlib/test/dnn.cpp
```diff
@@ -4672,12 +4719,14 @@ void test_embeddings()
             test_positional_encodings();
             test_embeddings();
             test_tril();
+            test_positional_encodings();
```
davisking, line 4722:
```suggestion
```
It's already being called here.

dlib/cuda/tensor_tools.h
```diff
@@ -158,28 +158,66 @@ namespace dlib { namespace tt
...
+                in lhs and rhs across all samples and channels.
+                - The operation is equivalent to performing the following for each sample
+                and channel:
+                    dest[s][k] = alpha * (lhs[s][k] * rhs[s][k]) + beta * dest[s][k]
+                Where [s][k] represents the 2D plane for sample s and channel k.
+
+        note
+            - The PLANE_WISE mode is particularly useful for operations like attention
+            mechanisms in neural networks, where you want to perform matrix multiplications
+            on 2D planes of 4D tensors while preserving the sample and channel dimensions.
```
davisking, line 220:
```suggestion
            - The PLANE_WISE mode is particularly useful for operations like attention
              mechanisms in neural networks, where you want to perform matrix multiplications
              on 2D planes of 4D tensors while preserving the sample and channel dimensions.
```
Here and elsewhere, please make sure the comments have indenting that lines up like how you see it in other parts of the library.

Cydral: It might also be related to the text editor I'm using. I'll keep an eye on that for future modifications. In any case, thank you for this update.

dlib/cuda/cublas_dlibapi.cpp
davisking: Yeah definitely use enums instead of ints :)

Cydral: I'm proposing to add a global enum in tensor_tools.h that will serve the various functions needing to differentiate their internal calculations based on tensor organization:
`enum class operation_mode { CHANNEL_WISE = 0, PLANE_WISE = 1 };`
I'll update the code accordingly to ensure clarity and consistency.

dlib/test/dnn.cpp
Cydral: Yes, it's fixed

dlib/dnn/layers_abstract.h
Cydral: I removed this duplicated declaration

dlib/cuda/tensor_tools.h
Cydral: I've changed it a bit to match the other notes

Cydral:
@Davis, I think I'm on the right track now (a lot of difficulty to find an enum shared by all the classes and accessible from the CPU and CUDA codes for example) and everything is working on my side now (including the dnn.cpp test). As indicated in the closed PR, to simplify the review, I've merged all the modifications (i.e. for the two new layers) via this single PR.
However, when I run dtest, I get this error, which seems to have nothing to do with my new layers (it might have something to do with convolution). Does this sound familiar?
Running test_dnn \

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!! TEST FAILED: test_dnn !!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Failure message from test:

Error occurred at line 1518.
Error occurred in file C:\Users\Shadow\source\repos\multm-prev-layer\dlib\test\dnn.cpp.
Failing expression was max(abs(mat(filter_gradient1)-mat(filter_gradient2))) < 1e-3.
0.00897217



Testing Finished
Total number of individual testing statements executed: 955
Number of failed tests: 1
Number of passed tests: 0

arrufat:
I've run the test on my GPU machine (`dtest --runall`) and they all pass.

Cydral:
> I've run the test on my GPU machine (`dtest --runall`) and they all pass.

OK, so maybe it's due to the GPU I'm using (an A4500); could it be possible to review again and merge all now? There are still other elements to work on ;-)

dlib/cuda/cudnn_dlibapi.cpp
```diff
@@ -1528,61 +1528,125 @@ namespace dlib
...
         )
         {
             DLIB_CASSERT(
-                  have_same_dimensions(dest,gradient_input) == true &&
-                  have_same_dimensions(dest,grad) == true );
-            if (dest.size() == 0)
-                return;
+                have_same_dimensions(dest, gradient_input) == true &&
+                have_same_dimensions(dest, grad) == true);
+            DLIB_CASSERT(mode == operation_mode::CHANNEL_WISE || mode == operation_mode::PLANE_WISE, "Invalid softmax mode");
```
davisking, line 1597:
```suggestion
```

dlib/cuda/cudnn_dlibapi.h
```diff
@@ -4,16 +4,28 @@
...
+    // ----------------------------------------------------------------------------------------
+
+    /*!
+        This enum is used to determine the mode of operation for certain functions
+        (such as gemm and softmax) in Dlib. It specifies whether the calculation
+        should be performed based on the matrix field in nr()xnc() or if the matrix
+        should be considered in num_samples()xk(). This helps in organizing tensor
+        computations more efficiently according to the required dimensions.
+    !*/
+    enum class operation_mode { CHANNEL_WISE = 0, PLANE_WISE = 1 };
```
davisking, line 26: This is nice.  Move it into some other file though since the docs in stuff like `layers_abstract.h` refer to it.  Probably dlib/cuda/tensor.h and dlib/cuda/tensor_abstract.h are the best place.

dlib/cuda/tensor_tools.h
```diff
@@ -165,21 +165,56 @@ namespace dlib { namespace tt
...
+            For CHANNEL_WISE mode:
+                - performs: dest = alpha*L*R + beta*mat(dest)
+                Where L, R, and D are as defined above.
+
+            For PLANE_WISE mode:
+                - Performs matrix multiplication for each corresponding 2D plane (nr x nc)
+                in lhs and rhs across all samples and channels.
+                - The operation is equivalent to performing the following for each sample
+                and channel:
+                    dest[s][k] = alpha * (lhs[s][k] * rhs[s][k]) + beta * dest[s][k]
```
davisking, line 212: Here and around here too, fix up the indenting :)

dlib/cuda/cpu_dlib.cpp
```diff
@@ -1620,122 +1620,176 @@ namespace dlib
...
+            const tensor& gradient_input,
+            operation_mode mode
         )
         {
-            DLIB_CASSERT(have_same_dimensions(grad,dest));
-            DLIB_CASSERT(have_same_dimensions(grad,gradient_input));
-            ttimpl::softmax_gradient(grad.nr()*grad.nc(), grad.k(), grad, dest, gradient_input);
+            DLIB_CASSERT(have_same_dimensions(grad, dest));
+            DLIB_CASSERT(have_same_dimensions(grad, gradient_input));
+            DLIB_CASSERT(mode == operation_mode::CHANNEL_WISE || mode == operation_mode::PLANE_WISE, "Invalid softmax mode");
```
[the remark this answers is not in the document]
Cydral: > Don't need that, since the enum can't take any other values :)

test removed.

dlib/cuda/cudnn_dlibapi.cpp
Cydral: Test removed also.

dlib/cuda/cublas_dlibapi.h
```diff
@@ -22,21 +22,52 @@ namespace dlib
...
-                    - D.nr() == L.nr() && D.nc() == R.nc()
-                      (i.e. dest must be preallocated and have the correct output dimensions)
-                    - L.nc() == R.nr()
-            ensures
+    /*!
+        requires
+            - The dimensions of lhs and rhs must be compatible for matrix multiplication.
+            The specific requirements depend on the mode:
+
+            For CHANNEL_WISE mode (default):
```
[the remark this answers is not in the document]
Cydral: The indentation should now be correct.

dlib/cuda/tensor_tools.h
Cydral: Fixing indentation

dlib/cuda/cudnn_dlibapi.cpp
```diff
@@ -1528,61 +1528,125 @@ namespace dlib
...
-            const tensor& src
+            const tensor& src,
+            operation_mode mode
         )
         {
-            DLIB_CASSERT(have_same_dimensions(dest,src));
-            if (src.size() == 0)
-                return;
+            DLIB_CASSERT(have_same_dimensions(dest, src));
+            DLIB_CASSERT(mode == operation_mode::CHANNEL_WISE || mode == operation_mode::PLANE_WISE, "Invalid softmax mode");
```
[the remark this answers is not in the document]
Cydral: Test removed

dlib/cuda/cudnn_dlibapi.h
Cydral: This may seem less "normal" as a location for this enum, but because of the some complex links between the different classes, it's the only place I've found that leads to visibility by all the other classes, whatever the compilation model. Putting it in tensor.h or its abstract class doesn't work, for instance.

davisking: Ah, yeah I see.  Put it in its own header then and #include it here.  I only care since cudnn_dlibapi.h is a private detail.  But this enum is exposed to users, so it's got to be public somewhere.  So in some header users are meant to read (and not full of stuff they aren't supposed to read in particular).

Cydral: OK, the enum has been moved to its own header, as proposed.

Cydral:
> Nice this is great. Fix up the minor stuff I commented on and I'll merge it right away :D

Done.

dlib/dnn/layers.h
```diff
@@ -3985,31 +4079,28 @@ namespace dlib
 
 // ----------------------------------------------------------------------------------------
 
+    template <unsigned long s_mode_>
```
davisking, line 4082:
```suggestion
    template <operation_mode s_mode_>
```
Change these to use `operation_mode` too though.  Like don't use an integer anywhere to represent the `operation_mode` but instead use `operation_mode` to represent the `operation_mode` :)

dlib/dnn/layers.h
```diff
@@ -3985,31 +4079,28 @@ namespace dlib
...
-        void setup (const SUBNET& /*sub*/)
-        {
-        }
+        void setup(const SUBNET& /*sub*/) {}
 
         void forward_inplace(const tensor& input, tensor& output)
         {
-            tt::softmax(output, input);
-        } 
+            tt::softmax(output, input, static_cast<operation_mode>(s_mode_));
```
davisking, line 4093:
```suggestion
            tt::softmax(output, input, s_mode_);
```
E.g. then this is just this.   There shouldn't be any `static_cast<operation_mode>()` happening.

Cydral: Thanks for updating the code accordingly.

dlib/dnn/layers.h
Cydral: Thanks for updating the code accordingly.

dlib/dnn/layers.h
davisking: No problem :)

PR #3028, opened 2024-10-17
Remove vscode stuff

arrufat:


davisking:
Yeah saw that go in.  I have mixed feelings about it.  If it's really for real super annoying to not have it it can be there. But better to not have it.  I figured @Cydral really wanted it since it was in several PRs :)   

In any case, I can be talked into putting it back if it's really worth it.  It does seem like user settings ought to be able to be elsewhere though.  I use vim and not vscode though so I don't have perspective on it.

arrufat:
Yeah, I don't use vs code either, but I figured that this kind of config should be personal. I'm not even sure about what those settings do.

Cydral:
On my end, I don't need it; I guess it's added automatically when I use the editor directly on GitHub...

arrufat:
What's surprising is that .gitignore contains a pattern for it.

PR #3039, opened 2024-12-27
Fix conversion from simd8f to simd8i for MSVC ARM on Windows

ebroglio:
This is needed because the conversion between float32x4_t to int32x4_t needs to be done explicit.

### DIFF
--- a/dlib/simd/simd8f.h
+++ b/dlib/simd/simd8f.h
@@ -106,8 +106,13 @@ namespace dlib
         inline operator simd8i::rawarray() const 
         { 
             simd8i::rawarray temp;
+#if defined(DLIB_HAVE_NEON)
+            temp.low = simd4i(vcvtq_s32_f32(_low));
+            temp.high = simd4i(vcvtq_s32_f32(_high));
+#else
             temp.low = simd4i(_low);
             temp.high = simd4i(_high);
+#endif
             return temp;
         }
 


dlib/simd/simd8f.h
```diff
@@ -106,8 +106,13 @@ namespace dlib
         inline operator simd8i::rawarray() const 
         { 
             simd8i::rawarray temp;
+#if defined(DLIB_HAVE_NEON)
+            temp.low = simd4i(vcvtq_s32_f32(_low));
+            temp.high = simd4i(vcvtq_s32_f32(_high));
+#else
```
davisking, line 112: What happens exactly?  It seems like this shouldn't be needed, since `simd4f` (which is what `_low` and `_high` are) has this already: `inline operator int32x4_t() const { return vcvtq_s32_f32(x); }`.  And then `simd4i` has has this constructor as well `inline simd4i(const int32x4_t& val):x(val) {}`.  So this change shouldn't be needed. But I guess visual studio is throwing some error?  What's the error exactly?

ebroglio: You are completely right. For whatever reason the operator `inline operator int32x4_t()` was missing on my side. Actually it compiled just fine, only it resulted in a crash at runtime in fhog.h. I will close this pull request.

ebroglio:
Closed because it is a problem on my machine.

PR #3043, opened 2025-01-11
Replace deprecated pkgutil.find_loader()

penguinpee:
That method has been deprecated in Python 3.12 and will be removed from
Python 3.14. Replace it with a direct call to
`importlib.util.find_spec()`, which `pkgutil.find_loader()` was wrapping
around.

### DIFF
--- a/tools/python/test/utils.py
+++ b/tools/python/test/utils.py
@@ -1,4 +1,4 @@
-import pkgutil
+import importlib
 import sys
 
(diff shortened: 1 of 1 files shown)


penguinpee:
For the removal changelog entry see:

https://docs.python.org/dev/whatsnew/3.14.html#pkgutil

davisking:
Thanks for the update :D

PR #3047, opened 2025-01-19
find blas and cudnn update for windows OS

hscodesolution:
cmake unable to detect blas, lapack, intel mkl and cudnn on my windows 11 64 bit OS. I fixed the issue by suggesting these changes in find_blas.cmake and find_cudnn.txt which will help cmake to find out proper path of library and include files.


davisking:
Nice, thanks for the PR :)

PR #3048, opened 2025-01-22
minor tweaks to the transformer example

arrufat:
Just a few minor tweaks to align the example's style with other examples in dlib.
@Cydral, please check if you agree. And great job, btw!


davisking:
Yeah at this point I could go either way.  Sometimes I `cout` and sometimes I `std::cout`.  Depends on the day :shrug: :D

PR #3052, opened 2025-02-03
Update minimum CMake version to 3.10.0

arrufat:
This should fix the warnings that appeared with CMake 3.31: Compatibility with CMake < 3.10 will be removed from a future version of CMake.

CMake 3.10.0 was released in November 2017, so this shouldn't be a hard requirement to meet.


davisking:
Works for me :D

PR #3055, opened 2025-02-09
Slice layer

kSkip:
I am training a U-Net using the architecture from [U-Net: Convolutional Networks for Biomedical Image Segmentation](https://arxiv.org/pdf/1505.04597). It requires cropping the tensors from the encoding stages to concatenate with the outputs of the decoding stages. I created the slice layer to implement this.

AFAICK, there are no layers in dlib that accommodate this. However, I probably should have asked before creating this one.

Some of my considerations:
- It is similar to extract, but instead it copies non-contiguous regions to perform a 3D crop. This is both different behavior and performance.
- I could have used an alias_tensor, similarly to extract, by aliasing every row. The CPU implementation may have been similar performance doing this. However, this is not ideal with CUDA because it requires a kernel invocation for every row. Since kernels on the same stream execute serially, the GPU throughput would not be saturated. So, I overloaded `copy_tensor` to copy non-contiguous rows with a single kernel call.
- I don't know if the public interface is the best, but it works for my case. I'm happy to hear any suggestions or improvements so this can be useful to others.

Thanks for taking a look!

### DIFF
--- a/dlib/cuda/tensor_tools.h
+++ b/dlib/cuda/tensor_tools.h
@@ -2344,6 +2344,27 @@ namespace dlib { namespace tt
         size_t sk, size_t snr, size_t snc,
         size_t k, size_t nr, size_t nc
     );
+    /*!
+        requires
+            - dest.num_samples() == src.num_samples()
+            - dest.k() - dk >= k
+            - dest.nr() - dnr >= nr
+            - dest.nc() - dnc >= nc
+            - src.k() - sk >= k
+            - src.nr() - snr >= nr
+            - src.nc() - snc >= nc
+            - is_same_object(dest,src) == false
+            - The memory areas of src and dest do not overlap.
+        ensures
+            - if (add_to) then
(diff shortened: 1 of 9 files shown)


dlib/cuda/tensor_tools.h
```diff
@@ -2334,6 +2334,17 @@ namespace dlib { namespace tt
...
+// ----------------------------------------------------------------------------------------
+
+    void copy_tensor(
+        bool add_to,
+        tensor& dest,
+        size_t dk, size_t dnr, size_t dnc,
+        const tensor& src,
+        size_t sk, size_t snr, size_t snc,
+        size_t k, size_t nr, size_t nc
+    );
```
davisking, line 2346: Add docs for this (like you see for the other functions in this file :) )

davisking (commented):
Nice yeah this seems like a useful thing to have :D

dlib/cuda/tensor_tools.h
kSkip: Oh, I forgot that. Added the doc string. Thanks!

davisking:
Thanks for the PR :)

PR #3056, opened 2025-02-15
Add Byte Pair Encoding (BPE) class for subword tokenization

Cydral:
#### Description:
This PR introduces a new `bpe_tokenizer` class to Dlib, implementing the **Byte Pair Encoding (BPE)** algorithm for subword tokenization. The BPE tokenizer is a widely used technique in natural language processing (NLP) for handling out-of-vocabulary words and reducing vocabulary size while maintaining text representation capabilities.

#### Key Features:
- **BPE Algorithm**: Implements the BPE algorithm as described in [Sennrich et al., 2016](https://arxiv.org/abs/1508.07909).
- **Special Tokens**: Supports predefined special tokens (e.g., `<text>`, `<url>`, `<image>`) for marking specific elements in the text.
- **Training and Encoding**: Provides methods for training the tokenizer on a text corpus and encoding/decoding text into subword tokens.
- **Serialization**: Supports saving and loading the tokenizer model and vocabulary for reuse.
- **Thread-Safe**: Utilizes multi-threading for efficient frequency statistics computation during training.

#### Usage:
```cpp
dlib::bpe_tokenizer tokenizer;
tokenizer.train(corpus_text, target_vocab_size, true); // Train on a text corpus
std::vector<int> tokens = tokenizer.encode("Sample text to tokenize."); // Encode text
std::string decoded_text = tokenizer.decode(tokens); // Decode tokens back to text

### DIFF
--- a/dlib/test/tokenizer.cpp
+++ b/dlib/test/tokenizer.cpp
@@ -377,14 +377,12 @@ namespace
 
         test.train(training_text, 300, true);
 
-        std::ofstream out_file("bpe_tokenizer_model.dat", std::ios::binary);
-        serialize(test, out_file);
-        out_file.close();
+        std::ostringstream out_stream;
+        serialize(test, out_stream);
 
         bpe_tok loaded_test;
-        std::ifstream in_file("bpe_tokenizer_model.dat", std::ios::binary);
-        deserialize(loaded_test, in_file);
-        in_file.close();
+        std::istringstream in_stream(out_stream.str());
+        deserialize(loaded_test, in_stream);
 
         std::vector<std::string> test_strings = {
             u8"\nThis is a test of the tokenisation process...\nimplemented in the Dlib library!\n", // English
--- a/dlib/test/tokenizer.cpp
+++ b/dlib/test/tokenizer.cpp
@@ -4,6 +4,7 @@
 
 #include <string>
 #include <sstream>
+#include <regex>
 
 #include <dlib/tokenizer.h>
 #include "tester.h"
@@ -350,6 +351,14 @@ namespace
 
     }
 
+    string postprocess_decoded_text(const string& decoded) {
+        string result = decoded;
+        result = regex_replace(result, std::regex("<text>"), "");
+        result = regex_replace(result, std::regex("</text>"), "\n");
+        if (!result.empty() && result.back() == '\n') result.pop_back();
+        return result;
+    }
+
     template <
         typename bpe_tok
     >
@@ -385,7 +394,7 @@ namespace
         deserialize(loaded_test, in_stream);
 
         std::vector<std::string> test_strings = {
-            u8"\nThis is a test of the tokenisation process...\nimplemented in the Dlib library!\n", // English
+            u8"This is a test of the tokenisation process...\nimplemented in the Dlib library!", // English
             u8"Ceci est un test du processus de\ntokenisation implémenté dans\nla bibliothèque Dlib!", // French
             u8"Dette er en test af tokeniseringsprocessen implementeret i Dlib-biblioteket!", // Danish
             u8"这是对Dlib库中实现的标记化过程的测试！" // Chinese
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: dlib/test/tokenizer.cpp [hunk 1]
```diff
@@ -350,9 +350,59 @@ namespace
...
+        test.train(training_text, 300, true);
+
+        std::ofstream out_file("bpe_tokenizer_model.dat", std::ios::binary);
+        serialize(test, out_file);
+        out_file.close();
+
+        bpe_tok loaded_test;
+        std::ifstream in_file("bpe_tokenizer_model.dat", std::ios::binary);
+        deserialize(loaded_test, in_file);
+        in_file.close();
```
davisking, line 387: This is a good thing to test.  But use the `std::ostringstream` and `std::istringstream` so the test doesn't end up leaving files around. With the stringstream it's all just in memory.

dlib/test/tokenizer.cpp
```diff
@@ -350,9 +350,59 @@ namespace
...
+        };
+
+        for (const auto& text : test_strings) {
+            std::vector<int> encoded = loaded_test.encode(text);
+            std::string decoded = loaded_test.decode(encoded);
+
+            std::cout << "Original: " << text << "\n";
+            std::cout << "Encoded: ";
+            for (int id : encoded) std::cout << id << " ";
+            std::cout << "\nDecoded: " << decoded << "\n----------------------------------------\n";
```
davisking, line 403: Do `DLIB_TEST(text == decoded)` right?  Don't cout anything.  Need to `DLIB_TEST()` something for the test to do anything.

dlib/tokenizer/bpe_tokenizer.h
```diff
@@ -0,0 +1,479 @@
...
+     *
+     * The implementation is based on the original BPE algorithm described in:
+     * - Sennrich, R., Haddow, B., & Birch, A. (2016). Neural Machine Translation of Rare Words with Subword Units.
+     *   In Proceedings of the 54th Annual Meeting of the Association for Computational Linguistics (ACL 2016).
+     *
+     * @note The tokenizer is designed to work with UTF-8 encoded text and supports a maximum token length of 8 bytes.
+     *       This limit can be adjusted by modifying the `MAX_TOKEN_LENGTH` constant.
+     *
+     */
+    class bpe_tokenizer
```
davisking, line 47: This is cool.  Add a bpe_tokenizer_abstract.h file and put the docs in there so it's like all the other docs in dlib.  I'll eventually link that into the dlib.net web page and docs and whatnot too.  And use the same comment/doc style as the other parts of the library (outlined in https://dlib.net/intro.html#notation but there are tons of examples in the library)

dlib/test/tokenizer.cpp
Cydral: The modification has been done... and it is indeed simpler this way.

dlib/test/tokenizer.cpp
Cydral: The decoded text is almost identical to the source text... except for the treatment of “\n” which correspond - for the provided implementation - to paragraph markers (so adding tokens between specific tokens corresponding to <text>...</text>). To apply a DLIB_TEST() test, I've added a short post-processing function however.

dlib/tokenizer/bpe_tokenizer.h
Cydral: I've taken your request into account and created an abstract header file (bpe_tokenizer_abstract.h) for the bpe_tokenizer class. The documentation follows Dlib's style guidelines, and the class is now ready for integration into the official documentation. Please, feel free to chain the abstract class wherever it fits best in the Dlib structure. Let me know if any adjustments are needed.

davisking:
Nice, this is great.  Sorry it took so long for me to get back to this.

Cydral:
No problem. I think this is another great new feature for our library.

davisking:
> No problem. I think this is another great new feature for our library.

Indeed 😁

PR #3058, opened 2025-02-24
Fix: Potential Vulnerability in Cloned Function

tabudz:
**Description**
This PR fixes a security vulnerability in start_input_tga() that was cloned from libjpeg-turbo but did not receive the security patch. The original issue was reported and fixed under https://github.com/libjpeg-turbo/libjpeg-turbo/commit/82923eb93a2eacf4a593e00e3e672bbb86a8a3a0.
This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2018-11212
https://github.com/libjpeg-turbo/libjpeg-turbo/commit/82923eb93a2eacf4a593e00e3e672bbb86a8a3a0

### DIFF
--- a/dlib/external/libjpeg/rdtarga.c
+++ b/dlib/external/libjpeg/rdtarga.c
@@ -365,7 +365,8 @@ start_input_tga (j_compress_ptr cinfo, cjpeg_source_ptr sinfo)
       width <= 0 || height <= 0 ||
       source->pixel_size < 1 || source->pixel_size > 4 ||
       (UCH(targaheader[16]) & 7) != 0 || /* bits/pixel must be multiple of 8 */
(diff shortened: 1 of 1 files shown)


davisking:
Nice, thanks for the fix.

tabudz:
Hi @davisking, thanks for merging my PR. Just want to let you know that I plan to report this as a CVE. Let me know if you have any concern!

DragonL641:
What is the point of this bug fix, in Line 365, the size of image is already checked before the changed line?

davisking:
Heh yeah, I just verified that the same patch was in the upstream repo.  But that's a slightly different version of the code and it's not needed here.  No worries though.

Issue #3045, opened 2025-01-16
[Bug]: Build break in LLVM v19


fs-eire:
### What Operating System(s) are you seeing this problem on?

Other (plase, specify in the Steps to Reproduce)

### dlib version

19.24.6

### Compiler

clang 19

### Expected Behavior

Build pass.

### Current Behavior

```
/sysroot/include/c++/v1/string:820:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned int>'
  820 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
/_deps/dlib-src/dlib/matrix/../unicode/unicode.h:718:19: note: in instantiation of template class 'std::basic_string<unsigned int>' requested here
  718 |     const ustring convert_to_utf32 (
      |                   ^
/sysroot/include/c++/v1/__fwd/string.h:23:29: note: template is declared here
   23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
```

in unicode/unicode.h, `unichar` is defined to `uint32`, and `std::char_traits<unichar>` is used but the template instantiation failed because LLVM libc++ v19 deprecated this behavior:

> The base template for std::char_traits has been removed in LLVM 19. If you are using std::char_traits with types other than char, wchar_t, char8_t, char16_t, char32_t or a custom character type for which you specialized std::char_traits, your code will stop working. The Standard does not mandate that a base template is provided, and such a base template is bound to be incorrect for some types, which could currently cause unexpected behavior while going undetected.

https://libcxx.llvm.org/ReleaseNotes/19.html#deprecations-and-removals

Simply replacing `using unichar = uint32;` to `using unichar = char32_t;` does not fix this build error because it causes other code to fail (eg serialize.h)

### Steps to Reproduce

Build with Clang 19

arrufat:
I've just tried to build dlib and `imglab`, which makes use of the Unicode stuff, with Clang 19.1.6 without issues. How are you building?

```
CC=clang CXX=clang++ cmake -B build -G Ninja
cmake --build build
```

fs-eire:
I am using Emscripten v4.0.0 to build ONNXRuntime, while dlib is an indirect dependency.

I checked the toolset version. The LLVM binaries version is [322eb1a92e6d4266184060346616fa0dbe39e731](https://github.com/llvm/llvm-project/commit/322eb1a92e6d4266184060346616fa0dbe39e731):

[...]

PR #3059, opened 2025-02-25
change the Unicode string default char type to char32_t

wenbingl:
Motivation: in some LLVM 19 based compiler (like emsdk v4.x), The base template for std::char_traits has been removed in LLVM 19. and the std::char_traits can only work with types other than char, wchar_t, char8_t, char16_t, char32_t. More discussion can be found in #3045 

Validation: 
    1. dlib unit test build pass, cannot run full test with or without the changes.
    2. emsdk 4.0.3 build pass


davisking:
Awesome, thanks for the update :D

Issue #1852, opened 2019-07-31
Choose backend (GPU/CPU)


xd009642:
So I couldn't find a way to do this in the documentation but I was wondering if there was a way to force dlib to use the CPU if it's built with CUDA enabled? I ask because I'm making a multithreaded application with dlib and only a maximum of 2 threads at a time can use CUDA (because of limitations of the GPU itself). So I'd like to in a thread get dlib to use the GPU in the thread if <2 threads are using the GPU and otherwise fallback to using the CPU. 

Also a slightly related question how would dlib cope with a machine with multiple GPUs installed all for compute purposes?

davisking:
There is no way to switch at runtime.  However you build it is the mode in
which it will run.  To assign things to different GPUs you use the normal
cudaSetDevice() method provided by the CUDA runtime.

xd009642:
Would there be any interest in adding the ability to run something on CPU if built with GPU support? Just because being able to handle this in a single binary would be preferential to building two different binaries linked to two different builds of dlib and coordinating between the two.

Cuda set device is super useful for the second question though cheers! Just saw on docs that it sets the device for the host thread 😊

davisking:
Sure, such an option would be cool.  So a PR that set that up would be
great.  It would probably best be accomplished via an API similar to
cudaSetDevice().  That is, calling some global function to set CPU or GPU,
for the current thread.

pliablepixels:
Is there a chance to reopen this? Maybe @xd009642 made some progress?
I am in a situation where I use dlib in GPU mode for inferencing, but depending on the training set, need to switch to CPU as GPU memory runs out.

pliablepixels:
Continuing on a closed thread @davisking unless you think this warrants a new issue.

I took at quick look at the code, it looks like CPU code is compiled out when `DLIB_USE_CUDA` is used so its not a simple matter of setting a runtime flag. Supporting switching to CPU when GPU is enabled would be a non trivial effort that would involve keeping both code bases compiled in. Is that right or am I going down the wrong path?

xd009642:
I didn't make any progress aside from looking at the code, unfortunately real-life interfered and I lost the need for this functionality at work. My approach though would have been to keep both code-bases compiling in as well as provide an option to keep the old behaviour (just in case any users had a requirement on the old behaviour for binary size etc).

I could potentially have another look at this over the Christmas holidays but I won't realistically get a chance before December 20th so you can always give it a shot if you wish

[...]

PR #3060, opened 2025-02-26
Support for runtime CPU/CUDA selection

kSkip:
This PR adds support for selecting either the CPU or CUDA implementation of the dnn module at runtime. As discussed in https://github.com/davisking/dlib/issues/1852, the `#ifdef DLIB_USE_CUDA` directives in `tensor_tools.cpp` just need to be replaced with a runtime branch like

```
if (use_cuda())
    cuda::function();
else
    cpu::function();
```

The thread local variable returned by `use_cuda()` is initialized to be consistent with the current behavior. If `DLIB_USE_CUDA=True` the function, `use_cuda()` will return `true` by default.

I found while making these changes that is was necessary to keep the preprocessor directives because the `cuda` namespace cannot be referenced when CUDA is not linked. Additionally, the pooling and convolution classes in `tensor_tools.h` have instances of both implementations as members. I couldn't find a cleaner way (with limited time) to handle those cases.


davisking:
Nice, let me test this a bit and look it over.

kSkip:
Hey @davisking! Do you have any feedback or get a chance to test this code?

davisking:
Yeah sorry. I know I'm super late to review this.  Looking at it now a bit.  Just pushed a merge commit with master into the PR too.

dlib/cuda/tensor_tools.h
```diff
@@ -20,6 +20,9 @@ namespace dlib
     bool dnn_prefer_fastest_algorithms();
     void set_dnn_prefer_fastest_algorithms();
     void set_dnn_prefer_smallest_algorithms();
+
+    bool use_cuda();
+    void set_use_cuda(bool flag);
```
davisking, line 25: Add any new public functions to the docs (the _abstract.h files).  These should be documented next to `set_dnn_prefer_smallest_algorithms()`, so over in `dlib/dnn/core_abstract.h`.

dlib/cuda/tensor_tools.cpp
```diff
@@ -77,12 +105,19 @@ namespace dlib { namespace tt
...
-        if (add_to)
-            out += sum_cols(pointwise_multiply(mat(lhs), mat(rhs))); 
+        if (use_cuda())
+        {
+            cuda::dot_prods(add_to, out, lhs, rhs);
+        }
         else
-            out = sum_cols(pointwise_multiply(mat(lhs), mat(rhs))); 
+        {
+#endif
```
davisking, line 114: So this PR is a good idea.  But can you change these so they are simpler?  The preprocessor macro logic is now pretty complicated and makes the code hard to read.

But it could be done like this instead:
```
    void dot_prods (
        resizable_tensor& out, 
        const tensor& lhs, 
        const tensor& rhs
    )    
    {    
        IF_DLIB_USE_CUDA(
            cuda::dot_prods(out, lhs, rhs);
        ) 

        IF_DLIB_NOT_USE_CUDA(
            out = sum_cols(pointwise_multiply(mat(lhs), mat(rhs))); 
        )    
    }   
```

Which would be a pretty trivial mechanical change from what dlib has now and doesn't really change the readability of the code.  And inside `IF_DLIB_USE_CUDA` you just have an `if (use_cuda())` wrapped inside a `#ifdef DLIB_USE_CUDA`.

That is, define these macros:
```
#ifdef DLIB_USE_CUDA
#define IF_DLIB_USE_CUDA(expression) if (use_cuda()) { expression }
#else
#define IF_DLIB_USE_CUDA(expression) 
#endif

#ifdef DLIB_USE_CUDA
#define IF_DLIB_NOT_USE_CUDA(expression) if (!use_cuda()) { expression }
#else
#define IF_DLIB_NOT_USE_CUDA(expression) expression
#endif
```
and then any code that looks like this on master:
```
#if DLIB_USE_CUDA 
    foo;
    bar;
    whatever;
#else
    foo_cpu;
    whatever2;
#endif   
```

Just turns into:
```
IF_DLIB_USE_CUDA(
    foo;
    bar;
    whatever;
)

IF_DLIB_NOT_USE_CUDA(
    foo_cpu;
    whatever2;
)
```

davisking:
Yeah sorry for the late review.  Just been real busy with other things.  Every day I see this in my email and think about how late I've been but hasn't made it up my todo list until today.  Haven't had a lot of free time lately :cry:

davisking (commented):
Ah I just realized this needs more work.  I've been going over it for the last hour and a bit.  I was updating the unit tests to work with and without a GPU.  But I just realized that even when `use_cuda()==false` it's still using cuda.  You can see this by running this for example:
```
CUDA_VISIBLE_DEVICES= ./dnn_face_recognition_ex ../faces/bald_guys.jpg 
```
with this change:
```
--- a/examples/dnn_face_recognition_ex.cpp
+++ b/examples/dnn_face_recognition_ex.cpp
@@ -80,6 +80,7 @@ std::vector<matrix<rgb_pixel>> jitter_image(
 
 int main(int argc, char** argv) try
 {
+    dlib::set_use_cuda(false);
     if (argc != 2)
     {
         cout << "Run this example by invoking it like this: " << endl;
```

You will get this error.  Since the `resizable_tensor` objects are all still calling `cudaMalloc` and allocating GPU memory.  So the tensor objects would need to be updated too in a similar way to the other code so as to not actually do any allocations when cuda is disabled.
```
Error while calling cudaGetDevice(&the_device_id) in file /home/davis/source/dlib/dlib/cuda/gpu_data.cpp:204. code: 100, reason: no CUDA-capable device is detected
```

That `CUDA_VISIBLE_DEVICES=` tells the cuda runtime to pretend there isn't any GPU attached to the machine.  That's really what you want to work here.  So that someone can run this stuff on a non-GPU machine even though it's compiled with cuda.

Anyway, I pushed the code I had.  If you update the tensors to also use `use_cuda()` then it should be good to go, unless I'm forgetting some other cuda tooling in there.  But I think that's it.  The key test is if prefixing programs with `CUDA_VISIBLE_DEVICES=` results in them running without error even though they are built with cuda support :D 

I updated the unit tests in a way that should work with `CUDA_VISIBLE_DEVICES=`, but like I was saying, I went to test that after doing so and was reminded about the tensor allocation stuff :(

davisking:
> > Since the resizable_tensor objects are all still calling cudaMalloc and allocating GPU memory.
> 
> Oh, good catch. I didn't check the behavior with `CUDA_VISIBLE_DEVICES=`. It seems like this can be accomplished by changing only the `gpu_data` class, unless I am missing something.
> 
> If we conditionally allocate the CUDA memory, we should also support the following scenario.
> 
> ```
> resizable_tensor data;
> 
> set_use_cuda(false);
> data.set_size(batch_size, channels, width, height);
> float* host_ptr = data.host();
> 
> // do something with host_ptr
> 
> set_use_cuda(true);
> float* device_ptr = data.device();
> 
> // do something with device_ptr
> ```
> 
> Here we must check if the device buffer needs to be allocated when the data currency is checked.

Yep, that would make it all work smoothly unless I'm forgetting another part of it as well. But I think that's the only bit remaining.

davisking:
> Also, do you still want to add the `do { } while(0)` to the macro?

Na, it's good how it is.

kSkip:
I think this is working now.

Let me know what you think about this behavior. A tensor allocated as host only (i.e. `use_cuda() == false`) must remain host only unless another call to `set_size` occurs. So, the situation I previously outline above https://github.com/davisking/dlib/pull/3060#issuecomment-2907793967 is actually illegal and throws an exception. To accomplish the same thing, you must create a new tensor.
```
resizable_tensor data;

set_use_cuda(false);
data.set_size(batch_size, channels, width, height);
float* host_ptr = data.host();

// do something with host_ptr

set_use_cuda(true);

{
    resizable_tensor temp_data(data);
    data.swap(temp_data);
}

float* device_ptr = data.device();

// do something with device_ptr
```
I decided that transformation between a tensor allocated with swap-able memory to one with device/page-locked memory pairs should be explicit. Alternatively, we would have to perform allocations and copies during methods calls like `device()`, which would lead to unexpected behavior, and just seemed like a bad idea when I started working on it.

I also added a couple of other changes.

- *device_id* -1 means it is a host only `gpu_data` instance. This helps select the appropriate memcpy method as well as provides a means to detect these instances.
- If the *device_id* for two `gpu_data` instances are both -1, then `std::memcpy` is used instead of `cudaMemcpy`. As a matter of fact, `cudaMemcpy`, even with `cudaMemcpyHostToHost`, actually fails if there are no devices.
- The curand generator must be initialized only if `is_available() == true`. I couldn't think of a better way to handle this. Creating the generator without any available devices fails.

On my machine, the tests all pass with `CUDA_VISIBLE_DEVICES=`, and do not appear to allocate any device memory.

kSkip:
@davisking What do you think about these changes to the cuda memory allocation?

I suppose the obvious alternative is to defer cuda allocation until a call to `device()` is made, but I wasn't sure about that.

davisking:
I just remembered this PR.  Is this working for you and you think ready to go?  Not sure what the state of it is.  Sorry for not being more attentive to this PR here :|

davisking:
Maybe the thing to do is to simplify this more.   E.g. this stuff isn't going to allow someone to switch cuda on and off for instance.   The thread local state and other bits of state will get all missed up in a number of different ways depending on the pattern/sequence of on/off attempts and other dlib API calls someone makes.

So instead of having an actual set_use_cuda() this could all be simplified by having a single global on off flag that is either on or off at program startup and is never allowed to change.  Basically the dlib equivalent to the `CUDA_VISIBLE_DEVICES` env var the cuda runtime has.   And maybe it's literally just controlled by an environment variable too.   So you can do `DLIB_DISABLE_CUDA_USE=1` as an env var and that toggles cuda use off.  And otherwise it's on (if you build with cuda and a GPU is available at program startup).

davisking:
Pushed a commit for what I mean.  Or basically what I mean, needs more review to be 100% sure it's right but I ran the example programs with and without that new env var and it all seems to work properly.  And `CUDA_VISIBLE_DEVICES` also toggles cuda use on and off too as you would hope.

kSkip:
Hey, no worries. Thanks for getting back! This branch is working fine (and correctly) for the project I need it for. However, I will need to check everything again because I just ran a major Linux distro upgrade and need to reconfigure my CUDA package repos before testing.

I understand your concern about the mutability of the `thread_local` variable. One thing that I need this for is a GIMP plugin I wrote that slices an image into tiles and processes them with a denoising U-Net. To maximize throughput, I create several model instances on *both* the GPU and CPU in different worker threads. Ideally, this work remains in the same process for simplicity.

It would be really nice if we could develop a solution that guarantees the immutability, but with thread level instead of process level granularity. Maybe this could be done by only allowing calls to `set_use_cuda()` before any other call to `use_cuda()` in that thread.

>  The thread local state and other bits of state will get all missed up in a number of different ways depending on the pattern/sequence of on/off attempts and other dlib API calls someone makes.

The situation with the `thread_local` flag isn't really that different from improper calls to `cuda::set_device()`. Another possible solution could be to unify the state modifications with `cuda::set_device()`, such that you could set CUDA device 0, 1, etc. or CPU "device". But that might require an API breaking change, so I'm not sure.

davisking:
Hmm yeah ok.  Well how about this, I'll give this PR in its current form another once over and then merge it into master.  And you can make another (now super tiny PR) that adds the ability to control this stuff in your plugin setup in a way that works for you and we can merge than in in a follow up PR.  I think it's fine to have so long as the API around controlling the switching is very clear about the limitations involved.  One of the most tricky ones is around the training code since that spawns threads.  There are also users that use threads too.  So it's just  got to be super clear how it can be correctly used, which we can figure out.  And for most users the global runtime switching is what they would use so hardly anyone would need to use the more "advanced" per thread switching API too, making it not a tempting thing to mess with for them and then accidentally get wrong.

kSkip:
Cool, that sounds like a good plan. Thanks again. I'll follow up with the next PR when I get a chance. Let me know if you need anything else from me for this one.

dlib/cuda/gpu_data.h
```diff
@@ -109,6 +119,7 @@ namespace dlib
 #ifndef DLIB_USE_CUDA
             DLIB_CASSERT(false, "CUDA NOT ENABLED");
 #endif
+            DLIB_CASSERT(device_id() >= 0, "This data is host only");
```
kSkip: Do we really need these runtime assertions now that we guarantee immutability of the cuda/cpu execution context?

davisking: yeah sorta. I'm cleaning this up a bit.  I think it's good to have an assert here, just not this exact one.

davisking:
Ha, been a year but this is finally merged.  Thanks for the PR :D It's a good one.

kSkip:
Awesome. Also, thanks for your help!

davisking:
> Awesome. Also, thanks for your help!

My pleasure :D

PR #3064, opened 2025-03-03
`throw()` -> `noexcept`

r-barnes:
`throw()` has been deprecated since C++11 and is removed in C++20. `noexcept` is accepted alternative.


davisking:
Thanks for the PR :D

PR #3067, opened 2025-03-14
Drop namespace std

penguinpee:
Solve build error when compiling with GCC 15.


penguinpee:
Having looked a bit closer at the other source files in `tools/python/src`, the issue could also be solved with the approach used in `matrix.cpp`. That is including `<dlib/string.h>` and replacing `use namespace std;` with `use std::string;` and `use std::ostringstream;`.

I'm open to suggestions (not a C/C++ expert, nor very familiar with the code base).

davisking:
Thanks for the PR :D 

What you did is a fine way to deal with the issue.

PR #3068, opened 2025-03-31
Add cmake to build time dependencies

Dobatymo:
Using cmake as a build time dependency, users can simply use `pip install dlib` to install dlib using a modern pip verion which uses build isolation by default.

Without this dependency (and without disabling build isolation) users would receive a "CMake is not installed on your system!" error.

EDIT: I just saw that this dependency was removed here... https://github.com/davisking/dlib/pull/2878
Since binary wheels are provided for almost all platforms, in my opinion it's better to have it as a build time dependecy. This way it can be installed with a clean environment and without having to consult the docs first.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools", "wheel", "cmake>=3.10.0"]
 build-backend = "setuptools.build_meta"
--- a/setup.py
+++ b/setup.py
@@ -252,7 +252,6 @@ setup(
     ext_modules=[CMakeExtension('_dlib_pybind11','tools/python')],
     cmdclass=dict(build_ext=CMakeBuild),
     zip_safe=False,
-    #install_requires=['cmake'], # removed because the pip cmake package is busted, maybe someday it will be usable.
     packages=find_packages(exclude=['python_examples']),
     package_dir={'': 'tools/python'},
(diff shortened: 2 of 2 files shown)


davisking:
Eh, this has been added and removed like 5 times now over the last many years.  Invariably it's fine on some platforms but then there are platforms where it literally is broken.  Maybe it's time for another attempt.  But what happens when it's bad is someone has cmake installed already, via some more official non-pip way of installing cmake (the most common way to install cmake is notably *not* with pip).  They then try to build dlib and pip pulls down yet another version of cmake. That new cmake version it pulls down doesn't work at all and dlib fails to build, or pip spends a long time compiling the new cmake that is not needed, or pip pulls down a new cmake and that new cmake somehow interferes with the existing official cmake and then users find that they can't even use make anymore themselves on the CLI.

All of that stuff has happened multiple times :(

Dobatymo:
Yeah sorry I should have search before I opened this PR. For me personally it works perfectly with multiple Python versions on Windows. But I can see there are problems for others and leaving the dependency out, provides a way for everyone to install it, albeit with some manual work, vs. making it difficult to fix for people who have problems when it's included. I will close the PR.

davisking:
Na it's ok.  I want it to work too.  I'm not sure what to do about it though.  It also works for me on the platforms I use, but invariably when this is added people start posting about how it breaks things for them :cry:

PR #3069, opened 2025-03-31
Fix SyntaxWarning in setup.py

Dobatymo:
Newer python versions raise a syntax warning when installing dlib.

```
<string>:235: SyntaxWarning: invalid escape sequence '\('
<string>:236: SyntaxWarning: invalid escape sequence '\('
<string>:237: SyntaxWarning: invalid escape sequence '\('
```


davisking:
Thanks for the PR :)

debnil-mercor:
Quick bump to the above - a new release with this commit would really be appreciated and fix breaking changes! @davisking

PR #3074, opened 2025-04-28
Add linear_ layer for neural networks

Cydral:
This PR adds a new `linear_` layer that provides an optimized linear transformation for multi-dimensional inputs. The layer preserves the spatial dimensions of the input tensor while transforming the feature dimension.

**Features**

- Supports both biased and unbiased linear transformations
- Optimized for maintaining tensor dimensions (except for feature dimension)
- Compatible with existing DNN API patterns and serialization

This implementation is more efficient than reshaping inputs to use fc_ layers when working with multi-dimensional data.
Similar functionality exists in other major frameworks like PyTorch (`nn.Linear`), which operates on the last dimension while preserving other dimensions.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2170,7 +2170,7 @@ namespace dlib
         unsigned long get_num_outputs() const { return num_outputs; }
         void set_num_outputs(long num)
         {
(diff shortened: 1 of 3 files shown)


davisking:
Nice, thanks for the PR 😁

PR #3076, opened 2025-05-06
Add `reshape_to` layer for flexible tensor reshaping/rescaling

Cydral:
This implements a new layer that reshapes or resizes tensors with two operation modes:

1. Pure Reshape Mode:

- When input/output element counts match: simple tensor reshaping without value changes.

3. Spatial Rescaling Mode (similar to `resize_to` existing layer):

- When only spatial dimensions change: uses bilinear interpolation while preserving channels.

Includes `flatten` alias for common flattening operations

### DIFF
--- a/dlib/test/dnn.cpp
+++ b/dlib/test/dnn.cpp
@@ -2538,7 +2538,19 @@ void test_embeddings()
             embeddings_<7, 12> l;
             auto res = test_layer(l);
             DLIB_TEST_MSG(res, res);
-        }        
+        }
+        {
+            print_spinner();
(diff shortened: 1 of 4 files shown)


dlib/test/dnn.cpp
```diff
@@ -5168,6 +5213,7 @@ void test_multm_prev()
             test_embeddings();
             test_tril();
             test_basic_tensor_ops();
+			test_resize_to();
```
davisking, line 5216:
```suggestion
            test_resize_to();
```

davisking (commented):
Nice, this is great.  Thanks for the PR :)

PR #3080, opened 2025-05-31
remove std::move where clang warns of pessimizing-move

ribbon-otter:
As I understand it, `std::move` achieves nothing on these lines, since they are already temporary objects, and can even prevent some optimizations.


davisking:
Just so, thanks for the PR :D

PR #3086, opened 2025-06-08
Minor optimization: `emplace_back` neighbors

reunanen:
I maintain an application that calls `label_connected_blobs` a lot. Noticed while profiling that the code that collects a `point`'s neighbors is a bit of a hotspot there, and it seems that this small (and hopefully rather uncontroversial) change makes a measurable (even if not huge) difference.

### DIFF
--- a/dlib/image_transforms/label_connected_blobs.h
+++ b/dlib/image_transforms/label_connected_blobs.h
@@ -26,7 +26,7 @@ namespace dlib
             for (long i = -2; i <= 2; ++i)
                 for (long j = -2; j <= 2; ++j)
                     if (i!=0||j!=0)
-                        neighbors.push_back(point(p.x()+i,p.y()+j));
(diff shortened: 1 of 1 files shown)


davisking:
Huh, really?  Both of those should end up turning into the same machine code.  It's cleaner the new way, so a good change regardless.  But are you sure it's really faster?  I would only expect a difference if compiled without compiler optimizations and/or with DLIB_ASSERT enabled.

reunanen:
> But are you sure it's really faster?

Hmm, good question. I did measure, and I did look at the profiler reports before and after the change – but not too rigorously perhaps.

Also, this (legacy) project uses an older version of the compiler, which may explain something (?). Also, optimizations should be enabled and assertions disabled, but I can't really say I triple-checked everything.

I could go back and do more measurements if needed, but I guess that's not really necessary.

davisking:
Na, no need. Just curious.

PR #3087, opened 2025-06-08
Use CUDAToolkit, fix #2833

### LINKED ISSUE #2833: Migrate to FindCUDAToolkit.cmake
### Main idea

NVIDIA considers FindCUDA.cmake (`find_package(CUDA)`) deprecated: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#importing-tarballs-into-cmake. The common approach to consume cudatoolkit today is to use FindCUDAToolkit.cmake (`find_package(CUDAToolkit COMPONENTS cudart cublas ...)`).

Aside from FindCUDA being obsolete, there are notable differences between the modules: e.g. `FindCUDAToolkit.cmake` allows its `CUDAToolkit_ROOT` variable to take the form of a list of paths, and this way it supports splayed installations of cudatoolkit, where different components (cudart, cublas, ...) may reside in different prefixes (as is the case e.g. in [nixpkgs](https://github.com/NixOS/nixpkgs/))

Thanks!

### Anything else?

_No response_

Tobias-Fischer:


Tobias-Fischer:
This has been working in conda-forge for quite some time without issues.

Tobias-Fischer:
CI fails because CUDA is now enforced, should be easy to set the relevant flag to `OFF`. Note that CUDA is currently silently failing in the current CI.

I won't have time to develop this further, but if someone wants to take over, please feel free to :)

Tobias-Fischer:
Bump

davisking:
How about https://github.com/davisking/dlib/pull/3090?  Is that variant working for you?   I haven't gone over it in detail yet but it looks reasonable on the surface.

Tobias-Fischer:
Looks good thanks!

PR #3089, opened 2025-06-16
fix the miss of including matlab ut library

gmm76:
'\dlib\matlab\cmake_mex_wrapper' missed including the library 'libut.lib', which leads to a link error of 'undefined reference to utIsInterruptPending'. The version fix the error by adding 2 lines (line 41 & 46), and append ' ${MATLAB_UT_LIBRARY}' in line 48 in '\dlib\matlab\cmake_mex_wrapper'


davisking:
Awesome, thanks for the PR :)

PR #3091, opened 2025-06-19
Fix numeric overflow when accessing pixel data of very large images

reunanen:
Problem: reading large images (in my case, it was a 40792x20480 RGBA PNG one) causes a crash, when the product _row number_ * _width step_ overflows a 32-bit integer

Solution: change the type of the width step to be `size_t` (and not `long`)


davisking:
Yep, agreed.  Should just be `size_t`.  Thanks for the PR :)

PR #3092, opened 2025-06-23
`throw()` -> `noexcept`

r-barnes:
`throw()` has been deprecated since C++11 and is removed in C++20. `noexcept` is the accepted alternative.

This modifies additional instances that weren't covered by the previously merged #3064 - apologies for missing those at the time.


davisking:
Thanks for the PR

Issue #2183, opened 2020-09-18
shape_predictor: allow dpoint as training input


ejoebstl:
Currently, the shape predictor training example [showcases how to provide](https://github.com/davisking/dlib/blob/master/examples/faces/training_with_face_landmarks.xml) training and test data using integer coordinates.

Hoever, I would like to train a shape predictor using float coordinate pairs. I hope to gain increased precision by doing so. Since the shape predictor is also giving float numbers as output, I would assume that this should be possible.

However, there is no documentation to be found on the input format.

I'm happy to extend the documentation as soon as I know how this works.

davisking:
This isn't supported by the current API.  There isn't any particularly good reason, it's just how the code is setup. It could use floating point coordinates.  I am somewhat skeptical that someone can label landmarks to sub-pixel accuracy though.  Do you really have a problem domain where humans, or whatever it is that generates the labels, somehow really knows the location of the landmarks to sub-pixel accuracy?

ejoebstl:
Thanks for getting back.

I am using [this](https://www.kaggle.com/selfishgene/exploring-youtube-faces-with-keypoints-dataset) dataset, which supplies float coordinates. The images are quite small, therefore I assumed it might make a difference.

davisking:
Were they annotated at a higher resolution and then downsampled?

In any case, if you want to put up a PR that switches the
full_object_detection to use dpoint instead of point (but in a way that's
reasonably backwards compatible so existing users are not bothered) that
would be cool with me.

guilbut:
Hi ! 
I am developing a software for which I need sub-pixel tracking with coordinates of points in float. Unfortunately the float rounding at the output of the shape predictor causes a jitter that is far too important for my needs. For the learning I have a database annotated very precisely in sub-pixel. I managed to give numpy matrices with floats as input (hoping they are not truncated internally ), but the shape predictor seems to only output integer points, contrary to the statement of "Since the shape predictor is also giving float numbers as output".
I unfortunately don't have the level in C ++ to do the modification cleanly and to offer a pool request. Does one of you two already tried to start modifying the code to allow dpoint as input to the training and as output of the predictor? 
Thanks !

davisking:
Yeah this would be a reasonable thing to upgrade. If someone wants to go replace `point` with `dpoint` in the right places and make any necessary attendant adjustments that would be cool.  It has to be done in a way that's backwards compatible with old code, but that shouldn't be hard though.

akulabs8:
[...]

PR #3095, opened 2025-07-23
full_object_detection: API Support for dpoint (#2183)

Infinage:
- Add constructor accepting parts as std::vector<dpoint>
- Internally switch parts representation to std::vector<dpoint>
- Update serialization logic to version 2 to support dpoint
- Deprecate constructor that takes std::vector<point>


davisking:
Nice, thanks for the PR.

Issue #3097, opened 2025-08-01
std::aligned_storage_t is deprecated in c++23


ribbon-otter:
### Main idea

Currently, dlib uses `std::aligned_storage_t` in two places: storage.h line [528](https://github.com/davisking/dlib/blob/fa163cbc07c6dc923ad4420df9553bf7a13dfe2f/dlib/any/storage.h#L528) and [line 776](https://github.com/davisking/dlib/blob/fa163cbc07c6dc923ad4420df9553bf7a13dfe2f/dlib/any/storage.h#L776). In C++23, aligned_storage_t is deprecated, by [p1413r3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1413r3.pdf)

They suggest that `std::aligned_storage_t<sizeof(T), alignof(T)> `is replaced with 
```C++
alignas(T) std::byte t_buff[sizeof(T)];
```

However, `std::byte` is introduced in C++17, and dlib supports all the way back to C++14. We could use `unsigned char`. The obvious alternative, `uint8_t` isn't mentioned by cppreference as safe with [reinterpret_cast](https://en.cppreference.com/w/cpp/language/reinterpret_cast.html).

Alternatively, dlib could just continue to use `std::aligned_storage_t ` despite the deprecation warnings when compiled with `-std=c++23` flag. 

### Anything else?

Would dlib be interested in a pull request which eliminated std::algined_storage_t and replaced it with `unsigned char`?

davisking:
Yeah good idea.  No reason not to switch to `alignas(T) unsigned char t_buff[sizeof(T)];`  That works just the same. The thing `std::byte` has going for it is it makes the intention clear of "hey this isn't trying to be any kind of string or character data!" is all.  But both are bytes.

You want to put up a PR that makes the change? :)

ribbon-otter:
This issue is now solved in master.

PR #3100, opened 2025-08-09
std::aligned_storage_t is deprecated in c++23

ribbon-otter:
This is the pull request which solves [this corresponding bug](https://github.com/davisking/dlib/issues/3097). 

Note that while trying to test it, I was unable to run all of the tests, as the makefile to build the tests would fail with 
```
../dlib/dnn/layers.h:2447:25: error: casting ‘dlib::alias_tensor_const_instance’ to ‘dlib::tensor&’ does not use 
‘dlib::alias_tensor_const_instance::operator const dlib::tensor&()’ [-Werror=cast-user-defined]
 2447 |             tt::gemm(1, (tensor&)sgi, 1, gi, false, w, true);
      |                         ^~~~~~~~~~~~
```

However, this would happen on master as well, so this isn't the fault of my patch.

### DIFF
--- a/dlib/any/storage.h
+++ b/dlib/any/storage.h
@@ -525,7 +525,7 @@ namespace dlib
             }
 
         private:
-            std::aligned_storage_t<Size, Alignment> data;
+            alignas(Alignment) unsigned char data[Size];
             void (*del)(storage_stack&)                         = nullptr;
             void (*copy)(const storage_stack&, storage_stack&)  = nullptr;
             void (*move)(storage_stack&, storage_stack&)        = nullptr;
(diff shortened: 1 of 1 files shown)


davisking:
Nice, thanks for the PR.

How did you hit that build error?  All the tests build on all the test platforms.  You trying to use literally the `dlib/test/makefile` file?  I just realized that file doesn't set C++14 (everything really uses the CMakeLists.txt file, the makefile is there just for convenient hacking sometimes and as an example, but it's out of date.  Fixing that...).

ribbon-otter:
I am using the cmake generated makefile. Some files compile but display the `-Wcast-user-defined` warning. A file later is apparently configured with `-Werror=cast-user-defined` and that causes an error when it is compiles. The easy "fix" would be to dig into the CMakeLists.txt files and disable the flag that sets Wcast-user-defined to be an error on just those later files.

Perhaps this is a compiler difference? Compilers do differ on what they warn about sometimes.

Perhaps the correct fix would be to change the [offending line](https://github.com/davisking/dlib/blob/master/dlib/dnn/layers.h#L2447) to which does explicitly call the user defined conversion function.

davisking:
Yeah presumably it would not warn if that line used a `static_cast` instead (which is a better way to do it).  Does switching to that make it not complain for you?  You want to send a PR for that if so? :)

PR #3101, opened 2025-08-10
Cache MKL plans

pfeatherstone:
This PR caches MKL plans. This substantially improves performance when using MKL sequential and MKL TBB. Not benchmarked with MKL RT but I expect the same results.

### DIFF
--- a/dlib/fft/mkl_fft.h
+++ b/dlib/fft/mkl_fft.h
@@ -2,17 +2,143 @@
 #define DLIB_MKL_FFT_H
 
 #include <type_traits>
+#include <memory>
+#include <unordered_map>
 #include <mkl_dfti.h>
 #include "fft_size.h"
 
-#define DLIB_DFTI_CHECK_STATUS(s) \
-    if((s) != 0 && !DftiErrorClass((s), DFTI_NO_ERROR)) \
-    { \
-        throw dlib::error(DftiErrorMessage((s))); \
-    }
+#ifdef DLIB_USE_MKL_WITH_TBB
+// This is a workaround to make libdlib link to libtbb.so explicitly and adding its path to the build tree RPATH.
+// This is because libmkl_tbb_threads.so depends on libtbb.so but dlib doesn't normally have any explicit symbols to libtbb.
+// If you don't do this, the runtime path of libtbb.so is stripped after build time.
+// Without this, at runtime you would get an error like "can't find libtbb.so"
+// And you would have to manually set it via LD_LIBRARY_PATH or something.
+// The better way to get around this is force cmake to add libtbb.so to RPATH. 
+// But i can't find a way to do it cleanly. Hopefully there isn't much performance impact here.
+extern "C" const char* TBB_runtime_version();
+#endif
 
 namespace dlib
 {
+    namespace mkl_details
+    {
+
+//----------------------------------------------------------------------------------------------------------------
+
+        constexpr auto check_status = [](auto s)
+        {
+            if(s != 0 && !DftiErrorClass(s, DFTI_NO_ERROR))
+                throw std::runtime_error(DftiErrorMessage(s));
+        };
+
+//----------------------------------------------------------------------------------------------------------------
+
+        struct mkl_deleter { void operator()(DFTI_DESCRIPTOR* h) {DftiFreeDescriptor(&h);} };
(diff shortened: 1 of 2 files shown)


pfeatherstone:
Need to add a Github Actions runner that actually tests all the MKL stuff. Probably one for each option: MKL RT, MKL SEQUENTIAL and MKL TBB. Currently i think there are some bugs in the cmake stuff.

pfeatherstone:
@davisking Ready for an initial review. Linking to libmkl_rt.so was broken. Mainly because by default it loads libmkl_intel_thread.so at runtime using `dlopen` which in turns requires openmp, specifically the intel version. If that's not in your link libraries it breaks at runtime. I think this behaviour is not desirable. What i found is if you're also linking to ffmpeg or other libraries, you're inadvertently linking to gomp. So you might not see the error. Though libmkl_intel_thread.so requires the intel openmp, not the GNU. So it was actually using the "wrong" symbols. So i changed the cmake BLAS stuff to link to one of 
- MKL sequential
- MKL tbb
- MLL thread (openmp)
With MKL thread you explicitly link to openmp at link-time. It doesn't use `dlopen`. I think that's better.
This required me to refactor the `find_blas.cmake` script which was a mess to be honest. I've also added some jobs to Github Actions to test all three version of MKL. The `build_cpp.yml` script needs some serious refactoring but that can be for another PR.

For the BLAS stuff i'm now using [cmake_dependent_option](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) which simplifies the super messy logic in the CMakeLists.txt script.

pfeatherstone:
@davisking Do you mind having a look at this? Or are you wanting to merge the CUDA stuff first? I would rather get this in before i touch cmake revamping stuff again as I would be duplicating certain things in here already.

dlib/cmake_utils/find_blas.cmake
```diff
@@ -194,290 +219,25 @@ if (UNIX OR MINGW)
...
-
-   if (NOT blas_found)
-      find_library(generic_blas blas PATHS ${extra_paths})
-      if (generic_blas)
-         set(blas_libraries ${generic_blas})
-         set(blas_found 1)
-         message(STATUS "Found BLAS library")
-      endif()
-      mark_as_advanced( generic_blas)
-   endif()
```
davisking, line 277: Why get rid of these ones?

pfeatherstone: In fact, if you use 
```
set(BLA_PREFER_PKGCONFIG 1)
find_package(BLAS)
```
it will first check with pkg-config then check the rest. I think that's a sufficient check if MKL isn't found. I'll do a commit and test. I'll see if i can get the windows build to use blas.

davisking: Eh. I would rather not remove stuff that isn't broken.  I have in the past used dlib for instance on systems where those blas libraries were what was available.  It wouldn't blow my mind if there are users out there that rely on those.  Intel MKL and openblas are the most popular right now.  But invariably there are sub-communities of people in their own worlds.

pfeatherstone: The CI/CD tests MKL (of course) and openblas and that's all happy. I've tested with atlas as well and it works on my machine. We could add more jobs to Github Actions if you want. Probs a good idea anyway. In fact, the unit tests haven't been working with MKL for some time now which shows how many people are actually properly testing dlib... With the updated `build_cpp.yml` it's super easy to add more configurations.  The default [FindBLAS](https://cmake.org/cmake/help/latest/module/FindBLAS.html) is a pretty good catch all. We can add back some of those fallbacks but there's quite a lot of code there to cater for maybe a 1 or 2 users world-wide. We should maybe embrace the KISS principle a bit more for this kind of stuff.

pfeatherstone: Added ATLAS job

davisking: Yeah alright, we can see what the default FindBLAS is like.  That stuff used to not be great but maybe it's dope now.

pfeatherstone:
@davisking hold fire on merging. I haven't got the windows jobs working correctly with openblas and MKL. `vcpkg` takes too long to install those deps and haven't figured out how to use it correctly with cmake.

pfeatherstone:
Right, i'm gonna need some help. I don't have a windows machine i can use for dev. I'm giving up trying to debug the windows builds simply by looking at github actions logs.

PR #3103, opened 2025-08-11
fix -Wcast-user-defined warning

ribbon-otter:
The tests compile and run mostly. I get deprecated warnings for something in ffmpeg which can an error on one of the final tests. I am not enough of an expert at cmake to understand why only some of the files are compiled with -Werror. 

Please look carefully at this pull request as I turned something that was `const` into non-`const`. I am not convinced it is correct. Though the c-style cast was likely modifying a `const` variable through a non-`const` reference anyway.

`static_cast` did not work. 

I am planning on adding a flag to the `CMakelist.txt` test file that lets me turn off `-Werror` so I can at least fully test changes before submitting them. Particularly while I am patching away warnings like this.

### DIFF
--- a/dlib/dnn/layers.h
+++ b/dlib/dnn/layers.h
@@ -2440,11 +2440,13 @@ namespace dlib
                     tt::assign_bias_gradient(pb, gi);
                 }
             }
-
-            const auto& prev_gradient = sub.get_gradient_input();
-            auto sgi = alias_tensor(prev_gradient.num_samples() * prev_gradient.k() * prev_gradient.nr(), num_inputs)(prev_gradient, 0);
+            
+            //prev_gradient is not const, so that sgi isn't const
+            //since sgi is used as a destination for tt::gemm
+            auto& prev_gradient = sub.get_gradient_input();
+            alias_tensor_instance sgi = alias_tensor(prev_gradient.num_samples() * prev_gradient.k() * prev_gradient.nr(), num_inputs)(prev_gradient, 0);
             auto w = weights(params, 0);
-            tt::gemm(1, (tensor&)sgi, 1, gi, false, w, true);
+            tt::gemm(1, sgi, 1, gi, false, w, true);
         }
 
         alias_tensor_instance get_weights() { return weights(params, 0); }


ribbon-otter:
I think the new warning/error happens at the end of cmake script when it then compiles examples.

ribbon-otter:
Specifically I get warnings/errors that [ch_layouts](https://ffmpeg.org/doxygen/trunk/structAVCodec.html#a3aa4a57d42f9b9dfc867fe3fbcad4539) is deprecated, in dlib/media/ffmpeg_muxer.h

pfeatherstone:
You'll need to submit an issue with a reproducible example for me to diagnose that. To be honest i think we should drop support for the older FFMpeg libs now.

arrufat:
I get a lot of deprecation warnings on Arch Linux with FFmpeg 7.1.1, the `ch_layouts` deprecation included, just by running:

```console
$ cd examples
$ cmake -G Ninja -B build
...
-- Searching for FFMPEG/LIBAV
-- Checking for modules 'libavdevice;libavfilter;libavformat;libavcodec;libswresample;libswscale;libavutil'
--   Found libavdevice, version 61.3.100
--   Found libavfilter, version 10.4.100
--   Found libavformat, version 61.7.100
--   Found libavcodec, version 61.19.101
--   Found libswresample, version 5.3.100
--   Found libswscale, version 8.3.100
--   Found libavutil, version 59.39.100
-- Found FFMPEG/LIBAV via pkg-config in `/usr/lib`
...
$ ninja -C build ffmpeg_webcam_face_pose_ex
...
dlib/media/ffmpeg_muxer.h:814:80: warning: ‘AVCodec::ch_layouts’ is deprecated [-Wdeprecated-declarations]
  814 |                         av_channel_layout_copy(&pCodecCtx->ch_layout, &pCodec->ch_layouts[0]);
      |                                                                                ^~~~~~~~~~
...
```

pfeatherstone:
Maybe the easiest thing to do is drop support for older versions. I'll have a look later though

pfeatherstone:
I'm willing to bet it's because the following isn't working as expected with all versions of ffmpeg

https://github.com/davisking/dlib/blob/4cf8f5759fe715b0defb0ba998ac1c354b814cca/dlib/media/ffmpeg_details.h#L67C1-L69C7

pfeatherstone:
Oh i see the problem. Yeah, I fixed this in my own repo ages ago. Sorry. I can submit a PR at some point. FFmpeg is quite annoying at constantly changing APIs. In my own private repos i only ever support the latest version since i always build from source direct from cmake (using ExternalProject_Add)

davisking:
Thanks for the PR, it's a reasonable way to avoid that warning :)

pfeatherstone:
I've started https://github.com/davisking/dlib/pull/3105 for ffmpeg 7 warnings

PR #3104, opened 2025-08-11
add option to disable -Werror on tests

ribbon-otter:
Adds a cmake option to disable -Werror on tests. This is helpful because when dealing with multiple disparate sources of warnings, it is nice to be able to compile and run tests so see if your fix for one, introduced new problems or not. 

-Werror defaults to on. Use `cmake .. -DWerror=OFF` inside the test build directory to disable.

### DIFF
--- a/dlib/test/CMakeLists.txt
+++ b/dlib/test/CMakeLists.txt
@@ -9,8 +9,12 @@ cmake_minimum_required(VERSION 3.10.0)
 set (target_name dtest)
 PROJECT(${target_name})
 
+option(Werror "Error on warnings" ON)
+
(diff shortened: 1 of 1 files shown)


ribbon-otter:
After thinking about this more, I am not convinced this is a useful feature to have, since passing an argument to cmake isn't much less difficult than just manually editing the CMakeLists.txt file temporarily when I need to. I won't feel bad if you reject this pull request.

davisking:
Yeah, someone can just edit the file instead.  And when there is some problem we should fix it anyway instead of flipping a switch to ignore it too :D 

Thanks for the PR though.

PR #3110, opened 2025-08-21
[unicode] fixed warning in convert_mbstring_to_wstring()

pfeatherstone:
This suppresses a warning that's been bugging me for a while.


pfeatherstone:
Also having functions returning `const std::string` or `const std::wstring` is pointless but i'm not fixing that.

davisking:
Sweet, thanks for the PR :D

PR #3111, opened 2025-08-27
Add Adaptive Computation Time (ACT) layer

Cydral:
Implement adaptive_computation_time_ layer following Graves (2016) "Adaptive Computation Time for Recurrent Neural Networks".

This layer enables neural networks to adaptively determine the number of computation steps to perform for each sequence position, spending more computation on difficult parts while quickly processing easier parts.

Key features:
- Full CPU and CUDA implementations
- Configurable halting threshold and ponder penalty for regularization  
- Support for depth-dependent gradient scaling
- Early termination optimization when all positions converge
- Template-based max_steps configuration (act4, act8, act16 aliases)

**Note:** This ACT implementation can serve as a building block for hierarchical reasoning mechanisms such as those described in "Hierarchical Reasoning Model" (Wang et al., 2025, arXiv:2506.21734) for advanced sequential reasoning tasks.


dlib/cuda/cuda_dlib.cu
```diff
@@ -2733,6 +2733,241 @@ namespace dlib
...
+                const size_t n = pos / seq_len;
+                const size_t s = pos % seq_len;
+
+                float logit = b_halt;
+
+                for (size_t c = 0; c < num_channels; ++c) {
+                    for (size_t d = 0; d < d_model; ++d) {
+                        const size_t in_idx = ((n * num_channels + c) * seq_len + s) * d_model + d;
+                        const size_t weight_idx = c * d_model + d;
+                        logit += input_data[in_idx] * W_halt[weight_idx];
```
davisking, line 2763: How about using `warp_reduce_atomic_add()` instead here?   It's probably *way* faster.

Take a look at `_cuda_dot_prods()` for an example of how to do it.  It's a relatively simple change but lets you use all the cuda cores.  http://devblogs.nvidia.com/parallelforall/faster-parallel-reductions-kepler/ is a good post explaining the warp reduce atomic add technique.

dlib/cuda/tensor_tools.h
```diff
@@ -2392,6 +2392,85 @@ namespace dlib { namespace tt
...
+        const tensor& halt_probs,
+        resizable_tensor& cumulative_halting,
+        resizable_tensor& remainders,
+        resizable_tensor& n_steps,
+        long batch_size,
+        long seq_len,
+        long d_model,
+        long num_channels,
+        float halt_threshold,
+        long current_step
```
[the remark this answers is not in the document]
Cydral: I have indeed commented on all 4 utility functions to clarify the use of parameters.

dlib/cuda/cuda_dlib.cu
Cydral: > How about using `warp_reduce_atomic_add()` instead here? It's probably _way_ faster.
> 
> Take a look at `_cuda_dot_prods()` for an example of how to do it. It's a relatively simple change but lets you use all the cuda cores. http://devblogs.nvidia.com/parallelforall/faster-parallel-reductions-kepler/ is a good post explaining the warp reduce atomic add technique.

Yes, I can indeed rewrite the kernel to optimise its operation using warp_reduce_atomic_add(). It's done. However, this is not possible for the following three. For the last utility function, we could completely flatten the index, but I am not convinced of the performance gain, rather than clarity in porting the code from CPU to CUDA. Honestly, I would prefer to leave them as they are if you agree...

davisking: Yeah that's fine.  Whatever you think is most appropriate.

Cydral: So let's keep it simple; in any case, based on the tests carried out, I don't really notice any improvement in performance, especially given the size of the matrices or tensors we usually use in Dlib.

davisking:
Nice, this is great.  Thanks for the PR :D

Cydral:
No problem. My next plan is to work on a native Transformer implementation to make them easier to use within Dlib.
In fact, I think I've brought almost everything needed: the existing template structure used in samples works well to maintain the library's style while allowing for flexible configuration (number of layers, vocab size, etc.). I'll also use this opportunity to add an image preprocessing layer, creating a unified framework for Transformers in Dlib.
After that, training a respectable-sized model will be the next challenge, of course... Perhaps, the place for another example to add ;-)

PR #3112, opened 2025-09-01
Improve the somewhat flaky `test_loss_multibinary_log` by avoiding samples very close to class boundaries

reunanen:
**Problem**: This test case (`test_loss_multibinary_log`) requires non-zero error, which may be difficult to obtain when samples can be arbitrarily close to the class boundaries.

**Solution**: If a newly generated sample is very close to a class boundary, simply try again.


davisking:
Thanks for the PR :D

PR #3113, opened 2025-09-01
Improve numerical stability of batch normalization on CPUs

reunanen:
**Problem**: Calculating mean and variance in the CPU implementation of batch normalization is prone to losing some numerical precision (when the data is not zero mean, for example).

**Solution**: Use Welford's algorithm that is numerically more stable.

Note that the CUDA implementation should already be doing this (or something similar at least), so this change should make the CPU implementation better match the CUDA one.


reunanen:
Note that [the CI error](https://github.com/davisking/dlib/actions/runs/17376948930/job/49325318105?pr=3113) is addressed separately in PR #3112.

davisking:

PR #3115, opened 2025-10-06
Implementation of ARC-AGI dataset loader

Cydral:
Introduce arc_agi_manager for loading and tokenizing ARC-AGI reasoning tasks.

This implementation provides:
- JSON-type parsing for ARC-AGI task files (training and evaluation sets)
- Grid tokenization with special markers for implicit dimension encoding
- Sliding window context generation for causal language model training
- Detokenization utilities for grid reconstruction from token sequences
- Serialization/deserialization support for faster dataset loading
- Task access by index or ID

The tokenization strategy encodes variable-size grids (1x1 to 30x30) using row-end markers, enabling transformer models to learn grid dimensions without explicit specification.
Context windows implement left-padding for proper causal attention masking.


davisking:
Awesome, thanks for another PR :D

Issue #1607, opened 2018-12-28
[request] NaN check for optimization algorithms?


Red-Portal:
I found out after painful debugging that dlib's `find_global_mean` fails with a segmentation fault when the observed values contain NaN.
What do you think about including a NaN checking procedure to optimization algorithms?
If you like the idea I'll work on it and commit to dlib.

meSajied:
I am trying to find the solution. I am looking for this bug in optimization folder. In which file I need to look for this bug?

PR #3142, opened 2026-04-12
Add DLIB_CASSERT for optimization algorithms

KamalElFeniche:
This PR adds an assert to check if the computed values are not NaNs. This is related to #1607.


davisking:
Thanks, but this might break existing users code.   E.g. someone could, by the current contract, have a function that is giving nans and be just ignoring it and taking the best result they found that wasn't nan.  Which by the current contract they are allowed to do.

So better to let someone who doesn't want their objective to ever be nan to add that assert to their own output if that's what they want.

davisking:
Er, wait, sorry, just read the other thread.  It segfaulted?  That segfault should just be fixed.  Can you post an example that causes the segfault?   Although that other thread is very old.  Did you observe a segfault?  What bad thing does this fix exactly?

PR #3150, opened 2026-06-27
Workaround for missing typename compiler error

### LINKED ISSUE #3149: [Bug]: gcc 15.2 typename in depedent scope problem

Linux (x86-64)

### dlib version

a41c2e65

### Python version

N/A

### Compiler

GCC 15.2

### Expected Behavior

no compilation failures

### Current Behavior

compilation of dlib fails with the following error
```
image_processing/scan_image_pyramid_tools.h:85:1: error: need ‘typename’ before ‘std::__cxx11::list<std::pair<__decay(long unsigned int&), __decay(const dlib::rectangle)> >::iterator’ because ‘std::__cxx11::list<std::pair<__decay(long unsigned int&), __decay(const dlib::rectangle)> >’ is a dependent scope [-Wtemplate-body]
   85 |                 list_type::iterator i = sorted_rects.begin();
```

### Steps to Reproduce

checkout master branch at a41c2e65 and compile with GCC 15.2

### Anything else?

I recently moved to Ubuntu 26.04 which uses GCC 15.2 as the default C++ compiler. This error appeared when compiling a fresh cmake config of a project that depends on dlib. I prefer to use the default GCC version for my Linux distro, so this is a problem.

Honestly, I think it may be a bug in GCC because the type it is complaining about is `list_type::iterator` where `list_type` is defined as `typedef std::list<std::pair<unsigned long, rectangle> > list_type;`. This is not a dependent type because it does not refer to any template parameters.

If I replace `list_type::iterator` with `auto`, it compiles fine. I can make a PR for this, but it seems like a dumb workaround.

kSkip:
GCC 15.2 complains about this iterator type missing the typename keyword as if it were a dependent type, but it is not. However, using auto works. Fixes https://github.com/davisking/dlib/issues/3149

### DIFF
--- a/dlib/image_processing/scan_image_pyramid_tools.h
+++ b/dlib/image_processing/scan_image_pyramid_tools.h
@@ -82,7 +82,7 @@ namespace dlib
                  itr < scanner.get_max_pyramid_levels() && cur.area() < max_area; 
                  ++itr)
             {
-                list_type::iterator i = sorted_rects.begin();
+                auto i = sorted_rects.begin();
                 while (i != sorted_rects.end())
                 {
                     const rectangle temp = move_rect(i->second, cur.tl_corner());


davisking:
Cool, thanks for the PR.  Something updated in github actions apparently because that windows-latest error is obviously unrelated to this PR.

PR #3152, opened 2026-07-05
Added support for unix sockets

kSkip:
I'm working on a Fast CGI server for my ML models and would like to integrate supporting changes into dlib's networking module. One of these changes is supporting socket type AF_UNIX, so web servers like nginx can bypass the network stack when communicating with the fcgi server. Another is a listener factory method that can accept connections on a user-provided socket. In many fast cgi setups, the application is expected to listen on stdin (which is bounded to a socket), so the user-provided socket is helpful here.

I implemented the required changes for both posix and win32 including tests. However, I'm not sure how to test the new `create_listener_from_socket` method. It would require platform specific code in the unit test.


dlib/sockets/sockets_kernel_abstract.h
```diff
@@ -126,6 +126,57 @@ namespace dlib
...
+        std::unique_ptr smart pointer instead of a C pointer.
+    !*/
+
+    int create_listener_from_socket (
+        listener*& new_listener,
+        connection::socket_descriptor_type sock
+    );
+    /*!
+        requires
+            - sock is either a ICP/IP socket or a Unix socket
```
davisking:
```suggestion
            - sock is either a TCP/IP socket or a Unix socket
```

davisking:
> I implemented the required changes for both posix and win32 including tests. However, I'm not sure how to test the new create_listener_from_socket method. It would require platform specific code in the unit test.

How are you making the socket you have when you call `create_listener_from_socket()` in your real application?  Can't just do basically that in the tests or even better have an abstracted generic method for doing it?

kSkip:
>How are you making the socket you have when you call create_listener_from_socket() in your real application?

For Fast CGI, you are usually not making the socket. Either systemd or another tool, like spawn-fcgi, will bind a socket to file descriptor 0 (stdin). So, I just pass 0 as the argument to `create_listener_from_socket()`.

kSkip:
>Can't just do basically that in the tests or even better have an abstracted generic method for doing it?

Yes, it's easy enough to just create, bind, and listen on a socket to use for the test, but I wasn't sure if putting a bunch of platform specific headers and an `ifdef _WIN32 ...` directive in test/sockets2.cpp would be the best way.

I could also add some public function, say `create_listening_socket()`, with two overloads that factors out the socket creation code in `socket_kernel_1[2].cpp`, and use that.

Let me know what you prefer, and thanks for reviewing!

davisking:
> > Can't just do basically that in the tests or even better have an abstracted generic method for doing it?
> 
> Yes, it's easy enough to just create, bind, and listen on a socket to use for the test, but I wasn't sure if putting a bunch of platform specific headers and an `ifdef _WIN32 ...` directive in test/sockets2.cpp would be the best way.
> 
> I could also add some public function, say `create_listening_socket()`, with two overloads that factors out the socket creation code in `socket_kernel_1[2].cpp`, and use that.
> 
> Let me know what you prefer, and thanks for reviewing!

Yeah make some public function that cleanly does it and use that in the tests.  That would be sweet.

kSkip:
I believe this is ready now. I added a public function called `create_listening_socket` which returns an error code and writes to a socket type argument. That socket can then be used in `create_listener_from_socket` to produce equivalent behavior to `create_listener`.

This also allows for using sockets obtained from external APIs in listener objects.

davisking:
Looks good.  I just pushed a commit to harden up some parts of it.

  The main changes are:

  - `create_listening_socket()` now initializes socket startup itself. Since this is a new public API, callers should be able to use it directly, including on Windows, without first calling another socket helper that happens to initialize Winsock.

  - Unix socket paths are now validated before use. Previously an overlong path would be silently truncated into `sun_path`, which could make dlib bind or connect to a different socket path than the caller
  requested. Now we return `OTHER_ERROR` instead.

  - The listener creation code no longer unlinks an existing filesystem entry at the requested Unix socket path. Automatically deleting whatever is at that path is risky because it might be a regular file or
  another process’s socket. The caller should own cleanup of the pathname they pass in.

  - The `sockets2` test now cleans up its own test socket path more portably with `std::remove()` rather than calling `unlink()` directly.


What do you think?  Still work for your use cases?

kSkip:
Yes, thank you. Everything is working great as far as I can tell. Removing the automatic unlink makes sense.

davisking:
Sweet, thanks for the PR then :D

PR #3154, opened 2026-07-17
Allow setting sockstreambuf buffer size and max putback

kSkip:
I would like to use the sockstreambuf with a larger buffer size, and think it would be nice to adjust the max put back as well. In addition, I also changed the buffer pointers to unique_ptr. I figured that was just pre-C++11 code.

### DIFF
--- a/dlib/sockstreambuf/sockstreambuf.h
+++ b/dlib/sockstreambuf/sockstreambuf.h
@@ -4,7 +4,10 @@
 #define DLIB_SOCKStREAMBUF_Hh_
 
 #include <iosfwd>
+#include <limits>
+#include <memory>
 #include <streambuf>
+#include "../assert.h"
 #include "../sockets.h"
 #include "sockstreambuf_abstract.h"
 #include "sockstreambuf_unbuffered.h"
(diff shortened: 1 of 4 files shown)


dlib/sockstreambuf/sockstreambuf.h
```diff
@@ -88,27 +103,17 @@ namespace dlib
...
-            catch (...)
-            {
-                if (out_buffer) delete [] out_buffer;
-                throw;
-            }
-            setp(out_buffer, out_buffer + (out_buffer_size-1));
-            setg(in_buffer+max_putback, 
-                 in_buffer+max_putback, 
-                 in_buffer+max_putback);
+            setp(out_buffer.get(), out_buffer.get() + (out_buffer_size-1));
```
kSkip, line 107: For here, should this be `out_buffer + out_buffer_size`? Because `epptr` is defined as "one past the end of the put area". If I understand this correctly, the effective buffer size will be 1 char smaller if we use `out_buffer_size-1`

davisking: Yeah epptr() is one past the put area, but the final allocated byte is deliberately excluded from that put area because this implementation uses it to hold the character that triggered overflow().  So it's fine, it's just a matter of how the other parts of the code define it.

davisking:
Looks good.  I just fixed `std::unique_ptr<char>` needing to be `std::unique_ptr<char[]>` and pushed some cleanup stuff.  Once all the tests run I'll merge it :D  

Thanks for another PR.

dlib/sockstreambuf/sockstreambuf.h
kSkip: Cool, thanks for explaining. I find the std::streambuf interface really confusing sometimes.

kSkip:
> I just fixed `std::unique_ptr<char>` needing to be `std::unique_ptr<char[]>`

Oh, good catch. Thanks for the help.

dlib/sockstreambuf/sockstreambuf.h
davisking: Yeah it's somewhat bizarre

PR #3155, opened 2026-07-29
pickup latest pybind11 and support free threaded

yokomichikt:
Add support for free threaded python.


davisking:
I just tried to fix some bugs in this PR but can't push.  So I moved this over to https://github.com/davisking/dlib/pull/3158 and will merge it when all the tests pass

PR #3158, opened 2026-08-09
Pybind11 free threading

davisking:
Updates to pybind11 v3.0.4 and declares that the gil doesn't need to be locked when python calls into dlib.

This is a copy of https://github.com/davisking/dlib/pull/3155 but with some issues fixed

Some stuff wasn't actually thread safe:
   - Made the jitter RNG thread_local in tools/python/src/numpy_returns.cpp:97.
   - Serialized concurrent inference on the same face-recognition or CNN detector instance in tools/python/src/face_recognition.cpp:72 and tools/python/src/ cnn_face_detector.cpp:52.

I also pulled the official pybind11 release.  What was here looks like one of their alpha snapshots.  We should use an official pybind11 release
In particular, I replaced the mixed development pybind11 snapshot with exact upstream pybind11 v3.0.4, commit d03662f0984f652b60e7ddce53d3868002275197. Also added provenance/update instructions in dlib/external/pybind11/README.dlib.

Then some other minor related cleanup


davisking:
All the tests are passing now @yokomichikt.  You want to try this out in your setup before I land it?  I'm pretty sure it works but you want to give it a test drive first? :D

yokomichikt:
I've verified pybind11-free-threading branch and it works very well. Thanks for quick response.

davisking:
No problem.  Thanks for the PR :D

